PR #23, opened 2014-09-29
Fix indentation error that might prevent check to be run

nbessi:


eLBati:
:+1: can the travis build be re-launched?

rvalyi:
:+1:  OK for the fix, but fragment_agreement test doesn't run pass yet...

nbessi:
Restoring constraint seems to break the test putting this in WIP

pedrobaeza:
It seems that this is not going to be continued, so I close.

PR #27, opened 2014-10-09
purchase_fiscal_position_update: port to v8 and new API

alexis-via:


rvalyi:
:+1: LGTM code review, no test

pedrobaeza:
:+1:

purchase_fiscal_position_update/purchase.py
```diff
@@ -51,16 +47,14 @@ def fiscal_position_change(
...
+                product = self.env['product.product'].browse(
+                    line.get('product_id'))
                 taxes = product.supplier_taxes_id
-                tax_ids = fp_obj.map_tax(
-                    cr, uid, fp, taxes, context=context)
-
-                line['taxes_id'] = [(6, 0, tax_ids)]
+                if fp:
+                    taxes = fp.map_tax(taxes)
+                line['taxes_id'] = [(6, 0, [tax.id for tax in taxes])]
```
yvaucher: taxes.ids ? instead of [tax.id for tax in taxes]

yvaucher:
Apart from my inline comment :+1:

purchase_fiscal_position_update/purchase.py
```diff
@@ -22,26 +22,22 @@
...
 
 
-class purchase_order(orm.Model):
+class purchase_order(models.Model):
     _inherit = "purchase.order"
 
-    def fiscal_position_change(
-            self, cr, uid, ids, fiscal_position, order_line,
-            context=None):
+    @api.multi
```
eLBati: Why not `@api.onchange`?

alexis-via:
@eLBati 

I don't know... I took example on odoo/addons/account/account_invoice.py line 1291 : product_id_change() has @api.multi

eLBati:
@alexis-via here is what I propose https://github.com/akretion/purchase-workflow/pull/1

lepistone:
For me it is good but not compulsory to always use new API idioms when porting modules, at least until it stabilizes a bit.
But @eLBati did the job so if you @alexis-via merge @elbati's branch into yours (or you say you have a reason not to do so), I'm happy to merge this PR.

alexis-via:
yes, I'll have a look+test and merge during the OCA code sprint at PyconFR.

alexis-via:
I have accepted the PR of @eLBati with a small modification (no trigger of the on_change with an order line is modified).

But, when working on this, I discovered a very bad bug in the code of the map_tax for the NEW API (the bug is not present in the code for the old API): https://github.com/odoo/odoo/issues/2261

I have made a PR to fix this bug on odoo v8. But, until this bug is fixed on the official branch or in OCB, it is NOT a good idea to use the map_tax function with the new API. So we should delay this merge until the bug is fixed.

pedrobaeza:
Yeah, we discovered it on spanish localization, because we have one fiscal position that maps two destination taxes for one source tax. Let's push to get this fixed.

alexis-via:
For the fix in map_tax with new API: 
Here is the PR on OCB: https://github.com/OCA/OCB/pull/63
Here is the PR on odoo official: https://github.com/odoo/odoo/pull/3316

eLBati:
:+1:

PR #33, opened 2014-10-21
[FIX] supplier invoice: avoid discount doubloon

rdeheele:
Standard discount is only visible when you active "Allow setting a discount on the sales order lines" in Sales Configuration.
If you active this option + install this addon : the discount is visible two times on invoice lines (only for supplier invoices).
This commit allows to display only one field in all cases.

### DIFF
--- a/purchase_discount/views/account_invoice_view.xml
+++ b/purchase_discount/views/account_invoice_view.xml
@@ -6,6 +6,9 @@
             <field name="model">account.invoice</field>
             <field name="inherit_id" ref="account.invoice_supplier_form"/>
             <field name="arch" type="xml">
+                <field name="discount" position="attributes">
+                    <attribute name="invisible">True</attribute>
+                </field>            
                 <field name="price_unit" position="after">
                     <field name="discount"/>
                 </field>


yvaucher:
Just a question. Why not using base discount field where it is and setting it's invisible param to always false ?

pedrobaeza:
You're right, but it's curious that discount doesn't appear in purchase orders and does on supplier invoices.

:+1:

lepistone:
Har, har! That's the second biggest dobloon I've ever seen! So you want to be a pirate, eh?

:+1:

rdeheele:
@yvaucher @pedrobaeza  mmh, the problem comes from standard addons.
The option should be more precise.
in standard addons, we should have two options, and the discount should be displayed conditionally in function of invoice type.
@lepistone Look behind you, a Three-Headed Monkey!

lepistone:
On a slightly lighter tone, the red build is unrelated to this change.

rdeheele:
I've posted a bug report here: https://github.com/odoo/odoo/issues/3227
but this commit could help some time.

PR #38, opened 2014-10-27
[ADD] purchase_budget_limit

clonedagain:
New module to block purchase orders when the budget is at risk of being exhausted

Supersedes this merge proposal from Launchpad: https://code.launchpad.net/~numerigraphe-team/purchase-wkfl/7.0-add-purchase_budget_limit/+merge/219829

"Here is a new module to block purchase orders when the budget is at risk of being exhausted.

Only validated budgets are checked.
Also, only invoiced purchases are counted (because only invoices are counted in the budget). I'll soon be publishing an additional module to count draft purchases in the budgets too.

Unfortunately, no automatic test bundled because the account_budget does not provide any either."

### DIFF
--- a/purchase_budget_limit/purchase.py
+++ b/purchase_budget_limit/purchase.py
@@ -1,0 +1,111 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    This module is copyright (C) 2014 Numérigraphe SARL. All Rights Reserved.
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+import time
+
+from openerp import netsvc
+from openerp.osv import orm
+from openerp.tools.translate import _
+
+
+class PurchaseOrder (orm.Model):
+    """Display a warning when confirming a Purchase if the budget is too low"""
+    _inherit = 'purchase.order'
+
+    # Initialization (no context)
+    def __init__(self, pool, cr):
+        """
+        Add a new state value.
+
+        Doing it in __init__ is cleaner than copying and pasting the field
+        definition in _columns and should be compatible with future/customized
+        versions.
+        """
+        # TODO: new API in v8 probably let us do it in a simpler way
+        super(PurchaseOrder, self).__init__(pool, cr)
+        super(PurchaseOrder, self).STATE_SELECTION.append(
+            ('over_budget', 'Over Budget'))
+
+    # Model methods (context except when called by the Workflow Engine)
+    def exhausted_budget_lines(self, cr, uid, ids, context=None):
+        """
+        Find the Budget Line which do not have enough funds left to pay the POs
+
+        @return: IDs of the Budget Lines
+        """
+        if not isinstance(ids, list):
+            ids = [ids]
+
+        budget_line_ids = set()
+        b_line_obj = self.pool['crossovered.budget.lines']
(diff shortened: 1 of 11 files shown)


purchase_budget_limit/wizard/purchase_budget.py
```diff
@@ -0,0 +1,62 @@
...
+        for po_id in context.get('active_ids'):
+            wf_service.trg_validate(
+                uid, 'purchase.order', po_id, 'purchase_confirm_overbudget',
+                cr)
+        return {'type': 'ir.actions.act_window_close'}
+
+    _defaults = {
+        'budget_line_ids': _get_budget_line_ids
+    }
+PurchaseBudget()
```
joaoalf: This is not necessary.

purchase_budget_limit/purchase.py
```diff
@@ -0,0 +1,116 @@
...
+        Doing it in __init__ is cleaner than copying and pasting the field
+        definition in _columns and should be compatible with future/customized
+        versions.
+        """
+        super(PurchaseOrder, self).__init__(pool, cr)
+        super(PurchaseOrder, self).STATE_SELECTION.append(
+            ('over_budget', 'Over Budget'))
+
+    # Workflow Action methods (no context)
+    def wkf_over_budget(self, cr, uid, ids):
```
joaoalf: If this action will always only set the state to 'over_budget', you can simply pass the the write({...}) to the workflow action. If you defined the method to allow others to overwride it, then it's ok.

joaoalf:
Thanks for the contribution. I just added some coments to your code that you might conside.

clonedagain:
Thanks for your suggestions @joaoalf.
Changed and rebased.

clonedagain:
Now conforms to PEP8.

max3903:
:+1:

jgrandguillaume:
Hi there,

Thanks for this contrib. I think this module serve a good purpose, but honestly, I'm not in favor of having a new state for this. Adding state is always a trouble when mixing with other modules (I mean unless you really don't have the choice to do so).

I would rather prefer 

a) or base this on a purchase_exception module (like the one existing in sale-workflow) => this module allow to raise various errors message on validation of SO. Make the same for PO would be nice. So we have the same way of managing all exceptions. It's also great because every company can then define his own exception. For SO, it's working really great ! This is my prefered option.

b) If too much work to deal with that new purchase_exception module (that I can understand), I'd prefer to have a boolean like "PO over actual budget" to deal with that instead of a new state introduced.

Would you consider this option ?

Regards,

Joël

clonedagain:
I'm not opposed to splitting the state part to another module, but I 
don't know the sale_exeption module so I'd need to see how much work 
that is.
That would probably have to wait until we port this module to v8 
however, because we don't have a development budget/time-slot for this 
at the moment, sorry.
I doubt moving the budget status to a checkbox would be a simpler change 
either: we would still need to inherit the transitions in the standard 
workflow.

purchase_budget_limit/purchase.py
```diff
@@ -0,0 +1,111 @@
...
+        for po_id in ids:
+            for ol in self.browse(cr, uid, po_id, context=context).order_line:
+                if ol.account_analytic_id:
+                    # should we check sub-accounts too?
+                    bl_ids = b_line_obj.search(
+                        cr, uid,
+                        [('analytic_account_id',
+                            '=', ol.account_analytic_id.id),
+                         ('date_from', '<=', today),
+                         ('date_to', '>=', today)],
```
ghost, line 68: I'm not sure "today" is the best date from which to compare the budget lines here. I think the date_order field on PurchaseOrder would more accurate.

jgrandguillaume:
Hi,

I'm coming back here because in the meanwhile, ew developed the purchase_exception in this huge PR https://github.com/OCA/purchase-workflow/pull/113/files

It adds improvements to FA, but also a new purchase exception module.

@lepistone Could you may be extract the purchase_exception from the above PR so we have one per module ?

I'm in favor of using an exception to deal with the budget check.

What do you think @clonedagain ?

Regards,

Joël

clonedagain:
Hi,
I have no strong opinion on the matter.
The deployment of our module failed (users never adopted it, analytic 
accounts scared them) and I'm unsure what we'll do during our migration 
to v8.
By all means do move forward and we'll catch up eventually when we get our acts together.
You may close our PR if you think it does not fit in the OCA project 
anymore.

lepistone:
I am closing this PR. Feel free to ping to reopen, or make a new one. Thanks!

PR #43, opened 2014-10-28
[8.0] add module purchase_origin_address

yvaucher:
In this module, we add a settable origin address on the purchase order that we copy on created picking in.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,12 @@ python:
   - "2.7"
 
 env:
-  - VERSION="8.0" ODOO_REPO="odoo/odoo" INCLUDE="framework_agreement"
-  - VERSION="8.0" ODOO_REPO="OCA/OCB" INCLUDE="framework_agreement"
-  - VERSION="8.0" ODOO_REPO="odoo/odoo" EXCLUDE="framework_agreement"
-  - VERSION="8.0" ODOO_REPO="OCA/OCB" EXCLUDE="framework_agreement"
+    - VERSION="8.0" ODOO_REPO="odoo/odoo" INCLUDE="framework_agreement"
+    - VERSION="8.0" ODOO_REPO="OCA/OCB" INCLUDE="framework_agreement"
+    - VERSION="8.0" ODOO_REPO="odoo/odoo" INCLUDE="purchase_rfq_bid_workflow,purchase_requisition_bid_selection,purchase_requisition_auto_rfq,purchase_requisition_auto_rfq_bid_selection"
+    - VERSION="8.0" ODOO_REPO="OCA/OCB" INCLUDE="purchase_rfq_bid_workflow,purchase_requisition_bid_selection,purchase_requisition_auto_rfq,purchase_requisition_auto_rfq_bid_selection"
+    - VERSION="8.0" ODOO_REPO="odoo/odoo" EXCLUDE="framework_agreement,purchase_rfq_bid_workflow,purchase_requisition_bid_selection,purchase_requisition_auto_rfq,purchase_requisition_auto_rfq_bid_selection"
+    - VERSION="8.0" ODOO_REPO="OCA/OCB" EXCLUDE="framework_agreement,purchase_rfq_bid_workflow,purchase_requisition_bid_selection,purchase_requisition_auto_rfq,purchase_requisition_auto_rfq_bid_selection"
 
 virtualenv:
   system_site_packages: true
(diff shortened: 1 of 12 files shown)


yvaucher:
something wrong in tests. Maybe some incompatibility with other modules -> WIP

yvaucher:
Last tests confirms an hidden incompatibility

.travis.yml
```diff
@@ -8,6 +8,8 @@ env:
   - VERSION="8.0" ODOO_REPO="OCA/OCB" INCLUDE="framework_agreement"
   - VERSION="8.0" ODOO_REPO="odoo/odoo" EXCLUDE="framework_agreement"
   - VERSION="8.0" ODOO_REPO="OCA/OCB" EXCLUDE="framework_agreement"
+  - VERSION="8.0" ODOO_REPO="odoo/odoo" INCLUDE="purchase_origin_address"
```
pedrobaeza: This is not needed. Module is tested on the lines with EXCLUDE option, because it tests all modules except the one in the option. INCLUDE changes behaviour, because it only tests modules put on the argument.

purchase_origin_address/README.rst
```diff
@@ -0,0 +1,11 @@
+Purchase Origin Address
+=======================
+
+This modules adds an origin address on the purchase order that will be
+transfered on picking in.
```
pedrobaeza: s/transfered/transferred

purchase_origin_address/model/purchase_order.py
```diff
@@ -0,0 +1,46 @@
...
+
+    origin_address_id = fields.Many2one(
+        'res.partner',
+        'Origin Address')
+
+    @api.multi
+    def onchange_partner_id(self, partner_id):
+        res = super(PurchaseOrder, self).onchange_partner_id(partner_id)
+        if not partner_id:
+            res['value']['origin_address_id'] = False
```
pedrobaeza: You can put directly:
`res['value']['origin_address_id'] = partner_id`

purchase_origin_address/model/stock_picking.py
```diff
@@ -0,0 +1,29 @@
...
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#
+from openerp import models, fields
+
+
+class StockPicking(models.Model):
+    _inherit = 'stock.picking'
+
+    origin_address_id = fields.Many2one(
```
pedrobaeza, line 27: What if you make this field related and you don't have to assign it? Or do you want to change it later?

.travis.yml
yvaucher: This is what I wanted to do. A test with only purchase_origin_address.

This WIP I wanted to check if on travis I had the same correct result as locally for the tests.

Now I have to figure out which is the guilt module. Unless I lazily adds purchase_origin_address tested in stand alone. But the incompatibility seams to ignore my additions. As partner is not forwarded

purchase_origin_address/README.rst
yvaucher: Thanks, good catch

.travis.yml
pedrobaeza: Sorry, I didn't see WIP tag. I have made comments anyway later on code to help you :wink:

purchase_origin_address/model/stock_picking.py
gurneyalex, line 27: That's the idea yes : a given supplier may have different addresses from which the shipment may come from and so we need a m2o and not a related field.

purchase_origin_address/tests/test_origin_address.py
```diff
@@ -0,0 +1,119 @@
...
+
+        po2.origin_address_id = False
+
+        pol_vals = self.order_line1_vals.copy()
+        pol_vals['order_id'] = po2.id
+        self.env['purchase.order.line'].create(pol_vals)
+
+        po2.signal_workflow('purchase_confirm')
+
+        self.assertEquals(po2.picking_ids.origin_address_id.id,
```
lepistone: assertFalse

yvaucher: Indeed

lepistone:
Hang on, I am improving one of the tests.

lepistone:
I refactored the tests.
 :+1: Thanks!

yvaucher:
@lepistone thanks

gurneyalex:
:+1:

PR #44, opened 2014-10-30
fix bid_tendering_mode in purchase requisition

gurneyalex:
set the field required in the model rather than in the view
add a default value for the field

### DIFF
--- a/purchase_requisition_bid_selection/model/purchase_requisition.py
+++ b/purchase_requisition_bid_selection/model/purchase_requisition.py
@@ -65,6 +65,8 @@ class PurchaseRequisition(models.Model):
         [('open', 'Open'),
          ('restricted', 'Restricted')],
         'Call for Bids Mode',
+        required=True,
+        default='open',
         help="- Restricted : you select yourself the bidders and generate a "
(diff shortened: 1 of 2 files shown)


lepistone:
Thanks! :+1:

purchase_requisition_bid_selection/model/purchase_requisition.py
```diff
@@ -65,6 +65,8 @@ class PurchaseRequisition(models.Model):
         [('open', 'Open'),
          ('restricted', 'Restricted')],
         'Call for Bids Mode',
+        required=True,
+        default = 'open',
```
LeartS: Nitpicking: to space or not to space? http://legacy.python.org/dev/peps/pep-0008/#other-recommendations

PCatinean:
:+1:

purchase_requisition_bid_selection/model/purchase_requisition.py
lepistone: your comment was not lost :smile_cat: but don't worry: travis is there checking those things.

pedrobaeza:
:+1:

PR #47, opened 2014-11-03
add new module purchase_requisition_delivery_address

yvaucher:
This is an extraction of dest_address_id field from module
purchase_requisition_bid_selection

### DIFF
--- a/purchase_requisition_delivery_address/__openerp__.py
+++ b/purchase_requisition_delivery_address/__openerp__.py
@@ -27,7 +27,6 @@
  'complexity': "normal",
  'images': [],
  'depends': ['purchase_requisition',
-             'purchase_delivery_address',
              ],
  'demo': [],
  'data': ['view/purchase_requisition.xml'],
--- a/purchase_requisition_delivery_address/__openerp__.py
+++ b/purchase_requisition_delivery_address/__openerp__.py
@@ -24,7 +24,7 @@
  'author': "Camptocamp",
  'category': "Purchase Management",
  'license': "AGPL-3",
- 'complexity': "normal",
+ 'complexity': "easy",
  'images': [],
  'depends': ['purchase_requisition',
              ],
--- a/purchase_requisition_delivery_address/__openerp__.py
+++ b/purchase_requisition_delivery_address/__openerp__.py
@@ -27,6 +27,7 @@
  'complexity': "easy",
  'images': [],
  'depends': ['purchase_requisition',
+             'stock_dropshipping',
              ],
  'demo': [],
  'data': ['view/purchase_requisition.xml'],
--- a/purchase_requisition_delivery_address/__openerp__.py
+++ b/purchase_requisition_delivery_address/__openerp__.py
@@ -1,0 +1,37 @@
+# -*- coding: utf-8 -*-
+#
+#
+#    Author: Yannick Vaucher
+#    Copyright 2014 Camptocamp SA
+#
+#    This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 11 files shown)


eLBati:
travis says `You try to install module 'purchase_requisition_delivery_address' that depends on module 'purchase_delivery_address'.\nBut the latter module is not available in your system.`

purchase_requisition_delivery_address/__openerp__.py
```diff
@@ -0,0 +1,37 @@
...
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#
+{'name': "Purchase Requisition - Delivery Address",
+ 'summary': "Adds delivery address on Purchase requisition",
+ 'version': "0.1",
+ 'author': "Camptocamp",
+ 'category': "Purchase Management",
+ 'license': "AGPL-3",
+ 'complexity': "normal",
```
lepistone: easy

lepistone: ..but I never understood what this field is for. No problem for me if you leave it to normal.

lepistone:
After discussion on #46, the module purchase_delivery_address will probably disappear, replaced by a fix showing the address fields that are hidden on the `purchase.order` and `stock.move`. I will take care of that soon.

As for this PR, I think we can just remove the dependency on purchase_delivery_address. Thanks!

purchase_requisition_delivery_address/model/purchase_requisition.py
```diff
@@ -0,0 +1,64 @@
...
+
+    @api.model
+    def _prepare_purchase_order(self, requisition, supplier):
+        values = super(PurchaseRequisition, self
+                       )._prepare_purchase_order(requisition, supplier)
+        values['dest_address_id'] = requisition.dest_address_id.id
+        return values
+
+    @api.onchange('dest_address_id')
+    def onchange_dest_address_id(self):
```
lepistone, line 39: I know this onchange is just extracted from another module that could have been ported by me, but I am not sure searching a warehouse with partner_id = the destination address and taking any picking type of that warehouse is a useful choice. We could instead look for any "dropshipping-type" picking type, i.e. a picking type that goes from a supplier location to a customer location. That is done here:

https://github.com/OCA/purchase-workflow/pull/46/files#diff-02d586cf1b8cef18fe6ed0146a6af1d8R25

What do you think about that?

yvaucher:
@eLBati sorry I forgot to mention https://github.com/OCA/purchase-workflow/pull/46

We will break that dependency thought.

purchase_requisition_delivery_address/model/purchase_requisition.py
```diff
@@ -0,0 +1,64 @@
...
+        types = PickType.search([
+            ('warehouse_id.partner_id', '=', self.dest_address_id.id)])
+
+        if types:
+            if self.picking_type_id in types:
+                return
+            picking_type_id = types[0].id
+        else:
+            raise exceptions.Warning(
+                'No picking types were fourd on warehouse. Please verify you '
```
yvaucher, line 55: s/fourd/found

yvaucher:
rebased

lepistone:
thanks :+1:

joaoalf:
The code looks ok to me. On the other hand the users of the module purchase_requisition_bid_selection will silently loose a field unless they install the purchase_requisition_delivery_address module. Maybe this could be better documented.

yvaucher:
@joaoalf I added a warning in modified module

Issue #22, opened 2014-09-26
purchase_landed_costs are red in 7.0


gurneyalex:

pedrobaeza:
I think this is not going to be fixed, so I close.

PR #51, opened 2014-11-05
update travis config

gurneyalex:
use UNIT_TEST=1 rather than maintain a complex include/exclude configuration


lepistone:
:+1:

pedrobaeza:
I see that you are changing all Travis config to this option, but I don't think that it's convenient, because one of the most powerful things we have with this is to see incompatibilities between modules. In this way, we are isolating each one and not seeing possible conflicts. This is a bad practise and will lead to a lot of problems in environments with some of these modules.

gurneyalex:
not all of them, only the ones I'm concerned with.

Then there are 2 kinds on incompatibilities:
- install time incompatibilities (and these are indeed not tested with the new setup), and we generally don't do much about these apart from separating the tests. 
- test time incompatibilities (tests not green for one module when another module is installed) and these are generally painfully worked around in the travis config with include / exclude rules

I find the second issue error prone and hard to maintain and I'd like to simplify the travis configurations.

gurneyalex:
regarding install time incompatibilities, in case we decide that we don't want incompatible modules within a project, then a solution could be to have travis_run_tests to install all modules together after testing each module separately and to exit with an error in case all modules cannot be installed together. But this is an issue to create on OCA/maintainer-quality-tools

bwrsandman:
While you're at this, please remove the .coveragerc file as it is now provided by MQT

yvaucher:
(Need fixing added to remove .coveragerc)

gurneyalex:
tests are red because of #22, don't merge this until this is fixed otherwise the other modules won't be tested at all.

bwrsandman:
:+1:

gurneyalex:
builds with this config take way to much time, and we are finding this difficult to cope with in OCA/vertical-ngo and we are reverting the change. I'm withdrawing this PR for now.

PR #53, opened 2014-11-05
fix get_cheapest_agreement_for_qty sorting

### LINKED ISSUE #52: framework agreement - get_cheapest_agreement_for_qty fails on sort
https://github.com/OCA/purchase-workflow/blob/8.0/framework_agreement/model/framework_agreement.py#L427

This return a record set. `sort()` won't work on it.

This raises AttributeError: 'framework.agreement' object has no attribute 'sort' when playing with some onchanges from `vertical-ngo/framework_agreement_sourcing`

yvaucher:
fixes #52

### DIFF
--- a/framework_agreement/model/framework_agreement.py
+++ b/framework_agreement/model/framework_agreement.py
@@ -401,10 +401,7 @@ class framework_agreement(models.Model):
                        ('draft', '=', False)]
         if qty:
             search_args.append(('available_quantity', '>=', qty))
-        agreement_ids = self.search(search_args)
-        if agreement_ids:
-            return agreement_ids
-        return None
+        return self.search(search_args)
 
     @api.model
     def get_cheapest_agreement_for_qty(self, product_id, date, qty,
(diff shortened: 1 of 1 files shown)


framework_agreement/model/framework_agreement.py
```diff
@@ -401,9 +401,9 @@ def get_all_product_agreements(self, product_id, lookup_dt, qty=None):
                        ('draft', '=', False)]
         if qty:
             search_args.append(('available_quantity', '>=', qty))
-        agreement_ids = self.search(search_args)
-        if agreement_ids:
-            return agreement_ids
+        agreements = self.search(search_args)
```
pedrobaeza: Why not returning directly the search result (it will be [] if none)?

yvaucher: it will be an empty record set instead of None

I didn't wanted to change the output but it might be better to work with record set

yvaucher:
@pedrobaeza fixed

pedrobaeza:
Thanks :+1:

lepistone:
:+1:

PR #65, opened 2014-11-25
Add module procurement_batch_generator

alexis-via:
During a demo for a potential customer, after showing all the great features of procurements/re-ordering rules/etc..., I had this simple question : is it possible to "simply" create POs from the list of products, by just looking at the stock levels ? I had to admit that this simple feature was not available (at least to my knowledge)... so I developed the module purchase_from_product_tree.

Extract from the module description:
This module adds a wizard on the product variants tree view to create new purchase orders. Some companies don't want to use the re-ordering rules to create purchase orders for some reasons (complexity, fast changes in demand, etc...). Some companies may prefer to look at their stock level and, from this analysis, decide which product variants they need to re-order. This module implements this scenario.


pedrobaeza:
Hi, Alexis, although the functionality is not exactly the same, you can request a procurement for each product variant from inventory tab, which ends (after running the scheduler) in a purchase order if the product has the route 'Buy' selected. The advantage of this approach is that is more general, because you are requesting a procurement to one warehouse, which can be satisfied replinishing from other warehouse, making a purchase or manufacturing.

Your approach have the advantage of being able to select more than one product at a time. Do you think it worths still?

jgrandguillaume:
Hi,

Thanks for the contribs. I must say I'm also joining Perdo's opinion here. I'd rather prefer the user to use the standard feature than adding a new module for that purpose.

Going through the procurement process is the proper way to deal with that I think, so I'm :-1: 

Regards,

Joël

alexis-via:
@pedrobaeza Thanks for pointing out this feature of requesting a procurement via the inventory tab in the product form. The goal of my module is to be able to handle several products at a time, instead of going on the form view of each product. Maybe we could convert this module to a module "procurement_from_product_tree" ; that way, it would generate procurements, that would then generate POs or MO depending on the configuration of the product. Do you think such a module would be interesting ?

pedrobaeza:
Yeah, you can present a wizard to select quantity for each selected product, defaulting to 1, and make this operation in batch mode. I see that interesting. It could be named `procurement_batch_generator` or something similar, but your name is also good.

alexis-via:
@pedrobaeza @jgrandguillaume 
I took into accout all your remarks and added a YAML test. I hope you will like this new version.
Maybe it would be better to propose it in another OCA project: stock-logistics-workflow ?

procurement_batch_generator/i18n/fr.po
```diff
@@ -0,0 +1,134 @@
...
+#. module: procurement_batch_generator
+#: view:procurement.batch.generator:procurement_batch_generator.procurement_batch_generator_form
+msgid "Cancel"
+msgstr "Annuler"
+
+#. module: procurement_batch_generator
+#: field:procurement.batch.generator,create_uid:0
+#: field:procurement.batch.generator.line,create_uid:0
+msgid "Created by"
+msgstr "Created by"
```
pedrobaeza: Why don't you translate this?

pedrobaeza:
A little remark, but the rest :+1: 

As this is procurement, it's the same bad place purchase nor stock. Maybe a new repo, @jgrandguillaume ?

procurement_batch_generator/wizard/procurement_batch_generator.py
```diff
@@ -0,0 +1,125 @@
...
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+from openerp import models, fields, api, _
+import openerp.addons.decimal_precision as dp
+from openerp.exceptions import Warning
+
+
+class ProcurementBathGenerator(models.TransientModel):
```
bealdav: Oops, ProcurementBathGenerator or ProcurementBatchGenerator

jgrandguillaume:
Hi,

Thanks Alexis for making the changes, this is a clean implementation now !

Regarding the repo, I think this one is fine IMO as it concerns the purchase process it make sens to have it here even if technically speaking it based on procurement object not purchase.

So I'm :+1: 

Regards,

Joël

PR #71, opened 2014-12-05
[7.0] Add module 'account_creditor_price difference'

JordiBForgeFlow:
When the supplier invoice is created from a purchase order, if the
destination of the product is a company location the application
will propose as debiting account the stock input account from the
product or product category, instead of the product, in order
to balance the credit that occurred when the product was received.

The stock input account is frequently called the Goods Received Not Invoiced
account (GRIN).

Furthermore, when the invoice is accepted, the application will identify
any differences between the invoice price and the product cost, and will
post the differences to a new Price Differences account defined in the
product or product category.

See the following blog post:

http://www.eficent.com/openerp_en/product-receptions-and-returns-with-real-time-inventory-valuation-in-odoo/

### DIFF
--- a/account_creditor_price_difference/__init__.py
+++ b/account_creditor_price_difference/__init__.py
@@ -1,0 +1,25 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    Copyright (C) 2014 Eficent (<http://www.eficent.com/>)
+#              Jordi Ballester Alomar <jordi.ballester@eficent.com>
+#
+#    This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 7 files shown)


lepistone:
Hi @jbeficent, thanks for the PR!

before going further, could you please:
- check if CLAbot is right and the you didn't sign a CLA
- add an automated test or two
- check why the travis build is red.

yvaucher:
Travis is red because there is some PEP8 to fix

pedrobaeza:
I close because there is no source branch now. @jbeficent, if you want again to make the PR...

PR #74, opened 2014-12-17
Added procurement_working_days module

npiganeau:
Hi, 

The aim of this module is to schedule lead times in working days rather than calendar days as it now in odoo.
It is a reply to https://github.com/odoo/odoo/issues/3452 
It acts on the following:
- Scheduling according to a "move" type procurement.rule: the delay is counted only on working days
- Scheduling of RFQ: po_lead and supplierinfo delay are counted only on working days
- Scheduling of MO: manufacturing_lead is counted only on working days

I pushed it here because I could not find a more appropriate OCA repository, although this is not only about purchase.


pedrobaeza:
Why do you close the PR?

npiganeau:
After double thinking, I think it is not quite finalized yet:
- It needs https://github.com/odoo/odoo/pull/4030 bugfix on the resource.calendar model to be pulled upstream to work correctly
- The module deals with scheduling of purchase orders, stock moves and production orders altogether.  It should probably be split in 3: one module for purchase orders, one module for mrp both depending on one module for stock moves
- To have the complete set, I should add a module to deal with working day scheduling of sale deliveries, also depending on the module for stock moves.

So I think I will work a bit more on it this way, before making a new PR. 

But then, my question will be to which OCA repository to push to, since this will be about stock, purchase, sale and mrp and there is no such oca repository for scheduling or procurement in general.

Now, if you think I should leave the PR open in the meantime, I could do that too.

PR #89, opened 2015-02-04
[ADD] [7.0] Add module 'purchase_invoice_control_check'.

JordiBForgeFlow:
Currently a user can create a PO containing services and set the invoice control 'Based on incoming shipments', and the application does not prevent the user from doing that.

Then, the user has no option in the 'Invoice Control' menu to invoice that line, and can only invoice based on a list of PO lines.

This modules introduces the following features:

Prevents a user from being able to confirm a Purchase Order with invoice control 'Based on incoming shipments' that contains services.

Makes it possible to cancel a Purchase Order in which the user created the PO containing services, with invoice control 'Based on incoming shipments' (in order to correct the PO's that were created incorrectly before this module was used).


pedrobaeza:
I would call it _purchase_service_, because the rule is to put singular forms, and there are several features, not only the cancel one. Also, a better place for this is the project-service repository I think.

JordiBForgeFlow:
@pedrobaeza I renamed to 'purchase_services_check' in a later commit, but agree to change to 'purchase_service_check' or 'purchase_service'. 

My understanding is that 'project-service' repo is to be used for service or project oriented companies. 
I'm proposing to add it to 'purchase-workflow' because potentially any kind of business eventually purchases services of any kind.
@dreispt perhaps you can let us know your opinion.

dreispt:
It is not desirable for the `project-service` to depend on purchase/stock modules. IMO this is the right place for this feature.

pedrobaeza:
OK, @dreispt and @jbeficent. We continue here with the review with the new name.

JordiBForgeFlow:
Renamed to 'purchase_invoice_control_service', using singular forms. I think that the new name describes more precisely what it does.

pedrobaeza:
If you rename it that way, I prefer then _purchase_invoice_control_check_, because it also includes the check for non existing product_id

JordiBForgeFlow:
Done. I also reworded the module description.

purchase_invoice_control_check/purchase.py
```diff
@@ -0,0 +1,74 @@
...
+                    raise orm.except_orm(
+                        _('Unable to cancel this purchase order.'),
+                        _('Invoices have already been created.'))
+        return super(purchase_order, self).action_cancel(cr, uid, ids,
+                                                         context=context)
+
+    def wkf_confirm_order(self, cr, uid, ids, context=None):
+        for order in self.browse(cr, uid, ids, context=context):
+            if order.invoice_method == 'picking':
+                for order_line in order.order_line:
```
pedrobaeza: Why not `if order.has_services()`?

purchase_invoice_control_check/purchase.py
```diff
@@ -0,0 +1,74 @@
...
+from openerp.osv import orm
+from openerp.tools.translate import _
+
+
+class purchase_order(orm.Model):
+
+    _inherit = 'purchase.order'
+
+    def has_services(self, cr, uid, ids, *args):
+        res = False
```
pedrobaeza: You can put this as a computed field

JordiBForgeFlow:
@odony and @jco-odoo are proposing to submit a PR to the Odoo core. Would you agree @pedrobaeza? Then maybe there's no need for this PR.

pedrobaeza:
Yeah, this is indeed a patch for something that must be on the core.

PR #90, opened 2015-02-04
8.0 fa fixes

gurneyalex:
one big bug fix: wrong keyword `copyable` instead of `copy` in field declaration

some docstring / help string / module description improvements.

### DIFF
--- a/framework_agreement/model/company.py
+++ b/framework_agreement/model/company.py
@@ -29,6 +29,6 @@ class res_Company(models.Model):
     one_agreement_per_product = fields.Boolean(
         'One agreement per product',
         help='If checked you can have only'
-        ' one framework agreement '
+        ' one framework agreement'
         ' per product at the same time.'
     )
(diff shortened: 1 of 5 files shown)


nbessi:
:+1:

framework_agreement/model/company.py
```diff
@@ -30,5 +30,5 @@ class res_Company(models.Model):
         'One agreement per product',
         help='If checked you can have only'
         ' one framework agreement '
-        ' per product at the same time'
+        ' per product at the same time.'
```
pedrobaeza, line 33: You need to change the translation template then.

nbessi, line 33: @pedrobaeza good catch I always forgot to do it. It would be nice to have some automated tools/check to enforce this.

pedrobaeza, line 33: Yeah, but I think it's difficult to detect that.

gurneyalex:
@pedrobaeza fixed, thanks

pedrobaeza:
Although there is still a double space, it's not so important.

:+1:

yvaucher:
Extra space removed :+1:

PR #102, opened 2015-02-25
[ADD] module 'purchase_delivery'

JordiBForgeFlow:
This module makes it possible to add delivery method to purchase orders and
can compute a shipping line in the purchase order or in the invoice when
created from an incoming shipment.

The application makes use of the same concepts of carrier and delivery grid
extending them as follows:
    \* Introduces origin countries, states and ZIP in the delivery grid.
    \* Uses the cost defined in the grid, instead of the sale price.
    \* Uses the vendor address to match with the existing grid based on the
    origin information defined in the grid.
    \* Uses the destination address (customer address for direct shipments
    and warehouse otherwise) to determine the grid, based on the destination
    parameters defined in the grid.

In case that the incoming shipment contains moves that have multiple
destination addresses, computes the average freight cost.

### DIFF
--- a/purchase_delivery/__init__.py
+++ b/purchase_delivery/__init__.py
@@ -1,0 +1,22 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    Copyright (C) 2015 Eficent (<http://www.eficent.com/>)
+#               <contact@eficent.com>
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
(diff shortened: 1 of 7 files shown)


lepistone:
Hi, some lint problems in the build.

gurneyalex:
@jbeficent lots of flake8 issues. Please fix them. 

https://travis-ci.org/OCA/purchase-workflow/jobs/52070415

JordiBForgeFlow:
Not relevant anymore.

pedrobaeza:
Well, the idea is still good. As you know, I have a bit different implementation, but with the same idea.

PR #107, opened 2015-03-03
[7.0] Add purchase_supplier_postage_free

guewen:
- Add a field "Postage free above amount" on the suppliers, which is an
  optional amount above which the supplier does not charge the shipping
  fees.
- Display this amount on the purchase orders as an indication
- Add a filter on the purchase orders showing the ones which have reached
  the postage free amount.


purchase_supplier_postage_free/__openerp__.py
```diff
@@ -0,0 +1,50 @@
...
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{'name': 'Purchase Supplier Postage Free',
+ 'version': '1.0',
+ 'author': 'Camptocamp',
```
gurneyalex: please add "Odoo Community Association (OCA)"

lepistone:
I'm restarting the build

lepistone:
I deleted a commit that added a dependency on `OCA/web`, that should not be needed. Let's see how the build goes. Otherwise :+1:

guewen:
I am about to port this module to 8.0

guewen:
> I deleted a commit that added a dependency on OCA/web, that should not be needed. Let's see how the build goes. Otherwise :+1:

The dependency on `web` is needed for `web_context_tunnel`.

pedrobaeza:
I think this is better treated using delivery method in purchases, that already includes free postage. An approximation has been already done here for v8: https://github.com/OCA/purchase-workflow/pull/157

gurneyalex:
runbot should be green once #163 is merged and this PR is rebased.

jgrandguillaume:
I'm :+1:  for the reasons given here: https://github.com/OCA/purchase-workflow/pull/159#issuecomment-131098621

guewen:
Rebased

purchase_supplier_postage_free/__openerp__.py
```diff
@@ -0,0 +1,50 @@
...
+ 'version': '1.0',
+ 'author': 'Camptocamp,Odoo Community Association (OCA)',
+ 'maintainer': 'Camptocamp',
+ 'license': 'AGPL-3',
+ 'category': 'Purchase Management',
+ 'depends': ['purchase',
+             'web_context_tunnel',
+             ],
+ 'description': """
+Purchase Supplier Postage Free
```
pedrobaeza, line 32: Can you please follow full OCA's description template?

pedrobaeza:
@jgrandguillaume, @guewen. OK, I think we can have both modules co-existing for the reasons you gave.

purchase_supplier_postage_free/__openerp__.py
guewen, line 32: Done

pedrobaeza:
I mention in 8.0 PR about renaming the module. What do you think?

pedrobaeza:
As this is very old and there's another PR for version 8, I'm going to close this PR, but if you still think this is useful, please reopen, but fixing Travis status.

PR #114, opened 2015-03-25
fix purchase cancelation

gurneyalex:
we need to call super().action_cancel  which will cancel the pickings and invoices if needed
and then trigger the workflow.


gurneyalex:
@jgrandguillaume your feedback on this is welcome.

lepistone:
:+1: . I restarted the build that was red because of github downtime.

guewen:
:+1:

PR #120, opened 2015-04-07
Fix tender workflow from selected to closed

yvaucher:
Action modal was calling a single write instead of triggering
signal_workflow


lepistone:
Absolutely. Thanks :+1:

PR #121, opened 2015-04-08
8.0 fix header cba report

yvaucher:
- Add few spacing as with no spacing, table content will be overlapped by header
- Add table header in thead so it is displayed on other pages


pedrobaeza:
:+1:

jgrandguillaume:
:+1:

PR #127, opened 2015-04-16
Create an attachement for CBA report on states Bids Selection and Bids S...

yvaucher:
...elected

### DIFF
--- a/purchase_requisition_bid_selection/report.xml
+++ b/purchase_requisition_bid_selection/report.xml
@@ -7,6 +7,7 @@
       id="action_report_purchase_requisitions"
       model="purchase.requisition"
       report_type="qweb-pdf"
+      attachment="(object.state in ('open', 'selected')) and ('CBA-' + (object.name or '') + '-' + time.strftime('%Y%m%d%H%M') + '.pdf')"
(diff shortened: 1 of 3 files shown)


gurneyalex:
:+1:

lepistone:
Thanks :+1: . It's ok like this, but maybe you could do a domain instead of computed fields, like this:

https://github.com/odoo/odoo/blob/8.0/addons/account_voucher/account_voucher.py#L311-314

yvaucher:
@lepistone Good suggestion, I'll change it with domains

purchase_requisition_bid_selection/report.xml
```diff
@@ -7,6 +7,7 @@
       id="action_report_purchase_requisitions"
       model="purchase.requisition"
       report_type="qweb-pdf"
+      attachment="(object.state in ('open', 'selected') and len(object.eligible_bid_ids)) and ('CBA-' + (object.name or '') + '-' + time.strftime('%Y%m%d%h%M') + '.pdf')"
```
yvaucher: Hours should be %H not %h

yvaucher:
@lepistone I Improved the way fields for reports where computed getting ride of those functions.

PR #130, opened 2015-04-29
[8.0]  Somes fixes on  purchase partial invoicing

adrienpeiffer:
-  Use context instead of write on purchase order line to invoice partial quantity.
-  Domain with fully_invoiced doesn't work if this field isn't store.
-  Reset invoiced flag to fully_invoiced before invoice generation. Without this fix, it is not possible to invoice an already invoiced purchase order line whose invoice's line was deleted because the invoiced flag remains true on the purchase order line.
- Purchase order should be not invoiced if all lines aren't fully invoiced

EDIT:
-  Decimal precision on invoiced_qty must be the same as product_qty on purchase order line

### DIFF
--- a/purchase_partial_invoicing/purchase.py
+++ b/purchase_partial_invoicing/purchase.py
@@ -109,9 +109,11 @@ class purchase_order(orm.Model):
     }
 
     def _prepare_inv_line(self, cr, uid, account_id, order_line, context=None):
+        if context is None:
+            context = {}
         res = super(purchase_order, self).\
             _prepare_inv_line(cr, uid, account_id, order_line, context=context)
-        if context is not None and context.get('partial_quantity_lines'):
+        if context.get('partial_quantity_lines'):
             partial_quantity_lines = context.get('partial_quantity_lines')
             if partial_quantity_lines.get(order_line.id):
                 res.update({'quantity':
--- a/purchase_partial_invoicing/purchase.py
+++ b/purchase_partial_invoicing/purchase.py
@@ -111,10 +111,12 @@ class purchase_order(orm.Model):
     def _prepare_inv_line(self, cr, uid, account_id, order_line, context=None):
         res = super(purchase_order, self).\
             _prepare_inv_line(cr, uid, account_id, order_line, context=context)
-        if context is not None and context.get('partial_quantity', False):
-            partial_quantity = context.get('partial_quantity')
-            if partial_quantity.get(order_line.id, False):
-                res.update({'quantity': partial_quantity.get(order_line.id)})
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Why not convert to new API and you gain the depends possibilities?

adrienpeiffer:
Hi Pedro,
This is for a possible more simple backport

purchase_partial_invoicing/purchase.py
```diff
@@ -51,17 +51,37 @@ def _all_invoices_approved(self, cursor, user, ids, name, arg,
                 res[line.id] = False
         return res
 
+    def _order_lines_from_invoice_line(self, cr, uid, ids, context=None):
+        result = {}
```
pedrobaeza: Use a set instead (although standard makes this way)

purchase_partial_invoicing/wizard/po_line_invoice.py
```diff
@@ -68,18 +71,23 @@ def makeInvoices(self, cr, uid, ids, context=None):
...
-            line.po_line_id.write({
-                'product_qty': line.invoiced_qty,
-            })
+            ] = line.invoiced_qty
+            if line.po_line_id.fully_invoiced:
+                invoiced_ids.append(line.po_line_id.id)
+            else:
+                not_invoiced_ids.append(line.po_line_id.id)
+        if len(not_invoiced_ids) > 0:
+            purchase_line_obj.write(cr, uid, not_invoiced_ids,
```
pedrobaeza, line 80: You can make the write directly, although it is []

purchase_partial_invoicing/purchase.py
adrienpeiffer: Thanks ! Can you give me a pointer to the standard ?

adrienpeiffer: After reflection, I think I misunderstood. Do you speak about the set ?

pedrobaeza: I mean something like this:

```
result = set()
for invoice in self.pool['account.invoice'].browse(cr, uid, ids,
                                                           context=context):
            for line in invoice.invoice_line:
                result.add(line.purchase_line_id.id)
return list(result)
```

adrienpeiffer: Thanks !

adrienpeiffer:
@pedrobaeza @sbidoul  Thanks for the review !

pedrobaeza:
:+1:

gurneyalex:
:+1:

sbidoul:
:+1:

PR #131, opened 2015-04-30
purchase_discount: Removes discount from supplier invoices form view

omar7r:
[FIX] purchase_discount: Removes discount from supplier invoices form view, because by default it exists


pedrobaeza:
OK, I see, it is activated when you mark "Use discounts on sales lines". A curious behaviour...

:+1:

guewen:
:+1:

yvaucher:
:+1:

PR #134, opened 2015-05-20
VCI - Allow to create Purchase invoice with invoice on incoming shipm…

yvaucher:
…ent policy

If `Configuration / Purchases`, "Default invoicing control method is
set to "Based on incoming shipments", obviously it was impossible to generate a
supplier invoice for a Purchase created by a Vendor Consignement Inventory sale.
Simply because there is no incoming shipment.
Here we change purchase policy on creation to ensure the invoice is created once
purchase is confirmed.


gurneyalex:
:+1:

jgrandguillaume:
:+1:

PR #136, opened 2015-05-24
8.0 - [ADD] product_supplierinfo_discount: Discounts in product supplier info

pedrobaeza:
# Discounts in product supplier info

This module allows to put a discount in the supplier info form, and propagate
it to purchase orders, having two advantages over pricelists:
- The discount is directly put on purchase orders instead of reducing the
  unit price.
- You can set prices and discounts on the same screen.
# Installation

This module requires _purchase_discount_ module, that it's also available in
this repository.
# Configuration

To see prices and discounts on supplier info view, you have to enable the
option "Manage pricelist per supplier" inside _Configuration > Purchases_
# Usage

Go to Purchase > Products, open one product, and edit or add a record on the
_Suppliers_ section of the _Procurements_. You will see in the prices section
in the down part a new column called _Discount (%)_. You can enter here
the desired discount for that quantity.

When you make a purchase order for that supplier and that product, discount
will be put automatically.
# Known issues / Roadmap
- The discount is always applied, independently if you have based
  your pricelist on other value than "Supplier Prices on the product form".


pedrobaeza:
Check the 7.0 version in #137

lepistone:
Would it be possible to start from your 7.0 branch?

pedrobaeza:
Actually I did it the contrary: make the 8.0 version, and then backport it to v7.

lepistone:
I was thinking if it would be best to make it so that the v7 branch can be easily merged into v8 ?

pedrobaeza:
But branches has differed too much apart from this module, isn't it? I can make a cherry-pick of the 7.0 version, and put over it the 8.0 version, but I'm not sure if this is going to make a difference.

lepistone:
ok then.

IMO the 7.0 -> 8.0 merge thing is cool if it is easy to do, otherwise let's forget about it.

jgrandguillaume:
Hi,

Thanks for the contribs ! :+1: even if I would have loved a little test.

Regards,

Joël

product_supplierinfo_discount/models/purchase_order.py
```diff
@@ -0,0 +1,66 @@
...
+        product_obj = self.env['product.product']
+        product_uom_obj = self.env['product.uom']
+        pl_pinfo_obj = self.env['pricelist.partnerinfo']
+        # Look for a possible discount
+        product = product_obj.browse(product_id)
+        from_uom = self.env.context.get('uom') or product.uom_id.id
+        qty_in_product_uom = qty
+        sinfos = supplierinfo_obj.search(
+            [('product_tmpl_id', '=', product.product_tmpl_id.id),
+             ('name', 'child_of', partner_id)])
```
StefanRijnhart: Same comment as on 7.0: either take child_of of the commercial_partner_id or compare to this specific partner only.

pedrobaeza:
@StefanRijnhart done

@jgrandguillaume, I have added tests

product_supplierinfo_discount/models/product_supplierinfo.py
```diff
@@ -0,0 +1,28 @@
...
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+from openerp import models, fields
+
+
+class PricelistPartnerInfo(models.Model):
+    _inherit = "pricelist.partnerinfo"
+
+    discount = fields.Float(string='Discount (%)', digits=(16, 2))
```
StefanRijnhart: Just noticed the hardcoded digits. Would make sense to use dp.get_precision('Discount'). This particular precision is added in the account module, so it should be available.

pedrobaeza:
@StefanRijnhart, thanks for noticing. Updated!

StefanRijnhart:
Thanks for the tests and the other updates! :+1:

PR #137, opened 2015-05-24
7.0 - [ADD] product_supplierinfo_discount: Discounts in product supplier info

pedrobaeza:
# Discounts in product supplier info

This module allows to put a discount in the supplier info form, and propagate
it to purchase orders, having two advantages over pricelists:
- The discount is directly put on purchase orders instead of reducing the
  unit price.
- You can set prices and discounts on the same screen.
# Installation

This module requires _purchase_discount_ module, that it's also available in
this repository.
# Configuration

To see prices and discounts on supplier info view, you have to enable the
option "Manage pricelist per supplier" inside _Configuration > Purchases_
# Usage

Go to Purchase > Products, open one product, and edit or add a record on the
_Suppliers_ section of the _Procurements_. You will see in the prices section
in the down part a new column called _Discount (%)_. You can enter here
the desired discount for that quantity.

When you make a purchase order for that supplier and that product, discount
will be put automatically.
# Known issues / Roadmap
- The discount is always applied, independently if you have based
  your pricelist on other value than "Supplier Prices on the product form".


pedrobaeza:
Check the corresponding v8 PR on #136

lepistone:
Here the discount is applied on an `onchange` of the order. Would it be better to do that in the pricelist object instead?

Could you add a test?

pedrobaeza:
No, it can't be, because the `_price_get` method only returns one value - the price. That's why product_visible_discount undo the operations to get the equivalent discount.

pedrobaeza:
Take also into account that this works independently from the pricelist: if you have a supplierinfo record with a discount, you get it here. It's something I mention on the known issues, but for the most of the cases, it's going to work.

lepistone:
I see!

product_supplierinfo_discount/models/purchase_order.py
```diff
@@ -0,0 +1,68 @@
...
+        product_obj = self.pool['product.product']
+        product_uom_obj = self.pool['product.uom']
+        pl_pinfo_obj = self.pool['pricelist.partnerinfo']
+        # Look for a possible discount
+        product = product_obj.browse(cr, uid, product_id, context=context)
+        from_uom = context.get('uom') or product.uom_id.id
+        qty_in_product_uom = qty
+        sinfo_ids = supplierinfo_obj.search(
+            cr, uid, [('product_id', '=', product.product_tmpl_id.id),
+                      ('name', 'child_of', partner_id)], context=context)
```
StefanRijnhart: Why the child_of of the actual partner? Intuitively, I would take the child_of of this partner's commercial_partner_id, but in line with standard Odoo I would 'abstract' from this particularity in the partner model and compare this partner only.

StefanRijnhart:
One small nit, but otherwise :+1:

product_supplierinfo_discount/models/purchase_order.py
pedrobaeza: I agree that it should be the childs of commercial_partner_id, but we can't abstract it from the partner model because you are only adding one supplierinfo for each company, not one per contact, so you need to be consistent if you make a purchase order for any of the contacts of the company. I'm changing the search right now.

pedrobaeza:
@StefanRijnhart, done!

StefanRijnhart:
Thanks! :+1:

pedrobaeza:
Another thumb up or may I merge as this one is old (in version and in time from publishing)?

StefanRijnhart:
If I count Joël's thumb-up on the 8.0 version, which is very similar, I guess we can merge this now.

pedrobaeza:
Thanks! Changing 8.0 version right now...

PR #141, opened 2015-06-09
In CBA report, show only as selected the confirmed lines

yvaucher:
We want to show what is really selected, relying only on bid_quantity is wrong as only confirmed line will be used for purchase.

### DIFF
--- a/purchase_requisition_bid_selection/view/report_purchaserequisition.xml
+++ b/purchase_requisition_bid_selection/view/report_purchaserequisition.xml
@@ -5,7 +5,7 @@
       <t t-call="report.internal_layout">
         <div class="page">
           <t t-set="num_of_suppliers" t-value="len(set(bid.partner_id for bid in o.purchase_ids))"/>
-          <t t-set="tendering_output" t-value="sum(l.quantity_bid * l.order_id.currency_id.compute(l.price_unit, l.order_id.company_id.currency_id, round=False) for bid in o.eligible_bid_ids for l in bid.order_line)"/>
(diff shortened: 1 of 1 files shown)


lepistone:
:+1:

purchase_requisition_bid_selection/view/report_purchaserequisition.xml
```diff
@@ -5,7 +5,7 @@
       <t t-call="report.internal_layout">
         <div class="page">
           <t t-set="num_of_suppliers" t-value="len(set(bid.partner_id for bid in o.purchase_ids))"/>
-          <t t-set="tendering_output" t-value="sum(l.quantity_bid * l.order_id.currency_id.compute(l.price_unit, l.order_id.company_id.currency_id, round=False) for bid in o.eligible_bid_ids for l in bid.order_line)"/>
+          <t t-set="tendering_output" t-value="sum(l.quantity_bid * l.order_id.currency_id.compute(l.price_unit, l.order_id.company_id.currency_id, round=False) for bid in o.eligible_bid_ids for l in bid.order_line if l.state == 'confirmed')"/>
```
gurneyalex, line 8: `l` for a 1-letter variable name is not the best choise, readability-wise.

I'm wondering if this `sum()` expression should not be rewritten as a single model method.

gurneyalex:
:+1: even if I'm not too fond of large Python expressions embedded in XML...

purchase_requisition_bid_selection/view/report_purchaserequisition.xml
yvaucher, line 8: I think this is a bit out of scope of this PR.
However, I agree we could re-factor some of these lines to have some field function doing the job.

PR #150, opened 2015-06-24
7.0 purchase reset invoice method

StefanRijnhart:
Port of https://code.launchpad.net/~therp-nl/purchase-wkfl/7.0-add_purchase_reset_invoice_method/+merge/209295

### DIFF
--- a/purchase_reset_invoice_method/model/purchase_reset_invoice_method.py
+++ b/purchase_reset_invoice_method/model/purchase_reset_invoice_method.py
@@ -1,0 +1,40 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    OpenERP, Open Source Management Solution
+#    This module Copyright (C) 2014 Therp BV (<http://therp.nl>).
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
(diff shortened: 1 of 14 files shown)


purchase_reset_invoice_method/model/purchase_reset_invoice_method.py
```diff
@@ -0,0 +1,40 @@
...
+        'invoice_method': fields.selection(
+            [('manual', 'Based on Purchase Order lines'),
+             ('order', 'Based on generated draft invoice'),
+             ('picking', 'Based on incoming shipments')],
+            'Invoicing Control'),
+        }
+
+    def do_reset(self, cr, uid, ids, context=None):
+        wizard = self.browse(cr, uid, ids[0], context=context)
+        return wizard.order_id.reset_invoice_method(
```
pedrobaeza, line 39: As a side note, this way of calling the method loses the context on v8 old API compatibility layer, so this will difficult the migration in case of using the context

purchase_reset_invoice_method/view/purchase_reset_invoice_method.xml
```diff
@@ -0,0 +1,28 @@
...
+                </footer>
+            </form>
+        </field>
+    </record>
+
+</data>
+</openerp>
+
+
+
```
pedrobaeza: Remove these ugly spaces/lines

pedrobaeza:
Maybe we can try to figure out the stock_move of the invoice lines in a migration script that runs on module installation?

purchase_reset_invoice_method/view/purchase_reset_invoice_method.xml
StefanRijnhart: Done.

StefanRijnhart:
Any hints in that direction? I have given it plenty of thought at the time but I could not figure that one out.

pedrobaeza:
We can use the source document to try to match the purchase order or the picking.

StefanRijnhart:
Does that not break when you combine two orders with the same product in a single invoice?

pedrobaeza:
Yeah, you're right, but it doesn't break, only it will not find any match. We can refine the script splitting by `;` and looking for each, matching lines with product. If we don't want to complicate it too much, the first version at least will match a great percentage of invoices, which is a improvement over nothing.

pedrobaeza:
Are you still interested in this, Stefan?

HviorForgeFlow:
> Are you still interested in this, Stefan?

@StefanRijnhart

PR #149, opened 2015-06-23
[ADD] test for missing unrevisioned_name

eLBati:

Issue #148, opened 2015-06-23
purchase_order_revision...issue question


jaronemo:
cancel purchase order.Number: PO1505050001
click new resvision button...show error..

![image](https://cloud.githubusercontent.com/assets/8476734/8302276/ce79f8c4-19c7-11e5-87d7-9cb11e92196f.png)

and odoo server log ...

![image](https://cloud.githubusercontent.com/assets/8476734/8302293/ecfe9732-19c7-11e5-9fd3-ef57d1b58ab7.png)

pedrobaeza:
This is due to there's no post-install script to assign `unrevisioned_name` field on existing purchase orders. @gurneyalex, @eLBati, can you check it?

eLBati:
@jaronemo I can't generate the error on a new DB. Could you provide the steps to reproduce it?

@gurneyalex however, I found a problem with `unrevisioned_name` as pointed out by https://github.com/OCA/purchase-workflow/pull/149: `unrevisioned_name` is always False. How is it supposed to be set?

jaronemo:
Dear sir

![image](https://cloud.githubusercontent.com/assets/8476734/8321814/4d0d7d3c-1a5e-11e5-82ad-6edfcd658554.png)

![image](https://cloud.githubusercontent.com/assets/8476734/8321817/6b4f23cc-1a5e-11e5-8667-3c5dacd23968.png)

UPDATE "purchase_order" SET "name"='False-01',"revision_number"=1 ... ==> Think this is the problem from any orders.

runing test yaml.....

![image](https://cloud.githubusercontent.com/assets/8476734/8321987/f74414a8-1a60-11e5-9b34-e89033bc1ed1.png)

gurneyalex:
@jaronemo I thing #151 may fix your issue (but you will need to manually update you database fix the `False` unrevisionned_name columns. 

Do you confirm?

jaronemo:
Dear Sir
1. uninstall purchase_order_revision module
2.  openerp-server -c /etc/odoo-server.conf -u all --stop-after-init -d DB
3. install purchase_order_revision
module is this:
![image](https://cloud.githubusercontent.com/assets/8476734/8345322/eed7f66c-1b1f-11e5-95b0-bb051d5e3626.png)

![image](https://cloud.githubusercontent.com/assets/8476734/8345278/3ebcb268-1b1f-11e5-98a7-715d60b1fd3b.png)

[...]

PR #151, opened 2015-06-24
purchase_order_revision: provide a value for unrevisioned_name

gurneyalex:
superseeds #149

### DIFF
--- a/purchase_order_revision/__init__.py
+++ b/purchase_order_revision/__init__.py
@@ -1,3 +1,3 @@
-# - -*- coding: utf-8 -*-
+# -*- coding: utf-8 -*-
 
 from . import model
--- a/purchase_order_revision/model/purchase_order.py
+++ b/purchase_order_revision/model/purchase_order.py
@@ -72,3 +72,12 @@ class purchase_order(models.Model):
         self.message_post(body=msg)
         old_revision.message_post(body=msg)
         return True
+
+    @api.model
+    def create(self, values):
+        if 'unrevisioned_name' not in values:
+            if values.get('name', '/') == '/':
+                seq = self.env['ir.sequence']
+                values['name'] = seq.next_by_code('purchase.order') or '/'
+            values['unrevisioned_name'] = values['name']
+        return super(purchase_order, self).create(values)
--- a/purchase_order_revision/test/purchase_order.yml
+++ b/purchase_order_revision/test/purchase_order.yml
@@ -15,16 +15,15 @@
         price_unit: 3297.20
         product_qty: 4.0
 -
-  I cancel the PO
+  I cancel the PO and create a new revision
 -
    !python {model: purchase.order}: |
+     po = self.browse(cr, uid, ref('purchase_order_1'))
+     old_name = po.name
      self.action_cancel(cr, uid, [ref('purchase_order_1')])
--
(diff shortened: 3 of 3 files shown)


pedrobaeza:
Where is the code for providing the value?

gurneyalex:
forgot to `git add`... Now fixed.

pedrobaeza:
I was thinking in a post_init hook for handling existing purchase orders, because this only handle new created purchase orders.

eLBati:
:+1: 
@pedrobaeza I think we don't need a post_init hook because I think there are not existing DB to fix: the current bug prevents the module to be used correctly, because it always writes 'False' in PO name

pedrobaeza:
I insist: we need a post_init_hook to handle existing orders. Imagine this scenario:
- You have already in your DB plenty of draft sales orders.
- You install the module.
- You cancel one of this sales order, and click on the new revision button.
- You will get the `False-01` sales order number, because this sales order doesn't have any value in unrevisioned_name field

eLBati:
Right @pedrobaeza 
See https://github.com/gurneyalex/purchase-workflow/pull/1
This should also fix @jaronemo's [problem](https://github.com/OCA/purchase-workflow/issues/148#issuecomment-117726467)

jaronemo:
Dear Sir..

test ..code ...still is falied...

problem is some..

eLBati:
@gurneyalex oops
`./purchase_order_revision/__init__.py:5:1: E302 expected 2 blank lines, found 1`

gurneyalex:
@eLBati fixed

lepistone:
@gurneyalex, as for the test failure, it is `purchase_requisition_multicurrency` that makes the field required. I don't know why tests passed before OCA/maintainer-quality-tools#212, but that may well be by accident.

pedrobaeza:
OK, we can move on with this PR, and let the other one be fixed by https://github.com/OCA/purchase-workflow/pull/164

:+1:

jaronemo:
Dear Sir

I re-established a test system, and at the same time today re-download purchase_order_revision verified.
It seems to have a normal job.
However, there is a problem spot for the first time why the establishment of a single number will still be False-01? Rather than follow the original single number PO00002-01 it?
My understanding is that with the actual situation should pic2 is the same only more in line with business processes

pic1:
image

pic2:
image

pedrobaeza:
This is due to the lack of a post-init hook that initializes unrevisioned_name field on existing orders. @gurneyalex, can you make it?

eLBati:
@jaronemo could you provide the steps to reproduce the problem on a new database? Thanks

pedrobaeza:
1. Create a purchase order without installing the module.
2. Install the module
3. Cancel the order and click on new revision

jaronemo:
Dear Sir
1. create new database 
2.create supplier
3.create product with this supplier 
4.create purchase order (confirm order status)
5.install module 
6.cancel purchase order 
7.touch revison button create new purchase ...

show false-01 number...this is wrong..

eLBati:
@gurneyalex @pedrobaeza uh it seems changes of my PR https://github.com/gurneyalex/purchase-workflow/pull/1 have been lost somewhere

Created new PR https://github.com/OCA/purchase-workflow/pull/183

PR #152, opened 2015-06-26
[FIX] Replace field discount instead of having 2 fields with the same…

vrenaville:
Replace field discount instead of having 2 fields with the same, it prevent that
the supplier invoice view is not correctly displayed if you have discount in sale order
activate.


gurneyalex:
:+1: and favorable to fast tracking as this is a bugfix.

nbessi:
:+1:

pedrobaeza:
But the current fix has not too much sense: replace discount field by another field discount. Isn't better to change the corresponding attributes?

vrenaville:
@pedrobaeza Thanks for the review, I have push a fix according to your comment

pedrobaeza:
Thanks :+1:

PR #40, opened 2014-10-27
[ADD] purchase_delivery_split_date

clonedagain:
Module to split purchase receptions by dates.
Supersedes this Launchpad proposal: https://code.launchpad.net/~numerigraphe-team/purchase-wkfl/7.0-add-purchase_delivery_split_date/+merge/211374

"This branch adds a new module which, when installed, creates one reception picking for each expected date on the purchase order.
This way stock workers have a clear view of which goods should be received when.

This modules uses the modular design of purchase orders to create several pickings rather that a single one: to the best of my knowledge, this should be compatible with the other modules in this branch."

PR #153, opened 2015-06-30
8.0 add purchase delivery split date

clonedagain:
New module which creates one reception picking for each expected date on the purchase order.
This way stock workers have a clear view of which goods should be received when.

This modules uses the modular design of purchase orders to create several pickings rather that a single one: to the best of my knowledge, this should be compatible with the other modules in this branch.

Proposed here for v8 with the same features as were proposed for v7 earlier: #40
Now comes with a unittest as demanded in https://github.com/OCA/purchase-workflow/pull/40#issuecomment-60617312.

### DIFF
--- a/purchase_delivery_split_date/README.rst
+++ b/purchase_delivery_split_date/README.rst
@@ -1,0 +1,40 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License AGPL-3
+
+Split Purchase Deliveries in one reception per expected date
+============================================================
+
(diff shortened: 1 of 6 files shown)


jgrandguillaume:
Hi,

Thanks for the contrib, very useful one. Functionally reviewed, I'm :+1: 

Regards,

Joël

clonedagain:
Merged with latest 8.0 : reviewers wanted.

clonedagain:
Rebased on latest 8.0 : reviewers still wanted.

gurneyalex:
:+1:

PR #155, opened 2015-07-07
[ADD] [7.0] Module 'purchase_order_cancel_on_invoice_method_manual'.

JordiBForgeFlow:
This module was written to extend the purchasing capabilities of Odoo.
If the state of the PO is 'approved', there is no transition foreseen in
order to cancel the PO.

The user might be blocked in the following situation:
- Create a purchase order with invoicing set as Based on incoming shipment
- Validate the purchase order, create the shipment
- Then, cancel it (the shipment)
- Return back in the purchase order, the PO should be in shipping exception
- Hit the "manually corrected"
- Then, try to cancel the PO: nothing happens.
# Known issues / Roadmap

This module only applies to verion 7.0. In version 8.0 the limitation was
resolved by the following commit:
https://github.com/odoo/odoo/commit/4a281754db610d004701ce8edd9aeed32e766af4


OSguard:
@jbeficent do you fix this?

pedrobaeza:
Closing as no answer. Feel free to reopen if needed.

PR #157, opened 2015-07-09
Purchase Delivery Module for Odoo v8

jpoa:
Module now works with Odoo 8. Removed the last view for picking as it seems to be already present (at least on my install), if later I see its needed it can be readded. At least this way some of the needed changes are already documented.

PR #159, opened 2015-07-21
Migrate purchase_supplier_postage_free to 8.0

guewen:
The 7.0 version is on #107


pedrobaeza:
I say the same as for 7.0: I prefer the more versatile and complete option of including delivery method in purchases, provided by this PR: https://github.com/OCA/purchase-workflow/pull/157

pedrobaeza:
:-1:

jgrandguillaume:
Hi,

Delivery methods in purchases are used to compute the proper amount for the delivery made from suppliers. You have various cases (that may vary depending on incoterms):
1. The goods are purchased transport included, then no transport costs nor line need to be added in PO. 
2. The goods are purchased, and you need to include a line for the shipping costs given by your supplier (in that case, no need for purchase_delivery, your supplier inform you how much it'll be right ?)
3. The goods are purchased "Exworks" and it is up to the company to organize the transport from the supplier to their own warehouse. 

In case 1) and 2) you don't really need purchase_delivery. For case 3) purchase_delivery will help you to estimate your shipping cost for this PO ( note that even then, your computation may vary from the reality).

**Anyway** before knowing how much are my shipping costs, I want to know in my procurement process when I reach the postage free amount. I most cases I faced, the process is as follow:
1. Sales generate needs and trigger PO through MTO or MTS rules
2. As long as PO for a given supplier are not confirm, quantities of goods are added in the draft PO, trying to group purchase as much as possible to lower the shipping costs per product
3. Procurement team needs to review and confirm PO based on various criteria to optimize the costs, one of them is to easily know if they reach the "PO amount" that give free postage fees.

This module propose an easy way to manage this use case:
1. Define free postage amount for each supplier
2. Have the missing amount to benefit from postage free in a blink for every PO
3. Filter on PO that have reach the postage free amount to confirm them

**Conclusion**

This module simple and efficient. For me this is the first module to be installed to manage this problematic, then purchase_delivery if you really have a need to "pre-compute" your postage fee per PO.

purchase_delivery is complementary with this module and bring more advance features. I see no point in blocking this PR, but would rather extends it in purchase_delivery (because you always want to identify PO that reach postage free, that's all the procurement "game").

I'm :+1:  because:
- I think lot's of people want just to know "Do I reach my postage free for this PO" .
- I see no reason to pull all the "complexity" of purchase_delivery (defining delivery method, grid per supplier etc..) just to know if your postage free has been reach
- Just been reviewed by two experienced purchase manager (>10 year) and they love it
- Simple is beautiful ;)

Regards,

Joël

pedrobaeza:
OK, thanks for the reasoning, @jgrandguillaume. I see it can coexist both, but don't you think to use "delivery" instead of postage can be better to unify nomenclature?

lepistone:
@pedrobaeza "postage" sounds more precise to me ("delivery" is not that clear), and in general I don't think it's worth it to rename modules for this.

purchase_supplier_postage_free/model/purchase_order.py
```diff
@@ -0,0 +1,79 @@
+# -*- coding: utf-8 -*-
+# -*- coding: utf-8 -*-
```
lepistone: duplicated magic comment

purchase_supplier_postage_free/model/res_partner.py
```diff
@@ -0,0 +1,41 @@
+# -*- coding: utf-8 -*-
```
lepistone, line 1: duplicated magic comment

lepistone:
:+1: with two minor remarks. Thanks!

gurneyalex (approved):

purchase_supplier_postage_free/model/purchase_order.py
```diff
@@ -0,0 +1,78 @@
...
+#
+
+import openerp.addons.decimal_precision as dp
+from openerp import models, fields, api
+
+
+class PurchaseOrder(models.Model):
+    _inherit = 'purchase.order'
+
+    free_postage = fields.Float(
```
legalsylvain, line 29: Hi.

I think that the free portage should not be computed, but filled with an @api.onchange on the partner_id. For the time being, if a user change the free_postage of a partner, it will change the free_postage of all the old purchase order, and the value free_postage_reached (stored) will change, that is wrong.
In many cases, free postage is a value that can change during the time.

Don't you think ?

(or maybe I missed something)

legalsylvain (changes requested):

pedrobaeza:
This is very old and Travis is red, so I close. If you are still interested, please feel free to reopen and fix things.

PR #160, opened 2015-07-21
Add purchase_order_type

guewen:
Add a configurable _type_ on the purchase orders.
This type can be used in filters and groupbys.


purchase_order_type/report/purchase_report.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+class purchase_report(models.AbstractModel):
+    _inherit = "purchase.report"
+
+    order_type = fields.Many2one(comodel_name='purchase.order.type',
+                                 readonly=True,
+                                 string='Type')
+
+    def init(self, cr):
```
guewen: Breaking compatibility :-(

angelmoya:
Tested on runbot, and not ok.

Type is readonly on purchase in draft state.

nhomar:
:+1: all looks perfect.

purchase_order_type/models/purchase_order.py
```diff
@@ -0,0 +1,32 @@
...
+
+from openerp import models, fields
+from openerp.addons.purchase.purchase import purchase_order
+
+
+class PurchaseOrder(models.Model):
+    _inherit = 'purchase.order'
+
+    order_type = fields.Many2one(comodel_name='purchase.order.type',
+                                 readonly=True,
```
pedrobaeza: If you see https://github.com/odoo/odoo/blob/8.0/addons/purchase/purchase.py#L206, states are defined for putting readonly=True, so default value should be False.

jgrandguillaume:
Hi,

When readonly of the field type has been removed, I'm :+1: 

Regards

guewen:
Thanks for the reviews! I changed the readonly option.

guewen:
Travis is red due to https://github.com/OCA/maintainer-quality-tools/issues/231

purchase_order_type/models/purchase_order.py
```diff
@@ -0,0 +1,32 @@
...
+from openerp import models, fields
+from openerp.addons.purchase.purchase import purchase_order
+
+
+class PurchaseOrder(models.Model):
+    _inherit = 'purchase.order'
+
+    order_type = fields.Many2one(comodel_name='purchase.order.type',
+                                 readonly=False,
+                                 states=purchase_order.READONLY_STATES,
```
lmignon, line 31: IMO it would be safer to restrict the deletion of purchase.order.type if it's still referenced in a purchase.order with `ondelete='restrict'`

guewen, line 31: Indeed better. Done

lmignon:
@guewen Thank you for the change. To be complete you should also add an 'active' field on `purchase.order.type`

guewen:
@lmignon good suggestion, done

pedrobaeza:
As we did in _sale_order_type_ module (https://github.com/OCA/sale-workflow/tree/8.0/sale_order_type), it can be also interesting to attach certain configuration to each type, so that it can be propagated to the purchase orders of that type, but for now I think it can be enough just mentioning this in the "Known issues / Roadmap" section.

guewen:
@pedrobaeza done

pedrobaeza:
:+1:

lmignon:
:+1:  (Code + functional)

pedrobaeza:
Uhm, now we have in Travis this error:

```
ImportError: No module named ldap.filter
```

pedrobaeza:
Travis problems fixed, so I merge.

PR #161, opened 2015-07-24
[ADD] add module to group procurement by analytic

laetitia-gangloff:
Group purchase order created from procurement by analytic account defined on sale order

depends of https://github.com/odoo/odoo/pull/7724

### DIFF
--- a/procurement_purchase_groupby_analytic/README.rst
+++ b/procurement_purchase_groupby_analytic/README.rst
@@ -1,0 +1,70 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Procurement Purchase Group by Analytic
+======================================
+
+This module groups purchase order created from procurement by analytic account.
+The analytic account used comes from project set on sale order.
+On purchase order line, set this analytic account.
+
+Installation
+============
+
+To install this module, you need to:
+
+* Click on install button
+
+Usage
+=====
+
+Example of use:
+
+Create a sale order with an analytic account, set some products with route buy.
+Run procurement, products are in the same purchase order (according other
+rules like partner).
+Create a second sale order, with another analytic account and set some products
+with route buy.
+Run procurement, 
+
+Current behavior (without this module): add products in existing
+purchase order.
+
+Excepted behavior (with this module): get a separate purchase order for
+these products (because it is not the same analytic account).
+
+For further information, please visit:
+
+* https://www.odoo.com/forum/help-1
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/purchase-workflow/issues>`_.
+In case of trouble, please check there if your issue has already been reported.
+If you spotted it first, help us smashing it by providing a detailed and welcomed feedback
+`here <https://github.com/OCA/purchase-workflow/issues/new?body=module:%20procurement_purchase_groupby_analytic%0Aversion:%208.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
+
+
+Credits
+=======
+
+Contributors
+------------
+
+* Laetitia Gangloff <laetitia.gangloff@acsone.eu>
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit http://odoo-community.org.
--- a/procurement_purchase_groupby_analytic/__init__.py
+++ b/procurement_purchase_groupby_analytic/__init__.py
@@ -1,0 +1,3 @@
+# -*- coding: utf-8 -*-
+
+from . import models
--- a/procurement_purchase_groupby_analytic/__openerp__.py
+++ b/procurement_purchase_groupby_analytic/__openerp__.py
@@ -21,7 +21,7 @@
 
 {
     "name": "Procurement purchase group by Analytic",
-    "version": "0.1",
+    "version": "8.0.1.0.1",
     'author': "Acsone, Odoo Community Association (OCA)",
     "category": "Other",
     "website": "http://www.acsone.eu",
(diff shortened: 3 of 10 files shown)


procurement_groupby_analytic/README.rst
```diff
@@ -0,0 +1,55 @@
...
+
+* https://www.odoo.com/forum/help-1
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/purchase-workflow/issues>`_.
+In case of trouble, please check there if your issue has already been reported.
+If you spotted it first, help us smashing it by providing a detailed and welcomed feedback
+`here <https://github.com/OCA/purchase-workflow/issues/new?body=module:%procurement_groupby_analytic%0Aversion:%200.1%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
```
pedrobaeza: Remove `%` before procurement word and version should be 8.0

procurement_groupby_analytic/__openerp__.py
```diff
@@ -0,0 +1,36 @@
...
+#
+##############################################################################
+
+{
+    "name": "Procurement group by Analytic",
+    "version": "0.1",
+    'author': "Acsone, Odoo Community Association (OCA)",
+    "category": "Other",
+    "website": "http://www.acsone.eu",
+    'summary': "Group purchase order by procurement",
```
pedrobaeza: This summary is not correct

procurement_groupby_analytic/models/procurement.py
```diff
@@ -0,0 +1,57 @@
...
+
+    @api.model
+    def _get_available_draft_po_domain(self, procurement, partner):
+        """ If account analytic is defined on procurement order
+            search draft purchase order on this criteria
+        """
+        available_draft_po_domain = super(
+            ProcurementOrder, self)._get_available_draft_po_domain(
+                procurement, partner)
+        available_draft_po_domain.append(
```
pedrobaeza: I don't think this should be the criteria to link sales orders with an analytic account with purchase orders. Follow the procurement chain for this.

pedrobaeza:
This makes something more than only grouping by analytic account, so it should be reflected on the README and the module name changed accordingly. Maybe _procurement_analytic_, but there are points commented inline that should be cleared before.

procurement_groupby_analytic/README.rst
laetitia-gangloff: Right for the % ! but I think it is %20 instead of %.

Version is not for the version of the module ?

pedrobaeza: No, because module version numbering is a totally mess. It's for recognizing which of the branches is the affected one. It is supposed that you're already running last available version (we should talked here better about revisions) of the module in the branch, and if not, the first suggestion made is that.

laetitia-gangloff:
My use case is the following :

I create a sale order with an analytic account, I set some products with route buy
When I run procurement, I want product in the same purchase order (according other rules like partner).
If I create a second sale order, with another analytic account and set some products with route buy
When I run procurement,

current behavior, is to add product in the existing purchase order.

excepted behavior, is to get a separate purchase order for these products (because it is not the same analytic account)

Maybe account_analytic_id can be a computed field (instead of related field), and in this module, the value comes from sale order.

pedrobaeza:
OK, now I understand, then a proper name would be _procurement_purchase_groupby_analytic_. This is a similar case like this module: https://github.com/odoomrp/odoomrp-wip/tree/8.0/procurement_purchase_no_grouping

nhomar:
2015-08-04 7:23 GMT-05:00 Laetitia Gangloff (ACSONE) <
notifications@github.com>:

> My use case is the following :
> 
> I create a sale order with an analytic account, I set some products with
> route buy
> When I run procurement, I want product in the same purchase order
> (according other rules like partner).
> If I create a second sale order, with another analytic account and set
> some products with route buy
> When I run procurement,
> 
> current behavior, is to add product in the existing purchase order.
> 
> excepted behavior, is to get a separate purchase order for these products
> (because it is not the same analytic account)
> 
> Maybe account_analytic_id can be a computed field (instead of related
> field), and in this module, the value comes from sale order.

Put specifically this use case in the readme it is a cool explanation.

## 

---

Saludos Cordiales

CEO at Vauxoo https://www.vauxoo.com Odoo's Gold Partner.

[image: --]
Nhomar Hernandez
[image: http://]about.me/nhomar
http://about.me/nhomar?promo=email_sig

laetitia-gangloff:
I renamed the module to procurement_purchase_groupby_analytic.
I updated readme to put the use case.
analytic_account_id is a computed field instead of a related field.

procurement_purchase_groupby_analytic/README.rst
```diff
@@ -0,0 +1,70 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Procurement Purchase Group by Analytic
+===========
```
pedrobaeza: This must have the same length of the title

procurement_purchase_groupby_analytic/README.rst
```diff
@@ -0,0 +1,70 @@
...
+
+This module group purchase order from sale order procurement by analytic.
```
pedrobaeza: s/group/groups

procurement_purchase_groupby_analytic/README.rst
```diff
@@ -0,0 +1,70 @@
...
+When purchase order is created from procurement order it sets analytic
+on purchase order line.
+
+Use case:
+
+Create a sale order with an analytic account, set some products with route buy.
```
pedrobaeza: Put this in the "Usage" section

procurement_purchase_groupby_analytic/README.rst
```diff
@@ -0,0 +1,70 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Procurement Purchase Group by Analytic
+===========
+
+This module group purchase order from sale order procurement by analytic.
+When purchase order is created from procurement order it sets analytic
+on purchase order line.
```
lmignon: Do you mean

> Once installed, when products with a 'buy' route are sold the generated purchase orders are grouped by supplier and analytic account.

laetitia-gangloff: All purchase orders created from a procurement order linked to a sale order with analytic account

laetitia-gangloff:
I tried to update readme.rst according to your comments.

PR #163, opened 2015-07-31
fix warning on view load

gurneyalex:
remove deprecated field  on ir.ui.view


pedrobaeza:
:+1:

lepistone:
:+1:
This fix is correct. There is another unrelated problem in a test that I'm going to address right now in another PR.

damdam-s:
:+1:

PR #164, opened 2015-07-31
fix travis build

gurneyalex:
test purchase_requisition_multicurrency together with
purchase_requisition_bid_selection


pedrobaeza:
:+1: if all green (obviously)

pedrobaeza:
Well, it's still the same. I think it's better to provide a default value for the required field. Don't you think?

jaronemo:
My test results are as follows, still problematic.Test different databases
image

gurneyalex:
thanks @damdam-s for the fix. 

:+1:

gurneyalex:
@damdam-s can you squash the first 3 commits?

lmignon:
:+1:

PR #165, opened 2015-07-31
[8.0] Fixes on purchase_partial_invoicing

adrienpeiffer:
- Avoid to copy cancelled_qty field
- Trigger purchase order workflow when when there is a quantity cancellation


pedrobaeza:
:+1:

eLBati:
LGTM :+1: 
Would it be cheap to add a test for this?

PR #166, opened 2015-08-03
[ADD] add module "purchase_order_reorder_lines" migrated from sale-wo…

damdam-s:
…rkflow

Depends on https://github.com/OCA/stock-logistics-workflow/pull/139/

### DIFF
--- a/purchase_order_reorder_lines/README.rst
+++ b/purchase_order_reorder_lines/README.rst
@@ -7,9 +7,11 @@ Purchase order lines with sequence number
 Provide a new field on the purchase order form, allowing to manage the order of the lines
 order.
 
-For further information, please visit:
+Installation
+============
 
-* https://www.odoo.com/forum/help-1
+This module depends on the following module(s) :
+ * stock_picking_reorder_lines (https://github.com/OCA/stock-logistics-workflow/)
 
 Known issues / Roadmap
 ======================
--- a/purchase_order_reorder_lines/README.rst
+++ b/purchase_order_reorder_lines/README.rst
@@ -1,0 +1,49 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Purchase order lines with sequence number
+=========================================
+
+Provide a new field on the purchase order form, allowing to manage the order of the lines
+order.
+
+For further information, please visit:
+
+* https://www.odoo.com/forum/help-1
+
+Known issues / Roadmap
+======================
+None known
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/purchase-workflow/issues>`_.
+In case of trouble, please check there if your issue has already been reported.
+If you spotted it first, help us smashing it by providing a detailed and welcomed feedback
+`here <https://github.com/OCA/purchase-workflow/issues/new?body=module:%20purchase_order_reorder_lines%0Aversion:%208.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
+
+
+Credits
+=======
+
+Contributors
+------------
+
+* Damien Crier <damien.crier@camptocamp.com>
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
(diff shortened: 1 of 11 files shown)


purchase_order_reorder_lines/README.rst
```diff
@@ -0,0 +1,49 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Purchase order lines with sequence number
+=========================================
+
+Provide a new field on the purchase order form, allowing to manage the order of the lines
+order.
+
+For further information, please visit:
```
pedrobaeza: Remove this part

purchase_order_reorder_lines/README.rst
```diff
@@ -0,0 +1,49 @@
...
+
+* https://www.odoo.com/forum/help-1
+
+Known issues / Roadmap
```
pedrobaeza, line 16: Include a section Installation explaining that the module stock_picking_reorder_lines is needed and point to OCA/stock-logistics-workflow repo.

purchase_order_reorder_lines/purchase.py
```diff
@@ -0,0 +1,94 @@
...
+from openerp.tools.safe_eval import safe_eval
+
+
+class PurchaseOrderLine(models.Model):
+    _inherit = 'purchase.order.line'
+    _order = 'order_id desc, sequence, id'
+
+    @api.model
+    def _get_sequence(self):
+        last_sequence = 0
```
pedrobaeza: I point to the same discussion in https://github.com/OCA/stock-logistics-workflow/pull/139#discussion-diff-36064784

purchase_order_reorder_lines/purchase.py
```diff
@@ -0,0 +1,98 @@
...
+            order,
+            order_line,
+            picking_id,
+            group_id)
+        if res:
+            res[0]['sequence'] = order_line.sequence
+        return res
+
+    @api.depends('order_line')
+    def _get_max_line_sequence(self):
```
nhomar, line 61: Can you extend with a little docstring why is this method for (functionally speaking)?

purchase_order_reorder_lines/purchase.py
```diff
@@ -0,0 +1,98 @@
...
+            [('invoice_id', 'in', invoice_ids)])
+        for invoice_line in invoice_lines:
+            order_lines = purchase_line_obj.search(
+                [('invoice_lines', '=', invoice_line.id)],
+                )
+            if not order_lines:
+                continue
+
+            if not invoice_line.sequence:
+                invoice_line.write({'sequence': order_lines.sequence})
```
nhomar: You are forcing the same sequence on invoices.

Frequently the invoices can have its own order (for example if the company want have an exact order of the physical invoice), I think it is _not_ a good Idea, did I interpret correctly the reason of this method?

damdam-s:
update :)

purchase_order_reorder_lines/__openerp__.py
```diff
@@ -0,0 +1,40 @@
...
+#
+{
+    'name': 'Purchase order lines with sequence number',
+    'version': '0.1',
+    'category': 'Purchase Management',
+    'author': "Camptocamp,Odoo Community Association (OCA)",
+    'website': 'http://www.camptocamp.com',
+    'depends': [
+        'purchase',
+        'stock_picking_reorder_lines',
```
nhomar, line 32: Hello.

Sorry for the outdated comment it pass away from me in the first review, why this dependency?

Put an order on the purchase should not depend from the picking.

I see more logic in that the method which depend of such module stay there then you can have 2 independent logics don't you think?.

nhomar:
You can see an independent implementation here:

https://github.com/Vauxoo/addons-vauxoo/blob/8.0/purchase_order_line_sequence/model/purchase.py

(we can move to this one one time is merged BTW).

Regards.

purchase_order_reorder_lines/__openerp__.py
damdam-s, line 32: Hello,
Original module in v7 already had this dependency.
It depends on  'stock_picking_reorder_lines' because that module add 'sequence' field in "stock.move" object that is used in this module when confirming a purchase.
Without this dependency, you cannot be sure there is a 'sequence' field in 'stock.move' object and Odoo server always raise a warning in the log saying "write to an unexisting field 'sequence' on object 'stock.move'"

gurneyalex:
:+1: thanks for porting

jgrandguillaume:
:+1:  Thanks for the work

PR #167, opened 2015-08-03
[RFR] Align various views of the framework agreement

StefanRijnhart:
[FIX] Use plural form in menu items
[RFR] Don't use spaces in view XML ID

### DIFF
--- a/framework_agreement/view/product_view.xml
+++ b/framework_agreement/view/product_view.xml
@@ -1,83 +1,14 @@
 <?xml version="1.0" encoding="utf-8"?>
 <openerp>
   <data>
-    <record id="agreement in product view" model="ir.ui.view">
+    <record id="agreement_in_product_view" model="ir.ui.view">
       <field name="name">agreement in product view</field>
       <field name="model">product.product</field>
       <field name="inherit_id" ref="product.product_normal_form_view"/>
       <field name="arch" type="xml">
         <field name="seller_ids" position="after">
           <group string="Framework agreements (LTA)" colspan="4">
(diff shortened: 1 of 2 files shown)


framework_agreement/view/product_view.xml
```diff
@@ -1,83 +1,14 @@
...
   <data>
-    <record id="agreement in product view" model="ir.ui.view">
+    <record id="agreement_in_product_view" model="ir.ui.view">
       <field name="name">agreement in product view</field>
       <field name="model">product.product</field>
       <field name="inherit_id" ref="product.product_normal_form_view"/>
       <field name="arch" type="xml">
         <field name="seller_ids" position="after">
           <group string="Framework agreements (LTA)" colspan="4">
-            <field name="framework_agreement_ids" nolabel="1">
```
pedrobaeza, line 11: Why this removing?

StefanRijnhart, line 11: I'm reusing the stand alone view, on which I hide the product field depending on the context.

StefanRijnhart, line 11: There were some problems with this embedded view. It did not have some of the fields, such as the required portfolio field so that I could not add a new agreement from this view.

pedrobaeza, line 11: OK, I see. Thanks for the explanation

pedrobaeza:
:+1:

StefanRijnhart:
Thanks for the quick review @pedrobaeza! However, I am retracting this PR in favour of #113

PR #169, opened 2015-08-05
[ADD] add module to request procurement from sale

laetitia-gangloff:
On sale order, at confirmation, it is possible to request procurement at the same time.

It is based on the wizard introduced in the module procurement_batch_generator.

### DIFF
--- a/sale_to_purchase_order/__openerp__.py
+++ b/sale_to_purchase_order/__openerp__.py
@@ -1,0 +1,37 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    Authors: Laetitia Gangloff
+#    Copyright (c) 2015 Acsone SA/NV (http://www.acsone.eu)
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
+    "name": "Sale to purchase order",
+    "version": "0.1",
+    'author': "Acsone, Odoo Community Association (OCA)",
+    "category": "Other",
+    "website": "http://www.acsone.eu",
+    "depends": ["sale",
+                "purchase",
+                "procurement_batch_generator",
+                ],
+    "data": ["views/sale_order_views.xml",
+             ],
+    "license": "AGPL-3",
+    "installable": True,
+    "application": False,
+}
--- a/sale_to_purchase_order/models/procurement_batch_generator.py
+++ b/sale_to_purchase_order/models/procurement_batch_generator.py
@@ -1,0 +1,87 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    Authors: Laetitia Gangloff
+#    Copyright (c) 2015 Acsone SA/NV (http://www.acsone.eu)
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+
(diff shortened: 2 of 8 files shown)


sale_to_purchase_order/__openerp__.py
```diff
@@ -0,0 +1,37 @@
...
+
+{
+    "name": "Sale to purchase order",
+    "version": "0.1",
+    'author': "Acsone, Odoo Community Association (OCA)",
+    "category": "Other",
+    "website": "http://www.acsone.eu",
+    "depends": ["sale",
+                "purchase",
+                "procurement_batch_generator",
```
nhomar, line 30: If you will generate procurements what about when the procurement is "Manufacture", should not this module have such cases covered? (It is a serious question i really ahve the doubt about it)

sale_to_purchase_order/models/procurement_batch_generator.py
```diff
@@ -0,0 +1,87 @@
...
+
+class ProcurementBatchGenerator(models.TransientModel):
+    _inherit = 'procurement.batch.generator'
+
+    @api.model
+    def _is_product_to_add(self, order_line):
+        """ You can add rules to determine if the product of the order line
+        should be in the default line
+        """
+        return True
```
nhomar, line 34: Can you extend the functional use case where this hook is necessary?

An example should be cool!

pedrobaeza:
Why not using MTO (make to order)?

nhomar:
@pedrobaeza 

There is cases where you need a manual an explicit control of such procurements.

Even this is an actual use case necesary for us and we did it before for v6.1

And BTW with mts_mto module plus this one plus the original mto we fill all gaps necesary to fill the procurements.

IMHO a good job.

nhomar:
Not testes, just read the Test cases which looks perfect, and runbot green plus a little comment on dependency it have a +1 for me.

Important test case generating the procurement with mrp will be necesary to fill such gap.

pedrobaeza:
I understand the need, @nhomar, but you already have a procurement created when you put MTO as the route, so I don't see the case. It would be good to have as documentation if there's a case.

nhomar:
2015-08-05 8:44 GMT-05:00 Pedro M. Baeza notifications@github.com:

> I understand the need, @nhomar https://github.com/nhomar, but you
> already have a procurement created when you put MTO as the route, so I
> don't see the case. It would be good to have as documentation if there's a
> case.

+1 for the use case.

## 

---

Saludos Cordiales

CEO at Vauxoo https://www.vauxoo.com Odoo's Gold Partner.

[image: --]
Nhomar Hernandez
[image: http://]about.me/nhomar
http://about.me/nhomar?promo=email_sig

sale_to_purchase_order/models/procurement_batch_generator.py
laetitia-gangloff, line 34: My use case is to exclude some products category.

But I can imagine other criteria like product supplier, ...

laetitia-gangloff:
The use case it to add the possibility to choose the quantity to reorder.

pedrobaeza:
In a MTO you can also do that going to the procurement order AFAIK

jgrandguillaume:
Hi,

I agree with @pedrobaeza here, in such a use case you should use MTO. IMO:
- You work with MTO route so your SO generate a MTO procurement that trigger a PO. The customer delivery is linked to the supplier reception
- You work with MTS route so your SO generate a MTS procurement that trigger a PO when min stock rules is reach. The customer delivery is NOT linked to the supplier reception

In both cases, before the scheduler runs, you can always modify the quantity in the procurement order or create a manual procurement order (from procurement menu or product inventory tab).

Why would you sometimes create the procurement from SO when you confirm it and sometimes not ? In my understanding, you use case should be covered by adding min stock rule (min=0;max=0) for every product.

Except to avoid creating min stock rules, I don't see the use of this module, but I may miss something ;) For that I'm :-1: 

Regards,

Joël

chrisdec:
Hi, 

The module like it is here is simply a shortcut to create procurement for products. I so agree with you that utility for OCA is not really clear but it is not exactly what I expected from this module (functional aspect). 

I explain my customer use case : 

He has some products which are clearly MTO products. Each time he sells the product, he wants to buy it to the supplier and of course, delivery order is linked to supplier reception. -->  I simply assigned MTO route to those products. 

He has a lot of MTS products with reordering rules (MIN/MAX). Most of the time, he sells it to the customer and rebuy it to the supplier when needed according to min/max rules --> No link between delivery orders and supplier reception. 

BUT sometimes, for those MTS products, when he sells it to the customer, he wants to buy the quantity to the supplier, he doesn’t want to take in the stock (no reservation). In this case, for this particular SO, the product is treated like a MTO product. Delivery order and supplier reception must be linked together. He will never deliver to the customer if he doesn’t receive product from supplier. 

I know we can put MTO route on SO lines to force it but for my customer it is quite “difficult”. Functionally, the idea behind “confirm and deliver” is to create a procurement which is like a MTO procurement. My delivery order will not be available until I receive products from supplier. In my opinion, In the “confirm and deliver” windows, MTS products appears but not MTO products. 

What do you think of this? This module could be a candidate to oca? 

We know that the module is not complete but we are working on it with @laetitia-gangloff and any ideas are welcome! 

FYI : Some reasons push my customer to use this functionality :  
-   A customer orders a huge quantity of product; he doesn’t want to empty his stock and refuse other “smaller” orders waiting for min/max rules procurement and supplier reception. The customer with huge quantity is the one who waits for supplier reception. 
-    He wants to empower commercial people. They are responsible of stock arrival before installation. 

Thanks in advance for your feedback, 

Regards, 

Christelle

laetitia-gangloff:
Hello,
I close this PR, we following your advice and use route on sale order to manage the case.

Thank you.

PR #170, opened 2015-08-06
[ADD] product_supplierinfo_import

Daniel-CA:
This module allows to load product.supplierinfo list from a csv/xml file.
Usually used to load/update from a file that a supplier send from his own products.

### DIFF
--- a/product_supplierinfo_import/README.rst
+++ b/product_supplierinfo_import/README.rst
@@ -1,0 +1,22 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Product supplierinfo import
+===========================
+This module allows to load product.supplierinfo list from a file.
+
+Known issues / Roadmap
+======================
+| The file format should have these colummns:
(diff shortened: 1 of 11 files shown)


lmignon:
By installing https://github.com/OCA/purchase-workflow/tree/8.0/product_by_supplier you can use the standard import/export functionality from Odoo

Daniel-CA:
Thanks for the info @lmignon, I have used it and works fine, but I have seen some lacks for example if you have the product code instead of the name of the product in your file or supplier code instead of the supplier name.
With this module you can also load products from different suppliers in the same file and also creates a `pricelist.partnerinfo` for the `product.supplierinfo` with its price.

lmignon:
@Daniel-CA IMO all your requirements are possible with the standard import/export. If you only have the product_code, it's easy to export the list of products (with their Id and the related code)  and replace in your file the code by the corresponding id.

Daniel-CA:
Yes, you can do some steps to extract data and import it again to map directly with the ids, this module just helps a not experimented user to import easily and in a fast way (imagine you have plenty of files for different suppliers every month). The module also gives a visual of lines that has not been added in the import.

lmignon:
@Daniel-CA If your volume is large you should use https://github.com/OCA/connector-interfaces/tree/8.0/base_import_async. IMO your module seems more linked to a specific need than a transverse need. Let's see what others think

PR #174, opened 2015-08-31
IMP purchase order policy

nicomacr:


pedrobaeza:
As we have did on _sale_order_type_ module (https://github.com/OCA/sale-workflow/tree/8.0/sale_order_type), I prefer to include this inside of the recently merged _purchase_order_type_ (https://github.com/OCA/purchase-workflow/tree/8.0/purchase_order_type), selecting at partner level this type, and defining in the type the order policy.

jjscarafia:
Are you sure? In sales order we have a similar module also
https://github.com/OCA/sale-workflow/tree/8.0/sale_partner_order_policy

It is a simple modification to allow using policies. I prefer not to force to use sale order types or purchase order types as they make bigger modifications. 
This is a simple and tiny module. Perhups could be also configured in x_order_type, but I would prefer to have it stand alone like in sales orders.

What do you think?

pedrobaeza:
That module was done before the other approach arrives, and it's a more completed approach to group all possible preferences in an order type than adding a bunch of fields in partner model, don't you think?

jjscarafia:
Ok, seams ok for me. I will as Nicolas to make a new PR adding policies on purchase_order_type module. That's ok?

pedrobaeza:
Cool, thanks!

pedrobaeza:
I close this in favor of https://github.com/OCA/purchase-workflow/pull/176

PR #176, opened 2015-09-01
FIX purchase_order_type to add policy to purchase orders

nicomacr:


purchase_order_type/models/purchase_order.py
```diff
@@ -19,15 +19,35 @@
...
 
-from openerp import models, fields
+from openerp import models, fields, api
 from openerp.addons.purchase.purchase import purchase_order
 
 
 class PurchaseOrder(models.Model):
     _inherit = 'purchase.order'
 
+    def _get_order_type(self):
```
pedrobaeza: Call if `_default_order_type` to be correctly identified

purchase_order_type/models/purchase_order.py
```diff
@@ -19,15 +19,35 @@
...
+        return self.env['purchase.order.type'].search([])[:1]
```
pedrobaeza: You can put argument `limit=1` instead

purchase_order_type/models/purchase_order.py
```diff
@@ -19,15 +19,35 @@
...
     order_type = fields.Many2one(comodel_name='purchase.order.type',
                                  readonly=False,
                                  states=purchase_order.READONLY_STATES,
                                  string='Type',
-                                 ondelete='restrict')
+                                 ondelete='restrict',
+                                 default=_get_order_type)
+
+    @api.multi
+    def onchange_partner_id(self, part):
```
pedrobaeza: Please don't change the API signature and call the argument `partner_id`.

purchase_order_type/models/purchase_order.py
```diff
@@ -19,15 +19,35 @@
...
+        res = super(PurchaseOrder, self).onchange_partner_id(part)
+        if part:
+            partner = self.env['res.partner'].browse(part)
+            res['value'].update({
```
pedrobaeza: You should include a check about the existence of res['value']:

```
res['value'] = res.get('value', {})
```

purchase_order_type/models/purchase_order.py
```diff
@@ -19,15 +19,35 @@
...
+                'order_type': partner.purchase_type.id
```
pedrobaeza: You shouldn't change to the first purchase type if the partner doesn't have any purchase type. Just let the previous (don't make any change).

pedrobaeza:
Thanks for the PR

max3903:
@nicomacr can you provide the steps to reproduce the bug ?

I tried to create a new request for quotation on runbot and I get this error message:

```
File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/3114265-176-0179ef/openerp/addons/purchase_order_reorder_lines/purchase.py", line 71, in _get_max_line_sequence
    max(purchase.mapped('order_line.sequence')) + 10
ValueError: max() arg is an empty sequence
```

pedrobaeza:
That it's because of https://github.com/OCA/purchase-workflow/blob/8.0/purchase_order_reorder_lines/purchase.py#L70, so uninstall _purchase_order_reorder_lines_ for testing this one and I'll provide a PR for fixing it.

pedrobaeza:
Great! :+1:, although a test for the added options would be very welcome

pedrobaeza:
@max3903, please check it #179 for the solution.

purchase_order_type/models/res_partner.py
```diff
@@ -0,0 +1,29 @@
...
+#
+
+from openerp import models, fields
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    purchase_type = fields.Many2one(
+        comodel_name='purchase.order.type', string='Sale Order Type')
```
max3903: string should be 'Purchase Order Type', no ?

Otherwise, I set a value on the supplier and created a PO for that supplier. The onchange set the same value on the PO field. So if this was the bug, it's now fixed.

:+1:

max3903:
CLA is OK

:+1:

PR #178, opened 2015-09-03
[IMP] README.md file

max3903:


README.md
```diff
@@ -1,12 +1,14 @@
 [![Build Status](https://travis-ci.org/OCA/purchase-workflow.svg?branch=7.0)](https://travis-ci.org/OCA/purchase-workflow)
```
pedrobaeza, line 2: Please add runbot badge

README.md
```diff
@@ -1,12 +1,14 @@
...
 
 This project aim to deal with modules related to manage purchase and their related workflow. You'll find modules that:
 
  - Allow to easily group purchase order
  - Choose the cheapest supplier
  - Add validation step
- - ...
+
+Translation Status
+----------------
```
pedrobaeza: Put the same length for the underlining

README.md
pedrobaeza, line 2: I'm wondering if we should rename this to "Travis Status" to differentiate it from the runbot one...

max3903, line 2: Well, it's another discussion. I am compliant with the template:
https://raw.githubusercontent.com/OCA/maintainer-quality-tools/master/sample_files/README.md

pedrobaeza, line 2: Yeah, I know, it's to know your opinion

max3903, line 2: :) I agree. Name of the tools would make more sense and be coherent with the other badges, but honestly, this is not even a cosmetic issue as it is an alternative to the image. I won't sleep any better if it changes.

yvaucher:
:+1:

PR #179, opened 2015-09-03
[FIX] purchase_reorder_lines: fix computation of max line sequence

pedrobaeza:


max3903:
:+1:

damdam-s:
:+1:

StefanRijnhart:
If I show the sequence as an integer field (without widget="handle") it works perfectly. When I test this using the handle widget, the default sequence value is ignored.

StefanRijnhart:
(slip of the keyboard, rest of comment:) Instead, when the handle widget is used, Odoo takes care of the correct increments itself, assigning the max sequence of the other lines + 1.

moylop260:
:+1:

StefanRijnhart:
@moylop260 just curious, did you verify the results of this change? From my testing it seemed that having such a default is meaningless. It would be nice to know if you got other results.

moylop260:
@StefanRijnhart 
Sorry, without test just code revision.

StefanRijnhart:
Setting to question then, waiting for @pedrobaeza's response.

pedrobaeza:
I have made again tests, and the problem is that handle widget makes thing partially right: when adding a line, it takes last **stored** sequence value and increments 1. But the key word is stored, because it doesn't take into account on-memory added lines, so if you add one line, save, and add another one, both lines will be correctly ordered, but if you add 2 lines and then save, both lines will have the same sequence number, so a reordering can happen depending on what is the second field for ordering (if it's id, no problem, but if it's name or something else, the problem is there).

This is something to fix on the widget itself, but as JS is my pending, this is a workaround, well this is a fix for the workaround that it's not correct.

StefanRijnhart:
Confirmed, thanks for the explanation! :+1:

PR #190, opened 2015-09-23
[FIX][purchase_partial_invoicing] Check access rights when validating…

andrius-preimantas:
… invoice

If user has access to account.invoice but not purchase.order then invoice
can not be validated. This is even more relevant when invoice is issued in
POS because lack of access rights restricts from finishing an order

### DIFF
--- a/purchase_partial_invoicing/purchase.py
+++ b/purchase_partial_invoicing/purchase.py
@@ -99,6 +99,10 @@ class AccountInvoice(models.Model):
     def invoice_validate(self):
         res = super(AccountInvoice, self).invoice_validate()
         purchase_order_obj = self.env['purchase.order']
+        # Users might not have access rights for Purchase Order (e.g. POS users)
+        if not purchase_order_obj.check_access_rights(
+                'read', raise_exception=False):
+            return res
         po_ids = purchase_order_obj.search([('invoice_ids', 'in', self.ids)])
         for purchase_order in po_ids:
             for po_line in purchase_order.order_line:


sbidoul:
Interesting bug. The solution you propose could silently create inconsistencies in the database if the user does not have enough access rights.

I suggest doing the search and loop on purchase orders as sudo, or better using https://github.com/OCA/server-tools/tree/8.0/base_suspend_security.

sbidoul:
cc/ @adrienpeiffer

andrius-preimantas:
@sbidoul , I'd like to disagree.

Using sudo is unacceptable in 2 ways:
1. As we write on PO line (hence `po_line.invoiced = False`) we would lose true value of `write_uid`. IMHO sudo() should be used only in very extreme situations;
2. There is no point of looking for Purchase Orders when validating Customer Invoice created from POS or Sales. So using sudo() would make a workaround for this issue. Basically we would be searching Purchase Orders for Invoices created from POS or Sales and would find nothing.

Using additional module `base_suspend_security` as a dependency would be even more irrelevant.

What do you think?
Cheers
Andrius

sbidoul:
> I'd like to disagree.

That's perfectly fine :smile: 

> 1. As we write on PO line (hence po_line.invoiced = False) we would lose true value of write_uid. IMHO sudo() should be used only in very extreme situations;

Agreed. I don't like sudo for that same reason. `base_suspend_security` exists as a solution to that problem. 

I understood (perhaps misunderstood?) your issue is that the write on po.line was failing when the user has no write access on purchase order lines. If that's the case, doing `po_line.invoiced = False` under `base_suspend_security` will solve the issue and preserve the write_uid. Doing the whole search and loop under `base_suspend_security` is even better in situations where the user has no read access at all on the purchase orders.

andrius-preimantas:
Actually it is failing here:

```
        purchase_order_obj = self.env['purchase.order']
        po_ids = purchase_order_obj.search([('invoice_ids', 'in', self.ids)]) <-- Fail
        for purchase_order in po_ids:
```

So do you suggest adding `base_suspend_security` to dependency list and use that one instead?

Cheers
Andrius

sbidoul:
> So do you suggest adding base_suspend_security to dependency list and use that one instead?

Yes. That search+loop must give the same result independently of the permissions of the user. I think it's the typical use case for `base_suspend_security`.

andrius-preimantas:
Yeah, but still adding a dependency which has no business correlation with `purchase_partial_invoicing` does not feel right. IMHO technical dependencies should be kept to minimum.

Thanks for your suggestion but my case needs permission propagation which is avoided when using base_suspend_security. So I'll just leave PR as it is and if you think it does not suit good purpose please close it :-)

Cheers
Andrius

adrienpeiffer:
Hi,
As specified by @sbidoul, your correction can cause some inconsistencies in the event that the invoice comes from a purchase order and that the user has no access rights on purchases.
There are already similar cases in Odoo (e.g. https://github.com/odoo/odoo/blob/8.0/addons/purchase/purchase.py#L1640)
Contrary to the implementation in the example, I also encourage the use of `base_suspend_security` to preserve correct references on `write_uid` and on tracker messages

adrienpeiffer:
An other solution based on `base_suspend_security` in #191

pedrobaeza:
I prefer the option with base_suspend_security

andrius-preimantas:
As @adrienpeiffer made PR based on hopefully better solution I guess I can close this one :-)

PR #192, opened 2015-09-24
Update product_supplierinfo_view.xml

thomasklosinsky:
Deleted the product field which is causing troubles creating new products and directly adding new product supplier. The field was obsolete.

### DIFF
--- a/product_by_supplier/views/product_supplierinfo_view.xml
+++ b/product_by_supplier/views/product_supplierinfo_view.xml
@@ -40,18 +40,6 @@
             </field>
         </record>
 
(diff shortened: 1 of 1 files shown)


pedrobaeza:
The reason why the field is there is because when you create a new product supplierinfo in form view from the menu "Product by supplier" (this can be achieved having the default grouping and clicking on New), there's no possibility to select the product.

The proper patch should be a primary view inheriting the main one and with less priority, that should be used in this specific menu point, but not on the one2many field.

Cedric-Pigeon:
Hello,

I suggest a patch in #195

PR #195, opened 2015-11-12
[8.0] Product By Supplier issue

Cedric-Pigeon:
Product By Supplier module add product_tmpl_id field (mandatory) on product_supplier_info form.

This can be disturbing when you want to create a product and specify the product supplier informations in the same transaction. The field added is mandatory and you are not able to fill it because the product does not exist yet.

In the same time, if you use the menu Product By supplier to create some product supplier informations, you must mention the related product.

The solution implemented by this fix:

Identify when you open the product supplier info form from the product form and drop the field from the view.

### DIFF
--- a/product_by_supplier/views/product_supplierinfo_view.xml
+++ b/product_by_supplier/views/product_supplierinfo_view.xml
@@ -1,6 +1,17 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <openerp>
     <data>
+        <record id="product_template_form_view" model="ir.ui.view">
+            <field name="name">product.template.common.form</field>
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Why don't you pass a context value in seller_ids field inside the product form, and change visibility and required attributes depending on this value? I see this solution cleaner and straight-forward

Cedric-Pigeon:
@pedrobaeza So simple... I didn't think about it. Let's do it ;-)

laetitia-gangloff:
:+1: functionnal

pedrobaeza:
Thanks!

:+1:

PR #196, opened 2015-11-18
[FIX][product_supplierinfo_discount] Fail when same products has mult…

andrius-preimantas:
…iple records of same supplier

Odoo does not restrict entering mulltiple product.supplierinfo records
for the same product and same partner.

### DIFF
--- a/product_supplierinfo_discount/models/purchase_order.py
+++ b/product_supplierinfo_discount/models/purchase_order.py
@@ -51,7 +51,10 @@ class PurchaseOrderLine(models.Model):
              ('name', 'child_of', partner.commercial_partner_id.id)])
         if not sinfos:
             return res
-        seller_uom = sinfos.product_uom.id or False
+        if len(sinfos) > 1:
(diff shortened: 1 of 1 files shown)


product_supplierinfo_discount/models/purchase_order.py
```diff
@@ -51,7 +51,10 @@ def onchange_product_id(
              ('name', 'child_of', partner.commercial_partner_id.id)])
         if not sinfos:
             return res
-        seller_uom = sinfos.product_uom.id or False
+        if len(sinfos) > 1:
```
pedrobaeza: Please change this to `seller_uom = sinfos[:1].product_uom.id`. It makes the same and it doesn't fail on empty values (not on v8).

andrius-preimantas: Cool, thanks!

max3903:
:+1:

pedrobaeza:
:+1:

PR #198, opened 2015-12-04
[ADD] product_supplierinfo_discount: computed field for price with discount

acysos:
Hello,

Add a new computed field that show the price with discount.

Greetings,
Ignacio Ibeas

### DIFF
--- a/product_supplierinfo_discount/README.rst
+++ b/product_supplierinfo_discount/README.rst
@@ -47,6 +47,7 @@ Contributors
 ------------
 
 * Pedro M. Baeza <pedro.baeza@serviciosbaeza.com>
+* Ignacio Ibeas <ignacio@acysos.com>
 
 Maintainer
 ----------
--- a/product_supplierinfo_discount/__openerp__.py
+++ b/product_supplierinfo_discount/__openerp__.py
@@ -29,6 +29,9 @@
     "category": "Purchase Management",
     "website": "www.serviciosbaeza.com",
     "license": "AGPL-3",
+    "contributors": [
+        "Ignacio Ibeas <ignacio@acysos.com>",
+    ],
     "depends": [
         'product',
         'purchase_discount',
--- a/product_supplierinfo_discount/__openerp__.py
+++ b/product_supplierinfo_discount/__openerp__.py
@@ -22,9 +22,10 @@
 
 {
     "name": "Discounts in product supplier info",
-    "version": "8.0.1.0.0",
+    "version": "8.0.2.0.0",
     "author": "Serv. Tecnol. Avanzados - Pedro M. Baeza, "
               "Antiun Ingeniería S.L., "
+              "Acysos S.L., "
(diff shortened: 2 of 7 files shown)


pedrobaeza:
Maybe you can provide the inverse method for putting the price with discount and let the system calculate the discount.

pedrobaeza:
Remember to bump the module version number

acysos:
The field it's not thinking like you say, but it's a good suggestion. This change preserve the original requirement and improve it.

But there is a problem, if you change the "discount", what price use? "price" or "price_with_disc". I have decided use "price". It's more logical put the price and the discount.

Greetings

product_supplierinfo_discount/README.rst
```diff
@@ -47,6 +47,7 @@ Contributors
 ------------
```
elicoidal, line 47: "This module allows to set up a discount in the supplier info form, and propagate
 it to the purchase orders, with two advantages over pricelists:
- The discount is directly added in the purchase orders instead of modifying the
  unit price.
- Prices and discounts can be set on the same screen."

product_supplierinfo_discount/README.rst
```diff
@@ -47,6 +47,7 @@ Contributors
...
```
elicoidal, line 48: "This module requires _purchase_discount_ module, also available in
 this repository."

product_supplierinfo_discount/README.rst
```diff
@@ -47,6 +47,7 @@ Contributors
...
 
 * Pedro M. Baeza <pedro.baeza@serviciosbaeza.com>
```
elicoidal, line 49: To access prices and discounts in the supplier info view, you have to enable the
 option "Manage pricelist per supplier" in _Configuration > Purchases_

product_supplierinfo_discount/README.rst
```diff
@@ -47,6 +47,7 @@ Contributors
...
+* Ignacio Ibeas <ignacio@acysos.com>
```
elicoidal, line 50: 1.Go to Purchase > Products, open one product,
2. edit or add a record on the _Suppliers_ section of the _Procurements_. 
3. In the prices section, a new column called _Discount (%)_ has been 
added where you can input  the desired discount for that quantity.
4.  When you make a purchase order for that supplier and that product, the discount
 will be automatically set.

product_supplierinfo_discount/README.rst
```diff
@@ -47,6 +47,7 @@ Contributors
...
```
elicoidal, line 51: The discount always applies whatever the setup of your pricelist ("Supplier Prices on the product form", etc.).

(NB: dont make a list of one item)

pedrobaeza:
@acysos, are you going to work on fixing this PR or should we close it?

StefanRijnhart:
Closing after 1,5 years without a response.

PR #200, opened 2015-12-13
Fix 9.0 version

sbidoul:


pedrobaeza:
This module hasn't been migrated to 9.0, so you should put installable=False flag instead of changing version number. It can be a bug in my migration script, that doesn't detect the missing key (or it was run before that patch).

sbidoul:
installable set to False

Note to self: do not attempt to cleanup other's houses on Sundays.

pedrobaeza:
:+1: Fast-tracking for avoiding errors.

PR #201, opened 2015-12-15
[FIX] reference to unknow partner field in purchase report;

legalsylvain:
this report use partner_address_id on purchase_order. This field doesn't exist if we install only the module 'purchase_discount'.


pedrobaeza (approved):
Very old PR and version, but it deserves to be merged as simple.

Travis error unrelated.

PR #203, opened 2016-01-12
8.0 add purchase secondary uom

scigghia:
This module add a secondary unit of measure (purchase unit of measure) on product and on purchase order line.
An user can define a particular unit of measure for particular product with a different measure category of default uom.
This module introduce a feature we can find on sale order, too.


pedrobaeza:
Hi, have you checked https://github.com/odoomrp/odoomrp-wip/tree/8.0/purchase_secondary_unit that makes the same thing? Maybe you can get some ideas or we can merge the code from both.

PR #210, opened 2016-02-12
FIX purchase_order_revision

eLBati:
1.create purchase order ..maybe: PO00001
2.confirm purchase order
3. cancle this purchase order .name: PO00001
4.new revision ....then create new purhcase order name i s PO00001-1
5.cancel this purchase order name PO00001-1
6.more->duplicate
7.new purchase order create...name is PO00002
8. confirm this order PO00002
9.cancel this order PO00002
10. new revision.....then error

Order Reference must be unique per Company

### DIFF
--- a/purchase_order_revision/model/purchase_order.py
+++ b/purchase_order_revision/model/purchase_order.py
@@ -38,7 +38,7 @@ class PurchaseOrder(models.Model):
     revision_number = fields.Integer('Revision',
                                      copy=False)
     unrevisioned_name = fields.Char('Order Reference',
-                                    copy=True,
+                                    copy=False,
                                     readonly=True)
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Why the `copy=` attribute is not working? It would be great to have a test to cover this use case.

eLBati:
@pedrobaeza because `unrevisioned_name` must be copied only when a new revision is created, not in the standard copy.

Added the test case

pedrobaeza:
:+1: 

Thanks!

omar7r:
:+1:

jaronemo:
Dears:
Updated Code.
It has completed testing on a new database and production database.
Thank you very much help!

PR #214, opened 2016-03-07
[MIG] purchase_discount: from 8.0 to 9.0

JonathanNEMRY:
- Fix xml_id as Odoo removed the referenced one (purchase.purchase_order_line_form --> purchase.purchase_order_line_form2)
- Add account_invoice_line.py: datas of the invoice are managed from the invoice/invoice_line with the onchange of purchase_order_id and no more from the source model of the invoice creation
- Pass _calc_line_base_price into api.multi as Odoo is now into new api
- Redefine _compute_amount to add a depends on new column 'discount'
- Remove  StockMove there is no more a _get_invoice_line_vals and invoice creation process from the stock.move
- Fix and improve tests
- Add fr translations
- OCA conventions update

PR #215, opened 2016-03-08
[CHG] Update Travis:

JonathanNEMRY:
- Removing all the EXCLUDE/INCLUDE lines that 8.0 branch has,
  because here that modules are not yet installable, and thus they give error.
  Only need to let 4 lines: Pylint, Transifex, OCB and Odoo (without EXCLUDE/INCLUDE).
- Use same format as other OCA repositories

This PR results from https://github.com/OCA/purchase-workflow/pull/214#issuecomment-193330319

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,7 @@
+language: python
+
+python:
+  - "2.7"
 sudo: false
 cache: pip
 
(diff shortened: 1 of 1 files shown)


JonathanNEMRY:
cc @pedrobaeza

pedrobaeza:
:+1: Thanks.

I can fast-track this for rebasing the other PR.

PR #217, opened 2016-03-14
[IMP] purchase_discount: load price with discount in stock_move when confirm a purchase order

agaldona:
Added new functionality, product unit price in stock move is taken with discount in order to correct the cost price calculus when it's method is average.

### DIFF
--- a/purchase_discount/models/purchase_order.py
+++ b/purchase_discount/models/purchase_order.py
@@ -48,3 +48,13 @@ class PurchaseOrder(models.Model):
             account_id, order_line)
         result['discount'] = order_line.discount or 0.0
         return result
+
+    @api.model
+    def _prepare_order_line_move(self, order, order_line, picking_id,
+                                 group_id):
+        res = super(PurchaseOrder, self)._prepare_order_line_move(
+            order, order_line, picking_id, group_id)
+        for vals in res:
+            vals['price_unit'] = (vals.get('price_unit', 0.0) *
+                                  (1 - (order_line.discount / 100)))
+        return res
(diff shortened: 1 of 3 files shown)


purchase_discount/models/purchase_order.py
```diff
@@ -48,3 +48,13 @@ def _prepare_inv_line(self, account_id, order_line):
...
         result['discount'] = order_line.discount or 0.0
         return result
+
+    @api.model
+    def _prepare_order_line_move(self, order, order_line, picking_id,
+                                 group_id):
+        res = super(PurchaseOrder, self)._prepare_order_line_move(
+            order, order_line, picking_id, group_id)
+        for vals in res:
+            vals['price_unit'] = (vals.get('price_unit', 0.0) *
```
pedrobaeza, line 58: Use directly `order_line.price_subtotal`

pedrobaeza:
My doubt is what happens with this for dropshipping sales orders. I think the price is not correctly set. The sequence is:
- Create a sales order with dropshipping products (with supplierinfo added).
- Confirm the sales order
- A purchase order is generated
- When you confirm the purchase order, both delivery order and reception pickings are created from the same purchase order. The delivery order stock.move should have the sales prices (with or without discount, as defined in sales), not the purchase one.

Can you check it?

purchase_discount/models/purchase_order.py
agaldona, line 58: `price_subtotal` is calculated by quantity and this is a unit price

pedrobaeza, line 58: OK, you're right

agaldona:
When a dropshipping is done both delivery order and reception picking are one, there is only one picking created.
In that case the unit price must be taken from sale order, that means that the cost price will be calculated with the sale price.

pedrobaeza:
Yeah, that's right. Thanks for seeing it.

:+1:

PR #221, opened 2016-04-05
update purchase_order.py line 69

mohamedhabibchallouf:
update onchange_product_id  standard_price when  pl_pinfo.min_quantity > qty_in_product_uom

### DIFF
--- a/product_supplierinfo_discount/models/purchase_order.py
+++ b/product_supplierinfo_discount/models/purchase_order.py
@@ -63,5 +63,6 @@ class PurchaseOrderLine(models.Model):
             if pl_pinfo.min_quantity <= qty_in_product_uom:
                 res['value']['discount'] = pl_pinfo.discount
             else:
-                break
+                res['value']['discount'] = 0.0
+                res['value']['price_unit'] = product.standard_price
         return res


product_supplierinfo_discount/models/purchase_order.py
```diff
@@ -63,5 +63,6 @@ def onchange_product_id(
             if pl_pinfo.min_quantity <= qty_in_product_uom:
                 res['value']['discount'] = pl_pinfo.discount
             else:
-                break
+                res['value']['discount'] = 0.0
+                res['value']['price_unit'] = product.standard_price
```
yvaucher, line 67: Correct me if I'm wrong but this seems to break the logic of finding the biggest min_quantity.

Price list info are searched with `order="min_quantity"` and normally in res['value']['discount'] we want the highest discount possible.

What you might want to do is find if no discount has been discovered. For that, you probably want to filter the search with a domain like `[('suppinfo_id', 'in', sinfo.ids), ('min_quantity', '<=', qty_in_product_uom)]` and if this search return you nothing, in that case you want to fall back to your `discount = 0.0` and you standard_price

yvaucher, line 67: You could even limit the search with a `limit=1` using the right order by

pedrobaeza:
I agree that this is not the intended, so I close.

PR #222, opened 2016-04-07
[ADD] module to manage analytic on purchase order

laetitia-gangloff:
This module allow user to set analytic account on purchase order instead of purchase order line.
When analytic account is set on purchase order, all lines of this purchase order will be on this analytic account.

### DIFF
--- a/purchase_analytic/README.rst
+++ b/purchase_analytic/README.rst
@@ -1,0 +1,76 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=================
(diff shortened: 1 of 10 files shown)


pedrobaeza:
This modules should go to OCA/account-analytic, as we have a transversal repo for all the analytic stuff.

laetitia-gangloff:
@pedrobaeza  done here : https://github.com/OCA/account-analytic/pull/55

yvaucher:
@pedrobaeza We already have `purchase_analytic_global` in 7.0 in this repo. In fact it is one of those modules which deserve 2 categories instead of one ...
On one hand we want it here with purchase stuff, on the other hand we want it in analytics as we had a department repository. (which we finally removed).
Maybe we want middle repositories in such case ? I don't like moving stuff around. We loose trace of it.

PR #228, opened 2016-05-10
[9.0] [port] purchase analytic global

yvaucher:
From 7.0


purchase_analytic_global/views/purchase.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+  <data noupdate="0">
```
cyrilgdn: Can be replaced by `<odoo>` (or at least remove noupdate="0")

cyrilgdn:
:+1:

guewen:
@yvaucher wouldn't it be better to have a computed field with:
- `compute`: if all the lines share the same account, return this account, else it is empty
- `inverse`: write the account in all the lines

This way, the field will never show a value totally unrelated to the lines

guewen:
:+1:

grindtildeath (approved):
:+1:

purchase_analytic_global/__openerp__.py
```diff
@@ -0,0 +1,19 @@
...
+{'name': 'Purchase - Analytic Account Global',
+ 'version': '9.0.1.0.0',
+ 'author': 'Camptocamp, Odoo Community Association (OCA)',
+ 'maintainer': 'Camptocamp',
+ 'license': 'AGPL-3',
+ 'category': 'Purchase Management',
+ 'complexity': "easy",
+ 'depends': ['purchase',
+             ],
+ 'website': 'http://www.camptocamp.com',
```
nikul-serpentcs: https://github.com/OCA/purchase-workflow

purchase_analytic_global/__openerp__.py
```diff
@@ -0,0 +1,19 @@
...
+ 'data': ['views/purchase.xml',
+          ],
+ 'test': [],
```
nikul-serpentcs: @yvaucher if not any file then remove key

nikul-serpentcs (commented):
netpicking

i-vyshnevska (approved):

yvaucher:
Travis is broken with :disappointed: 
```
createdb: could not connect to database template1: could not connect to server: No such file or directory

	Is the server running locally and accepting

	connections on Unix domain socket "/var/run/postgresql/.s.PGSQL.5432"?
```

PR #230, opened 2016-05-11
[8.0] new module to manage minimum purchase amount for free shipping

Cedric-Pigeon:


moylop260:
Could you add a newline?

``` bash
purchase_free_shipping/__init__.py:1: [W7908(missing-newline-extrafiles), ] views/purchase_view.xml Missing newline
purchase_free_shipping/__init__.py:1: [W7908(missing-newline-extrafiles), ] views/res_partner_view.xml Missing newline
```

More info https://travis-ci.org/OCA/purchase-workflow/jobs/129411970#L426

Cedric-Pigeon:
@moylop260 Done.

moylop260:
Thanks
👍

Cedric-Pigeon:
@JonathanNEMRY :+1: (code review)

gurneyalex:

purchase_free_shipping/models/purchase.py
```diff
@@ -0,0 +1,14 @@
...
+from openerp import models, fields
+
+
+class PurchaseOrder(models.Model):
+    _inherit = 'purchase.order'
+
+    free_shipping_amount = fields.Float(
+        string='Free Shipping Minimum Amount',
+        related_sudo=False,
+        related='partner_id.free_shipping_amount',
```
legalsylvain, line 13: Sorry if my question is not relevant, but if we change the amount on partner, it will change the value on all the old purchase order, doesn't it ? if yes, I think we should consider better an onchange method.

legalsylvain (commented):
This module looks great. However, don't you have the need handle a product (for shipping). We could imagine that if the amount is not enough, the product is added in a new purchase order line, and if the amount is  enough the line is removed. 
Don't you think ?

legalsylvain (changes requested):
Hi This module seems to be a duplicate of an existing module available here.

https://github.com/OCA/purchase-workflow/pull/159/files

I propose to close this PR, as the other module is older. (exist since V7), and the PR is older.

regards

PR #236, opened 2016-06-02
[8.0] purchase_delivery_split_date - improve extensibility

JordiBForgeFlow:
This PR extends the existing module in order to bring in extra extensibility, so that you can define
in other modules new criteria to split incoming shipments.

The ability to flexibly define criteria to split incoming shipments in purchase orders is very important. Other criteria include, for example to split based on a destination location that could be indicated in the purchase order line, in situations where purchasing wants to specify in the PO lines the final destinations of the items that will be received, withing a warehouse (e.g. iMac is for "Office" location, iPhone is for "Maitenance" location). Additional modules will be submitted as separate PR's.


jgrandguillaume:
Dear @jbeficent ,

That's a very good feature ! I just have one question : Why didn't you use the method:

def _picking_assign(self, cr, uid, move_ids, procurement_group, location_from, location_to, context=None):

In my understanding, when stock.moves are "assigned" they try to figure out in which picking they should go. If they found no existing picking that match, it creates one.

So my wonder is, do you think we can use this mechanism instead of coding a split for PO ? 

My idea is to have a generic module that explain the group key to use to find out the proper picking... This would even bring this feature on SO.. And it'll be more generic, for example it can then also work with push rules...

My2cents,

Regards,

Joël

JordiBForgeFlow:
@jgrandguillaume Sorry for the late reply! The problem is that in PO's the picking is manually created by the PO, not by `_picking_assign`. See https://github.com/odoo/odoo/blob/8.0/addons/purchase/purchase.py#L915

In v9 it gets even worse because `_create_stock_moves` requires a picking to be provided: https://github.com/odoo/odoo/blob/9.0/addons/purchase/purchase.py#L353

So, there's not much we can do other than manually splitting.

JordiBForgeFlow:
cc @aheficent. Can you review?

JordiBForgeFlow:
@darshan-serpent can you review?

JordiBForgeFlow:
@clonedagain what do you think about this?

serpentcs-dev1:
LGTM :+1:

LoisRForgeFlow (approved):
code review :+1: 

Runbot was not working so forced rebuild.

AaronHForgeFlow:
Should it create as many pickings as different scheduled dates?

image

andhit-r:
I encounter same problem with @aheficent . Tried to test it on 8.0 runbot. Same result. So i guest the problem not in this PR. I will investigate further.

MiquelRForgeFlow:
Hi @andhit-r , have you found the problem?

PR #237, opened 2016-06-02
[ADD] [8.0] purchase_location_by_line

JordiBForgeFlow:
# 
# Purchase Location by Line

Purchase Order Lines.


JordiBForgeFlow:
It is related to https://github.com/OCA/purchase-workflow/pull/236

JordiBForgeFlow:
cc @jgrandguillaume We discussed about this in Odoo Experience. Your review is welcome!

PR #242, opened 2016-07-01
[FIX] add read access ACLs to request user and request manager on

JonathanNEMRY:
- purchase.order and purchase.order.line
- purchase.requisition and purchase.requisition.line

That is made to avoid access error to purchase request and purchase request line on users that are request user only or request manager only

see:

selection_027


JordiBForgeFlow:
:+1: Thanks for spotting.

pedrobaeza:
:+1:

JonathanNEMRY:
thanks :smile:

PR #247, opened 2016-07-26
[8.0] [purchase_request_to_rfq] - Fix in create RFQ

JordiBForgeFlow:
This PR contains the following fixes / improvements:
- When the RFQ was created from a purchase request line, if the product has a minimum defined in the product.supplierinfo, a RFQ line was created with quantity ordered = minimum. If a second request line was later added, and the sum of the request lines was < RFQ line quantity, no link was established between the RFQ line and the new request line.
- Tests are added to include the scenario of creating an RFQ for a purchase request that contains a product that has minimums.


JordiBForgeFlow:
cc @JonathanNEMRY

JonathanNEMRY:
Thanks for beeing notified! I l 'll test it and review it
cc @chrisdec

andhit-r (approved):
:+1:

AaronHForgeFlow (approved):
Code Review + Tested as user. :+1: 

Anyway as a user would be a bad practice to create to Purchase Request Line for the same product and the same supplier, wouldn't it?

PR #250, opened 2016-08-01
8.0 fix remove and ignore en translation from transbot jne

JonathanNEMRY:
- Remove added en.po from all the repository
- Add en.po into the gitignore to avoid future addition from transbot


pedrobaeza:
This doesn't work I think, as the addition is made directly. You should change https://github.com/OCA/maintainer-tools/blob/master/tools/tx_pull.py instead. But I'm seeing I'm already take that into account: https://github.com/OCA/maintainer-tools/blob/master/tools/tx_pull.py#L229, so this shouldn't be working.

JonathanNEMRY:
:disappointed: 
Thanks for your explaination @pedrobaeza  ....then what do you suggest to avoid this en.po?

pedrobaeza:
Explore what is reason behind the bypass is not working (maybe the language code provided is not `en`, but another...). You can do that debugging the script on your local.

sbidoul:
I suggest we merge this to see if the en.po file reappears over the week-end, because by reading tx_pull.py I really don't understand how it could create en.po...

pedrobaeza:
We should try with another repository (a fork for example) instead of this one, as the diff can be very large.

sbidoul:
The version of tx_pull that is used by the transbot is out of date.

pedrobaeza:
That should be the reason then. Who can switch it? @gurneyalex or @max3903?

sbidoul:
I'll do it today. 

Then we'll need a mass commit to remove all those en.po...

pedrobaeza:
I'm afraid so :cry:

gurneyalex:
@sbidoul do you need help with the mass commit?

sbidoul:
I updated `tx_pull` on transbot. @gurneyalex I created a new virtualenv and kept the other one as `env.old`, so it's easy to revert.

I tested on `l10n_belgium`, it looks ok.

sbidoul:
I ran the mass commit script on branches 7.0, 8.0, 9.0 for l10n-belgium, l10n-luxemburg and purchase-workflow. I'll launch the rest next week.

JonathanNEMRY:
Great!

PR #251, opened 2016-08-04
[ADD][9.0] Module 'product_supplierinfo_editable'

JordiBForgeFlow:
# 
# Product Supplerinfo Editable

This module provides the framework to define if a Supplier Pricelist is
editable. This module used as standalone will not provide any visual effect.
# Configuration

Other modules should implement method `_compute_is_editable` of model
`product.supplierinfo` introduced here to define when is the pricelist not
editable.


elicoidal:
Not sure about the usage but :+1:

PR #256, opened 2016-08-10
[FIX] purchase_partial_invoicing rst syntax errors

sbidoul:


sbidoul:
Hm. The rst error was red so I thought it was the one breaking the lint check. But it's not.

@moylop260 do you know which one is the real error?

pedrobaeza:
I think is this one:

```
purchase_picking_state/__openerp__.py:4: [C8101(manifest-required-author), ] Missing author required "Odoo Community Association (OCA)" in manifest file
```

moylop260:
sorry for delay... verifying...
(Note: We need this information from MQT, errors as real errors... I'll work on this change...)

moylop260:
I re-bulid travis and now is green

moylop260:
Merging since that is a small RST syntax fix (without changes of logic)
Thanks

PR #189, opened 2015-09-22
[ADD] 8.0 purchase request modules

JordiBForgeFlow:
This PR adds a new entity 'Purchase Request'.

The following modules are added:

purchase_request
purchase_request_to_requisition
purchase_request_to_rfq
# Purchase Request

You use this module if you wish to give notification of requirements of
materials and/or external services and keep track of such requirements.

Requests can be created either directly or indirectly. "Directly" means that
someone from the requesting department enters a purchase request manually.

The person creating the requisition determines what and how much to order,
and the requested date.

"Indirectly" means that the purchase request initiated by the application
automatically, for example, from procurement orders.

A purchase request is an instruction to Purchasing to procure a certain
quantity of materials services, so that they are is available at a
certain point in time.

A line of a requisition contains the quantity and requested date of the
material to be supplied or the quantity of the service to be performed. You
can indicate the service specifications if needed.

It is possible to print the purchase request.
# Purchase Request to Calls for Bid

This module introduces the following features:

The possibility to create new Calls for Bids or update existing Bids from Purchase Request Lines.
The possibility to create Purchase Requests on the basis of Procurement Orders.
# Purchase Request to RFQ

This module adds the possibility to create or update Requests for
Quotation (RFQ) from Purchase Request Lines.

PR #259, opened 2016-08-11
[9.0][ADD] purchase request

AaronHForgeFlow:
# Purchase Request

You use this module if you wish to give notification of requirements of materials and/or external services and keep track of such requirements.

This is a port from v8 to v9 The PR to v8 is here: https://github.com/OCA/purchase-workflow/pull/189


AaronHForgeFlow:
cc @jbeficent

alan196:
👍

AaronHForgeFlow:
Thanks a lot @jbeficent

JordiBForgeFlow:
:+1: Functionally tested. 

cc @JonathanNEMRY @sbidoul Can you help me to get this one merged?

sbidoul:
@adrienpeiffer will work on it this week.

JonathanNEMRY:
@jbeficent I've already started to review(reading the source code) this one and I will end this job soon (probably this evening or tomorrow evening) by testing it on runbot too

purchase_request/views/purchase_request_view.xml
```diff
@@ -0,0 +1,423 @@
...
+                                                           class="oe_inline"
+                                                           attrs="{'readonly': [('is_editable','=', False)]}"/>
+                                                    <field name="product_uom_id"
+                                                           groups="product.group_uom"
+                                                           class="oe_inline"
+                                                           attrs="{'readonly': [('is_editable','=', False)]}"/>
+                                                </div>
+                                                <field name="analytic_account_id"
+                                                       groups="analytic.group_analytic_accounting"
+                                                       domain="[('type','not in',('view','template'))]"
```
adrienpeiffer: `type` field no longer exists en 9.0

purchase_request/security/purchase_request.xml
```diff
@@ -0,0 +1,91 @@
...
+    <record id="purchase_request_modify_rule" model="ir.rule">
+        <field name="name">Modify Purchase Request</field>
+        <field name="model_id" ref="model_purchase_request"/>
+        <field name="groups" eval="[(6,0, [ref('group_purchase_request_user')])]"/>
+        <field name="perm_read" eval="True"/>
+        <field name="perm_write" eval="True"/>
+        <field name="perm_create" eval="True"/>
+        <field name="perm_unlink" eval="True"/>
+        <field name="domain_force">[('requested_by','=',user.id)]</field>
+    </record>
```
adrienpeiffer, line 65: Should add a rule that disables this one for Purchase Request Manager.
Currently, managers don't see all the requests

purchase_request/security/purchase_request.xml
```diff
@@ -0,0 +1,91 @@
...
+    <record id="purchase_request_line_modify_rule" model="ir.rule">
+        <field name="name">Modify Purchase Request Line</field>
+        <field name="model_id" ref="model_purchase_request_line"/>
+        <field name="groups" eval="[(6,0, [ref('group_purchase_request_user')])]"/>
+        <field name="perm_read" eval="True"/>
+        <field name="perm_write" eval="True"/>
+        <field name="perm_create" eval="True"/>
+        <field name="perm_unlink" eval="True"/>
+        <field name="domain_force">[('request_id.requested_by','=',user.id)]</field>
+    </record>
```
adrienpeiffer, line 98: Same remark

purchase_request/models/purchase_request.py
```diff
@@ -0,0 +1,251 @@
...
+    def _default_picking_type(self):
+        type_obj = self.env['stock.picking.type']
+        company_id = self.env.context.get('company_id') or \
+            self.env.user.company_id.id
+        types = type_obj.search([('code', '=', 'incoming'),
+                                 ('warehouse_id.company_id', '=', company_id)])
+        if not types:
+            types = type_obj.search([('code', '=', 'incoming'),
+                                     ('warehouse_id', '=', False)])
+        return types[:1]
```
adrienpeiffer, line 45: Is it not better to add a boolean field on stock.picking.type to set the default picking type for purchase request (With a unique constraint) ?

purchase_request/models/purchase_request.py
```diff
@@ -0,0 +1,251 @@
...
+                rec.is_editable = False
+            else:
+                rec.is_editable = True
+
+    @api.multi
+    def _compute_supplier_id(self):
+        for rec in self:
+            if rec.product_id:
+                for product_supplier in rec.product_id.seller_ids:
+                    rec.supplier_id = product_supplier.name
```
adrienpeiffer: Can you clarify the use of this field?

purchase_request/models/purchase_request.py
```diff
@@ -0,0 +1,251 @@
...
+        company_id = self.env['res.company']._company_default_get(self._name)
+        return self.env['res.company'].browse(company_id.id)
+
+    @api.model
+    def _get_default_requested_by(self):
+        return self.env['res.users'].browse(self.env.uid)
+
+    @api.model
+    def _get_default_name(self):
+        return self.env['ir.sequence'].get('purchase.request')
```
adrienpeiffer, line 33: If you generate the sequence as a default value that may involve many sequence gap.
Currently, if you click on create button, a sequence is generated. If you click on discard just after, the create sequence is lost.

purchase_request/models/purchase_request.py
aheficent: I think this is only for filtering the view

purchase_request/models/purchase_request.py
AaronHForgeFlow: So, the sequence in the name should be obtained when creating the record, shouldn't it? Anyway if the PR is deleted the sequence is lost.

purchase_request/security/purchase_request.xml
AaronHForgeFlow: You're right. Needs to be fixed. Thanks a lot @adrienpeiffer

purchase_request/views/purchase_request_view.xml
AaronHForgeFlow: You're right, I'll remove the domain filter

purchase_request/models/purchase_request.py
AaronHForgeFlow: I suggest that we leave it as it is because this is how it's done in Purchase orders and if you want you can override this method.

purchase_request/models/purchase_request.py
AaronHForgeFlow: This is information for the Purchasing Department so they are able to know what is the supplier for that product.

purchase_request/models/purchase_request.py
JordiBForgeFlow: @adrienpeiffer This is the standard approach in other parts of Odoo. We adopted as much as possible the best practices. If a number is lost in the sequence, so be it. Only in invoices it can really a problem to have a gap in the invoice sequence.

In general, as this module has already been merged in 8.0, I suggest not to change the behaviour in 9.0, unless it is really mandatory by the change of design in 9.0.

adrienpeiffer: @aheficent @jbeficent I find it still a little strange but it is not mandatory to change that.  this is a simple remark during browsing this one.

purchase_request/models/purchase_request.py
adrienpeiffer: yes but in your loop, you consider only the last supplier in `seller_ids`. So, I don't really understand this computed field

purchase_request/views/purchase_request_view.xml
```diff
@@ -100,7 +100,6 @@
                                                 </div>
                                                 <field name="analytic_account_id"
                                                        groups="analytic.group_analytic_accounting"
-                                                       domain="[('type','not in',('view','template'))]"
```
adrienpeiffer: I think you should consider `account_type` in this domain

purchase_request/security/purchase_request.xml
```diff
@@ -64,6 +64,16 @@
...
 
+    <record id="purchase_request_modify_manager_rule" model="ir.rule">
+        <field name="name">Purchase Request Manager</field>
+        <field name="model_id" ref="model_purchase_request"/>
+        <field name="groups" eval="[(6,0, [ref('group_purchase_request_manager')])]"/>
+        <field name="perm_read" eval="True"/>
+        <field name="perm_write" eval="True"/>
+        <field name="perm_create" eval="True"/>
+        <field name="perm_unlink" eval="True"/>
+    </record>
```
adrienpeiffer, line 75: Isn't necessary to put `<field name="domain_force">[(1, '=', 1)]</field>` ?

AaronHForgeFlow: For me it's working as it is with no error. I think is not needed

adrienpeiffer: Nice !

purchase_request/models/purchase_request.py
AaronHForgeFlow: You're right, now it's fixed. Thank you!

adrienpeiffer:
:+1: (Code review & functional tests)

purchase_request/demo/purchase_request_demo.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2016 Eficent Business and IT Consulting Services S.L.
+     License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl-3.0) -->
+<openerp>
+    <data noupdate="0">
```
JonathanNEMRY: Most of the time when using noupadate= into an odoo demo file, that's to specify it is noupdate="1" as https://github.com/odoo/odoo/blob/8.0/addons/account/demo/account_minimal.xml#L3

purchase_request/i18n/purchase_request.pot
```diff
@@ -0,0 +1,484 @@
+# Translation of Odoo Server.
+# This file contains the translation of the following modules:
```
JonathanNEMRY: As @pedrobaeza said and as it is currently beeing voted into the contributor mailing list, the .pot are no more necessay so you can remove it without any problems

purchase_request/models/purchase_request.py
```diff
@@ -0,0 +1,251 @@
...
+                       track_visibility='onchange')
+    origin = fields.Char('Source Document', size=32)
+    date_start = fields.Date('Creation date',
+                             help="Date when the user initiated the "
+                                  "request.",
+                             default=fields.Date.context_today,
+                             track_visibility='onchange')
+    requested_by = fields.Many2one('res.users',
+                                   'Requested by',
+                                   required=True,
```
JonathanNEMRY, line 78: I think that we can add an ondelete cascade here too as the field is required what do you think?

sbidoul: @JonathanNEMRY :-1: deleting a user must raise an error, not delete everything the user has done.

JonathanNEMRY: OK

purchase_request/models/purchase_request.py
```diff
@@ -0,0 +1,251 @@
...
+                                 ('warehouse_id.company_id', '=', company_id)])
+        if not types:
+            types = type_obj.search([('code', '=', 'incoming'),
+                                     ('warehouse_id', '=', False)])
+        return types[:1]
+
+    @api.multi
+    @api.depends('name', 'origin', 'date_start',
+                 'requested_by', 'assigned_to', 'description', 'company_id',
+                 'line_ids', 'picking_type_id')
```
JonathanNEMRY: Why so much fields into this depends? Shoudn't it be only the fields `state` as it is the only checked fields into the compute method?

purchase_request/i18n/purchase_request.pot
AaronHForgeFlow: Done. Thank you @JonathanNEMRY

purchase_request/tests/test_purchase_request.py
```diff
@@ -0,0 +1,37 @@
...
+            'product_id': self.env.ref('product.product_product_13').id,
+            'product_uom_id': self.env.ref('product.product_uom_unit').id,
+            'product_qty': 5.0,
+        }
+        self.purchase_request_line.create(vals)
+        purchase_request.button_to_approve()
+        self.assertEqual(
+            purchase_request.state, 'to_approve',
+            'Should be in state to_approve')
+        purchase_request.button_draft()
```
JonathanNEMRY, line 39: maybe add an assertEquals with this field is_editable? to check that it is not editbale anymore with state modified

purchase_request/demo/purchase_request_demo.xml
AaronHForgeFlow: So noupdate="1" will not update the data, will it?

JonathanNEMRY: noupdate="1" into demo datas means that if the users play with datas and you update the implied module then the datas of the users will be safe otherwise reset with the xml content. This could be usefull for demo datas and some others real datas like ir.parameter

purchase_request/tests/test_purchase_request.py
AaronHForgeFlow: Done. Thanks a lot @JonathanNEMRY

purchase_request/models/purchase_request.py
AaronHForgeFlow: Yes. It works if it depends just in the state field

JonathanNEMRY:
Great @aheficent !  Nothing more to say with the code review and I tested it on runbot too  8.0. so :+1:

AaronHForgeFlow:
Thanks a lot @adrienpeiffer  and @JonathanNEMRY. BTW do you know what should I do to pass the codecov checks?

JonathanNEMRY:
No problems @aheficent thank you for your work.
About the codecov I don't think it is relevant in this specific situation as you removed some code (with the state for exemple) and so it pass into a more litle part of the code during its test process but absolutely not test less source code.
Moreover as there are 3 :+1: this one is now ready to be merged

AaronHForgeFlow:
Good to know, thank you for explanation @JonathanNEMRY

PR #260, opened 2016-08-11
 [UPT][8.0]Delete the purchase request line if the procurement is cancelled

AaronHForgeFlow:
When the procurement creates a purchase request and later the procurement it is cancelled, the corresponding purchase request line is removed. If the Purchase request has not lines anymore it is also deleted.

This is done only if the purchase request is in state draft or rejected.

### DIFF
--- a/purchase_request_procurement/models/procurement.py
+++ b/purchase_request_procurement/models/procurement.py
@@ -4,7 +4,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
 from openerp import _, api, fields, models
-from openerp.exceptions import ValidationError
+from openerp.exceptions import Warning as UserError
 
 
 class Procurement(models.Model):
@@ -76,7 +76,7 @@ class Procurement(models.Model):
         result = super(Procurement, self).propagate_cancel(procurement)
         # Remove the reference to the request_id from the procurement order
         request = procurement.request_id
-        procurement.write({'request_id': None})
+        procurement.write({'request_id': False})
         # Search for purchase request lines containing the procurement_id
         request_lines = self.env['purchase.request.line'].\
             search([('procurement_id', '=', procurement.id)])
@@ -84,11 +84,10 @@ class Procurement(models.Model):
         # or reject
         for line in request_lines:
             if line.request_id.state not in ('draft', 'reject'):
-                raise ValidationError(_('Can not cancel this procurement as '
-                                        'the related purchase request is in '
-                                        'progress confirmed already. '
-                                        'Please cancel the purchase request '
-                                        'first.'))
+                raise UserError(_('Can not cancel this procurement as the '
+                                  'related purchase request is in progress '
+                                  'confirmed already. Please cancel the '
+                                  'purchase request first.'))
             else:
                 line.unlink()
         # If the purchase request has not lines, delete it as well
--- a/purchase_request_procurement/models/procurement.py
+++ b/purchase_request_procurement/models/procurement.py
@@ -82,13 +82,15 @@ class Procurement(models.Model):
             search([('procurement_id', '=', procurement.id)])
         # Remove the purchase request lines, if the request is not draft
         # or reject
-        for line in request_lines:
-            if line.request_id.state not in ('draft', 'reject'):
-                raise UserError(_('Can not cancel this procurement as the '
-                                  'related purchase request is in progress '
-                                  'confirmed already. Please cancel the '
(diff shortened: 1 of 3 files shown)


AaronHForgeFlow:
cc @jbeficent

JordiBForgeFlow:
cc @chrisdec @JonathanNEMRY @sbidoul @kittiu Can you provide your feedback?

purchase_request_procurement/models/procurement.py
```diff
@@ -69,3 +70,28 @@ def _run(self, procurement):
...
+        # Remove the reference to the request_id from the procurement order
+        request = procurement.request_id
+        procurement.write({'request_id': None})
+        # Search for purchase request lines containing the procurement_id
+        request_lines = self.env['purchase.request.line'].\
+            search([('procurement_id', '=', procurement.id)])
+        # Remove the purchase request lines, if the request is not draft
+        # or reject
+        for line in request_lines:
+            if line.request_id.state not in ('draft', 'reject'):
```
andhit-r: You should put it before the loop. It does not need to be repeted

purchase_request_procurement/models/procurement.py
```diff
@@ -4,6 +4,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
 from openerp import _, api, fields, models
+from openerp.exceptions import ValidationError
```
andhit-r: use **from openerp.exceptions import Warning as UserError** to comply with CONTRIBUTING.md

purchase_request_procurement/models/procurement.py
```diff
@@ -69,3 +70,28 @@ def _run(self, procurement):
             procurement.message_post(body=_("Purchase Request extended."))
             return True
         return super(Procurement, self)._run(procurement)
+
+    @api.model
+    def propagate_cancel(self, procurement):
+        result = super(Procurement, self).propagate_cancel(procurement)
+        # Remove the reference to the request_id from the procurement order
+        request = procurement.request_id
+        procurement.write({'request_id': None})
```
andhit-r: Use False instead None

purchase_request_procurement/tests/test_purchase_request_procurement.py
```diff
@@ -0,0 +1,72 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 Eficent Business and IT Consulting Services S.L.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp.tests import common
+from openerp import fields
+from openerp.exceptions import ValidationError
```
andhit-r: use **from openerp.exceptions import Warning as UserError** to comply with CONTRIBUTING.md

purchase_request_procurement/tests/test_purchase_request_procurement.py
```diff
@@ -0,0 +1,72 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 Eficent Business and IT Consulting Services S.L.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp.tests import common
```
andhit-r: use _from openerp.tests.common import TransactionCase_ to comply with latest template

andhit-r (changes requested):
Some comments inline

purchase_request_procurement/models/procurement.py
LoisRForgeFlow: Why are you preferring False over None?

purchase_request_procurement/models/procurement.py
JonathanNEMRY: This "if" restriction could also be used into a filtered on request_lines in order to loop only into the wanted records

AaronHForgeFlow: you are right, thank you @andhit-r @JonathanNEMRY

AaronHForgeFlow: There's a request_state field in the line indeed, so no need to access the request

LoisRForgeFlow (approved):
code review, LGTM now :+1:

sbidoul (approved):
(code review)

AaronHForgeFlow:
@andhit-r can you please review? Thanks

lreficent:
This has change significantly in v9, I don't know if this PR is still correct.

LoisRForgeFlow:
This has changed significantly in v9, I don't know if this PR is still correct.

PR #268, opened 2016-09-04
[9.0] [NEW] new purchase_buyer_contact 

legalsylvain:
module to register some info the supplier want to know (customer code, contact)

### DIFF
--- a/purchase_buyer_contact/README.rst
+++ b/purchase_buyer_contact/README.rst
@@ -1,0 +1,19 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Purchase Buyer Contact
+=========================
+
+This modules allow you to define the name, phone and email of the person in charge of purchase in your company.
+You can define this information against the res.company object.
+All res.partner created will then have the same info by default, but you can modify it.
+You can also define against the res.partner the code the supplier uses for you.
+
+Credits
+=======
+
+Contributors
+------------
+
+* Julien Weste
+
--- a/purchase_buyer_contact/models/res_company.py
+++ b/purchase_buyer_contact/models/res_company.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) 2016-Today: La Louve (<http://www.lalouve.net/>)
+# @author: Julien Weste
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp import fields, models
+
+
+class ResCompany(models.Model):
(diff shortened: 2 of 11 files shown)


purchase_buyer_contact/README.rst
```diff
@@ -0,0 +1,19 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Purchase Buyer Contact
+=========================
+
+This modules allow you to define the name, phone and email of the person in charge of purchase in your company.
+You can define this information against the res.company object.
+All res.partner created will then have the same info by default, but you can modify it.
+You can also define against the res.partner the code the supplier uses for you.
```
JordiBForgeFlow, line 10: Do you mean the code that a supplier uses to represent your company? This does not seem to be related to the concept of Purchase Buyer Contact. Can you please move it to another module? I'm also not convinced that this is the right technical solution to define how to enter the name that a supplier uses for your company.

JordiBForgeFlow:
I think that this module intends to solve a real need, but the approach does not seem correct to me.
- If you change one day the buyer in the company, you'll have to update all the partners. 
- If you change buyers in time, you have no history on the PO on who was the buyer.

This is generally approached as follows:
1. Define the buyer in the company
2. Define the buyer in the partner - will be filled on exception only
3. Create a Purchase Order. The buyer is defaulted from the company
4. Enter the supplier on a Purchase Order. The buyer is defaulted from the supplier, if one is provided.
5. Manually change the buyer in the Purchase Order (optionally)

When the Purchase Order is printer, the buyer details are included.

purchase_buyer_contact/models/res_company.py
```diff
@@ -0,0 +1,19 @@
...
+    _inherit = 'res.company'
+
+    # Column Section
+    buyer_partner_id = fields.Many2one('res.partner')
+    buyer_name = fields.Char(
+        related='buyer_partner_id.name', store=True, readonly="1")
+    buyer_email = fields.Char(
+        related='buyer_partner_id.email', store=True, readonly="1")
+    buyer_phone = fields.Char(
+        related='buyer_partner_id.phone', store=True, readonly="1")
```
JordiBForgeFlow, line 19: Why do you include these stored fields? It's not advisable to do so. The user can just click on the partner to view the contact details.

jweste: @jbeficent : thanks for your review! I'll be taking this into account as soon as possible.

susport (commented):
This module depend of purchase. View error in runbot, not find the purchase.group_purchase_manager

PR #273, opened 2016-09-23
[ADD] Migrated purchase_order_line_sequence to v9.

SerpentCS:


pedrobaeza:
Please don't use plurals in module name, so I think we can rename it to `purchase_order_line_reorder` or `purchase_order_line_sequence`

purchase_order_line_sequence/models/invoice.py
```diff
@@ -0,0 +1,17 @@
...
+# © 2015 Serpent Consulting Services Pvt. Ltd. - Sudhir Arya
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
+
+from openerp import fields, models
+
+
+class AccountInvoiceLine(models.Model):
+    _inherit = 'account.invoice.line'
+
+    sequence = fields.Integer(related='purchase_line_id.sequence', store=True)
```
JordiBForgeFlow: The field `sequence` was dropped in v9. IMHO it's better to add a module 'account_invoice_line_sequence' that puts it back. That way we achieve better modularity.

MiquelRForgeFlow:
In the repository, you should add <code>account-invoicing</code> module in the <code>oca_dependencies.txt</code> file. This will allow runbot install the dependencies of your PR.

AaronHForgeFlow:
This is waiting for that: https://github.com/OCA/stock-logistics-workflow/pull/262

AaronHForgeFlow:
Hi @darshan-serpent Can you rebase this branch? It's giving conflicts when merging. Besides, dependency is already merged: https://github.com/OCA/stock-logistics-workflow/pull/385

serpentcs-dev1:
I have tried a lot of times to rebase the Serpent/9.0 branch but it has a lot of other commits which are not even to the OCA/9.0. 

As you see the commits are made directly to the 9.0 branch which is not the good branching habit adapted. So that's the issue.

Maybe then I'll try to create a new branch/PR itself.

AaronHForgeFlow:
@darshan-serpent I did the thing. I created a new PR #485 Better close this PR then

PR #277, opened 2016-10-01
[FIX] add store=True

fmdl:
You use api.depends to update the status of purchase picking, with that you can use store='true'. Like that you cangroup_by picking_status in tree view.

### DIFF
--- a/purchase_picking_state/purchase.py
+++ b/purchase_picking_state/purchase.py
@@ -39,6 +39,7 @@ class PurchaseOrder(models.Model):
 
     picking_state = fields.Selection(
         string="Picking status", readonly=True,
+        store=True,
         compute='_compute_picking_state',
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

sla-subteno-it:
👍

andhit-r (approved):
:+1:

fmdl:
this commit generate this error, when you change the state of move or picking, and system is very slow.

if you remove  picking_ids.state in  @api.depends('picking_ids', 'picking_ids.state'), the error disappear, but the compute don't works.

ERROR openerp.osv.expression: Non-stored field purchase.order.picking_ids cannot be searched.

pedrobaeza:
Please propose a new PR for amending this.

PR #280, opened 2016-10-10
10.0 Port purchase_fiscal_position_update

alexis-via:


angelmoya:
I did similar work to mig to 9.0, you can see it on #281

purchase_fiscal_position_update/README.rst
```diff
@@ -0,0 +1,56 @@
...
+
+No specific configuration needed. This module uses the standard
+configuration of the fiscal positions.
+
+Usage
+=====
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/purchase-workflow/10.0
```
angelmoya: You need to set repo_id:
:target: https://runbot.odoo-community.org/runbot/142/10.0

purchase_fiscal_position_update/models/purchase.py
```diff
@@ -0,0 +1,23 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models, api
+
+
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+
+    @api.onchange('fiscal_position_id')
+    def fiscal_position_change(self):
```
angelmoya, line 14: you can call to onchange_product_id, I solved on version 9.0 on two lines

```
    for line in self.order_line:
        line.onchange_product_id()
```

alexis-via: For me, it's a mistake to call onchange_product_id because it resets the price/description/... to the default value etc... And that's quite heavy too !

pedrobaeza: For me it's not, because any modification in the onchange code via inherited modules is not reflected here, so you will need glue modules. It's also not too heavy working on cached recordset. But it's needed a change in the approach (it seems awkward, but it's the way to go to keep the best compatibility, and it's fast):

```
for line in self.order_line:
    temp_line = line.new(line._convert_to_write(line._cache))
    temp_line.onchange_product_id()
    line.taxes_id = temp_line.taxes_id
```

alexis-via: @pedrobaeza Your code proposal is better. @angelmoya 's code was not right. If an inherited module want to change something, he should do it in the map_tax() method of account.fiscal.position, not here. And by the way, your proposal is 4 lines of code ; mine is exactly the same, so it's not simpler. It just seems crazy for me to call onchange_product_id() which itself calls _suggest_quantity() (which reads supplierinfo !) and _onchange_quantity() (which calls 3 other methods !)

angelmoya: OK @alexis-via use cache is not so easy, so I copy your method
@pedrobaeza your code is not working because previous fiscal position is on cache

pedrobaeza (approved):

PR #281, opened 2016-10-11
9.0 mig purchase fiscal position update

angelmoya:


purchase_fiscal_position_update/models/purchase.py
```diff
@@ -0,0 +1,16 @@
...
+
+from openerp import models, api, _
+
+
+class purchase_order(models.Model):
+    _inherit = "purchase.order"
+
+    @api.onchange('fiscal_position_id')
+    def onchange_fiscal_position_id(self):
+        for line in self.order_line:
```
pedrobaeza, line 18: Do:

```
for line in self.order_line:
    temp_line = line.new(line._convert_to_write(line._cache))
    temp_line.onchange_product_id()
    line.taxes_id = temp_line.taxes_id
```

pedrobaeza (changes requested):
Change in the approach (only drafted - to be tested)

pedrobaeza:
Check also Travis status

purchase_fiscal_position_update/models/purchase.py
angelmoya: This way is not working, and using cache is not direct, so it's more clear do like @alexis-via on #280

purchase_fiscal_position_update/models/purchase.py
```diff
@@ -0,0 +1,24 @@
...
+
+from openerp import models, api, _
+
+
+class purchase_order(models.Model):
+    _inherit = "purchase.order"
+
+    @api.onchange('fiscal_position_id')
+    def fiscal_position_change(self):
+        '''Function executed by the on_change on the fiscal_position_id field
```
pedrobaeza: Double quote instead of simple for docstring according PEP257

pedrobaeza (changes requested):
Include tests

purchase_fiscal_position_update/models/purchase.py
angelmoya: Fixed

angelmoya:
Tests added

pedrobaeza:
Travis is still red

pedrobaeza (approved):

pedrobaeza:
As this one is the same as 10.0 version and it has another author, I merge both.

PR #284, opened 2016-10-14
[IMP][8.0][purchase_delivery_split_date] Fix lint error W7936

andhit-r:


pedrobaeza (approved):
Fast-tracking as obvious patch

PR #291, opened 2016-10-21
[9.0] [purchase_request_to_rfq] - Fixes error that made it impossible to create request lines

JordiBForgeFlow:
cc @aheficent @JonathanNEMRY @adrienpeiffer

### DIFF
--- a/purchase_request_to_rfq/models/purchase_request.py
+++ b/purchase_request_to_rfq/models/purchase_request.py
@@ -18,8 +18,8 @@ class PurchaseRequestLine(models.Model):
     @api.depends('purchase_lines')
     def _compute_is_editable(self):
         super(PurchaseRequestLine, self)._compute_is_editable()
-        self.filtered(lambda p: p in self and p.purchase_lines).write(
-            {'is_editable': False})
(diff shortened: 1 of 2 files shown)


Garamotte:
Using the create/write/unlink methods in computed fields is a bad practice, as these methods should not modify values in the database (these methods are called as onchanges, when the computed field is displayed in a view).

You should use a for loop instead (either using the `update` method, or a direct affectation).

JordiBForgeFlow:
Thanks @sylvain-garancher !

adrienpeiffer (approved):
:+1:

JonathanNEMRY:
:+1:

JonathanNEMRY (approved):

aheficent:
I also did not realize of this bug before. With this fix it's working fine. :+1:  Tested on runbot + code review

AaronHForgeFlow (approved):
I also did not realize of this bug before. With this fix it's working fine. :+1: Tested on runbot + code review

Garamotte (approved):
Thanks.

@jbeficent I just created a PR on your own branch, to add a unit test for this bug.

PR #294, opened 2016-10-27
[8.0][purchase_request][FIX] Description of the purchase request line is not passed to PO line

JordiBForgeFlow:
What this PR does: When a po line is to be created, first look for an existing po line that also matches in name. Otherwise create a new line, with name to be the same as the purchase request line.


andhit-r (approved):
:+1:

JordiBForgeFlow:
cc @aheficent @darshan-serpent can you review?

AaronHForgeFlow (approved):
Tested on runbot :+1:

pedrobaeza (approved):

PR #295, opened 2016-10-28
[FIX] Fix renamed attribute

Garamotte:
The `digits_compute` attribute has been renamed to `digits`.

Using the old name causes a traceback with recent Odoo code.
This is probably a bug in Odoo, but the best is to use the new name.


pedrobaeza (approved):

adrienpeiffer (approved):
:+1: Thanks Sylvain

AaronHForgeFlow (approved):
:+1: Thanks @sylvain-garancher @adrienpeiffer

PR #296, opened 2016-11-02
[8.0][purchase_request][IMP] Creating rfq without selecting product

levkar:
This fix enables creation of RFQ without product. This is useful when product is created right before confirming order instead of creating before RFQ.

### DIFF
--- a/purchase_request_to_rfq/wizard/purchase_request_line_make_purchase_order.py
+++ b/purchase_request_to_rfq/wizard/purchase_request_line_make_purchase_order.py
@@ -92,19 +92,27 @@ class PurchaseRequestLineMakePurchaseOrder(models.TransientModel):
         po_line_obj = self.env['purchase.order.line']
         product_uom = self.env['product.uom']
         product = item.product_id
-        default_uom_po_id = product.uom_po_id.id
-        qty = product_uom._compute_qty(item.product_uom_id.id,
-                                       item.product_qty,
-                                       default_uom_po_id)
-        supplier_pricelist = \
-            po.partner_id.property_product_pricelist_purchase \
-            and po.partner_id.property_product_pricelist_purchase.id or False
-        vals = po_line_obj.onchange_product_id(
-            supplier_pricelist, product.id, qty, default_uom_po_id,
-            po.partner_id.id, date_order=False,
-            fiscal_position_id=po.partner_id.property_account_position.id,
-            date_planned=item.line_id.date_required,
-            name=False, price_unit=False, state='draft')['value']
+        if product:
+            default_uom_po_id = product.uom_po_id.id
+            qty = product_uom._compute_qty(item.product_uom_id.id,
+                                           item.product_qty,
+                                           default_uom_po_id)
+            supplier_pricelist = \
+                po.partner_id.property_product_pricelist_purchase \
+                and po.partner_id.property_product_pricelist_purchase.id or False
+            vals = po_line_obj.onchange_product_id(
+                supplier_pricelist, product.id, qty, default_uom_po_id,
+                po.partner_id.id, date_order=False,
+                fiscal_position_id=po.partner_id.property_account_position.id,
+                date_planned=item.line_id.date_required,
+                name=False, price_unit=False, state='draft')['value']
+        else:
+            vals = {
+                'name': item.name,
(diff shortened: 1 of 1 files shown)


AaronHForgeFlow:
This would be in the same line as creating RFQ manually without order_lines. It is possible to do that. However, in my opinion the product should be a required field. It makes no sense to create a PO if I don't know what I'm going to buy.

levkar:
This depends on when and who creates the products. You assume purchase
request manager is doing it or the product is already known and created.

Assume a pencil is requested and company needs exact pencil code to create
it and it will be learnt after rfq. Rfq without a product is standard in
odoo.

Let me know if this makes sense so that i will close this pr.

2 Kas 2016 17:00 tarihinde "Aarón Henríquez" notifications@github.com
yazdı:

AaronHForgeFlow:
@levkar Don't close it, it's just one opinion. What you said makes sense. But how do you know that it's a pencil what are you going to buy?  Do you mean that the PO Manager is the one that goes to the Purchase Request Line and create the RFQ selecting there the product based on the description field?

levkar:
@aheficent 

There are two managers and two users here;
1) Purchase Request User  -> created by purchase_request module
2) Purchase Request Manager  -> created by purchase_request module
3) Purchase User -> created by purchase module
4) Purchase Manager -> created by purchase module

1 creates purchase request
2 changes purchase request to RFQ 
3 changes RFQ to purchase order
4 approves purchase order

My point is that if i can make 3rd and 4th step without selecting a product, why am i supposed to select a product in 2nd step?

AaronHForgeFlow:
Because you cannot create the RFQ from the wizard if the Purchase Request Manager don't select the product in the wizard that creates the RFQ. Am I wrong?

levkar:
Yes exactly. This PR is all about creating RFQ without selecting product. 

I haven't looked at requisition. That may require the same thing as well.

AaronHForgeFlow:
All right, so I did something wrong because I tested on runbot and I couldn't do what you are describing. 

First, I created a Purchase Request with a line and that has no products. Then, I approved.

Then I go to the line and I see that the description is  in the specifications field. All right.

Then I go to the "Create RFQ" wizard and click on "Create RFQ". Then the system tells me that the following fields are invalid: uom and items

Am I doing something wrong?

levkar:
It works here below I created a RFQ without product.

http://3271443-8-0-48123e.runbot2.odoo-community.org/web?db=3271443-8-0-48123e-all#id=15&
view_type=form&model=purchase.order&menu_id=313&action=362

You need to fill quantity (and uom with this module).

AaronHForgeFlow:
Do you mean that I have to create the RFQ from outside the Purchase Request Line?

levkar:
1- create a purchase request with no product and press request approval
button
2- go to purchase request lines menu and select created purchase request
line
3- from top menu select create rfq
4- select a supplier and press create button

4 Kas 2016 15:37 tarihinde "Aarón Henríquez" notifications@github.com
yazdı:

AaronHForgeFlow:
Hello @levkar That's exactly what I did but I get the following when I click on "Create RFQ": "The following fields are invalid: uom and items"

It's the PR00004 here: http://3271317-296-288096.runbot1.odoo-community.org/web#id=3&view_type=form&model=purchase.request.line&menu_id=332&action=402

levkar:
I couldn't login to your link. Uom becomes required with this pull request.
You can try from the link i sent before.

7 Kas 2016 10:12 tarihinde "Aarón Henríquez" notifications@github.com
yazdı:

AaronHForgeFlow:
I managed to create the RFQ now @levkar  The problem was I was not setting the quantity in the Purchase Request. As the the description is passed from the Purchase Request to the RFQ lines the Purchase Manager would know what product it is. I think this can be useful for some cases.

JordiBForgeFlow:
@levkar In v9 onwards the product_id is a required field in RFQ. See: https://github.com/odoo/odoo/blob/9.0/addons/purchase/purchase.py#L528

IMHO we should not encourage in v8 a behaviour that is discontinued in v9. You can always build your own customization in v8 on top of the standard OCA module.

levkar:
Thanks  @jbeficent 
I will make an extension module for it.

Issue #298, opened 2016-11-07
product_supplierinfo_discount improvment


legalsylvain:
hi, 

I just saw that this great module was ported in V9.0.

I was thinking to add an extra feature on res.partner. A new field named 'global_discount', that could be the default value on the product supplier info, when created.
What do you think about such feature ? 

thanks.

CC @pedrobaeza, @JonathanNEMRY

JonathanNEMRY:
Hello @legalsylvain 
I am not sure about the fields' name but the default value appears to me as a good idea :+1:

legalsylvain:
Ok, thanks for your feedback ! I'll do that. What about :
- supplier_discount
- recurring_discount

regards.

JonathanNEMRY:
`recurring_discount` sounds good to me :smiley:

pedrobaeza:
Better to be called `default_supplierinfo_discount`. I made something similar with a default discount on sales, but making the view totally inheritable: https://github.com/OCA/partner-contact/blob/9.0/partner_default_sale_discount/models/sale_order.py#L18

legalsylvain:
Done in recent version.

PR #301, opened 2016-11-09
[ADD] default_supplierinfo_discount on res.partner model

legalsylvain:
* Add default_supplierinfo_discount field on res.partner.
* this field will be used as the default value, when creating new supplierinfo line for the given supplier.

Misc : Add / improve description.

Do : #298

CC : @JonathanNEMRY, @pedrobaeza. 

Thanks for your review.

**Note** : @pedrobaeza, I tested the code (https://github.com/OCA/partner-contact/blob/9.0/partner_default_sale_discount/models/sale_order.py#L18) but I finally managed it with a simple onchange function. What to you think about that ? It seems to be that it is inheritable too.

reference of your comment : https://github.com/OCA/purchase-workflow/issues/298#issuecomment-259359443


JonathanNEMRY (approved):
Thanks for the work!
Code review :+1: 
Juste one point: could you make a litle test for the `onchange_name` method? As it is `onchange` I will not blocking this PR for it but as we already have the tests folder with existing tests inside it could be easy to improve it. thanks in advance! :smile: 
Jonathan

product_supplierinfo_discount/models/product_supplierinfo.py
```diff
@@ -12,3 +13,11 @@ class ProductSupplierInfo(models.Model):
 
     discount = fields.Float(
         string='Discount (%)', digits_compute=dp.get_precision('Discount'))
+
+    @api.onchange('name')
+    @api.multi
+    def onchange_name(self):
+        for supplierinfo in self:
```
pedrobaeza: Do directly:

```
for supplierinfo in self.filtered('name'):
```

and you save one indentation level.

pedrobaeza (approved):
Little change. This case allows the onchange, as you are not copying previously the value to the product form, so it's good with this indeed.

product_supplierinfo_discount/models/product_supplierinfo.py
legalsylvain: Done.

legalsylvain:
@JonathanNEMRY : test added. thanks for your review.

Note : I squashed the commit to make more easy futur git cherry-pick against OCA / 10 branch.

susport (approved):
👍

JonathanNEMRY (approved):

legalsylvain:
self merge is not a good practice, but it seems that this PR is ready now.

kind regards.

pedrobaeza:
No need for that, although coming from a trusty source like yours, it can be allowed without problems :wink:

PR #304, opened 2016-11-25
[ADD] Purchase request mts mto rule

alan196:
I this PR I created a module to ADD compatibility of purchase_request_procurement with stock_mts_mto_rule module of stock-logistic-warehouse repo.

This override method _prepare_purchase_request_line method to call method get_mto_qty_to_order from https://github.com/OCA/stock-logistics-warehouse/blob/9.0/stock_mts_mto_rule/model/procurement.py#L37-L52 in order to verify stock before creating the PR


alan196:
Runbot fails because of the dependency of stock-logistics-warehouse repo. Somebody can help me to add that dependency to this repo?... or use runbot_travis2docker

purchase_request_mts_mto_rule/models/procurement.py
```diff
@@ -0,0 +1,21 @@
...
+
+from openerp import api, models
+
+
+class ProcurementOrder(models.Model):
+    _inherit = 'procurement.order'
+
+    @api.model
+    def _prepare_purchase_request_line(self, purchase_request, procurement):
+        return {
```
moylop260, line 13: What about use `super` and overwrite just the key required?

purchase_request_mts_mto_rule/models/procurement.py
```diff
@@ -0,0 +1,21 @@
...
+            'product_id': procurement.product_id.id,
+            'name': procurement.product_id.name,
+            'date_required': procurement.date_planned,
+            'product_uom_id': procurement.get_mto_qty_to_order(),
```
moylop260, line 17: @julioserna
What do you think about this change?

PR #306, opened 2016-11-29
Add module procurement_select_route

ghost:


pedrobaeza:
The normal way to handle the scenario you describe is to make routes selectable on sale order lines, and select from there which one you want. It's even easier, as you don't need to switch your view to procurements. Please check this option.

On this, I don't see the meaning for this module.

ghost:
@pedrobaeza I don't agree. The user who creates the sale order is not necessarily the one who decides which route must be used. In the case of my client, the warehouse manager is the one who decides the route to use, but he is not the one who confirms the sale.

pedrobaeza:
OK, you're right. Please add this rationale to the README.

ghost:
@pedrobaeza ok done

PR #307, opened 2016-12-06
[FIX][purchase_request] Give access on stock.move for purchase request user.

adrienpeiffer:
Without this access, search more action on product select box from purchase.request.line raise an exception.


JordiBForgeFlow:
I cannot reproduce the error on Runbot, in v10 branch, with stock module installed.

adrienpeiffer:
Strange. I reproduced this one on runbot with demo user who has only purchase request user as security group

AaronHForgeFlow (approved):
@adrienpeiffer I reproduced the steps and found the same issue. I was getting an error when clicking on the product as a demo user. 

With your fix this is not happening anymore so :+1: 

Thank you!

JordiBForgeFlow (approved):

PR #309, opened 2016-12-16
[8.0] [ADD] purchase_partner_default_journal

hbrunn:
Default purchase journal for partners
=====================================

This module was written to make it more convenient to work with multiple
purchase journals by assigning a default journal per partner.

Configuration
=============

To configure this module, you need to:

- go to Settings / Technical / Parameters / Company properties
- search for 'Default purchase journal' and set the default for all partners
- on a specific partner, set the field `Purchase journal` in the `Accounting` tab


pedrobaeza:
For avoiding a lot of field in the partner, maybe it's better to add the journal to the purchase_order_type module (https://github.com/OCA/purchase-workflow/tree/8.0/purchase_order_type) and this way and manage that through one data: the type.

What do you think?

hbrunn:
for this module, I don't think but implement exactly what the customer wants ;-)
this is the pendant of https://github.com/OCA/sale-financial/pull/17, for which no such construction exists to my knowledge

pedrobaeza:
OK, it was just a quick suggestion.

purchase_partner_default_journal/models/purchase_order.py
```diff
@@ -0,0 +1,28 @@
+# -*- coding: utf-8 -*-
+# © 2016 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openerp import api, models
+
+
+class PurchaseOrder(models.Model):
+    _inherit = 'purchase.order'
+
+    def onchange_partner_id(self, cr, uid, ids, partner_id, context=None):
```
susport, line 10: Why this method is not in new api?

hbrunn: because you get weird results if you override an old style onchange with v8 api

gurneyalex:

PR #310, opened 2016-12-17
[9.0] [MIG] purchase_order_type module

cubells:
- Updated README
- Updated views
- Removed invoice_method
- Updated onchange_partner_id method

cc @Tecnativa

### DIFF
--- a/purchase_order_type/README.rst
+++ b/purchase_order_type/README.rst
@@ -6,7 +6,9 @@
 Purchase Order Type
 ===================
 
-Adds a configurable *type* on the purchase orders.
+Adds a configurable *type* on the purchase orders, allowing to predefine
+some purchase order fields attached to this type.
+
 This type can be used in filters and groupbys.
 
 Configuration
@@ -14,10 +16,17 @@ Configuration
 
 To configure this module, you need to:
 
-* Go to **Purchases > Configuration > Purchase types**
-* Modify / create the purchase order types
-* You can also predefine a purchase order type in the partner if you go to tab
-  **Sales & Purchases** and select one.
+#. Go to **Purchases > Configuration > Purchase types**
+#. Modify / create the purchase order types
+#. Assign (optionally) a default incoterm for this type.
+#. Assign (optionally) a default picking type for this type (only incoming
+   types).
+
+You can also predefine a purchase order type in the partner:
+
+#. Open a partner in **Purchases > Purchase > Vendors**.
+#. Go to the **Sales & Purchases** page.
+#. Select one type on the field.
 
 Usage
 =====
--- a/purchase_order_type/README.rst
+++ b/purchase_order_type/README.rst
@@ -1,6 +1,8 @@
 .. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
-    :alt: License: AGPL-3
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
 
+===================
 Purchase Order Type
 ===================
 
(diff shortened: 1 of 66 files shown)


purchase_order_type/README.rst
```diff
@@ -1,6 +1,8 @@
 .. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
-    :alt: License: AGPL-3
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
 
+===================
 Purchase Order Type
 ===================
```
pedrobaeza: Include in the configuration section that you can predefine a purchase order type in the partner.

Include in the usage section that the type will be autofilled when selecting a partner that has a type set in its data.

purchase_order_type/models/purchase_order.py
```diff
@@ -36,19 +20,10 @@ def _default_order_type(self):
...
-                    'order_type': partner.purchase_type.id,
-                })
-        return res
-
-    @api.one
-    @api.onchange('order_type')
-    def onchange_order_type(self):
-        self.invoice_method = self.order_type.invoice_method
+    @api.onchange('partner_id', 'company_id')
+    def onchange_partner_id(self):
```
pedrobaeza: If you don't want to call super as this is an isolate operation, you can do:
```
    @api.onchange('partner_id', 'company_id')
    def onchange_partner_id_purchase_order_type(self):
        if self.partner_id.purchase_type:
            self.order_type = self.partner_id.purchase_type.id
```
Although you don't do this change, change the `if` part to the new API possibilities (no need to check the existence of each field before accessing the subfields).

purchase_order_type/models/purchase_order_type.py
```diff
@@ -1,42 +1,13 @@
...
-#    You should have received a copy of the GNU Affero General Public License
-#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#
-#
+# Copyright 2015 Guewen Baconnier <guewen.baconnier@camptocamp.com>
+# Copyright 2016 Vicent Cubells <vicent.cubells@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from openerp import models, fields, api
+from openerp import models, fields
```
pedrobaeza: Order alphabetically

pedrobaeza (changes requested):
For giving any functionality added, you can add the stock picking type (only incoming ones), and incoterm information to purchase order types, and pass this information when changing the order type.

pedrobaeza:
What about the last point?:
```
For giving any functionality added, you can add the stock picking type (only incoming ones), and incoterm information to purchase order types, and pass this information when changing the order type.
```

purchase_order_type/models/purchase_order.py
```diff
@@ -24,3 +24,9 @@ def _default_order_type(self):
     def onchange_partner_id_purchase_order_type(self):
         if self.partner_id.purchase_type:
             self.order_type = self.partner_id.purchase_type.id
+
+    @api.onchange('order_type')
+    def onchange_purchase_order_type(self):
+        if self.order_type:
+            self.incoterm_id = self.order_type.incoterm_id.id
```
pedrobaeza: This should be both also conditional. If the value is set, then you overwrite whatever it has. If not, you leave the previous value.

pedrobaeza (approved):
You can add also instructions for the 2 new fields, but not blocking. 9.0 commits to be squashed on merge.

jjscarafia (approved):

PR #311, opened 2016-12-27
[9.0] migrate module "purchase_delivery_split_date"

JordiBForgeFlow:
Includes improvements proposed to v8 here: https://github.com/OCA/purchase-workflow/pull/236

PR #313, opened 2016-12-29
[9.0] purchase_location_by line

JordiBForgeFlow:
Purchase Location by Line
=========================

Purchase Order Lines.

The corresponding PR for v8 can be found here: https://github.com/OCA/purchase-workflow/pull/237


JordiBForgeFlow:
cc @darshan-serpent @aheficent

JordiBForgeFlow:
Depends on https://github.com/OCA/purchase-workflow/pull/311

AaronHForgeFlow:
I expect 3 shipments for this PO but I'm getting 4.

image

LoisRForgeFlow:
Same as @aheficent We expected the two PO for Chicago and same date to be together.

gmeficent:
@aheficent, it may be that it compares the minutes, so the grouping is technically correct, the dates minutes should be the same or truncated.

gmeficent:
@aheficent, Confirmed, if you choose the same date down to the minute it works
image

AaronHForgeFlow:
You're right! @gmeficent <code>date_planned</code> is type Datetime in v9. Thank you!

LoisRForgeFlow (approved):
LGTM now! :+1:

PR #316, opened 2017-01-03
[10.0] Port purchase_discount

### LINKED ISSUE #265: Purchase discount not taken into account in the purchase report
The module purchase_discount does not overwrite the view of the purchase.report, and therefore the discount is not taken into account into the price. 
purchase_report.py, line 99
sum(l.price_unit/cr.rate*l.product_qty)::decimal(16,2) as price_total,

alexis-via:
Fixes #265

Note that I still discourage users from adopting purchase_discount, because it changes the meaning of the price_unit field, which makes the module incompatible with all the modules that use the field price_unit in it's original meaning.

But I tried the known issues of the module on purchase reports and stock valuation.


pedrobaeza:
Please import latest commits from 9.0 using https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-10.0 and check Travis status

pedrobaeza:
And, hey, congratulations for your successfully deployment of v10 on January!

alexis-via:
@pedrobaeza I checked that the code didn't change between 9.0 and 10.0, but I forgot to check translations. It's done now.

pedrobaeza:
Please check PEP8:
```
purchase_discount/report/purchase_report.py:35:80: E501 line too long (108 > 79 characters)
purchase_discount/report/purchase_report.py:36:80: E501 line too long (113 > 79 characters)
purchase_discount/report/purchase_report.py:38:80: E501 line too long (134 > 79 characters)
purchase_discount/report/purchase_report.py:39:80: E501 line too long (206 > 79 characters)
purchase_discount/report/purchase_report.py:40:80: E501 line too long (106 > 79 characters)
purchase_discount/report/purchase_report.py:41:80: E501 line too long (188 > 79 characters)
purchase_discount/report/purchase_report.py:51:80: E501 line too long (84 > 79 characters)
purchase_discount/report/purchase_report.py:52:80: E501 line too long (160 > 79 characters)
purchase_discount/report/purchase_report.py:55:80: E501 line too long (82 > 79 characters)
purchase_discount/report/purchase_report.py:56:80: E501 line too long (120 > 79 characters)
purchase_discount/report/purchase_report.py:57:80: E501 line too long (85 > 79 characters)
purchase_discount/report/purchase_report.py:60:80: E501 line too long (93 > 79 characters)
```
And I would thank if you squash all the "OCA Transbot" commits in one.

purchase_discount/models/account_invoice_line.py
```diff
@@ -1,21 +1,14 @@
 # -*- coding: utf-8 -*-
```
legalsylvain: git mv purchase_discount/models/account_invoice_line.py purchase_discount/models/account_invoice.py

purchase_discount/views/purchase_discount_view.xml
```diff
@@ -1,6 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
 <odoo>
-  <record model="ir.ui.view" id="purchase_discount_order_line_form2">
-    <field name="name">purchase_discount.order.line.form2 (purchase_discount)</field>
+  <record model="ir.ui.view" id="purchase_order_line_form2">
+    <field name="name">purchase_discount.order.line.form2</field>
```
legalsylvain, line 4: name field is unnecessary for views. (it is autogenerated).

legalsylvain (commented):
Hi Alexis, 
Thanks for porting this usefull module. Some remarks inline. LGTM otherwise. good news to have fixed the report issue.

leemannd:
Thank you for the port. Nothing to add from the comments up. :+1:

leemannd:
You will find a small path for the pep8 [PR on your branch](https://github.com/akretion/purchase-workflow/pull/2/files)

leemannd (approved):

StefanRijnhart (approved):
Tested and code reviewed. Thanks!

pedrobaeza:
Please squash all "OCA Transbot" commits together

pedrobaeza:
This is still red on Travis + need commit squashing for "OCA Transbot". Can you please do it?

eLBati:
Fixing lint with https://github.com/akretion/purchase-workflow/pull/3
Then, :+1:

rvalyi:
@pedrobaeza I merged @eLBati PEP 8 PR and squashed the "OCA Transbot" commits together.

alexis-via:
I forgot to mention it here sorry, but there is a bug in the current version of the module which gives random results on the amount without tax of the purchase order line. And the fix I made on June 23 fixes some bugs, but not all. I think it's due to the code of _compute_amount on purchase.order.line https://github.com/akretion/purchase-workflow/blob/10-port-purchase_discount/purchase_discount/models/purchase_order.py#L50  which writes price_unit before calling super() and re-write it after. I'm pretty sure it causes some race condition problems.

The other problem is the fact that this module changes the meaning of the "price_unit" field from net price to gross price.

I have 1 small company using this module in production and they are very angry about it.

I think I'll completely re-write this module with a different approach where the meaning of the price_unit field is not changed and where we don't use the "trick" to re-write price_unit before and after calling super().

chienandalu:
Hi, @alexis-via. I'm creating a new module that extends this PR to have three successive discounts (https://github.com/OCA/purchase-workflow/issues/432). I didn't come across tax problems on my tests. Can you give me a tip on how to replicate that kind of issues?

StefanRijnhart:
You could try calling _compute_amount's super() one by one, passing actual unit price * quantity as base amount in the context https://github.com/odoo/odoo/blob/10.0/addons/account/models/account.py#L767. You probably need to assign the resulting record updates back to the original records from before the context change.

purchase_discount/models/purchase_order.py
```diff
@@ -3,14 +3,46 @@
...
+            })
 
 
 class PurchaseOrderLine(models.Model):
     _inherit = "purchase.order.line"
 
-    @api.depends('discount')
+    @api.depends('product_qty', 'price_unit', 'taxes_id', 'discount')
     def _compute_amount(self):
         prices = {}
```
chienandalu: @alexis-via I agree with @StefanRijnhart . When you call self super, some cache issues were happening that are solved calling from inside the recordset loop. Indeed, you could put this into a method and use it to extend the other ones instead of overriding them. Take a look at my PR if you want to (https://github.com/OCA/purchase-workflow/pull/432) :slightly_smiling_face: 

```
    @api.depends('product_qty', 'price_unit', 'taxes_id', 'discount')
    def _compute_amount(self):
        for line in self:
            # super() update resets fields from cache in some circumstances. 
            # It's convenient to have this assignation by default so other
            # modules can extend this method without issues.
            price = line.price_unit
            if line.discount:
                line.price_unit *= (1 - line.discount / 100.0)
            super(PurchaseOrderLine, line)._compute_amount()
            if line.discount:
                line.price_unit = price
```

purchase_discount/models/purchase_order.py
```diff
@@ -24,9 +56,29 @@ def _compute_amount(self):
...
 
     _sql_constraints = [
         ('discount_limit', 'CHECK (discount <= 100.0)',
          'Discount must be lower than 100%.'),
     ]
+
+    @api.multi
+    def _get_stock_move_price_unit(self):
+        # method copy-pasted from odoo/addons/purchase/models/purchase.py
+        self.ensure_one()
```
pedrobaeza: This is breaking inheritance also. This is worse than the previous method.

purchase_discount/models/purchase_order.py
```diff
@@ -3,14 +3,46 @@
...
+import odoo.addons.decimal_precision as dp
+
+
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+
+    @api.depends('order_line.price_total')
+    def _amount_all(self):
+        # Method copy-pasted from odoo/addons/purchase © Odoo S.A.
+        for order in self:
```
pedrobaeza: This can't be changed without calling super. You are breaking inheritance here.

purchase_discount/report/purchase_report.py
```diff
@@ -0,0 +1,107 @@
...
+from odoo import api, models, tools
+
+
+class PurchaseReport(models.Model):
+    _inherit = "purchase.report"
+
+    @api.model_cr
+    def init(self):
+        tools.drop_view_if_exists(self._cr, 'purchase_report')
+        self._cr.execute("""
```
pedrobaeza: This is also breaking inheritance. Do something similar to what I have done in https://github.com/OCA/purchase-reporting/blob/9.0/purchase_reporting_weight/reports/purchase_report.py

pedrobaeza (changes requested):

pedrobaeza:
I have improved the PR respecting inheratibility and adding tests for the untested parts (move price unit and purchase report).

pedrobaeza (approved):

purchase_discount/models/purchase_order.py
```diff
@@ -1,32 +1,80 @@
...
+    def _amount_all(self):
+        orders2recalculate = self.filtered(lambda x: (
+            x.company_id.tax_calculation_rounding_method ==
+            'round_globally' and any(x.mapped('order_line.discount'))
+        ))
+        for order in orders2recalculate:
+            vals = {}
+            for line in order.order_line.filtered('discount'):
+                vals[line] = line.price_unit
+                line.price_unit *= (1 - line.discount) / 100.0
```
chienandalu: @pedrobaeza If you'd took this calculation into a separte method it'd make extensibility way more simple. Something of the like:
```
    def _compute_discount(self, price):
        return price * (1 - self.discount or 0.0 / 100.0)
```

pedrobaeza: OK, I'll do that.

pedrobaeza:
@chienandalu I have changed the PR for having inheritable methods. Now you only need to overwrite `_get_discounted_price_unit` on purchase.order.line and `_get_discounted_price_unit_exp` on purchase.report for changing the way the discounted unit price is computed. If you want to add discount2 and discount3 to the purchases report, just overwrite `_select_purchase_discount` and `_group_by_purchase_discount`.

purchase_discount/models/purchase_order.py
```diff
@@ -1,32 +1,92 @@
...
+
+    def _get_discounted_price_unit(self):
+        """Inheritable method for getting the unit price after applying
+        discount(s).
+
+        :rtype: float
+        :return: Unit price after discount(s).
+        """
+        self.ensure_one()
+        if self.discount:
```
chienandalu: This is going to avoid reseting a discount with value 0.

PR #318, opened 2017-01-10
[9.0][IMP][purchase_discount] Make stock move cost be aware of discount

pedrocs-exo:
This functionality is implemented in 8.0 but not in subsequent versions. It makes the stock moves created from purchase receptions reflect the line discount in its cost.


StefanRijnhart:
An alternative fix was merged here: https://github.com/OCA/purchase-workflow/pull/373. Sorry we missed your (earlier) fix!

PR #319, opened 2017-01-18
[10.0] [ADD] add new module to allow invoicing from purchase order lines

Cedric-Pigeon:
With this new module, you will be allowed to select purchase order lines from several purchase orders to create an invoice.

- [x] This PR depends on https://github.com/odoo/odoo/pull/15104 (merged)


pedrobaeza:
We have already developed a module for invoicing from purchase orders: https://github.com/OCA/account-invoicing/pull/180. Although it doesn't allow to fine grain at purchase line level, you can remove later the undesired lines. You should check that one, and 2 options:

* Include the line invoicing behavior there.
* Follow the other approach, invoicing full order and removing later the lines you don't want. This has the advantage that you don't need any Odoo patch.

Cedric-Pigeon:
@pedrobaeza 

From my point of view both can exist as they don't have the same approach. They are complementary.

When a supplier invoices a lot of lines coming from several orders, it is not user-friendly to select all and remove the useless ones.

pedrobaeza:
As said then, you have to add your functionality to the existing PR. I will review it and merge it without problems.

Cedric-Pigeon:
@pedrobaeza 
I don't understand your problem to have 2 modules as they don't overlap features.

adrienpeiffer:
Personally, I prefer one module per functionality.

pedrobaeza:
Well, I also don't understand why splitting the same kind of features in 2 modules. Anyway, if you still want to have it separately, please do the PR to account-invoicing repo, and rename to purchase_order_line_invoicing (without `s`).

Cedric-Pigeon:
No problem to rename it but why changing repo as it already exists module related to invoicing in this one ? I mean purchase_partial_invoicing and purchase_partner_invoice_method.

pedrobaeza:
Well, you know that repositories overlap frequently, but IMO all invoicing related features should go in OCA/account-invoicing (from sale, purchase, stock...), or that repo will be very empty and this one very crowded. It's not an absolute truth, of course.

Cedric-Pigeon:
This is complex debat. My point of view is that a module depending exclusively on purchase has nothing to do in a repo called account_*

pedrobaeza:
Well, clearly it depends on account, as you are creating an account.invoice, but I don't want to open a breach here. My opinion is the other, and our PR is there, but you can decide on your own.

purchase_order_line_invoicing/wizards/purchase_order_line_invoicing_wizard.py
```diff
@@ -30,8 +30,8 @@ def _prepare_default_line_vals(self, purchase_line):
             'price_subtotal': purchase_line.price_subtotal,
             'qty_received': purchase_line.qty_received,
             'qty_invoiced': purchase_line.qty_invoiced,
-            'invoice_qty':
-                purchase_line.qty_received - purchase_line.qty_invoiced,
+            'invoice_qty': max(
```
rousseldenis, line 34: @benwillig Is this correct as qty_invoiced can be negative in case of in_refund (see https://github.com/odoo/odoo/blob/d8e49d3886ab7c5443b9173bd9b3e45e05b72c6c/addons/purchase/models/purchase.py#L552) ?

benwillig: Indeed the quantity invoiced field can be negative in case of refunds. But I don't think user wants to invoice negative quantities. If we want to tell the user that there was refunds, we should display the quantity invoiced field in the wizard.

rousseldenis:
@Cedric-Pigeon @pedrobaeza @sbidoul  Could we move forward on this as it is very old. IMHO putting it here is not a problem as the entry point is the purchase itself

pedrobaeza:
OK, go on then

HviorForgeFlow (approved):

purchase_order_line_invoicing/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Purchase Order Line Invoicing',
+    'summary': """
+        Invoice purchase order lines""",
+    'version': '10.0.1.0.0',
+    'license': 'AGPL-3',
+    'author': 'ACSONE SA/NV,Odoo Community Association (OCA)',
+    'website': 'https://acsone.eu',
```
nikul-serpentcs: https://github.com/OCA/purchase-workflow

purchase_order_line_invoicing/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Purchase Order Line Invoicing',
+    'summary': """
+        Invoice purchase order lines""",
```
nikul-serpentcs: allows to invoice lines of multiple purchase orders

nikul-serpentcs (commented):
minor change

Cedric-Pigeon:
@nikul-serpentcs Comments attended, thanks.

sbejaoui (approved):

PR #320, opened 2017-01-22
[9.0] [FIX] purchase_order_variant_mgmt: Related fields on purchase.order.line trigger a writes in related models

sergio-teruel:
cc @Tecnativa


pedrobaeza (approved):
The same as in sales counterpart. Fast-tracking it.

PR #321, opened 2017-01-23
[8.0][FIX] add store=True

chafique-delli:
Add store=True for picking_state field to be able to filter the purchase orders on the status of picking.


bealdav:
:+1:

pedrobaeza (approved):

PR #332, opened 2017-02-13
[9.0][FIX]purchase_request_to_rfq: Fix invisible domain in PO line tree views

zakiuu:
Fix invisible domain in PO line tree views by s/('purchase_request_lines', '=', [])/('purchase_request_lines', '=', False)

### DIFF
--- a/purchase_request_to_rfq/views/purchase_order_view.xml
+++ b/purchase_request_to_rfq/views/purchase_order_view.xml
@@ -13,7 +13,7 @@
                        position="inside">
                     <field name="purchase_request_lines" invisible="1"/>
                     <button string="Purchase Request lines"
-                        attrs="{'invisible': [('purchase_request_lines', '=', [])]}"
+                        attrs="{'invisible': [('purchase_request_lines', '=', False)]}"
                         name="action_openRequestLineTreeView"
                         type="object"
                         icon="gtk-open"/>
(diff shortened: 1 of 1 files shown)


zakiuu:
Hi @adrienpeiffer can you please review the PR?

purchase_request_to_rfq/views/purchase_order_view.xml
```diff
@@ -13,7 +13,7 @@
                        position="inside">
                     <field name="purchase_request_lines" invisible="1"/>
                     <button string="Purchase Request lines"
-                        attrs="{'invisible': [('purchase_request_lines', '=', [])]}"
```
adrienpeiffer, line 16: @zakiuu This one works as well in 10.0. I'll check in 9.0 and 10.0

adrienpeiffer:
@zakiuu This feature works fine in version 9.0 and 10.0. You can check on runbot. So, I don't really understand your issue. I check the runbot build of your PR and your change breaks the visibility condition behavior.

zakiuu:
@adrienpeiffer Yeah Indeed, I was wrong in my understanding of the issue I had,
When breaking the `editable="bottom"` of the order line tree view on purchase order, the `{'invisible': [('purchase_request_lines', '=', [])]}` couldn't be evaluated until I add the `<field name="purchase_request_lines"` to the order line form view.
I don't know if I should do the fix here or close this PR?

PR #336, opened 2017-02-14
[9.0] [ADD][WIP] modules 'purchase_return' and 'purchase_invoice_status'

JordiBForgeFlow:
Purchase Return
===============

This module adds the possibility to create and process Return Purchase Orders.

These are orders that users can create to process the return of items to a
vendor.

* A return PO can be created with or without reference to an existing PO.

* Once a return PO is confirmed, an outbound picking is created, so as to
  ship the products back to the vendor.

* A user can create a refund bill referencing a return PO.

Purchase Invoice Status
=======================

This module extends the functionality of purchase orders in order to handle
the invoice status in purchase orders equivalently of how it is done in sales
 orders.


purchase_invoice_status/README.rst
```diff
@@ -0,0 +1,60 @@
...
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=======================
+Purchase Invoice Status
+=======================
+
+This module extends the functionality of purchase orders in order to handle
+the invoice status in purchase orders equivalently of how it is done in sales
+ orders.
```
pedrobaeza: Remove extra space. What this means?

purchase_invoice_status/models/account_invoice.py
```diff
@@ -0,0 +1,33 @@
...
+
+from openerp import api, models, fields
+
+
+class AccountInvoice(models.Model):
+    _inherit = "account.invoice"
+
+    @api.onchange('state', 'partner_id', 'invoice_line_ids')
+    def _onchange_allowed_purchase_ids(self):
+        '''
```
pedrobaeza: This should be triple double quotes

pedrobaeza (commented):
Why a separate module purchase_invoice_status? Are you planning to use this with another module?

gurneyalex:
sill WIP ?

PR #341, opened 2017-02-20
[9.0] [purchase_request] Remove uom from onchange method.

gmeficent:
[FIXED] When using the multiple units option for purchase request the units of measure resets when changing from the purchase request line form.


JordiBForgeFlow:
@gmeficent is under @Eficent ECLA

JordiBForgeFlow (approved):

LoisRForgeFlow (approved):
:+1:

PR #342, opened 2017-02-20
[9.0][FIX][purchase_request_to_rfq] Line Description.

LoisRForgeFlow:
Description of Purchase Request Lines not updating as expected. With this fix, the supplier information is used to update the description of the lines in the wizard.

Also applies the equivalent fix of https://github.com/OCA/purchase-workflow/pull/341.


LoisRForgeFlow:
cc @jbeficent

LoisRForgeFlow:
Covered with the fixes of https://github.com/OCA/purchase-workflow/pull/346. This is not needed anymore.

Issue #344, opened 2017-02-21
[9.0][purchase_request_to_rfq] When creating a PO line from a PR line the supplier pricelists and taxes are not determined in the PO line


JordiBForgeFlow:
See below. The first line in the picture was created from a PR. The second line was created manually.

![image](https://cloud.githubusercontent.com/assets/7683926/23160545/38c40cda-f827-11e6-9f80-4ab4cc7feae6.png)

JordiBForgeFlow:
@lreficent can you review?

lreficent:
the manually created one should be the same as the other, doesn't it?

LoisRForgeFlow:
PR created @jbeficent

PR #346, opened 2017-02-21
[9.0][FIX][purchase_request_to_rfq] taxes not determined in the PO line.

LoisRForgeFlow:
Solves https://github.com/OCA/purchase-workflow/issues/344.

### DIFF
--- a/purchase_request_to_rfq/wizard/purchase_request_line_make_purchase_order.py
+++ b/purchase_request_to_rfq/wizard/purchase_request_line_make_purchase_order.py
@@ -220,7 +220,7 @@ class PurchaseRequestLineMakePurchaseOrder(models.TransientModel):
             # po line
             domain = self._get_order_line_search_domain(purchase, item)
             available_po_lines = po_line_obj.search(domain)
-            if available_po_lines:
+            if available_po_lines and not item.keep_description:
                 po_line = available_po_lines[0]
                 new_qty = pr_line_obj._calc_new_qty(line, po_line=po_line)
                 if new_qty > po_line.product_qty:
@@ -232,6 +232,8 @@ class PurchaseRequestLineMakePurchaseOrder(models.TransientModel):
             else:
                 po_line_data = self._prepare_purchase_order_line(purchase,
                                                                  item)
+                if item.keep_description:
+                    po_line_data['name'] = item.name
                 po_line_obj.create(po_line_data)
             res.append(purchase.id)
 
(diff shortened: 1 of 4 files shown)


JordiBForgeFlow:
@adrienpeiffer @JonathanNEMRY This is an important improvement.

JordiBForgeFlow (approved):

lasley (commented):
Code looks good. Can you please update the test to include logic for taxes?

adrienpeiffer:
Hi,
I'm not sure I understand the problem ...
Is there a problem when you create a purchase order from scratch (Without PR).
If there is no problem, the solution is directly on the purhase module and it should not need all this code

JordiBForgeFlow:
@adrienpeiffer The problem is that when you create a PO from a PR, and the supplier has a price agreement (supplier prices defined in the product), these are not considered. Same with taxes.

adrienpeiffer:
BTW during the migration on version 10.0 I modified this method of preparation of value. This fixed several problems on UOM, price and other things. You can view the commits from the migration. There may be things to backport if you want.

JordiBForgeFlow:
@adrienpeiffer OK, we will then review in 10.0. Thanks!

adrienpeiffer:
Especially, I launch onchange directly on a memory record.

purchase_request_to_rfq/wizard/purchase_request_line_make_purchase_order.py
```diff
@@ -138,37 +135,40 @@ def _prepare_purchase_order_line(self, po, item):
...
+            date=po.date_order and po.date_order[:10],
+            uom_id=item.product_id.uom_po_id)
+        price_unit = self.env['account.tax']._fix_tax_included_price(
+            seller.price, item.product_id.supplier_taxes_id, taxes_id) \
+            if seller else 0.0
+        if (price_unit and seller and po.currency_id and
+                seller.currency_id != po.currency_id):
+            price_unit = seller.currency_id.compute(
+                price_unit, po.currency_id)
         vals = po_line_obj.onchange_product_id()
```
adrienpeiffer: So, I think the problem is the onchange isn't launched on a memory record.
See https://github.com/odoo/odoo/blob/9ec95b2a682ca885c0796c8e299ac720e8b7e5d8/addons/purchase/models/purchase.py#L748
I fix that in my PR for the migration to 10.0

adrienpeiffer (changes requested):

purchase_request_to_rfq/wizard/purchase_request_line_make_purchase_order.py
JordiBForgeFlow: Thanks for pointing this out!

AaronHForgeFlow (approved):
Code review + functional test. :white_check_mark:

adrienpeiffer:
@aheficent As I mentionned above. It's better to call the onchange on purchase order line properly. It should make the job.

LoisRForgeFlow:
@adrienpeiffer what do you mean? We added your changes, or am I wrong?  :confused:

MiquelRForgeFlow:
Now https://github.com/OCA/purchase-workflow/pull/349 is full added here.

JordiBForgeFlow:
Superseded by https://github.com/OCA/purchase-workflow/pull/350

Issue #345, opened 2017-02-21
[9.0][purchase_request] Approver should appear in the tree view


JordiBForgeFlow:
This field is important for approvers

JordiBForgeFlow:
cc @lreficent

LoisRForgeFlow:
PR created.

PR #347, opened 2017-02-21
[9.0][FIX][purchase_request] Add Approver to tree view.

LoisRForgeFlow:
Attend issue https://github.com/OCA/purchase-workflow/issues/345.


yvaucher (approved):

JordiBForgeFlow:
Please add the approver also in the purchase.request.line, as a related field, and also add it to the corresponding tree view.

LoisRForgeFlow:
@jbeficent It already exists in line and in the view (with different string though):

image

JordiBForgeFlow (approved):

PR #349, opened 2017-02-23
[9.0] [purchase_request_to_rfq] - Fix in create RFQ

JordiBForgeFlow:
Port of https://github.com/OCA/purchase-workflow/pull/247 to 9.0

- When the RFQ was created from a purchase request line,
if the product has a minimum defined in the product.supplierinfo,
a RFQ line was created with quantity ordered = minimum.
If a second request line was later added, and the sum of
the request lines was < RFQ line quantity, no link was
established between the RFQ line and the new request line.
- Tests are added to include the scenario of creating an
RFQ for a purchase request that contains a product that has minimums.

### DIFF
--- a/purchase_request_to_rfq/models/purchase_request.py
+++ b/purchase_request_to_rfq/models/purchase_request.py
@@ -80,7 +80,8 @@ class PurchaseRequestLine(models.Model):
             or False
 
     @api.model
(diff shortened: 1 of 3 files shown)


JordiBForgeFlow:
Superseded by https://github.com/OCA/purchase-workflow/pull/350

PR #354, opened 2017-03-02
[8.0] [FIX] change depends on order_line.move_ids in purchase_picking_state module

chafique-delli:


chafique-delli:
Travis test is red because the test failed when loading the purchase_request_to_rfq module.
I installed the 'purchase_request_to_rfq' module on a database where only the Odoo 8 base modules are installed.
In a terminal, by running the tests of the 'purchase_request_to_rfq' module, the test fails with the same error message.

sebastienbeau:
The issue with travis already exist in the main branch so this is not related to your test
:+1:

hparfr:
👍

PR #355, opened 2017-03-02
[9.0][FIX][purchase_delivery_split_date]Datetime issue.

LoisRForgeFlow:
As found by @gmeficent in https://github.com/OCA/purchase-workflow/pull/313 in v9 the <code>date_planned</code> is now datetime field so the module <code>purchase_delivery_split_date</code> was not working as in v8. This PR fix this issue.


JordiBForgeFlow (approved):
:+1: Technical and functional

AaronHForgeFlow (approved):
:+1: Thanks

gmeficent (commented):
Tested on Runbot, acts according to the description :+1:

PR #360, opened 2017-03-13
[9.0][purchase_request] adds done status

gmeficent:
It adds a done state into the purchase request in order to better track the finished work.

### DIFF
--- a/purchase_request/data/purchase_request_data.xml
+++ b/purchase_request/data/purchase_request_data.xml
@@ -26,5 +26,11 @@
             <field name="description">Purchase Request Rejected</field>
         </record>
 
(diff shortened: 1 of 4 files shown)


JordiBForgeFlow (approved):

JordiBForgeFlow:
cc @adrienpeiffer This is an important addition. Similar to purchase order, the user can decide to manually change the status of the purchase request to "Done". In that status, it is not expected that any further activity will occur. So the purchaser can forget about this PR, because it has been completely satisfied.

LoisRForgeFlow (approved):
Functional and technical review

JordiBForgeFlow:
@gmeficent Can you propose an equivalent PR in v10?

PR #358, opened 2017-03-10
[10.0] [MIG] Purchase Request Procurement (purchase_request_procurement)

ThomasBinsfeld:
Migration of module purchase_request_procurement to 10.0
Depends on #424

PR #364, opened 2017-03-15
[10.0][mig] purchase_request_procurement

LoisRForgeFlow:
Changes done:
* Basic migration (odoo tag, bump version...)
* Adapt methods to new API.
* Remove yml test.
* Squash old OCA Transbot commits into one.


florian-dacosta:
Isn't it a duplicate with https://github.com/OCA/purchase-workflow/pull/358 ?

LoisRForgeFlow:
Oh, yes you are right, I've only checked the migration issue and the module was not assigned there.

PR #366, opened 2017-03-17
[10.0][ADD] Add purchase_manual_service_quantity module

adrienpeiffer:
Currently, when a purchase order is confirmed with some service's product, all quantities are automatically considered as received. This module adds an option on service's products to allow to set manually the quantity received directly on purchase order lines for these product. Therefore, this field becomes editable.


JordiBForgeFlow:
Very interesting!

In SAP there used to be Service Entry Sheets, that would be a concept similar to incoming pickings, but just for services. See https://www.tutorialspoint.com/sap_mm/sap_mm_service_management.htm

Service Entry Sheets allow you not only to indicate that the service was provided, but also when and how much, and would also allow to include unplanned services.

purchase_manual_service_quantity/models/purchase_order_line.py
```diff
@@ -0,0 +1,32 @@
...
+            if line.product_id.type in ['service'] and\
+                    line.product_id.purchase_manual_received_qty:
+                line.qty_received = line.qty_received_manually
+            else:
+                super(PurchaseOrderLine, self)._compute_qty_received()
+
+    def _set_qty_received(self):
+        for line in self:
+            if line.product_id.type in ['service'] and \
+                    line.product_id.purchase_manual_received_qty:
```
benwillig: I think there should be a method or a computed field which tell if we can manually set received quantity on the line. Actually, there is the same condition in compute and set method.

benwillig (changes requested):

gurneyalex (approved):

purchase_manual_service_quantity/tests/test_purchase_manual_service_quantity.py
```diff
@@ -0,0 +1,66 @@
...
+        execute_onchange(
+            self.cls_purchase_order_line, purchase_line_vals, onchanges)
+        self.purchase_line = self.cls_purchase_order_line.create(
+            purchase_line_vals)
+
+    def test_1(self):
+        self.common_test()
+        self.assertAlmostEqual(self.purchase_line.qty_received, 0, places=2)
+        self.purchase.button_confirm()
+        with self.assertRaises(UserError), self.cr.savepoint():
```
gurneyalex: this will actually not happen : this exception is not raised in the code (there is an unused import on UserError). 

Should the test be fixed or the code?

gurneyalex (changes requested):
I also created https://github.com/acsone/purchase-workflow/pull/9 to fix lint issues

gurneyalex:
new pr https://github.com/acsone/purchase-workflow/pull/10

purchase_manual_service_quantity/models/purchase_order_line.py
```diff
@@ -0,0 +1,34 @@
...
+            else:
+                super(PurchaseOrderLine, self)._compute_qty_received()
+
+    def _inverse_qty_received(self):
+        for line in self:
+            if line.product_id.type in ['service'] and \
+                    line.product_id.purchase_manual_received_qty:
+                line.qty_received_manually = line.qty_received
+
+    qty_received_manually = fields.Float(string="Manually Received Qty",
```
rousseldenis: Please put fields definition on top, see https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#models

rousseldenis: Do these fields require decimal precision ?

rousseldenis (changes requested):
Code review. Minor comments

gurneyalex (approved):

purchase_manual_service_quantity/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+
+{
+    'name': 'Purchase Manual Service Quantity',
+    'summary': """
+        Allow to Manually set received quantities for product service on
+        purchase order""",
+    'version': '10.0.1.0.0',
+    'license': 'AGPL-3',
+    'author': 'ACSONE SA/NV,Odoo Community Association (OCA)',
+    'website': 'http://acsone.eu',
```
tschanzt: I think this should point to the github repo.

purchase_manual_service_quantity/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+    'depends': [
+        'purchase',
+    ],
+    'data': [
+        'views/product_template.xml',
+        'views/purchase_order.xml',
+    ],
+    'demo': [
```
tschanzt: you can drop the empty key.

tschanzt (commented):
only some nitpicking. looks good otherwise.

purchase_manual_service_quantity/models/purchase_order_line.py
sbidoul: I don't think it's necessary here.

tschanzt (approved):

sbidoul:
Code review comments handled. This should be good for merging.

purchase_manual_service_quantity/README.rst
```diff
@@ -0,0 +1,56 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
```
simahawk, line 1: png

purchase_manual_service_quantity/README.rst
```diff
@@ -0,0 +1,56 @@
...
+
+To use this module, you need to:
+
+#. Configure service product to allow to proceed to the manual reception of the quantities directly on the order form
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/142/10.0
+
+Bug Tracker
```
simahawk, line 22: pls, update README according to latest https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

purchase_manual_service_quantity/views/purchase_order.xml
```diff
@@ -0,0 +1,26 @@
...
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='order_line']//field[@name='qty_received']" position="before">
+                <field name="purchase_manual_received_qty" invisible="1" />
+            </xpath>
+            <xpath expr="//field[@name='order_line']//field[@name='qty_received']" position="attributes">
+                <attribute name="attrs">{'readonly': [('purchase_manual_received_qty', '=', False)]}</attribute>
+            </xpath>
+        </field>
+    </record>
+
```
simahawk, line 21: enough space? :smile:

simahawk (changes requested):
nitpicking remarks, LGTM

PR #370, opened 2017-03-29
10.0 add purchase allowed product

mourad-ehm:
This module adds a restriction in purchase and supplier invoices that has the mark "Order and invoice only supplied products" checked for allowing to select only products that can be supplied by this supplier.

The restriction can be set by default for each partner, and it's propagated
to their supplier invoices and purchase, but it can be change for each invoice and purchase if you want.


purchase_allowed_product/models/product.py
```diff
@@ -0,0 +1,27 @@
...
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    @api.model
+    def search(self, args, offset=0, limit=None, order=None, count=False):
+        restrict_supplier_id = self.env.context.get(
+            'restrict_supplier_id', False)
+        use_only_supplied_product = self.env.context.get(
+            'use_only_supplied_product', False)
```
bealdav: it seems second arg False is not mandatory here and above because it's the default one

purchase_allowed_product/models/product.py
```diff
@@ -0,0 +1,27 @@
...
+        if use_only_supplied_product:
+            seller = self.env['res.partner'].browse(restrict_supplier_id)
+            seller = seller.commercial_partner_id if seller.\
+                commercial_partner_id else seller
+            supplierinfos = self.env['product.supplierinfo'].search(
+                [('name', '=', seller.id)])
+            args.append(
+                ('product_tmpl_id', 'in',
+                    [x.product_tmpl_id.id for x in supplierinfos]))
```
bealdav: Not completely sure but performance impact here for some implementations. 
May be we could add a logger here when supplierinfos is really big to explain that to keep good performance, a specific strategy must be define

purchase_allowed_product/models/purchse_order.py
```diff
@@ -0,0 +1,22 @@
...
+        help="If checked, only the products provided by this supplier "
+             "will be shown.")
+
+    @api.onchange('partner_id')
+    def partner_id_change(self):
+        only_supp_prod = self.partner_id.use_only_supplied_product
+        if not only_supp_prod:
+            only_supp_prod = self.partner_id.commercial_partner_id.\
+                use_only_supplied_product
+        self.use_only_supplied_product = only_supp_prod
```
bealdav: Maybe a mixin model could reduce the duplicate code in this file ?

purchase_allowed_product/models/product.py
mourad-ehm: Thanks @bealdav for review,
It right that this add a new query. But It is a simple query that must be executed in few millisecond even for thousands of products. I added a index on name of supplierinfo  to improve query performance. If somme one propose an other way the make this query, I w'll take it.

bealdav: ok but as seen with @sebastienbeau there is a problem because in product.supplierinfo there are product_id and product_tmpl_id field. In your search there is only search on product_tmpl_id. The model changed in v10

mourad-ehm: I used product_id first. But even if odoo adds this field in product.supplierinfo, it is never set in DB. 
I tested with the option "Product Variants" without any effect. product_id is always null in database  and  the query returns no product if we use product_id in domain. 
I assume that the editor does not go to the end of the logic to set the supplier info on product variant. They only corrected a "syntax" error where produtuct_id referenced the product templete

EBII:
seems ok

bealdav:
LGTM :+1:

purchase_allowed_product/models/product.py
mourad-ehm: I changed domain to take in to account the product_id in product.supplierinfo search.

mourad-ehm:
Hi @yvaucher, Could merge this PR please ?

PR #373, opened 2017-04-03
[IMP] purchase_discount: fix stock get_price_unit

cubells:
There's the concept of unit price in stock moves that it's used for stock valuation, and this one doesn't have the correct price when using discounts before this commit.

cc @Tecnativa

### DIFF
--- a/purchase_discount/__openerp__.py
+++ b/purchase_discount/__openerp__.py
@@ -10,7 +10,7 @@
               "Tecnativa, "
               "ACSONE SA/NV,"
               "Odoo Community Association (OCA)",
-    "version": "9.0.1.0.1",
+    "version": "9.0.1.1.0",
     "category": "Purchase Management",
     "depends": [
         "stock",
--- a/purchase_discount/models/purchase_order.py
+++ b/purchase_discount/models/purchase_order.py
@@ -28,3 +28,17 @@ class PurchaseOrderLine(models.Model):
         ('discount_limit', 'CHECK (discount <= 100.0)',
          'Discount must be lower than 100%.'),
     ]
+
+    @api.multi
+    def _get_stock_move_price_unit(self):
+        """Get correct price with discount replacing current price_unit
+        value before calling super and restoring it later for assuring
+        maximum inheritability.
+        """
+        if self.discount:
+            price_unit = self.price_unit
+            self.price_unit *= (100 - self.discount) / 100
+        price = super(PurchaseOrderLine, self)._get_stock_move_price_unit()
+        if self.discount:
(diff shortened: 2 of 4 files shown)


purchase_discount/models/purchase_order.py
```diff
@@ -28,3 +28,24 @@ def _compute_amount(self):
         ('discount_limit', 'CHECK (discount <= 100.0)',
          'Discount must be lower than 100%.'),
     ]
+
+
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+
+    @api.model
+    def _prepare_inv_line(self, account_id, order_line):
```
pedrobaeza: This method is not needed on v9

purchase_discount/__openerp__.py
```diff
@@ -9,6 +9,7 @@
               "Acysos S.L., "
               "Tecnativa, "
               "ACSONE SA/NV,"
+              "AvanzOSC, S.L., "
               "Odoo Community Association (OCA)",
     "version": "9.0.1.0.1",
```
pedrobaeza: Change version number

purchase_discount/__openerp__.py
```diff
@@ -9,6 +9,7 @@
               "Acysos S.L., "
               "Tecnativa, "
               "ACSONE SA/NV,"
+              "AvanzOSC, S.L., "
```
pedrobaeza: This little patch shouldn't be included in authors.

purchase_discount/models/purchase_order.py
```diff
@@ -28,3 +28,17 @@ def _compute_amount(self):
...
+        value before calling super and restoring it later for assuring
+        maximum inheritability.
+        """
+        if self.discount:
+            price_unit = self.price_unit
+            self.price_unit *= (100 - self.discount) / 100
+        price = super(PurchaseOrderLine, self)._get_stock_move_price_unit()
+        if self.discount:
+            self.price_unit = price_unit
+        return price
```
sergio-teruel, line 44: Is multi?? Maybe you're missing a loop...

sergio-teruel (changes requested):

purchase_discount/models/purchase_order.py
pedrobaeza: No, it isn't. super method has self.ensure_one()

sergio-teruel: Oks..

sergio-teruel (approved):

carlosdauden (approved):

PR #375, opened 2017-04-04
[9.0][FIX][purchase_request_to_rfq] purchased quantity to use the PO UoM

JordiBForgeFlow:
We have identified that in the cases where you wanted to purchase from a purchase request, where the product had a purchase uom different from the base UoM, the quantity that was transferred to the PO was the original from the purchase request, but the unit was the purchase UoM.

E.g. your purchase request is un Unit(s)
The product has Purchase UoM Dozen(s)

your purchase request is 12 unit(s). Your po is created with 12 Dozen(s)

What is expected:
your purchase request is 12 unit(s). Your po is created with 1 Dozen(s)

The PR contains an additional test case that covers this scenario.

### DIFF
--- a/purchase_request_to_rfq/models/purchase_request.py
+++ b/purchase_request_to_rfq/models/purchase_request.py
@@ -95,10 +95,10 @@ class PurchaseRequestLine(models.Model):
     @api.model
     def _calc_new_qty(self, request_line, po_line=None, cancel=False,
                       new_pr_line=False):
-        uom = request_line.product_uom_id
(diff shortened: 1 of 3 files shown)


JordiBForgeFlow:
cc @Eficent

LoisRForgeFlow (approved):
Travis is complaining. Despite of that, LGTM

JordiBForgeFlow:
@mreficent can you review this PR?

MiquelRForgeFlow (approved):
soft code review :+1:

PR #385, opened 2017-04-27
[ADD][10.0] [purchase_request] adds done status

gmeficent:
It adds a done state into the purchase request in order to better track the finished work.

@Eficent

### DIFF
--- a/purchase_request/models/purchase_request.py
+++ b/purchase_request/models/purchase_request.py
@@ -9,7 +9,8 @@ _STATES = [
     ('draft', 'Draft'),
     ('to_approve', 'To be approved'),
     ('approved', 'Approved'),
-    ('rejected', 'Rejected')
+    ('rejected', 'Rejected'),
+    ('done', 'Done')
 ]
 
 
@@ -48,7 +49,7 @@ class PurchaseRequest(models.Model):
     @api.depends('state')
     def _compute_is_editable(self):
         for rec in self:
-            if rec.state in ('to_approve', 'approved', 'rejected'):
+            if rec.state in ('to_approve', 'approved', 'rejected', 'done'):
                 rec.is_editable = False
             else:
                 rec.is_editable = True
@@ -62,6 +63,8 @@ class PurchaseRequest(models.Model):
                 return 'purchase_request.mt_request_approved'
             elif 'state' in init_values and rec.state == 'rejected':
                 return 'purchase_request.mt_request_rejected'
+            elif 'state' in init_values and rec.state == 'done':
+                return 'purchase_request.mt_request_done'
         return super(PurchaseRequest, self)._track_subtype(init_values)
 
     name = fields.Char('Request Reference', size=32, required=True,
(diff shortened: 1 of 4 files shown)


purchase_request/models/purchase_request.py
```diff
@@ -30,7 +31,7 @@ def _get_default_requested_by(self):
 
     @api.model
     def _get_default_name(self):
-        return self.env['ir.sequence'].next_by_code('purchase.request')
+        return self.env['ir.sequence'].get('purchase.request')
```
LoisRForgeFlow: `get()` is deprecated, `next_by_code()` is the way to go and was fine I think

purchase_request/models/purchase_request.py
```diff
@@ -9,7 +9,8 @@
     ('draft', 'Draft'),
     ('to_approve', 'To be approved'),
     ('approved', 'Approved'),
-    ('rejected', 'Rejected')
+    ('rejected', 'Rejected'),
+    ('done', 'Done')
```
LoisRForgeFlow, line 13: `state = fields.Selection(selection_add=[('done', 'Done')])` is a better way to do it, so in case another module also adds states it won't break.

gmeficent: It's not an inherit it's the original class.

MiquelRForgeFlow (approved):
LGTM :+1:

purchase_request/models/purchase_request.py
```diff
@@ -111,7 +114,7 @@ def copy(self, default=None):
         self.ensure_one()
         default.update({
             'state': 'draft',
-            'name': self.env['ir.sequence'].next_by_code('purchase.request'),
+            'name': self.env['ir.sequence'].get('purchase.request'),
```
AaronHForgeFlow: Why this?

AaronHForgeFlow (approved):
LGTM :+1:

purchase_request/models/purchase_request.py
```diff
@@ -154,6 +157,12 @@ def button_rejected(self):
             rec.state = 'rejected'
         return True
 
+    @api.multi
+    def button_done(self):
+        for rec in self:
```
zakiuu: `return self.write({'state': 'done'})` would be much effecient

zakiuu:
@gmeficent thx!

LoisRForgeFlow (approved):
:+1:

LoisRForgeFlow:
@zakiuu could you update your review?

zakiuu (approved):
LGTM

purchase_request/models/purchase_request.py
```diff
@@ -135,35 +135,25 @@ def write(self, vals):
...
-        return True
+        return self.write({'state': 'approved'})
 
     @api.multi
     def button_rejected(self):
-        for rec in self:
-            rec.state = 'rejected'
-            rec.line_ids.do_cancel()
-        return True
+        self.line_ids.do_cancel()
```
LoisRForgeFlow: `self.mapped('line_ids').do_cancel()` and same fix in L138

PR #387, opened 2017-05-02
Update README.md

blutecsolutions:
Free of Paiment corrected -> Free of Payment


pedrobaeza:
This shouldn't be update here, but on the `__openerp__.py` file of the involved module. Later, this will be automatically updated.

MiquelRForgeFlow:
This PR should be closed as derived from what @pedrobaeza said.

PR #390, opened 2017-05-08
V10 purchase isolated quotation pyn

phongyanon:
fix error code test from flake8


purchase_isolated_quotation/__init__.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
+#
+#
+#    Copyright (C) 2012 Ecosoft (<http://www.ecosoft.co.th>)
+#
```
LoisRForgeFlow: change to short headers following [OCA template](https://github.com/OCA/maintainer-tools/tree/master/template/module)

purchase_isolated_quotation/__manifest__.py
```diff
@@ -0,0 +1,44 @@
...
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#
+{
+    'name': 'Purchase - Separate Quote and Order',
+    'version': '10.0.0.1.0',
+    'author': 'Ecosoft,Odoo Community Association (OCA)',
+    'category': 'Purchase',
+    'description': """
+This module separate quotation and purchase order
```
LoisRForgeFlow: Add a README file

LoisRForgeFlow (changes requested):

kittiu:
@andhit-r commented taht **purchase_order_type** already introduce field name **order_type**. Its better you avoid to use that name. 

@phongyanon I think as it won't be more than 2 type (Order/Quotation), should we use boolean field instead? i.e., isolated_order = True for Purchase Order and False for Quotation.

By doing so, when integrate with other module, i.e., when purchase.order is created by call for bid, the default isolated_order = False and it become Quotation (and we want to start with Quotation as default)

phongyanon:
I've got some exception while run --test-enable, How to fix this?
selection_039

kittiu:
Hi, we are trying to finish this one. The code is done, but the Travis check is not listed (only runbot).
Any clue? Thanks!

purchase_isolated_quotation/__init__.py
```diff
@@ -0,0 +1,5 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) 2012 Ecosoft (<http://www.ecosoft.co.th>)
```
kittiu: @phongyanon I think we can remove the copyright here. Already have the contributor list in README

elicoidal:
Strange indeed: try to trigger the travis directly in travis for this repo. Maybe it will reconnect it again

kittiu:
Don't know how to do it :p.
How about this, @phongyanon can you resubmit your code with new Branch and new PR?

elicoidal:
check this: https://travis-ci.org/OCA/purchase-workflow/pull_requests

kittiu:
Hi, I find the link of the Travis CI here. https://travis-ci.org/OCA/purchase-workflow/builds/238005243
But I have no option to rebuild it. @phongyanon can you?

pedrobaeza:
You have to login with GitHub on Travis and you'll be able to.

kittiu:
Hello again, I want to see if it is ok now, as the travis already green?

purchase_isolated_quotation/models/purchase.py
```diff
@@ -0,0 +1,83 @@
...
+        copy=False,
+    )
+    order_id = fields.Many2one(
+        'purchase.order',
+        string='Order Reference',
+        readonly=True,
+        ondelete='restrict',
+        copy=False,
+    )
+    state2 = fields.Selection(
```
legalsylvain: better to name it quotation_state ?

phongyanon: i agree with you, it will be meaningful name.

Issue #396, opened 2017-05-12
[9.0][purchase_request_procurement] Procurement not cancelled when a purchase request is cancelled


JordiBForgeFlow:
We need to change the purchase_request_procurement module, so that cancelling a purchase request will cancel the procurement as well. But then you should NOT be able to reset the purchase request back to draft, if the related procurements are cancelled.

Also, you should not be able to cancel a procurement if a subsequent purchase request exists. The user needs to first cancel it through the procurement.

JordiBForgeFlow:
cc @lreficent

sbidoul:
cc/ @benwillig @gva-acsone

LoisRForgeFlow:
This comment describes the [final approach](https://github.com/OCA/purchase-workflow/pull/398#issuecomment-303458774) followed.

benwillig:
@lreficent @jbeficent Can I start to migrate to 10.0 merged changes ?

JordiBForgeFlow:
@benwillig Yes, you can

PR #398, opened 2017-05-15
[9.0][IMP] purchase_request_procurement: cancel procurements

LoisRForgeFlow:
This PR responds to https://github.com/OCA/purchase-workflow/issues/396

Depends on https://github.com/OCA/purchase-workflow/pull/412

**Todo:**

* [x] : Update README
* [x] : Fix tests

### DIFF
--- a/purchase_request_procurement/models/procurement.py
+++ b/purchase_request_procurement/models/procurement.py
@@ -2,24 +2,15 @@
 # Copyright 2016 Eficent Business and IT Consulting Services S.L.
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl-3.0).
 
-from openerp import _, api, fields, models
-from openerp.exceptions import ValidationError
+from openerp import api, fields, models, _
 
 
 class Procurement(models.Model):
     _inherit = 'procurement.order'
 
-    request_id = fields.Many2one('purchase.request',
-                                 ondelete='restrict',
-                                 string='Latest Purchase Request')
-
-    @api.multi
-    def copy(self, default=None):
-        if default is None:
-            default = {}
-        self.ensure_one()
-        default['request_id'] = False
-        return super(Procurement, self).copy(default)
+    request_id = fields.Many2one(
+        comodel_name='purchase.request', ondelete='restrict',
+        string='Latest Purchase Request', copy=False)
 
     @api.model
     def _prepare_purchase_request_line(self, purchase_request, procurement):
@@ -62,11 +53,11 @@ class Procurement(models.Model):
             pr = self._search_existing_purchase_request(procurement)
             if not pr:
                 request_data = self._prepare_purchase_request(procurement)
-                req = request_obj.create(request_data)
+                pr = request_obj.create(request_data)
                 procurement.message_post(body=_("Purchase Request created"))
-                procurement.request_id = req.id
+                procurement.request_id = pr.id
             request_line_data = self._prepare_purchase_request_line(
-                req, procurement)
+                pr, procurement)
             request_line_obj.create(request_line_data),
             procurement.message_post(body=_("Purchase Request extended."))
             return True
@@ -77,24 +68,15 @@ class Procurement(models.Model):
         result = super(Procurement, self).propagate_cancels()
         # Remove the reference to the request_id from the procurement order
         for procurement in self:
-            request = procurement.request_id
-            procurement.write({'request_id': None})
             # Search for purchase request lines containing the procurement_id
-            request_lines = self.env['purchase.request.line'].\
-                search([('procurement_id', '=', procurement.id)])
-            # Remove the purchase request lines, if the request is not draft,
-            # to be approved or rejected.
-            for line in request_lines:
-                if line.request_id.state not in ('draft',
-                                                 'to_approve', 'rejected'):
-                    raise ValidationError(_('Can not cancel this procurement '
-                                            'as the related purchase request '
-                                            'is in progress confirmed already.'
-                                            ' Please cancel the purchase'
-                                            ' request first.'))
-                else:
-                    line.unlink()
-            # If the purchase request has not lines, delete it as well
-            if len(request.line_ids) == 0:
-                request.unlink()
+            #  and cancel them.
+            proc_request_lines = self.env['purchase.request.line'].search([
+                ('procurement_id', '=', procurement.id)])
+            if proc_request_lines and not self.env.context.get(
+                    'from_purchase_request'):
+                proc_request_lines.do_cancel()
+                for prl in proc_request_lines:
+                    prl.message_post(
+                        body=_("Related procurement has been cancelled."))
+            procurement.write({'request_id': None})
         return result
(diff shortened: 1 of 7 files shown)


LoisRForgeFlow:
cc @jbeficent

gva-acsone (approved):
Functional test 👍

JordiBForgeFlow:
@lreficent Can you add a test to cover this new scenario?

purchase_request_procurement/models/procurement.py
```diff
@@ -84,12 +84,13 @@ def propagate_cancels(self):
...
             # or reject
             for line in request_lines:
-                if line.request_id.state not in ('draft', 'reject'):
-                    raise ValidationError(_('Can not cancel this procurement '
-                                            'as the related purchase request '
-                                            'is in progress confirmed already.'
-                                            ' Please cancel the purchase'
-                                            ' request first.'))
+                if (line.request_id.state not in ('draft', 'reject') and not
+                        self._context.get('from_purchase_request')):
```
benwillig: Don't we need to check that request state is not in 'draft' or '__**rejected**__'? I think that ```reject``` state doesn't exists. Then I guess you won't have to use ```from_purchase_request``` in the context.

LoisRForgeFlow: You are right about the typo error, It something that was in the original module, I take the ocassion to fix. Good catch!

In the other hand, `from_purchase_request` context has to be checked because this is what makes this error raise if you try to directly cancel the procurement and allows you to cancel when rejecting a purchase request. The condition has to be OR though.

LoisRForgeFlow:
@jbeficent tests added.

purchase_request_procurement/models/procurement.py
benwillig: If you cancel a procurement which is linked to a draft purchase request, an error will be raised. Are you sure this is the behavior we want?

LoisRForgeFlow: Look the issue https://github.com/OCA/purchase-workflow/issues/396:

> Also, you should not be able to cancel a procurement if a subsequent purchase request exists.   The user needs to first cancel it through the procurement.

LoisRForgeFlow: what do you think @sbidoul @jbeficent ?

sbidoul: @jbeficent the last sentence of #396 is not clear to me. @jbeficent can you reformulate it?

JordiBForgeFlow:
@sbidoul @benwillig after further brainstorming this with @lreficent we concluded that it would be best to:

1. Avoid deleting request lines. Instead introduce a '**cancelled**' state/boolean in the purchase request line.
2. When a procurement is **cancelled**, the PR lines that are linked to it will be cancelled
3. When a PR line is **cancelled**, the linked procurements that are not yet cancelled will become cancelled.
4. When a PR is **rejected**, the PR lines are **cancelled** (and the linked procurements will also be cancelled).
5. When a PR changes from **rejected** to **draft** the lines will change to non cancelled (lines with cancelled procurement remain cancelled)
6. When a PO is cancelled, the procurements are cancelled, and the subsequent PR lines will be **cancelled**.
7. A user cannot create RFQ / Bids from cancelled PR lines.

Note that a PR can be approved but still have multiple lines **cancelled**

The lines that are **cancelled** appear in grey color, so that the user is conscious.

This would require a change to the purchase_request, purchase_request_procurement, purchase_request_to_rfq, purchase_request_to_requisition, purchase_request_to_procurement

LoisRForgeFlow:
Rebased to solve conflicts.

MiquelRForgeFlow (approved):
LGTM! :+1:

benwillig:
@lreficent @jbeficent 

1. Would it be possible to reset/reject a PR if linked objects (procurement or PO) are already in done state ?
2. What happen if we reconfirm a procurement or a PO which are linked to cancelled purchase request lines?

Anyway I think that adding a state on purchase request line is a good idea.

LoisRForgeFlow:
@benwillig Those are good points.

> Would it be possible to reset/reject a PR if linked objects (procurement or PO) are already in done state ?

I think that is better not to allow that. What do you think?
>What happen if we reconfirm a procurement or a PO which are linked to cancelled purchase request lines?

In the case of procurement, which I'm working now in, when you cancel the procurement the system will set the purchase request in the procurement to none. This way, if you reconfirm the procurement, the system will create a new purchase request (with new PR line). Do you agree?

AaronHForgeFlow:
> In the case of procurement, which I'm working now in, when you cancel the procurement the system will set the purchase request in the procurement to none. This way, if you reconfirm the procurement, the system will create a new purchase request (with new PR line).

@lreficent I agree with that. If the procurement. Using the same Purchase Request would be confusing.

benwillig:
@lreficent 

> Would it be possible to reset/reject a PR if linked objects (procurement or PO) are already in done state ?

Indeed, I think we should block it.

> In the case of procurement, which I'm working now in, when you cancel the procurement the system will set the purchase request in the procurement to none. This way, if you reconfirm the procurement, the system will create a new purchase request (with new PR line). Do you agree?

As @aheficent said, It would be confusing. I think it's a good idea to create a new purchase request. The first purchase request line (which has been cancelled) will still have the link to the procurement so we can keep an history.

LoisRForgeFlow:
Depends on https://github.com/OCA/purchase-workflow/pull/412

Added changes to attend discussion.

please have a look @benwillig @sbidoul

purchase_request_procurement/models/purchase_request.py
```diff
@@ -0,0 +1,29 @@
...
+                  'done procurement orders.'))
+        return super(PurchaseRequest, self).button_rejected()
+
+    @api.multi
+    def button_draft(self):
+        if any([s == 'done' for s in self.mapped(
+                'line_ids.procurement_id.state')]):
+            raise UserError(
+                _('You cannot set back to draft a purchase request with lines '
+                  'related to done procurement orders.'))
```
benwillig, line 31:
```python
      if any([s == 'done' for s in self.mapped(
                 'line_ids.procurement_id.state')]):
```

Maybe we could move this part in two methods called ```reject_allowed``` and ```draft_allowed```? Then it would be easy to change the condition in a submodule.

LoisRForgeFlow: I think that if you change the condition you will also want to change the UserError msg as they are related. Don't you think?

benwillig: Yes indeed. Then maybe we should have two methods which have the condition and two other methods that check the condition and raise an error. As you wish. But I think it would be better to at least, move thoses check in a specific method.

LoisRForgeFlow: Ok, I understand your point. ~I will attend it now.~ It's fixed now!

JordiBForgeFlow (approved):

AaronHForgeFlow (approved):
Tested on runbot :+1:

LoisRForgeFlow:
@benwillig Could you update your review?

benwillig (approved):
Tested on runbot :+1:

JordiBForgeFlow:
@lreficent can you squash the commits?

LoisRForgeFlow:
@jbeficent Done!

PR #399, opened 2017-05-17
[9.0][FIX]purchase_request_to_rfq: Fix default_get on wizard

zakiuu:
* Current behavior: in `default_get` method of the wizard converting a purchase request lines to a purchase order (Transient model `purchase.request.line.make.purchase.order`), we read the `active_ids` in the context. If the `active_ids` isn't provided in the context a `KeyError` is raised.
* The fix: To fix this issue i've used the `get()` fonction to retrieve the `active_ids` in the context.
* The expected behavior: With this fix if the `active_ids` isn't provided the method return the `res` of the super method.

### DIFF
--- a/purchase_request_to_rfq/wizard/purchase_request_line_make_purchase_order.py
+++ b/purchase_request_to_rfq/wizard/purchase_request_line_make_purchase_order.py
@@ -91,9 +91,8 @@ class PurchaseRequestLineMakePurchaseOrder(models.TransientModel):
         res = super(PurchaseRequestLineMakePurchaseOrder, self).default_get(
             fields)
         request_line_obj = self.env['purchase.request.line']
-        request_line_ids = self.env.context['active_ids'] or []
-        active_model = self.env.context['active_model']
-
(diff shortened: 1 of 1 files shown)


ThomasBinsfeld (approved):

sbidoul (approved):
No need for  `, False` in get. Otherwise ok, just in case the wizard is invoked without active_ids.

LoisRForgeFlow (approved):
You are right, but, how can a user invoke the wizard without selecting some records?

AaronHForgeFlow (approved):
:+1: @lreficent I would like to know how to do that also

zakiuu:
@lreficent @aheficent if you do create wizard using an API or inside your models without a `with_context(active_ids=[some_ids])` you will get a `KeyError`. As a use case you want to convert the request lines to purchases using a cron job

LoisRForgeFlow:
@zakiuu yes, that could be a case. Thanks!

gmeficent:
This can be merged already @jbeficent, @pedrobaeza.

PR #401, opened 2017-05-22
[10.0][ADD] product_customer_code_purchase

angedras:
Adds product customer code purchase module for v10.0

Depends on [#236](https://github.com/OCA/product-attribute/pull/236)


eLBati:
I would close this.
See https://github.com/OCA/product-attribute/pull/236#issuecomment-304812958

PR #402, opened 2017-05-23
[10.0][FIX] [product_by_supplier] Fix view for product.supplierinfo for showing Product

Roodin:
The actual inherited view for product.supplierinfo duplicates the product_tmpl_id field so it is not correctly shown

### DIFF
--- a/product_by_supplier/views/product_supplierinfo_view.xml
+++ b/product_by_supplier/views/product_supplierinfo_view.xml
@@ -62,10 +62,9 @@
         <field name="inherit_id" ref="product.product_supplierinfo_form_view"/>
         <field name="priority">99</field>
         <field name="arch" type="xml">
-            <xpath expr="//field[@name='product_name']" position="before">
-                <field name="product_tmpl_id" string="Product"
-                invisible="context.get('from_product_form')"
-                required="not context.get('from_product_form')"/>
+            <xpath expr="//field[@name='product_tmpl_id']" position="attributes">
+                <attribute name="invisible">context.get('from_product_form')</attribute>
+                <attribute name="required">not context.get('from_product_form')</attribute>
             </xpath>
         </field>
     </record>


pedrobaeza:
If the field is duplicated, it's not enough to simply hide it.

product_by_supplier/views/product_supplierinfo_view.xml
```diff
@@ -62,10 +62,9 @@
         <field name="inherit_id" ref="product.product_supplierinfo_form_view"/>
         <field name="priority">99</field>
         <field name="arch" type="xml">
-            <xpath expr="//field[@name='product_name']" position="before">
-                <field name="product_tmpl_id" string="Product"
-                invisible="context.get('from_product_form')"
-                required="not context.get('from_product_form')"/>
+            <xpath expr="//field[@name='product_tmpl_id']" position="attributes">
+                <attribute name="invisible">context.get('from_product_form')</attribute>
```
MiquelRForgeFlow, line 66: Shouldn't this line be <code>>context.get('visible_product_tmpl_id', True) or context.get('from_product_form')<</code>?

zaoral:
> If the field is duplicated, it's not enough to simply hide it.

@pedrobaeza IMHO as far as I know If you try to add the same field twice in the view this will make that the view stays buggy and will not show the field value properly. If the field is hidden will still result in a buggy view....

So adding the needed field attributes using xpath is the correct way.

pedrobaeza:
I don't want to add the field 2 times. What I want is the contrary: to not have that field on the view that is used for the one2many.

pedrobaeza:
Ah, and good news: on v11 you can add the same field several times, and all occurrences will be synced.

zaoral:
> I don't want to add the field 2 times. What I want is the contrary: to not have that field on the view that is used for the one2many.

Ok. I think we are taking of two different things

First, In this case you dont want to have the field twice in the view. but that is what is happening since in version 9.0 the product_tmpl_id is in the original view. the generic one that is a primary view. 

As you can see as follow:

https://github.com/odoo/odoo/blob/9.0/addons/product/product_view.xml#L799

image

Before the change, you were adding again the product_tmpl_id, in this PR case instead of re adding the field is overwritten its attribute without duplicating the field . 

Second, if you need to generate a different view because you will like to show this information different from a o2m field then that is another thing that should be made in a new PR.

But I think that this fix is right taking into account what I mentioned first, the PR name well.. maybe is not the proper one but is a fix that is needed to do not add a bug to the view

zaoral:
NOTE: this also apply for version 9.0 https://github.com/OCA/purchase-workflow/pull/502

pedrobaeza:
Please use same approach as in https://github.com/OCA/purchase-workflow/pull/502

pedrobaeza:
I have implemented the same approach like #502 in 1953f50496249f87510cbe2f81c4ee45f59d96c8, so closing this.

PR #404, opened 2017-05-30
[FIX] purchase_request: avoid to have a loop on UOM change.

adrienpeiffer:
cc\ @lreficent @jbeficent


LoisRForgeFlow (approved):
Makes sense for me

AaronHForgeFlow (approved):
I agree with this :+1:

MiquelRForgeFlow (approved):
LGTM! :+1:

PR #407, opened 2017-06-05
10.0 product supplierinfo discount

StefanRijnhart:
Apart from migrating this module to 10.0, this PR suggests to handle an issue that already existed in 9.0 in https://github.com/OCA/purchase-workflow/pull/407/commits/61a32a0726: because Odoo automatically creates supplierinfo entries when a purchase order is confirmed, the purchase order lines' discounts should be added to these entries.


JonathanNEMRY:
@StefanRijnhart can you rebase it? The depends is now merged so I think travis should be happy with that! :smile:

legalsylvain (approved):
Code review. (only : 9c219cf1ce217fc17d9f3f7ef04946137213fbba)
no test.

StefanRijnhart:
@JonathanNEMRY thanks for noting! Now rebased.

legalsylvain:
Hi. I'm just thought about a potential improvment.

For the time being : 
- without this module, in product.supplierinfo, the 'price' field is the purchase price, (so the "Net price").
- with this module, this field become the gross price, and we add a field 'discount'. So we change the meaning of this field.

**The Idea** : 
- keep the 'price' as the "Net Price", readonly, depends('gross_price', 'discount') 
- Add a new field "gross_price".
- Add the field "discount" as this module does it.

- Optionally, do the same thing in purchase.order.line. (so : we could remove the following lines in the readme file : https://github.com/OCA/purchase-workflow/blob/10.0/purchase_discount/README.rst

> With this module, the price_unit field of purchase order line stores the gross price instead of the net price, which is a change in the meaning of this field. So this module breaks all the other modules that use the price_unit field with it's native meaning.


**Pro** : 
- keep the meaning of the original field.
- having both informations for the end users. (gross / net prices).
- Futur possibility to handle rounding problem more easily. `round(gross_price * (1-discount/100)) * qty != round(gross_price * qty) * (1-discount/100)`

**Cons** : 
- More work for this PR.
- migration script required to managed correctly transition between 10.0 (or -) and 11.0 (or +)

regards.
(really not a blocking point regarding this PR)

StefanRijnhart:
@legalsylvain I don't think I agree with your conclusion that this module changes the semantics of the field. Is it not the case that without this module, the price field represents __both__ the gross price as well as the net price? You may as well add a new field for the net price instead of adding a new field for the gross price. But IMHO it is out of scope for this migration.

legalsylvain:
Hi @StefanRijnhart. 

>  I don't think I agree with your conclusion that this module changes the semantics of the field

Well, for the supplierinfo, it is a matter of PoV. I can understand your. About purchase.order.line, well it is written in the current readme file, that module breaks the meaning of the price_unit field.  Maybe it's wrong, but in that case, it should be removed. Don't you think ?

> But IMHO it is out of scope for this migration.

Of course, as said, not a blocking point, just a question !

rousseldenis:
@legalsylvain The question had to be asked. I think these concepts have to be clear for user, at least.

pedrobaeza:
The problem is that the same field, `price_unit`, is used for both gross value and net value in `purchase` core module, so it uses it for some operations as gross value and another as net value. Adding the discount, we have to interfere in all the operations where the giving meaning is the net one for adding the discount in that logic. The perfect solution would be to have both values unfolded in core (for v12 obviously), and we change the way the net value is computed in `purchase_discount`, `purchase_triple_discount`, etc.

Any way, `purchase_discount` handles already the only case that is related in core with net price, which is the purchase value put in the stock move for average cost price computation.

lukebranch:
@StefanRijnhart ,

I'm going to work on an 11.0 port of this shortly (fundamentally just removing procurement.order). I just wanted to check, are there any changes planned on this PR before it's merged into 10.0 that I should take into consideration?

StefanRijnhart:
@lukebranch no, I think this is working fine as it is now. Maybe if you review we can have it merged!

lukebranch:
@StefanRijnhart ,

Brilliant, thanks for the reply. I'll take a look at this over the weekend and get back to you.

legalsylvain:
@lukebranch  did you take the time to review this PR ?
regards.

apesquero:
I don't now why close the other PR if this PR is down!

StefanRijnhart:
Hi @apesquero what is down about this PR?

apesquero:
Nothing, I think this PR is perfect, but it does not finish being published.

pedrobaeza:
Then make your review here and follow the process, or use it.

StefanRijnhart:
@pedrobaeza as the original author, can I have your additional approval on the added change in https://github.com/OCA/purchase-workflow/commit/61a32a0726?

product_supplierinfo_discount/models/purchase_order.py
```diff
@@ -1,67 +1,42 @@
...
+        """
+        Check if a discount is defined into the supplier info and if so then
+        apply it to the current purchase order line
+        """
+        res = super(PurchaseOrderLine, self)._onchange_quantity()
+        if self.product_id:
+            product_supplierinfo = self.product_id._select_seller(
+                partner_id=self.partner_id, quantity=self.product_qty,
+                date=self.order_id.date_order and
+                self.order_id.date_order[:10],
```
rousseldenis: You cannot do that. Use 

```python
fields.Date.to_string(fields.Date.from_string(self.order_id.date_order))
```

StefanRijnhart:
@apesquero still looking forward to your review! Reviews are the only thing that can get any PR merged.

product_supplierinfo_discount/models/purchase_order.py
```diff
@@ -1,67 +1,42 @@
...
+
+    @api.multi
+    def _add_supplier_to_product(self):
+        """ Insert a mapping of products to discounts to be picked up
+        in supplierinfo's create() """
+        self.ensure_one()
+        discount_map = dict(
+            [(line.product_id.product_tmpl_id.id, line.discount)
+             for line in self.order_line.filtered('discount')])
+        return super(PurchaseOrder, self.with_context(
```
rousseldenis, line 20: As with_context() recreate new environment, it is strongly encouraged to avoid this to pass parameters. Performance reasons

But this works. Improvement in another version.

rousseldenis (changes requested):
Minor comments.
One can have performance impacts. Keep eye on it

apesquero:
@StefanRijnhart,

I'm sorry, I'm new, it's possible that I did not express myself well, I think your code is correct. I have no review!

product_supplierinfo_discount/models/purchase_order.py
StefanRijnhart: @rousseldenis sure, this will lead to a new read() on the purchase order and its lines but I'm a bit surprised you mention it as a bad practice. Have you got a link to guidelines from Odoo or the OCA that mention with_context as such?

StefanRijnhart:
@apesquero in that case, please go to the Review Changes on the Files Changed tab on this PR and select 'Approve'. Thanks!

product_supplierinfo_discount/models/purchase_order.py
rousseldenis: Nope, that is based on experienced cases.

For this, I don't have an immediate solution, I didn't study it deeply.

product_supplierinfo_discount/models/purchase_order.py
StefanRijnhart: Fixed!

product_supplierinfo_discount/models/purchase_order.py
```diff
@@ -1,67 +1,44 @@
...
-                break
+    @api.onchange('product_qty', 'product_uom')
+    def _onchange_quantity(self):
+        """
+        Check if a discount is defined into the supplier info and if so then
+        apply it to the current purchase order line
+        """
+        res = super(PurchaseOrderLine, self)._onchange_quantity()
+        if self.product_id:
+            date = False
```
rousseldenis: @StefanRijnhart Nearly, you have to define it to None. Either function _select_seller will crash (Look at it - same case here https://github.com/OCA/stock-logistics-warehouse/pull/351).

StefanRijnhart: Thanks for spotting this! Changed to None and I fixed another instance of date_order[10] in this code.

rousseldenis (approved):

apesquero (approved):

StefanRijnhart:
Thanks for your reviews everyone!

pedrobaeza:
I'm late to the party, but yes, very good addition in 61a32a0, @StefanRijnhart

PR #410, opened 2017-06-09
[9.0][purchase_location_by_line] Fixes error when picking is created

JordiBForgeFlow:
Prior to this fix users were required to manually enter the destination location in all the PO lines. Otherwise an error would occur at the time of creating the picking.

Extended the test cases to cover the changes made.

### DIFF
--- a/purchase_location_by_line/models/purchase.py
+++ b/purchase_location_by_line/models/purchase.py
@@ -35,13 +35,24 @@ class PurchaseOrderLine(models.Model):
         additional keys or replace them by others."""
         key = super(PurchaseOrderLine, self)._get_group_keys(order, line,
                                                              picking=picking)
-        return key + ({'location_dest_id': line.location_dest_id},)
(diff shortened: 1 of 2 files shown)


JordiBForgeFlow:
@mreficent, @lreficent, @gmeficent can you review?

MiquelRForgeFlow (approved):
I approve, but I prefer the following approach:
1. Define `order_id._get_destination_location()` as `default` for the `location_dest_id` field.
2. Add a migration script to fill the `location_dest_id` field whenever it's empty.

Perhaps when migrating the `purchase_location_by_line` module to v10 then we might apply this approach.

JordiBForgeFlow:
Thanks @mreficent for your proposal.

LoisRForgeFlow (approved):
Code review. LGTM :+1:

gmeficent (approved):
Code and functional review :+1:

PR #411, opened 2017-06-14
[ADD][8.0]product_by_supplir: Add active field

maurochip:
Added active field related to the product active field.
I think it is useful because we will only see the active
products.


pedrobaeza:
This affects the lines shown on Suppliers list, so looking at the inactive product will shown the list empty. You should put other name and add a filter (maybe default filter) with the field, but not call it `active`.

gurneyalex:

PR #412, opened 2017-06-15
[9.0][IMP] purchase_request: add cancelled state to lines.

LoisRForgeFlow:
related to https://github.com/OCA/purchase-workflow/issues/396.

A new state is added to lines to enhance the behavior and flow of purchase_request and its submodules.

### DIFF
--- a/purchase_request/views/purchase_request_view.xml
+++ b/purchase_request/views/purchase_request_view.xml
@@ -82,7 +82,7 @@
                                     <field name="analytic_account_id"
                                            groups="analytic.group_analytic_accounting"/>
                                     <field name="date_required"/>
-                                    <field name="cancelled"/>
+                                    <field name="cancelled" invisible="1"/>
                                     <field name="is_editable" invisible="1"/>
                                 </tree>
                                 <form>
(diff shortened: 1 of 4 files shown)


LoisRForgeFlow:
cc @benwillig @aheficent @jbeficent @sbidoul @gva-acsone

AaronHForgeFlow (approved):
:+1: I had to uninstall purchase_request_procurement to cancel the procurement and test properly

purchase_request/views/purchase_request_view.xml
```diff
@@ -249,6 +252,7 @@
                     <field name="analytic_account_id"
                            groups="analytic.group_analytic_accounting"/>
                     <field name="supplier_id"/>
+                    <field name="cancelled"/>
```
benwillig: @lreficent Can we make the field "cancelled" invisible in all tree view? The line is already grey when it has been cancelled so I guess we can hide it.

benwillig (changes requested):

purchase_request/views/purchase_request_view.xml
LoisRForgeFlow: I'm not sure that the user would realize about why the line is gray, that's why I prefer to show the field but yes, maybe you are right and I have to trust more the user :sweat_smile:

benwillig (approved):

MiquelRForgeFlow (approved):
LGTM :+1:

JordiBForgeFlow (approved):

MiquelRForgeFlow (commented):
Please add migration scripts :octocat:

PR #414, opened 2017-06-16
[9.0][IMP]purchase_request_to_procurement: adapt to new cancelled state in lines.

LoisRForgeFlow:
Depends on https://github.com/OCA/purchase-workflow/pull/412. and related to https://github.com/OCA/purchase-workflow/issues/396.

### DIFF
--- a/purchase_request_to_procurement/models/purchase_request.py
+++ b/purchase_request_to_procurement/models/purchase_request.py
@@ -75,6 +75,7 @@ class PurchaseRequestLine(models.Model):
             'product_qty': self.product_qty,
             'product_uom': self.product_uom_id.id,
             'date_planned': self.date_required,
+            'request_id': self.request_id.id,
         }
         return vals
 
--- a/purchase_request/models/__init__.py
+++ b/purchase_request/models/__init__.py
@@ -3,3 +3,4 @@
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl-3.0).
 
 from . import purchase_request
+from . import procurement_order
--- a/purchase_request/models/procurement_order.py
+++ b/purchase_request/models/procurement_order.py
@@ -1,0 +1,13 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Eficent Business and IT Consulting Services S.L.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl-3.0).
+
+from openerp import fields, models
+
+
+class Procurement(models.Model):
+    _inherit = 'procurement.order'
+
+    request_id = fields.Many2one(
+        comodel_name='purchase.request', ondelete='restrict',
+        string='Latest Purchase Request', copy=False)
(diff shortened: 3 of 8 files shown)


MiquelRForgeFlow (approved):
LGTM :+1:

AaronHForgeFlow (approved):
:+1: Thanks!

purchase_request_to_procurement/models/purchase_request.py
```diff
@@ -73,6 +73,7 @@ def _prepare_procurement_order_values(
             'product_qty': self.product_qty,
             'product_uom': self.product_uom_id.id,
             'date_planned': self.date_required,
+            'request_id': self.request_id.id,
```
benwillig, line 78: Just an idea, but maybe we could link the purchase request line directly to the procurement. What do you think @lreficent ? If we have a purchase request which contains two lines with the same product, we can't know what is the true origin of the procurement orders.

And I think that ```request_id``` field is declared in ```purchase_request_procurement``` so we can't use it here.

LoisRForgeFlow: The purchase request lines have a field `procurement_id` that unequivocally gives you the origin. While in the procurement order the link is with the last purchase request created/modified where the line was added, this is like this just to follow the same behavior as with purchase orders (the link in the procurement is to a PO not a PO line).

For your second point, I had already realized of that, but If you use this module alone without `purchase_request_procurement` the only issue is that in the logs a warning would appear saying that the value 'request_id' is useless (kind of, don't know now the exact msg :wink:). But when you use this module in conjunction with `purchase_request_procurement` it makes sense. I rather prefer this than to add a new dependency just for this.

What do you think?

benwillig: Yes you are right. The link already exists, I missed that.

For the second point, I don't like to set a value on a field which might not exist ^^, in my opinion this is a programming error (even if there is only a warning). But as I said, this is just my opinion. And ```purchase_request_to_procurement``` should not inherit from ```purchase_request_procurement``` so we can't add this dependency.

LoisRForgeFlow: One solution that comes to my mind is to move the definition of `request_id` field from `purchase_request_procurement` to `purchase_request`.

benwillig: Yes indeed, this could be a solution.

LoisRForgeFlow: Some more people agreeing to do this?

cc @jbeficent @sbidoul

LoisRForgeFlow: Finally I moved the field definition to purchase_request.

LoisRForgeFlow:
@benwillig Could you update your review? I think this need to move forward

benwillig (approved):
LGTM :+1:

PR #415, opened 2017-06-21
[10.0][MIG] purchase_location_by_line

MiquelRForgeFlow:
Standard migration.


rami-wafaie (approved):
:+1:

pedrobaeza:
Isn't going to be this change to use pull rules?

purchase_location_by_line/views/purchase_view.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0"?>
+<odoo>
+
+    <record id="purchase_order_form" model="ir.ui.view">
+        <field name="name">purchase.order.form</field>
+        <field name="model">purchase.order</field>
+        <field name="inherit_id" ref="purchase.purchase_order_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="/form/sheet/notebook/page/field/tree/field[@name='date_planned']" position="after">
+                <field name="location_dest_id"/>
```
rousseldenis, line 10: @mreficent There is an inconsistency here but maybe it's wanted.
On your purchase order, you have the picking type (Deliver To) field which belongs to a warehouse.
You have to define a domain here to restrict location selection according to that 'Deliver To' field.

rousseldenis (changes requested):
Code review + test on runbot

purchase_location_by_line/views/purchase_view.xml
MiquelRForgeFlow: It's wanted, and you should consider that a domain is already defined as <code>[('usage', 'in', ['internal', 'transit'])]</code>.

rousseldenis: @mreficent I saw that one.

What I wanted to say is there can be an inconsistency between the location**s** defined on each purchase order line and the 'Deliver To' field on purchase order.

It can be weird for user to have 'Deliver To' : 'Chicago/Stock' and one line with 'Chicago/Stock' and one line with 'WH/Stock' for example (I reproduced it on runbot) as both locations are in different warehouses...

MiquelRForgeFlow: Yes, but that's on purpose. The usecase would be for instance a PO with lines that must be sent to 'Location/receptions' and 'Different_parent_locations/office'. It can be confusing, that's right, but it's the better approach we found last time.

rousseldenis: Ok, I understand. This should be clearer in documentation (as I understood first it was to dispatch products on several different children locations - like putaway strategies).

LoisRForgeFlow: @mreficent please add some lines to describe this in the documentation as suggested, I agree that might be a little confusing.

MiquelRForgeFlow: Done

rousseldenis (approved):

purchase_location_by_line/README.rst
```diff
@@ -0,0 +1,54 @@
...
+Purchase Location by Line
+=========================
+
+Allows to define a specific destination location on each Purchase Order line.
+When the PO is confirmed, it will generate one Incoming Shipment per
+combination of destination location and expected date indicated in the
+Purchase Order Lines.
+
+This module is used to send the products to different locations, that could
+not be children of the default location of the same PO picking type. Not be confused.
```
AaronHForgeFlow: Do you mean that users can only put children locations of the default location of the picking type?

MiquelRForgeFlow: Reading comprehension :astonished:

AaronHForgeFlow: :stuck_out_tongue:

LoisRForgeFlow:
runbot life is shorter than a fly's lately :pensive: Forced rebuilt

JordiBForgeFlow (approved):

LoisRForgeFlow (approved):
Tested on runbot :+1:

PR #419, opened 2017-06-22
[9.0][IMP] purchase_request_to_rfq: cancelating procurements in hands of purchase_request

LoisRForgeFlow:
The aim of this PR is to modify the cancellation of purchase orders (POs) to not propagate the cancel signal to related procurements and the chain of moves if the origin of a PO line is a purchase request. 

This scenario only applies if you have also `purchase_request_procurement`installed. To complete the cancellation as it is done by default the user has to also cancel the purchase request with the fixes proposed in https://github.com/OCA/purchase-workflow/pull/398

### DIFF
--- a/purchase_request_to_rfq/models/purchase_order.py
+++ b/purchase_request_to_rfq/models/purchase_order.py
@@ -85,8 +85,7 @@ class PurchaseOrder(models.Model):
         """
         po_from_request = self.filtered(
             lambda po: po.mapped('order_line.purchase_request_lines'))
-        remaining_po = self.filtered(
-            lambda po: not po.mapped('order_line.purchase_request_lines'))
+        remaining_po = self - po_from_request
         po_from_request.cancel_po_from_request()
         return super(PurchaseOrder, remaining_po).button_cancel()
 
@@ -124,10 +123,10 @@ class PurchaseOrder(models.Model):
                     req.message_post(body=message, subtype='mail.mt_comment')
             # Search for lines not created from a Purchase request and end
             # the process
-            no_request_po_lines = order.order_line.filtered(
-                lambda pol: not pol.purchase_request_lines)
             # TODO: remove PO from procurement ?!
             if not self.env.context.get('cancel_procurement'):
+                no_request_po_lines = order.order_line.filtered(
+                    lambda pol: not pol.purchase_request_lines)
                 procurements = no_request_po_lines.mapped('procurement_ids')
                 procurements.filtered(lambda r: r.state not in (
                     'cancel', 'exception') and r.rule_id.propagate).write(
--- a/purchase_request_to_rfq/models/purchase_order.py
+++ b/purchase_request_to_rfq/models/purchase_order.py
@@ -185,3 +185,22 @@ class PurchaseOrderLine(models.Model):
              )
         message += '</ul>'
         return message
+
+    @api.multi
+    def _po_line_unlink_purchase_request_message_content(self):
+        self.ensure_one()
+        title = _('Line %s of Purchase Order %s was removed') % (
+            self.name, self.order_id.name)
+        message = '<b>%s</b><br/>' % title
+        return message
+
+    @api.multi
+    def unlink(self):
+        for line in self:
+            message = \
+                line._po_line_unlink_purchase_request_message_content()
+            requests = line.purchase_request_lines.mapped('request_id')
+            for req in requests:
+                req.message_post(body=message, subtype='mail.mt_comment')
+
+        return super(PurchaseOrderLine, self).unlink()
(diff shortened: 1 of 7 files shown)


AaronHForgeFlow (commented):
Tested on runbot. Works smoothly :cool: 
Would you add some tests?

LoisRForgeFlow:
@aheficent Yes, it is in my to do list :sweat_smile:

JordiBForgeFlow (approved):

JordiBForgeFlow:
@gva-acsone @benwillig Do you agree? Once merged we plan to inmediately forward the proposed changes to 10.0

lreficent:
Also related to issue https://github.com/OCA/purchase-workflow/pull/419

LoisRForgeFlow:
Also related to issue https://github.com/OCA/purchase-workflow/issues/396

LoisRForgeFlow:
To Do: 
* [ ] : Add more tests.
* [x] : Bump version.
* [x] : Update README.

LoisRForgeFlow:
Test are working in my local, don't know why they are failling in this built :confused:

purchase_request_to_rfq/models/purchase_order.py
```diff
@@ -75,6 +76,71 @@ def button_confirm(self):
...
+    def button_cancel(self):
+        """Filters the Purchase Orders (POs):
+        * POs with purchase request(s) related go to
+        cancel_po_from_request method.
+        * POs with no purchase request related are returned to super.
+        """
+        po_from_request = self.filtered(
+            lambda po: po.mapped('order_line.purchase_request_lines'))
+        remaining_po = self.filtered(
+            lambda po: not po.mapped('order_line.purchase_request_lines'))
```
benwillig: We loop twice to test the same condition. Maybe we can use following?

```python
remaining_po = self - po_from_request
```

purchase_request_to_rfq/models/purchase_order.py
```diff
@@ -75,6 +76,71 @@ def button_confirm(self):
...
+                lambda pol: pol.purchase_request_lines)
+            for line in request_po_lines:
+                message = \
+                    line._po_line_purchase_request_cancel_message_content()
+                requests = line.purchase_request_lines.mapped('request_id')
+                for req in requests:
+                    req.message_post(body=message, subtype='mail.mt_comment')
+            # Search for lines not created from a Purchase request and end
+            # the process
+            no_request_po_lines = order.order_line.filtered(
```
benwillig: I think that this line should be under
 ```python 
if not self.env.context.get('cancel_procurement'):
```

File: purchase_request_to_rfq/models/purchase_order.py [hunk 3]
```diff
@@ -75,6 +76,71 @@ def button_confirm(self):
...
+            if not self.env.context.get('cancel_procurement'):
+                procurements = no_request_po_lines.mapped('procurement_ids')
+                procurements.filtered(lambda r: r.state not in (
+                    'cancel', 'exception') and r.rule_id.propagate).write(
+                    {'state': 'cancel'})
+                procurements.filtered(lambda r: r.state not in (
+                    'cancel', 'exception') and not r.rule_id.propagate).write(
+                    {'state': 'exception'})
+                moves = procurements.filtered(
+                    lambda r: r.rule_id.propagate).mapped('move_dest_id')
```
benwillig, line 138: You loop 3 times on ```procurements```. I don't know if the performance loss is significative. What do you think @sbidoul?

purchase_request_to_rfq/models/purchase_order.py
LoisRForgeFlow: that's a good one, I'll add!

purchase_request_to_rfq/models/purchase_order.py
LoisRForgeFlow: sure, you're right!

MiquelRForgeFlow:
The tests are failing...

LoisRForgeFlow:
@mreficent I know they are, but at the moment I had to keep this aside. As soon as I have some spare time I will come back to this. If you can't wait, feel free to update the tests yourself, it would be appreciated. :smiley:

lreficent:
This is very weird, I was trying to fix the tests but they were actually running OK in my local. Once pushed, indeed, travis fails because of tests. Looking into it in a [non-practical way](https://github.com/OCA/purchase-workflow/pull/419/commits/25ea8644339fa23c626405749a30bd6d4f3a6231) I found that a procurement is not creating any PO line, which should (totally standard) and in fact occurs even if I do manually through OCA runbot... I really have no idea how to get to the issue and adding commits and waiting for CI is definitely not efficient way to solve it... :sweat_smile:

LoisRForgeFlow:
@aheficent @benwillig travis is now green, could you update your reviews?

MiquelRForgeFlow (approved):

AaronHForgeFlow (approved):
:+1: Thank you!

PR #422, opened 2017-06-26
[10.0][MIG] purchase_date_planned_manual

MiquelRForgeFlow:
Standard migration.


elicoidal (approved):

purchase_date_planned_manual/__openerp__.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
```
LoisRForgeFlow: rename to `__manifest__.py` please

LoisRForgeFlow:
I just realized of one issue:
1. Create procurement for imac with scheduled date July 1st. PO assigned.
2. Create procurement for imac with scheduled date July 7th. Same PO is assigned but the line is merged with the previous one (so scheduled date in PO will be July 1st).

cc @jbeficent

MiquelRForgeFlow:
@lreficent , to fix this issue you mentioned, we would need to add the full method `make_po` defined in `purchase` module with a minor change, but this method is too long.

May Odoo accepts to improve the method: https://github.com/odoo/odoo/pull/18467

purchase_date_planned_manual/models/purchase_order.py
```diff
@@ -0,0 +1,67 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Eficent Business and IT Consulting Services S.L.
+#   (http://www.eficent.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models, _
+from datetime import datetime
```
congdpt: All python native imports first, then odoo
See OCA contributing guidelines

purchase_date_planned_manual/tests/test_date_planned_manual.py
```diff
@@ -0,0 +1,104 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Eficent Business and IT Consulting Services S.L.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.tests.common import TransactionCase
+from odoo import fields
```
congdpt: All python native imports first, then odoo
See OCA contributing guidelines

congdpt (changes requested):

AaronHForgeFlow:
@mreficent Maybe making this module depending on https://github.com/OCA/purchase-workflow/tree/9.0/procurement_purchase_no_grouping (after its migration) would solve the issue commented by @lreficent

lasley:
Any updates on this?

LoisRForgeFlow:
@lasley We are still waiting some reaction [here](https://github.com/odoo/odoo/pull/18467), but maybe we have to start thinking about moving forward w/out it :sweat_smile:

lasley:
@lreficent - Thanks for the update. Maybe we'll give that one a few more days now that we've pinged on it. Realistically though, it's an IMP, so I don't think the likelihood of merge on a production branch is high.

LoisRForgeFlow:
@mreficent @lasley As they don't agree to add the hook... which do you think would be the better approach?

lasley:
@lreficent What about https://github.com/OCA/purchase-workflow/pull/422#issuecomment-328048901?

LoisRForgeFlow:
@lasley @aheficent Not sure, that could be a functional fix for sure, but I don't like the idea to add `procurement_purchase_no_grouping` as a dependency and use a script to add the 'No line grouping' option to all product groups. Also, I would be ok to merge lines with same planned date I think.

hparfr (approved):
Thanks

purchase_date_planned_manual/models/purchase_order.py
```diff
@@ -3,11 +3,11 @@
 #   (http://www.eficent.com)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
-from openerp import api, fields, models, _
 from datetime import datetime
 from dateutil.relativedelta import relativedelta
-from openerp.tools import DEFAULT_SERVER_DATETIME_FORMAT
-from openerp.exceptions import UserError
+from odoo import api, fields, models, _
+from odoo.tools.misc import DEFAULT_SERVER_DATETIME_FORMAT
```
pedrobaeza: Use `fields.Datetime.from_string()` instead

pedrobaeza (changes requested):

purchase_date_planned_manual/models/purchase_order.py
MiquelRForgeFlow: done

LoisRForgeFlow (approved):
:+1:

bealdav (approved):

congdpt (approved):

purchase_date_planned_manual/models/purchase_order.py
```diff
@@ -0,0 +1,63 @@
...
+    def action_delayed_line(self):
+        raise UserError(_('This line is scheduled for: %s. \n However it is '
+                          'now planned to arrive late.') % self.date_planned)
+
+
+class PurchaseOrder(models.Model):
+    _inherit = 'purchase.order'
+
+    @api.multi
+    def _inverse_date_planned(self):
```
rousseldenis: @mreficent You can remove as https://github.com/odoo/odoo/commit/a572b8115cdaa3a28f5c977a6006ddfa95647987

purchase_date_planned_manual/models/purchase_order.py
```diff
@@ -0,0 +1,63 @@
...
+            order_date = fields.Datetime.from_string(line.order_id.date_order)
+            po_lead = line.order_id.company_id.po_lead
+            delta = po_lead + seller.delay if seller else po_lead
+            date_expected = fields.Datetime.to_string(
+                order_date + relativedelta(days=delta))
+            line.predicted_arrival_late = (
+                date_expected > line.date_planned and line.order_id.state ==
+                'draft')
+
+    predicted_arrival_late = fields.Boolean(
```
rousseldenis: @mreficent Please put field definition on top and quote compute method

rousseldenis (changes requested):
Code review

purchase_date_planned_manual/models/purchase_order.py
MiquelRForgeFlow: why? aesthetic change?

rousseldenis: No I'm not so mad :smile:  but because of https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#models

rousseldenis:
By the way I will propose a PR on Odoo to recompute date_planned on lines if date_order is changed (onchange) on draft PO and if purchase lines have no procurement.

rousseldenis:
https://github.com/odoo/odoo/pull/26697

MiquelRForgeFlow:
@rousseldenis we should wait odoo feedback of your PR or we could merge this PR as it has already 5 approvals?

rousseldenis:
Of course not. It was for information purpose. We can move on this

rousseldenis (approved):

hparfr:
Is it ok for the merge @pedrobaeza  ? Thanks

pedrobaeza (approved):

PR #416, opened 2017-06-21
[9.0][IMP] purchase_request: add migration script

LoisRForgeFlow:
Adds a post-migration script to set to cancelled all old purchase request lines belonging to a purchase request in state 'rejected'.

Improves https://github.com/OCA/purchase-workflow/pull/412

PR #424, opened 2017-06-29
[10.0][IMP] purchase request cancelled state

### LINKED ISSUE #396: [9.0][purchase_request_procurement] Procurement not cancelled when a purchase request is cancelled

benwillig:
Fixes #396.

From the work of #412 and #416.

### DIFF
--- a/purchase_request/views/purchase_request_view.xml
+++ b/purchase_request/views/purchase_request_view.xml
@@ -251,7 +251,7 @@
             <field name="model">purchase.request.line</field>
             <field name="priority" eval="20"/>
             <field name="arch" type="xml">
-                <form string="Purchase Request Line" version="7.0">
+                <form string="Purchase Request Line">
                     <sheet>
                         <h1>
                             <field name="request_id"/>
--- a/purchase_request/views/purchase_request_view.xml
+++ b/purchase_request/views/purchase_request_view.xml
@@ -2,7 +2,7 @@
 <!-- Copyright 2016 Eficent Business and IT Consulting Services S.L.
      License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl-3.0) -->
 <odoo>
-    <data>
+
         <record model="ir.ui.view" id="view_purchase_request_form">
             <field name="name">purchase.request.form</field>
             <field name="model">purchase.request</field>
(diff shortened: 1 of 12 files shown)


benwillig:
@lreficent Your review is welcome.

LoisRForgeFlow (approved):
LGTM :+1: 

Thanks for the work!

benwillig:
@lreficent I added a commit. I think we should not copy cancelled field when duplicating a line.

LoisRForgeFlow:
@benwillig mmm I was about to tell that it wasn't very important since the only way to duplicate lines is duplicating the PR itself, but you can actually duplicate a line :open_mouth:. I think that shouldn't be possible, don't  you think? every created/new line must pass through the approval mechanism.

benwillig:
The case when you only duplicate the line may not appear often. But as you said, when you duplicate a PR, it copies the lines. So if a user duplicate a rejected PR, PR lines shouldn't be cancelled.

LoisRForgeFlow:
@benwillig this is fixed in v9 here: https://github.com/OCA/purchase-workflow/pull/428

MiquelRForgeFlow (approved):
LGTM :+1:

ThomasBinsfeld (approved):

purchase_request/views/purchase_request_view.xml
```diff
@@ -247,7 +251,7 @@
             <field name="model">purchase.request.line</field>
             <field name="priority" eval="20"/>
             <field name="arch" type="xml">
-                <form string="Purchase Request Line" version="7.0">
+                <form string="Purchase Request Line">
```
LoisRForgeFlow, line 254: I would update to `<form string="Purchase Request Line" duplicate="false">`

LoisRForgeFlow (commented):
The fix related to my https://github.com/OCA/purchase-workflow/pull/424#issuecomment-312221065 is partially attended. I point what is missing in this review

PR #430, opened 2017-07-13
[9.0][ADD] procurement_purchase_no_grouping

pedrobaeza:
No grouping on procurement purchases
====================================

This module allows to not group generated purchase orders from procurements. The grouping behaviour can be configurable at product category level.

Configuration
=============

Go to each product category, and select one of these values in the field "Procured purchase grouping":

* *Standard grouping (default)*: With this option, procurements will generate purchase orders as always, grouping lines and orders when possible.
* *No line grouping*: With this value, if there are any open purchase order for the same supplier, it will be reused, but lines won't be merged.

It comes from odoomrp-wip project.

cc @Tecnativa


carlosdauden (approved):
:+1: LGTM

sergio-teruel (approved):

LoisRForgeFlow (approved):
Functional + code review :+1:

elicoidal (approved):

PR #432, opened 2017-07-19
[ADD] purchase_triple_discount: New module

chienandalu:
Depends on:

- [x] account_invoice_triple_discount: https://github.com/OCA/account-invoicing/pull/294
- [x] purchase_discount: https://github.com/OCA/purchase-workflow/pull/316

Purchase Triple Discount
========================

This module allows to have three successive discounts on every purchase order line.

Usage
=====

Create a new purchase order and add discounts in any of the three discount fields given. They go in order of precedence so discount 2 will be calculated over discount 1 and discount 3 over the result of discount 2. For example, let's divide by two on every discount:

Unit price: 600.00 ->

  - Disc. 1 = 50% -> Amount = 300.00
  - Disc. 2 = 50% -> Amount = 150.00
  - Disc. 3 = 50% -> Amount = 75.00

You can also use negative values to charge instead of discount:

Unit price: 600.00 ->

  - Disc. 1 = 50% -> Amount = 300.00
  - Disc. 2 = -5% -> Amount = 315.00

cc @Tecnativa

### DIFF
--- a/purchase_triple_discount/models/purchase_order.py
+++ b/purchase_triple_discount/models/purchase_order.py
@@ -1,0 +1,42 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Tecnativa - David Vidal
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+from odoo.addons import decimal_precision as dp
+
+
+class PurchaseOrderLine(models.Model):
+    _inherit = "purchase.order.line"
+
+    @api.depends('discount2', 'discount3')
+    def _compute_amount(self):
+        super(PurchaseOrderLine, self)._compute_amount()
+
+    discount2 = fields.Float(
+        'Disc. 2 (%)',
+        digits=dp.get_precision('Discount'),
+        default=0.0,
+    )
+
+    discount3 = fields.Float(
+        'Disc. 3 (%)',
+        digits=dp.get_precision('Discount'),
+        default=0.0,
+    )
+
+    _sql_constraints = [
+        ('discount2_limit', 'CHECK (discount2 <= 100.0)',
+         'Discount 2 must be lower than 100%.'),
+        ('discount3_limit', 'CHECK (discount3 <= 100.0)',
+         'Discount 3 must be lower than 100%.'),
+    ]
+
+    def _get_discounted_price_unit(self):
+        price_unit = super(
+            PurchaseOrderLine, self)._get_discounted_price_unit()
+        if self.discount2:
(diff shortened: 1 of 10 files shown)


purchase_triple_discount/models/purchase_order_line.py
```diff
@@ -0,0 +1,59 @@
...
+            # Keep these values for restore later
+            price = line.price_unit
+            discount = line.discount
+            line.price_unit *= (1 - (line.discount or 0.0) / 100.0)
+            line.price_unit *= (1 - (line.discount2 or 0.0) / 100.0)
+            line.price_unit *= (1 - (line.discount3 or 0.0) / 100.0)
+            # Avoid recalculate discount
+            line.discount = 0.0
+            # Call a super that uses price_unit to compute subtotal
+            call_super()
```
StefanRijnhart: Seems to me you are calling the super on all the lines every time.

chienandalu: Is what I want to. I need to make sure that's the object I'm sending to super. Otherwise I was having weird cache issues (fields reseting values) here: https://github.com/odoo/odoo/blob/10.0/addons/purchase/models/purchase.py#L517

Not to mention that I need to bridge purchase_discount methods so discount is not calculated twice.

Maybe I could convert this into a singleton method and loop it (or not) from the outside

pedrobaeza:
Please check https://github.com/OCA/purchase-workflow/pull/316 for seeing how I have finally made this to work and use the same approach.

chienandalu:
@pedrobaeza Way better now. I just had to adjust the cases when no discount is set but discount2 o discount3 are.

purchase_triple_discount/models/purchase_order.py
```diff
@@ -0,0 +1,42 @@
...
+
+class PurchaseOrderLine(models.Model):
+    _inherit = "purchase.order.line"
+
+    @api.depends('discount2', 'discount3')
+    def _compute_amount(self):
+        super(PurchaseOrderLine, self)._compute_amount()
+
+    discount2 = fields.Float(
+        'Discount 2 (%)',
```
pedrobaeza: For making this label smaller to fit in the editable tree view, better to put `Disc. 2 (%)`

pedrobaeza (approved):
Little nitpicking.

chienandalu:
@pedrobaeza Done :smile:

pedrobaeza:
Please rebase now that the dependencies are merged.

chienandalu:
@pedrobaeza All green now

pedrobaeza:
@luismontalba please review

luismontalba (approved):
Tested on runbot

PR #433, opened 2017-07-19
[9.0][REF]purchase_request: subscribe assigned_to user using a separate method

zakiuu:
Currently in the purchase request module, when assigning a user to `assigned_to` field, that user will be subscribed automatically to the document. This behavior is done by overriding the `create` and `write` method using `message_subscribe_users()`.
The issue with this implementation is the `assigned_to user`  is subscribed to all the subtypes of the purchase request and it doesn't give us an override point to change that behavior.
As a solution for this issue, I propose to use a separate method to subscribe the` assigned_to`  user called `_subscribe_assigned_to_user` and call that method in the `create` and `write` methods. This way we can easily change the subscription behavior by overriding this new method.

### DIFF
--- a/purchase_request/models/purchase_request.py
+++ b/purchase_request/models/purchase_request.py
@@ -178,7 +178,8 @@ class PurchaseRequest(models.Model):
         for rec in self:
             if not rec.assigned_to:
                 continue
-            rec.message_subscribe_users(user_ids=[rec.assigned_to.id])
+            rec.message_subscribe_users(
+                user_ids=[rec.assigned_to.id], subtype_ids=None)
 
 
 class PurchaseRequestLine(models.Model):
--- a/purchase_request/models/purchase_request.py
+++ b/purchase_request/models/purchase_request.py
@@ -122,15 +122,14 @@ class PurchaseRequest(models.Model):
     def create(self, vals):
         request = super(PurchaseRequest, self).create(vals)
         if vals.get('assigned_to'):
-            request.message_subscribe_users(user_ids=[request.assigned_to.id])
+            request._subscribe_assigned_to_user()
         return request
 
     @api.multi
     def write(self, vals):
         res = super(PurchaseRequest, self).write(vals)
-        for request in self:
-            if vals.get('assigned_to'):
-                self.message_subscribe_users(user_ids=[request.assigned_to.id])
+        if vals.get('assigned_to'):
+            self._subscribe_assigned_to_user()
         return res
 
     @api.multi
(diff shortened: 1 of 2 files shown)


sbidoul (approved):

ThomasBinsfeld (approved):
LGTM

LoisRForgeFlow (approved):
LGTM

purchase_request/models/purchase_request.py
```diff
@@ -173,6 +172,14 @@ def check_auto_reject(self):
             if not pr.line_ids.filtered(lambda l: l.cancelled is False):
                 pr.write({'state': 'rejected'})
 
+    @api.multi
+    def _subscribe_assigned_to_user(self):
```
lmignon: @zakiuu It would be nice to add the parameter `subtype_ids=None` since it's also supported by `message_subscribe_users`. This parameter will ease the override of this method when you want to subscribe the assigned user to specific subtypes.

lmignon (changes requested):

purchase_request/models/purchase_request.py
```diff
@@ -178,7 +178,8 @@ def _subscribe_assigned_to_user(self):
         for rec in self:
             if not rec.assigned_to:
                 continue
-            rec.message_subscribe_users(user_ids=[rec.assigned_to.id])
+            rec.message_subscribe_users(
+                user_ids=[rec.assigned_to.id], subtype_ids=None)
```
zakiuu: @lmignon Done!

purchase_request/models/purchase_request.py
lmignon: @zakiuu Thank you for the change but it's not what I requested :smirk: 
```python
@api.multi
def _subscribe_assigned_to_user(self, subtype_ids=None):
    """If the assigned to user is set on the PR, subscribe him."""
    for rec in self:
        if not rec.assigned_to:
            continue
        rec.message_subscribe_users(
            user_ids=[rec.assigned_to.id], subtype_ids=subtype_ids)
```

lmignon (changes requested):

zakiuu:
@lmignon Sorry I've miss understood, fixed!

lmignon (approved):
Thank you @zakiuu

ThomasBinsfeld (approved):
This one seems ready to merge :+1:

AaronHForgeFlow (approved):
:+1:

pedrobaeza:
Can I squash? Please change module version number

zakiuu:
@pedrobaeza version updated, and yes you can squash

PR #435, opened 2017-07-22
[9.0][FIX] procurement_purchase_no_grouping: PO total amount don't computed

carlosdauden:
@Tecnativa


pedrobaeza (changes requested):
Please change version number

pedrobaeza (approved):

rafaelbn (approved):

PR #426, opened 2017-07-07
[9.0][FIX] purchase_open_qty

MiquelRForgeFlow:
Makes it compatible with https://github.com/OCA/account-invoicing/pull/285.

PR #439, opened 2017-07-25
[10.0][MIG] purchase_open_qty

MiquelRForgeFlow:
Standard migration.

Includes:

- [x]  Cherry-pick of cec018359db643a52d2fd1f98d46f836d041c2a7 (https://github.com/OCA/purchase-workflow/pull/426)
- [x] Cherry-pick of 0d705dfe666f7f610806540aafe5c7297a4e5c4a (https://github.com/OCA/purchase-workflow/pull/447).


florian-dacosta:
Hello,
Thanks for this module, it seems usefull, however, I wonder about canceled backorders.
From what I read of the code, if we cancel a backorder of an incoming shipment, the qty_to_receive will be for ever positive on both the purchase line and the purchase order.
I think it is an issue, and with time, the number will grow. Indeed, there are some business where it is really common to not receive the exact ordered quantity.

I was wondering maybe add some code to put the qty_to_receive to 0.0 in case the purchase order state is 'done' (indeed if the user manually put the order to done state, it should mean it is totally over I guess)

Or, since the field already depends on move_ids, we could check if all the related stock.move are done or cancel => qty_to_receive would be 0, maybe it is even cleaner.

What do you think? (I would be glad to contribute this kind of improvment to the version 9 module)
@mreficent @lreficent @jbeficent

MiquelRForgeFlow:
@florian-dacosta , I think what you said is not a problem of this module. Our module is just calculating the `qty_to_receive` from the `qty_received`. And is a problem of Odoo if for any situation, including yours, the `qty_received` is wrong. Maybe you should do a PR fixing that in Odoo or do a new module.

florian-dacosta:
@mreficent I disagree, I do think it is a problem of the `qty_to_receive` field and not `qty_received
Indeed, if you have a purchase order with one line of 50 unit of a product.
You receive 35 products, and then, cancel the backorder.
We have `qty_received` = 35 => it is correct
`qty_to_receive` = 15 => not correct, it should be 0.

I guess I could do a new module on top of this one to change this behavior, and I probably will if this is not a feature wanted for this module. However, I don't see why it would be bad to be in this module?

Have you uses cases for which this behavior is not wanted?

LoisRForgeFlow:
@mreficent @florian-dacosta I agree with Florian that this has to be considered in this module, and also I rather go for the option to automatically check for pending moves. Note that in v10 the status 'done' doesn't exist anymore and it is replaced by a lock/unlock flow, in which a locked PO would not mean that no actions are pending.

MiquelRForgeFlow:
@florian-dacosta we thought it is best if qty_to_receive is just computed as the qty of the stock moves that are pending to receive. That way the data offered to the puchase user is in line with what the warehouse can see in the pickings.

For example, you purchase 10, then warehouse receives 9 and does not create a backorder but completes the incoming shipment. Qty to receive would be 0.

In the scenario that you pointed out, warehouse receives 9, creates backorder for 1, so qty_to_receive=1 and then when they cancel the backorder, qty_to_receive=0.

florian-dacosta:
Well I think it is great!
If you want, I can make a PR, at least for version 9 (But I won't be available before 7 days)

JordiBForgeFlow:
We are fixing for v9

JordiBForgeFlow:
and will cherry pick for v10

MiquelRForgeFlow:
@florian-dacosta @lreficent it's fixed now 😉

JordiBForgeFlow (approved):

LoisRForgeFlow (approved):
technical and functional review. This is great now :+1:

purchase_open_qty/models/purchase_order.py
```diff
@@ -0,0 +1,87 @@
...
+            else:
+                line.qty_to_invoice = line.product_qty - line.qty_invoiced
+
+    @api.depends('order_id.state', 'move_ids.state', 'product_qty')
+    def _compute_qty_to_receive(self):
+        for line in self:
+            total = 0.0
+            for move in line.move_ids.filtered(
+                    lambda m: m.state not in ('cancel', 'done')):
+                if move.product_uom != line.product_uom:
```
florian-dacosta, line 32: I believe this case is not possible, so it seems useless

purchase_open_qty/init_hook.py
```diff
@@ -0,0 +1,56 @@
...
+            COMMENT ON COLUMN purchase_order_line.qty_to_invoice IS
+            'Qty to Bill';
+            """)
+
+    logger.info('Computing values for fields qty_to_receive and qty_to_invoice'
+                ' on purchase_order_line')
+    cr.execute(
+        """
+        UPDATE purchase_order_line pol
+        SET qty_to_receive = pol.product_qty - pol.qty_received,
```
florian-dacosta: This is not good anymore.
You can make a query like this : 
```
UPDATE purchase_order_line SET qty_to_receive = foo.qty
FROM (SELECT purchase_line_id, sum(product_uom_qty) as qty FROM stock_move WHERE purchase_line_id IS NOT NULL AND state not in ('cancel', 'done') GROUP BY purchase_line_id) as foo
WHERE purchase_order_line.id = foo.purchase_line_id
```
So it updates all purchase line still waiting for products. 
The other lines are ignored and will have the 0.0 value, which is fine

purchase_open_qty/init_hook.py
```diff
@@ -0,0 +1,56 @@
...
+            qty_to_invoice = pol.product_qty - pol.qty_invoiced
```
florian-dacosta: It is wrong if the purchase_method of the product is `receive`
I guess you could split it in 2 queries
UPDATE purchase_order_line SET qty_to_invoice = purchase_order_line.qty_received - purchase_order_line.qty_invoiced FROM product_product p JOIN product_template t ON p.product_tmpl_id = t.id WHERE t.purchase_method = 'receive'

And another one similar but with `qty_to_invoice = pol.product_qty - pol.qty_invoiced` and `WHERE t.purchase_method != 'receive'`

florian-dacosta (changes requested):

purchase_open_qty/models/purchase_order.py
MiquelRForgeFlow: This line of code is taken from https://github.com/OCA/OCB/blob/10.0/addons/purchase/models/purchase.py#L568.

purchase_open_qty/init_hook.py
MiquelRForgeFlow: Thank you, we didn't thought in the hook's query.

florian-dacosta (approved):

purchase_open_qty/init_hook.py
```diff
@@ -0,0 +1,79 @@
...
+        WHERE t.purchase_method = 'receive'
+        """
+    )
+    cr.execute(
+        """
+        UPDATE purchase_order_line pol
+        SET qty_to_invoice = pol.product_qty - pol.qty_invoiced
+        FROM product_product p
+        JOIN product_template t ON p.product_tmpl_id = t.id
+        WHERE t.purchase_method != 'receive'
```
florian-dacosta, line 65: it missed AND pol.product_id = p.id

purchase_open_qty/init_hook.py
```diff
@@ -0,0 +1,79 @@
...
+
+    logger.info('Computing values for fields qty_to_receive and qty_to_invoice'
+                ' on purchase_order_line')
+    cr.execute(
+        """
+        UPDATE purchase_order_line pol
+        SET qty_to_invoice = pol.qty_received - pol.qty_invoiced
+        FROM product_product p
+        JOIN product_template t ON p.product_tmpl_id = t.id
+        WHERE t.purchase_method = 'receive'
```
florian-dacosta, line 56: It misses AND pol.product_id = p.id
Without it, it creates huge performance problem

florian-dacosta (commented):
Actually, I approved too fast, I see a problem now...
sorry for that

florian-dacosta:
@lreficent @mreficent Can you check this?
Same problem on v9 branch

MiquelRForgeFlow:
Ok, fixed now here: https://github.com/OCA/purchase-workflow/pull/450
Please review that.

Thanks @florian-dacosta

Issue #440, opened 2017-07-27
[BUG] purchase_request_procurement: not able to reject a request


LoisRForgeFlow:
**Scenario:**
* purchase_request_procurement and mrp_production_request are installed.
* User A is a Purchase Request (PR) User but has no access to Manufacturing Request (MR).

**Issue:**
User A is not able to reject a PR, a similar to this raises:
![image](https://user-images.githubusercontent.com/23449160/28663582-17b38068-72be-11e7-9cd6-624f641fe623.png)

The issue is related to the propagation of the cancellation through the procurement. When a PR user cancels a PR the system looks into the related procurements for the PR request lines to be cancelled, but, also for MRs. Given that the PR users have no access to even read MRs, the above error raises (regardless in the vast majority of the cases the result of this search for MRs related will be none).

This would also happen the other way around: a MR user wouldn't be able to reject a MR if he/she has no access to PR lines.

LoisRForgeFlow:
Closing as this is already solved.

PR #441, opened 2017-07-27
[9.0][FIX] purchase_request_procurement: interfering mrp_production_request

LoisRForgeFlow:
Fixes issue https://github.com/OCA/purchase-workflow/issues/440.

### DIFF
--- a/purchase_request_procurement/models/procurement.py
+++ b/purchase_request_procurement/models/procurement.py
@@ -70,13 +70,14 @@ class Procurement(models.Model):
         for procurement in self:
             # Search for purchase request lines containing the procurement_id
             #  and cancel them.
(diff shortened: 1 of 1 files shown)


AaronHForgeFlow (approved):
LGMT, just one question. Why do you need sudo()?

LoisRForgeFlow:
@aheficent have a look to issue in the description. With sudo() you let the system search and perform actions in a model even if such model is not accessible for the user that triggered the action.

AaronHForgeFlow:
I see @lreficent. Thanks

MiquelRForgeFlow (approved):

JordiBForgeFlow:
@lreficent Can you make sure that this is also fixed in 10.0?

PR #442, opened 2017-07-27
[9.0][FIX] purchase_request: only allow picking type operation 'incoming'

LoisRForgeFlow:
Purchase Request (PR) should only allow picking type with *Suppliers* as *Type of Operation*. This may avoid error when creating Purchase Orders from PRs.

cc @jbeficent

### DIFF
--- a/purchase_request/models/purchase_request.py
+++ b/purchase_request/models/purchase_request.py
@@ -2,7 +2,8 @@
 # Copyright 2016 Eficent Business and IT Consulting Services S.L.
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl-3.0).
 
-from openerp import api, fields, models
+from openerp import api, fields, models, _
+from openerp.exceptions import UserError
 import openerp.addons.decimal_precision as dp
 
 _STATES = [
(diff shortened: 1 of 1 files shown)


AaronHForgeFlow (approved):
:+1: It makes sense to me

purchase_request/models/purchase_request.py
```diff
@@ -67,6 +68,14 @@ def _track_subtype(self, init_values):
                 return 'purchase_request.mt_request_done'
         return super(PurchaseRequest, self)._track_subtype(init_values)
 
+    @api.multi
+    @api.constrains('picking_type_id')
+    def _check_picking_type_id(self):
+        for rec in self:
+            if rec.picking_type_id.code != 'incoming':
+                raise UserError(_(
```
zakiuu: Maybe ValidationError?

LoisRForgeFlow: why do you think is better?

zakiuu: it's just more significant, on a constraint a ValidationError is more appropriate than a UserError

zakiuu (approved):
LGTM

MiquelRForgeFlow (approved):

PR #443, opened 2017-08-11
[10.0][ADD] Module Purchase Order Block

rgarnau:
This module allows you to block the validation of a Purchase Order providing a block reason.

@jbeficent @etobella @aheficent @darshan-serpent

This module is still 'Work in progress'.

### DIFF
--- a/purchase_order_approval_block/models/purchase_order.py
+++ b/purchase_order_approval_block/models/purchase_order.py
@@ -1,0 +1,72 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Eficent Business and IT Consulting Services S.L.
+# Copyright 2017 Serpent Consulting Services Pvt. Ltd.
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
+
+from odoo import api, fields, models, _
+
+
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+
+    approval_block_id = fields.Many2one(
+        comodel_name='purchase.approval.block.reason',
+        string='Approval Block Reason',
+    )
+    approval_blocked = fields.Boolean(
+        'Approval Blocked',
+        compute='_compute_approval_blocked',
+    )
+
+    @api.multi
+    def _compute_approval_blocked(self):
+        for rec in self:
+            if rec.approval_block_id:
+                rec.approval_blocked = True
+
+    @api.model
+    def create(self, vals):
+        po = super(PurchaseOrder, self).create(vals)
+        if 'approval_block_id' in vals and vals['approval_block_id']:
+            po.message_post(body=_('Order \"%s\" blocked with reason'
+                                   ' \"%s\"') % (po.name,
+                                                 po.approval_block_id.name))
+        return po
+
+    @api.multi
+    def write(self, vals):
+        res = super(PurchaseOrder, self).write(vals)
+        for po in self:
+            if 'approval_block_id' in vals and vals['approval_block_id']:
+                po.message_post(
+                    body=_('Order \"%s\" blocked with reason \"%s\"') % (
+                        po.name, po.approval_block_id.name))
+            elif 'approval_block_id' in vals and not vals['approval_block_id']:
+                po.message_post(
+                    body=_('Order \"%s\" approval block released.') % po.name)
+        return res
+
+    @api.multi
+    def button_approve(self, force=False):
+        for rec in self:
+            if rec.approval_block_id:
+                rec.button_release_approval_block()
+        return super(PurchaseOrder, self).button_approve(force=force)
+
+    @api.multi
+    def button_release_approval_block(self):
+        for order in self.with_context(force_po_approval_block_release=True):
+            order.approval_block_id = False
+        return True
+
+    @api.multi
+    def _check_order_release(self):
+        self.ensure_one()
+        if self.approval_block_id:
+            return True
+
+    @api.multi
+    def button_confirm(self):
+        self.filtered(lambda o: o.state in ['draft', 'sent'] and
+                      o.approval_blocked).write({'state': 'to approve'})
+        return super(PurchaseOrder, self).button_confirm()
--- a/purchase_order_approval_block/README.rst
+++ b/purchase_order_approval_block/README.rst
@@ -1,0 +1,94 @@
+.. image:: https://img.shields.io/badge/licence-LGPL--3-blue.svg
+    :alt: License LGPL-3
+
+=============================
+Purchase Order Approval Block
+=============================
+
+This module allows you to block the approval of an RFQ when an Approval
+Block Reason has been provided. Upon confirmation of an RFQ the orders will be
+waiting for approval by a Manager.
+
+Configuration
+=============
+
+* Go to ‘Purchases / Configuration / Purchase Approval Block Reasons’ and create
+  the blocking reasons as needed, providing a name and a description. A field
+  ‘Active’ allows you to deactivate the reason if you do not plan to use it
+  any more.
+* Assign the security group 'Release blocked RFQ' to users that should be able
+  to release the block. Users in group 'Purchase / Managers' are by default
+  assigned to this group.
+
+Usage
+=====
+
+Set the Purchase Approval Block
+-------------------------------
+
+#. Go to ‘Purchases / Purchase / Requests for Quotation’
+#. Create a new RFQ and indicate the approval block reason (found in the
+   right hand side of the screen, below the order date).
+
+Search existing RFQ
+-------------------
+
+There is a filter ‘Blocked’ to search for orders that are blocked for approval.
+It is also possible to search RFQ’s with a specific block reason.
+
+Confirm the RFQ
+---------------
+
+#. Press the button ‘Confirm’. If there’s an approval block, the order will
+   be set to status 'To Approve'. You will then need to request a Purchase
+   Manager to approve it.
+
+Release the purchase approval block
+-----------------------------------
+
+While the RFQ is in draft, members of security group ‘Release blocked RFQ’,
+can see a button ‘Release Approval Block’. From this point and on, anyone
+seeing that RFQ will be able to validate it.
+
+Notifications to followers
+--------------------------
+
+Followers of the RFQ receive notifications when an approval block has been
+set or released.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/142/10.0
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/purchase-workflow/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smashing it by providing a detailed and welcomed feedback.
+
+Credits
+=======
+
+Contributors
+------------
+
+* Jordi Ballester Alomar <jordi.ballester@eficent.com>
+* Roser Garcia <roser.garcia@eficent.com>
+* Darshan Patel <darshan.patel.serpentcs@gmail.com>
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit https://odoo-community.org.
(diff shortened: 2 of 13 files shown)


purchase_order_block/README.rst
```diff
@@ -0,0 +1,102 @@
...
+* Assign the security group “Release blocked RFQ” to users that should be able
+  to release the block.
+
+
+Usage
+=====
+
+Set the purchase block
+----------------------
+# Go to ‘Purchases / Purchase / Requests for Quotation’
```
LoisRForgeFlow: You need a space above this and to use `#.` for automatic numbered lists

purchase_order_block/README.rst
```diff
@@ -0,0 +1,102 @@
...
+  it.
+
+Search existing RFQ
+-------------------
+There is a filter ‘Blocked’ to search for orders that are blocked.
+It is also possible to search RFQ’s with a specific block reason.
+
+Release the purchase block
+--------------------------
+# Access to an existing RFQ that is blocked.
```
LoisRForgeFlow: same as above.

purchase_order_block/models/purchase_order.py
```diff
@@ -0,0 +1,55 @@
...
+        else:
+            po.empty_check = True
+        return po
+
+    @api.multi
+    def button_release_block(self):
+        for order in self:
+            order.\
+                message_post(body=_('Order \"%s\" with block \"%s\" has now '
+                                    'released') % (order.name,
```
LoisRForgeFlow: "is now released" or "has been now released" right? :thinking:

purchase_order_block/models/purchase_order.py
```diff
@@ -0,0 +1,55 @@
...
+                                                   order.po_block_id.name))
+            order.released = True
```
LoisRForgeFlow: Isn't it better to just remove the `po_block_id`? setting it to false. Any order without a `po_block_id` is released

purchase_order_block/models/purchase_order.py
```diff
@@ -0,0 +1,55 @@
...
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+
+    po_block_id = fields.Many2one(
+        comodel_name='purchase.block.reason',
+        string='Purchase Block Reason')
+
+    released = fields.Boolean(copy=False, readonly=True)
+
+    empty_check = fields.Boolean(copy=False, readonly=True)
```
LoisRForgeFlow: I don't see any real benefit of using these two fields

purchase_order_block/views/purchase_order_view.xml
```diff
@@ -0,0 +1,44 @@
...
+     License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl). -->
+
+<odoo>
+    <record id="purchase_order_form" model="ir.ui.view">
+        <field name="name">purchase.order.form.block.reason</field>
+        <field name="model">purchase.order</field>
+        <field name="inherit_id" ref="purchase.purchase_order_form"/>
+        <field name="arch" type="xml">
+            <field name="date_order" position="after">
+                <field name="po_block_id" attrs="{'readonly': [('id','!=', False)], 'invisible':['|', ('released','=', True), ('empty_check','=', True)]}"/>
```
LoisRForgeFlow: I would always show the po_block_id so you can see whether it is set or not

purchase_order_block/views/purchase_order_view.xml
```diff
@@ -0,0 +1,44 @@
...
+                <field name="released" invisible='True'/>
+                <field name="empty_check" invisible='True'/>
```
LoisRForgeFlow: Showing always the po_block_id will make these two fields unneded.

LoisRForgeFlow (commented):
Code review

purchase_order_block/views/purchase_order_view.xml
serpentcs-dev1: Agree! Removing this both boolean fields will make code cleaner and faster.

`released` If released it will make the Block Reason M2O field invisible. 
and `empty_check` will make the field invisible if it is empty upon record save.

@rgarnau What do you think? do you agree to keep the `po_block_id` field visible all the time?!

rgarnau: @darshan-serpent Oke, let's leave it visible and if it is remaining blank it should be editable while the PO is in draft. This way we give more flexibility choosing a blocking rule. Could you do this?

purchase_order_block/models/purchase_order.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+
+    po_block_id = fields.Many2one(
+        comodel_name='purchase.block.reason',
+        string='Purchase Block Reason')
+
+    released = fields.Boolean(copy=False, readonly=True)
```
JordiBForgeFlow: why use this field? It is released/unblocked if `po_block_id` is null.

purchase_order_block/models/purchase_order.py
JordiBForgeFlow: I agree. Please remove them.

purchase_order_block/views/purchase_order_view.xml
JordiBForgeFlow: We only need:
` <field name="po_block_id" attrs="{'readonly': ['|', ('id','!=', False), ('state', '!=', 'draft')]"/>`

why make it invisible?

JordiBForgeFlow (changes requested):

purchase_order_block/models/purchase_order.py
serpentcs-dev1: That was the requirement earlier. But now the requested changes are already made and have sent a request to the current Eficent branch

AaronHForgeFlow:
Maybe it'd be good to show an error when trying to modify the block reason? :thinking:  

Update: Or better allow to change the block reason but adding the changes to the chatter...

purchase_order_approval_block/models/purchase_order.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+
+    approval_block_id = fields.Many2one(
+        comodel_name='purchase.approval.block.reason',
+        string='Approval Block Reason',
+    )
+    approval_blocked = fields.Boolean('Approval Blocked',
```
LoisRForgeFlow: Why again a boolean? I don't get it, is it needed for another purpose? In this module seems useless.

AaronHForgeFlow: Probable it's there for technical purposes. It's used in view's domains

LoisRForgeFlow: I see, to not be able to change ever the block reason once it is set? seems reasonble. Is that the desired behavior?

rgarnau: yes, that's it.

LoisRForgeFlow (approved):
Tested on runbot and works great! :+1:

purchase_order_approval_block/models/purchase_order.py
```diff
@@ -0,0 +1,74 @@
...
+        for rec in self:
+            if rec.approval_block_id:
+                rec.button_release_approval_block()
+        return super(PurchaseOrder, self).button_approve(force=force)
+
+    @api.multi
+    def button_release_approval_block(self):
+        for order in self:
+            order.with_context(
+                force_po_approval_block_release=True).approval_block_id = False
```
rousseldenis: Why this value in context as it is not used anywhere ?

If you want to add a parameter, I'd prefer having a function `release_approval_block` which do the assignation to False with the parameter force.

Don't forget you create a new environment at each time.

purchase_order_approval_block/models/purchase_order.py
```diff
@@ -0,0 +1,74 @@
...
+                                      compute='_compute_approval_blocked')
+
+    @api.multi
+    def _compute_approval_blocked(self):
+        for rec in self:
+            if rec.approval_block_id:
+                rec.approval_blocked = True
+
+    @api.model
+    def create(self, vals):
```
rousseldenis, line 28: Instead of using write and create, use the `_track_subtype` function of mail.thread like in https://github.com/odoo/odoo/blob/8fbee7978f154bfd7d1424709cd66796cb3d82f3/addons/account/models/account_invoice.py#L1134

purchase_order_approval_block/models/purchase_order.py
```diff
@@ -0,0 +1,74 @@
...
+
+    @api.multi
+    def _check_order_release(self):
+        self.ensure_one()
+        if self.approval_block_id:
+            return True
+
+    @api.multi
+    def button_confirm(self):
+        for order in self:
```
rousseldenis: I'd prefer using filtered in an inline function:

`self.filtered(lambda o: o.state not in ['draft', 'sent'] and o.approval_blocked).write({'state': 'to_approve'})`

rousseldenis (changes requested):
Code review.

purchase_order_approval_block/models/purchase_order.py
rgarnau: Hi @rousseldenis,
This module is used as the base of #446. For this reason and from my point of view, it is necessary that it is in the create method that we send the notifications as in the other module this gets more restrictive (notifications are not send that frequent).

purchase_order_approval_block/models/purchase_order.py
rgarnau: In this case this line does not do exactly the same as the method of the module: both conditions don't have to be true to change the state to approve.

rousseldenis: Oups, `state in ['draft', 'sent']`

purchase_order_approval_block/models/purchase_order.py
rgarnau: This is used here https://github.com/OCA/purchase-workflow/pull/446/files#diff-a03a69efc043be23d1167165b82a60ebR25

The only option to avoid this would be to add a boolean field 'Unblock Approval forced' in the PO. When a person adds a block again, this 'Unblock Approval forced' would be reset. Then we would be able to follow your recommendation, but the drawback could be that we store an extra field. On the other hand, perhaps it is a good idea for traceability purposes that you know what Orders were eventually approved, where the block was released under the forced unblock button.

If you want to further clarify the use of the modules we can jump to a Hangouts.

purchase_order_approval_block/models/purchase_order.py
rgarnau: This looks better! I've applied the changes.

rgarnau:
@rousseldenis - attended review

rgarnau:
@rousseldenis - Could you update your review?

purchase_order_approval_block/models/purchase_order.py
rousseldenis: I understand your purpose.

But in my point of view this is a little bit dirty to put such things in create and write.

In the track_subtype, why not using a method that can be surcharged to launch or not the notification ?

But it's my opinion :-)

purchase_order_approval_block/models/purchase_order.py
rousseldenis: I'm ok with this but you have to put with_context on self:

`for order in self.with_context(...):`

My remark was mainly on the new environment creation at each iteration

rgarnau: I've applied the changes. However, doing some research about with_context(), I don't see that it is really creating a new environment every time: https://www.odoo.com/documentation/saas-13/reference/orm.html

rousseldenis: Yes I'll do it

That's where Odoo is tricky with its documentation!

https://github.com/odoo/odoo/blob/8165018058f59eedecc2fa0b9bde93e505d56c53/odoo/models.py#L4890

rousseldenis (approved):
Code review

purchase_order_approval_block/models/purchase_order.py
rgarnau: Okay, I see now. Thank you for your feedback. I'll keep this in mind! :+1:

bealdav:
Nice module, thanks. 

But have you considered to use https://github.com/OCA/server-tools/tree/10.0/base_exception to reach this feature ?

base_exception is used by sale_exception.

Sorry I have seen your module too late. FYI

rgarnau:
Hi @bealdav 
I didn't know about this module but it looks very interesting! Thanks, I will keep this in mind for further implementations :+1:

bealdav:
OK, but could put your mind also in the README § Roadmap ? :-)

Example: "Before v11 migration consider to use base_exception module and create picking_exception module (similarly to sale_exception) instead of this one"

PR #445, opened 2017-08-28
[10.0][FIX] purchase_request: duplicate lines should be disallowed.

LoisRForgeFlow:
Fixes what was missing after merging in https://github.com/OCA/purchase-workflow/pull/424. See https://github.com/OCA/purchase-workflow/pull/424#pullrequestreview-58956754.

cc @jbeficent @mreficent @benwillig @ThomasBinsfeld

### DIFF
--- a/purchase_request/views/purchase_request_view.xml
+++ b/purchase_request/views/purchase_request_view.xml
@@ -251,7 +251,7 @@
             <field name="model">purchase.request.line</field>
             <field name="priority" eval="20"/>
             <field name="arch" type="xml">
-                <form string="Purchase Request Line">
+                <form string="Purchase Request Line" duplicate="false">
                     <sheet>
                         <h1>
                             <field name="request_id"/>


benwillig:
@lreficent  I'm not really sure it's needed. The purpose of this fix was to avoid to copy a line that has been cancelled ?

LoisRForgeFlow:
@benwillig  In the comment I pointed out in my review the important part is this one:

> you can actually duplicate a line :open_mouth:. I think that shouldn't be possible, don't you think? every created/new line must pass through the approval mechanism.

It has nothing to do with the cancelled state, I'm just saying that a line cannot be created directly in any way, and duplicating was a way to do so. Therefore, I think that we should not allow it in the purchase request lines view.

MiquelRForgeFlow (approved):

ThomasBinsfeld (approved):

JordiBForgeFlow (approved):

PR #447, opened 2017-08-31
[9.0][FIX][purchase_open_qty] fixing methods

MiquelRForgeFlow:
This PR fixes:
- method `qty_to_invoice` for cases which the purchase_method is 'receive'.
- method `qty_received` is changed to `qty_to_receive`, and calculated directly from the stock_moves associated.
- added migration script for the changes produced in the init_hook

### DIFF
--- a/purchase_open_qty/models/purchase_order.py
+++ b/purchase_open_qty/models/purchase_order.py
@@ -10,23 +10,38 @@ import openerp.addons.decimal_precision as dp
 class PurchaseOrderLine(models.Model):
     _inherit = 'purchase.order.line'
 
-    @api.depends('product_qty', 'qty_invoiced')
+    @api.depends('product_qty', 'qty_invoiced',
+                 'invoice_lines.invoice_id.state')
     def _compute_qty_to_invoice(self):
         for line in self:
-            line.qty_to_invoice = line.product_qty - line.qty_invoiced
+            if line.product_id.purchase_method == 'receive':
+                qty = line.qty_received - line.qty_invoiced
+                if qty >= 0.0:
+                    line.qty_to_invoice = qty
+                else:
+                    line.qty_to_invoice = 0.0
+            else:
+                line.qty_to_invoice = line.product_qty - line.qty_invoiced
 
     @api.depends('order_id.state', 'move_ids.state', 'product_qty')
-    def _compute_qty_received(self):
-        super(PurchaseOrderLine, self)._compute_qty_received()
+    def _compute_qty_to_receive(self):
         for line in self:
-            line.qty_to_receive = line.product_qty - line.qty_received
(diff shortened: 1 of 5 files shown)


LoisRForgeFlow (approved):
Tested on runbot :+1:

purchase_open_qty/models/purchase_order.py
```diff
@@ -10,23 +10,36 @@
...
 
-    @api.depends('product_qty', 'qty_invoiced')
+    @api.depends('product_qty', 'qty_invoiced',
+                 'invoice_lines.invoice_id.state')
     def _compute_qty_to_invoice(self):
         for line in self:
-            line.qty_to_invoice = line.product_qty - line.qty_invoiced
+            if line.product_id.purchase_method == 'receive':
+                qty = line.qty_received - line.qty_invoiced
+                if qty >= 0.0:
```
AaronHForgeFlow, line 19: Why this check?

florian-dacosta (approved):

JordiBForgeFlow (approved):

florian-dacosta (approved):
Thanks for the reactivity

JordiBForgeFlow (approved):

purchase_open_qty/models/purchase_order.py
```diff
@@ -10,23 +10,38 @@
 class PurchaseOrderLine(models.Model):
     _inherit = 'purchase.order.line'
 
-    @api.depends('product_qty', 'qty_invoiced')
+    @api.depends('product_qty', 'qty_invoiced',
+                 'invoice_lines.invoice_id.state')
```
florian-dacosta, line 14: In case of purchase_method = 'receive', it won't recompute.
I guess we have to add `qty_received` so the computation is triggered by every cases.
At the contrary, I am not sure the `invoice_lines.invoice_id.state` is useful?

florian-dacosta (commented):
I see one more problem I had not detect before...
Hopefully the last one!
@mreficent

purchase_open_qty/models/purchase_order.py
MiquelRForgeFlow: <code>qty_received</code> is another compute, so we need to add all the fields of the _compute_qty_received()'s depends.

florian-dacosta: Are you sure? 
Here is an example on Odoo's code of a compute field depending on another one : 
https://github.com/odoo/odoo/blob/9.0/addons/account/models/account_move.py#L339

LoisRForgeFlow: just for have everything tied so anyone can follow the threads: this is fixed in https://github.com/OCA/purchase-workflow/pull/451

PR #451, opened 2017-09-12
[9.0][FIX][purchase_open_qty] wrong api.depends

MiquelRForgeFlow:
Please, merge this fast-forward.

cc: @jbeficent @florian-dacosta

### DIFF
--- a/purchase_open_qty/models/purchase_order.py
+++ b/purchase_open_qty/models/purchase_order.py
@@ -11,7 +11,8 @@ class PurchaseOrderLine(models.Model):
     _inherit = 'purchase.order.line'
 
     @api.depends('product_qty', 'qty_invoiced',
-                 'invoice_lines.invoice_id.state')
+                 'invoice_lines.invoice_id.state',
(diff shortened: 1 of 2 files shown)


florian-dacosta:
I did test with these fields and just with `qty_received` instead of` 'order_id.state', 'move_ids.state', 'move_ids'` and both work fine.
You already depend on a computed field `qty_invoiced`
I think it is better, but I am no expert in computed fields....
If you only depend on ` 'order_id.state', 'move_ids.state', 'move_ids'` I wonder if it can be a source of problem.
For instance, the qty_to_invoice is recomputed before the qty_received field (because they depends on the same fields). The qty_to_invoice will have the wrong value no?
If it depend directly on qty_received, this problem can't happen, unless it is really not recommended to depend on a stored computed field.

MiquelRForgeFlow:
@florian-dacosta Comment attended 😉

florian-dacosta (approved):

LoisRForgeFlow (approved):
Tested on runbot and LGTM

PR #453, opened 2017-09-18
[9.0][FIX] purchase_discount: Fix calculation order taxes with globally round method

sergio-teruel:
cc @Tecnativa


pedrobaeza:
Please check how is this done on v10 because I think this one is too much complicated.

sergio-teruel:
It is the same

sergio-teruel:
Sorry it is the same in core....

pedrobaeza:
It's not the same. Please check it again. You are not being very friendly with inheritance.

PR #455, opened 2017-09-19
[9.0][FIX] purchase_discount: Fix calculation order taxes with globally round method

sergio-teruel:
cc @Tecnativa


pedrobaeza:
We are still loosing the inheritability...

sergio-teruel:
Ok. I Will adapt it as v10...

pedrobaeza (approved):
Checking twice, a similar technique has been finally adopted in v10 because of weird results in ORM. So I approve and merge this to solve in all previous version the issue.

This doesn't happen fortunately in v11.

PR #456, opened 2017-09-22
[10.0][ADD] Module which allow to indicate cancelled quantities on purchase order lines

benwillig:
Allow purchase users to define cancelled quantity on purchase order lines.
* Quantity field becomes readonly when a PO is confirmed and its value is stored in an other field
* A new field named 'Cancelled quantity' will update the quantity based on the initially ordered quantity


purchase_cancel_qty/views/purchase_order.xml
```diff
@@ -0,0 +1,21 @@
...
+
+    <record model="ir.ui.view" id="purchase_order_form_view">
+        <field name="name">purchase.order.form (in purchase_cancel_qty)</field>
+        <field name="model">purchase.order</field>
+        <field name="inherit_id" ref="purchase.purchase_order_form"/>
+        <field name="arch" type="xml">
+
+            <xpath expr="//field[@name='order_line']/tree/field[@name='product_qty']" position="after">
+                <field name="initially_ordered_qty" invisible="not context.get('show_purchase', False)"/>
+                <field name="cancelled_qty" invisible="not context.get('show_purchase', False)"/>
```
sbidoul: would it make sense to hide `product_qty` when `show_purchase` is True?

benwillig: If we do that, user will have to calculate himself what quantity is actually processed.

purchase_cancel_qty/README.rst
```diff
@@ -0,0 +1,76 @@
...
+------------
+
+* Benjamin Willig <benjamin.willig@acsone.eu>
+
+Funders
+-------
+
+The development of this module has been financially supported by:
+
+* ACSONE SA/NV (https://acsone.eu)
```
sbidoul: Please remove the Funders section, this work has been financed by our client.

benwillig: Done.

purchase_cancel_qty/README.rst
```diff
@@ -0,0 +1,69 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+===================
+Purchase Cancel Qty
+===================
+
+Allow purchase users to define cancelled quantity on purchase order lines.
+* Quantity field becomes readonly when a PO is confirmed and its value is stored in an other field
```
LoisRForgeFlow: Add a blank line above, otherwise this is not correctly displayed.

purchase_cancel_qty/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+    'author': 'ACSONE SA/NV,Odoo Community Association (OCA)',
+    'website': 'https://acsone.eu',
+    'depends': [
+        'purchase',
+    ],
+    'data': [
+        'views/purchase_order.xml',
+    ],
+    'demo': [
+    ],
```
LoisRForgeFlow: remove empty tags.

LoisRForgeFlow (commented):
Please check travis + some comments in line

gva-acsone:
Functional test: 

1. Negative amounts shouldn't be allowed in the 'cancelled quantity' field.
2. The two collumns (quantity ordered and quantity cancelled) should be added on the purchase order lines too.
3. I would change the order of the collumns so that the collumns that impact the 'quantity' collumn appear first. So the order I would put would be: quantity ordered, quantity cancelled, quantity, quantity received, quantity billed
4. On Runbot, RFQ cannot be confirmed, which makes the testing on runbot impossible.
Message: 
  File "/.repo_requirements/odoo/odoo/http.py", line 640, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/.repo_requirements/odoo/odoo/http.py", line 677, in dispatch
    result = self._call_function(**self.params)
  File "/.repo_requirements/odoo/odoo/http.py", line 333, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/.repo_requirements/odoo/odoo/service/model.py", line 101, in wrapper
    return f(dbname, *args, **kwargs)
  File "/.repo_requirements/odoo/odoo/http.py", line 326, in checked_call
    result = self.endpoint(*a, **kw)
  File "/.repo_requirements/odoo/odoo/http.py", line 935, in __call__
    return self.method(*args, **kw)
  File "/.repo_requirements/odoo/odoo/http.py", line 506, in response_wrap
    response = f(*args, **kw)
  File "/home/odoo/odoo-10.0/addons/web/controllers/main.py", line 889, in call_button
    action = self._call_kw(model, method, args, {})
  File "/home/odoo/odoo-10.0/addons/web/controllers/main.py", line 877, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/.repo_requirements/odoo/odoo/api.py", line 689, in call_kw
    return call_kw_multi(method, model, args, kwargs)
  File "/.repo_requirements/odoo/odoo/api.py", line 680, in call_kw_multi
    result = method(recs, *args, **kwargs)
  File "/home/odoo/build/OCA/purchase-workflow/purchase_request_to_rfq/models/purchase_order.py", line 74, in button_confirm
    res = super(PurchaseOrder, self).button_confirm()
  File "/home/odoo/odoo-10.0/addons/purchase/models/purchase.py", line 347, in button_confirm
    order.button_approve()
  File "/home/odoo/build/OCA/purchase-workflow/purchase_fop_shipping/models/purchase.py", line 41, in button_approve
    result = super(PurchaseOrder, self).button_approve(force=force)
TypeError: button_approve() got an unexpected keyword argument 'force'

Thanks

sbidoul:
@benwillig could you rebase on OCA/10.0, I think you started from an old version which has the error shown in travis and @gva-acsone 's test.

benwillig:
@sbidoul I'm not sure if rebasing would solve the problem. ```purchase_fop_shipping``` add a kwargs named ```force```. Then I have to add ```**kwargs``` in ```button_approve``` method definition.

purchase_cancel_qty/models/purchase_order_line.py
```diff
@@ -0,0 +1,55 @@
...
+        inverse='_inverse_cancelled_qty',
+        track_visibility='onchange',
+        digits=dp.get_precision('Product Unit of Measure'),
+        states=CANCELLED_QTY_EDITABLE_STATES)
+
+    _sql_constraints = [
+        (
+            'cancelled_qty_gt_zero',
+            'check(COALESCE(cancelled_qty, 0) >= 0)',
+            _("Cancelled quantity must be greater or equal than 0.")
```
sbidoul: greater or equal *to*

purchase_cancel_qty/README.rst
benwillig: Indeed, thanks

purchase_cancel_qty/__manifest__.py
benwillig: Removed it.

purchase_cancel_qty/models/purchase_order_line.py
benwillig: Corrected.

purchase_cancel_qty/models/purchase_order.py
```diff
@@ -0,0 +1,25 @@
...
+
+from odoo import api, models
+
+
+class PurchaseOrder(models.Model):
+
+    _inherit = 'purchase.order'
+
+    @api.multi
+    def button_approve(self, **kwargs):
```
sbidoul: keep the same signature as the original method, ie `button_approve(self, force=False)`

purchase_cancel_qty/models/purchase_order_line.py
```diff
@@ -0,0 +1,55 @@
...
+        inverse='_inverse_cancelled_qty',
+        track_visibility='onchange',
+        digits=dp.get_precision('Product Unit of Measure'),
+        states=CANCELLED_QTY_EDITABLE_STATES)
+
+    _sql_constraints = [
+        (
+            'cancelled_qty_gt_zero',
+            'check(COALESCE(cancelled_qty, 0) >= 0)',
+            _("Cancelled quantity must be greater or equal to 0.")
```
sbidoul: Why not do this in `_check_cancelled_qty`?

purchase_cancel_qty/models/purchase_order.py
benwillig: My bad, I didn't see this kwarg was added in the base module.

purchase_cancel_qty/models/purchase_order_line.py
benwillig: To prevent users to bypass the ORM, but in this case, I should have done the other check (cancelled_qty <= ordered_qty) as a sql constraint too.

purchase_cancel_qty/models/purchase_order_line.py
```diff
@@ -0,0 +1,50 @@
...
+    product_qty = fields.Float(states=PRODUCT_QTY_READONLY_STATES)
+    ordered_qty = fields.Float(
+        string="Ordered Qty", readonly=True, copy=False,
+        digits=dp.get_precision('Product Unit of Measure'))
+    cancelled_qty = fields.Float(
+        string="Cancelled Qty", readonly=True, copy=False,
+        inverse='_inverse_cancelled_qty',
+        track_visibility='onchange',
+        digits=dp.get_precision('Product Unit of Measure'),
+        states=CANCELLED_QTY_EDITABLE_STATES)
```
sbidoul: Ah, one last minor style nitpicking:

```python
product_qty = fields.Float(
    states=PRODUCT_QTY_READONLY_STATES,
)
ordered_qty = fields.Float(
    string="Ordered Qty", 
    readonly=True, 
    copy=False,
    digits=dp.get_precision('Product Unit of Measure'),
)
cancelled_qty = fields.Float(
    string="Cancelled Qty", 
    readonly=True, 
    copy=False,
    inverse='_inverse_cancelled_qty',
    track_visibility='onchange',
    digits=dp.get_precision('Product Unit of Measure'),
    states=CANCELLED_QTY_EDITABLE_STATES,
)
```

purchase_cancel_qty/models/purchase_order_line.py
sbidoul: I tend to prefer python constraints because they allow for more explicit error messages.

purchase_cancel_qty/models/purchase_order_line.py
sbidoul: This style makes it easier to spot all fields, and makes diffs easier to read when adding/modifying fields.

sbidoul (approved):

sbidoul:
@gva-acsone runbot should be ok now.

gva-acsone:
Functional test: ok!

LoisRForgeFlow:
Just one question, I don't want to block this. What is the purpose of this cancelled qty field in purchase state? When you have confirmed the order, the picking is already created and then if you modify the cancelled qty the picking will still have the full qty, which would require a manual action by the user to reduce the qty in the picking and then reject the backorder creation.

gva-acsone:
This 'cancelled quantity field' has been added to indicate that a supplier will not deliver the whole quantity we ordered. The problem with the standard way of working (changing the quantity directly on the confirmed PO) is that there is no tracking of the quantity initially ordered to the supplier. The PO will only indicate the new quantity that will be delivered by the supplier, so the initially ordered quantity and the cancelled quantity cannot be used for controlling purposes or reporting purposes. 

* Why not only indicate on the picking that the whole quantity will not be delivered by our supplier (by not creating a backorder for the remaining quantity)? 
1. Because this doesn't work for products which are 'services'.  There is no picking created for services so the only way to work in the standard version is to change the ordered quantity on the PO. This brings us back to our controlling and reporting issues.

* As you mention, for products which are 'stockable products', users will need to indicate that the whole quantity will not be delivered and cancel the backorder. However, as the picking is currently created as soon as the PO is confirmed, this means that in the actual version of Odoo, users already have to do those two steps when they modify the ordered quantity on the PO. So we don't make this process more difficult than in the standard version of Odoo. 

Hope this answers your question,

Cheers!

LoisRForgeFlow:
@gva-acsone Ok, I understand now. Thanks for the explanation :+1:

LoisRForgeFlow (approved):
Tested on runbot, works as expected.

benwillig:
I identified an error, if you add a line when PO has been confirmed, ```ordered_qty``` field will always be 0.

gva-acsone:
The 'Quantity' field on a new line on a confirmed PO (Purchase order state) is not editable after you have added a new PO line and saved.
Therefore, you could just state that when you save your new line on your PO, the ordered quantity of your NEW line (and only your new line) takes the value of the 'Quantity' field.

AaronHForgeFlow (commented):
It works as expected. However, if I receive less qty in the picking and create no backorder or if the backorder is cancelled I should see some qty cancelled in the PO. IMHO.

sbidoul:
@aheficent that would be nice to have indeed. That said, when you do that in standard Odoo, the quantity in the PO does not change either, you must do it manually.

So I'd suggest to clarify the module description and put that in the roadmap section.

purchase_cancel_qty/models/purchase_order_line.py
```diff
@@ -0,0 +1,58 @@
...
+    @api.multi
+    def _inverse_cancelled_qty(self):
+        for rec in self:
+            rec.product_qty = rec.ordered_qty - rec.cancelled_qty
+
+    @api.multi
+    @api.constrains('cancelled_qty', 'ordered_qty')
+    def _check_cancelled_qty(self):
+        for rec in self:
+            if rec.cancelled_qty < 0:
```
ThomasBinsfeld: @benwillig Float comparisons must be done with `float_compare`

purchase_cancel_qty/models/purchase_order_line.py
```diff
@@ -0,0 +1,58 @@
...
+                raise ValidationError(
+                    _("Cancelled quantity must be greater or equal to 0."))
+            if rec.cancelled_qty > rec.ordered_qty:
```
ThomasBinsfeld: @benwillig Same (float_compare)

purchase_cancel_qty/models/purchase_order.py
```diff
@@ -0,0 +1,25 @@
...
+    @api.multi
+    def button_approve(self, force=False):
+        res = super(PurchaseOrder, self).button_approve(force=force)
+        for rec in self:
+            rec._update_lines_ordered_qty()
+        return res
+
+    @api.multi
+    def _update_lines_ordered_qty(self):
+        self.ensure_one()
```
ThomasBinsfeld: @benwillig Why `self.ensure_one()` ? You could iterate here instead of calling the method once for each record.

purchase_cancel_qty/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+
+{
+    'name': 'Purchase Cancel Quantity',
+    'summary': """
+        Allow purchase users to define cancelled quantity on
+        purchase order lines""",
+    'version': '10.0.1.0.0',
+    'license': 'AGPL-3',
+    'author': 'ACSONE SA/NV,Odoo Community Association (OCA)',
+    'website': 'https://acsone.eu',
```
ThomasBinsfeld: https://github.com/OCA/purchase-workflow

ThomasBinsfeld (changes requested):

benwillig:
Updated my PR based on your comments @ThomasBinsfeld.

I also fixed the use case when a user adds a line on a PO which has been confirmed. The fix is not really sexy but it's the only way I found at the moment.

rousseldenis (approved):
LGTM. Code review

For roadmap, this could have a 'Cancel quantity reason' that would appear in chatter.

ThomasBinsfeld (approved):

purchase_cancel_qty/README.rst
```diff
@@ -0,0 +1,70 @@
...
+===================
+Purchase Cancel Qty
+===================
+
+Allow purchase users to define cancelled quantity on purchase order lines.
+
+* Quantity field becomes readonly when a PO is confirmed and its value is stored in an other field
+* A new field named 'Cancelled quantity' will update the quantity based on the initially ordered quantity
+
+Installation
```
elicoidal, line 14: you can remove this section

purchase_cancel_qty/README.rst
```diff
@@ -0,0 +1,70 @@
...
+=============
+
+No configuration needed.
+
+Usage
+=====
+
+To use this module, you need to:
+
+* Go on the PO form view
```
elicoidal, line 29: Better use enumerated list for procedures with `#. `

purchase_cancel_qty/__init__.py
```diff
@@ -0,0 +1,2 @@
+from . import models
```
elicoidal, line 1: dont you need to declare `utf-8`

elicoidal (approved):
some non blocking details

benwillig:
@pedrobaeza I think it can be merged.

pedrobaeza:
OK, I'm merging it blindly as I haven't reviewed it

/ocabot merge

PR #458, opened 2017-09-22
[10.0][FIX]account_type does was removed from analytic_account

AaronHForgeFlow:
Fixes this https://github.com/OCA/purchase-workflow/issues/457

PR #459, opened 2017-09-22
Remove account_type from domain

yk2kus:
field is not available in V10


yk2kus:
https://github.com/OCA/purchase-workflow/pull/458

PR #470, opened 2017-10-18
[11.0][MIG] purchase discount: Migration to 11.0

pedrobaeza:
Standard migration

@Tecnativa


chienandalu:
Seems like there's some issue with runbot. It seems like it's well built but it's not possible to use it :confused: 

image

legalsylvain (approved):
:+1: 
thanks. code review. no test.

pedrobaeza:
I have launched a rebuild.

purchase_discount/__manifest__.py
```diff
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 # © 2004-2009 Tiny SPRL (<http://tiny.be>).
-# © 2014-2015 Serv. Tecnol. Avanzados - Pedro M. Baeza
+# © 2014-2017 Tecnativa - Pedro M. Baeza
```
lasley: Not blocking, just questioning - can you copyright something with an entity that didn't exist on the date the copyright started?

lasley (approved):
Changes look good, thanks @pedrobaeza 

TIL about `list(dict.keys())` in py3

LoisRForgeFlow:
@pedrobaeza @lasley maybe a stupid question about using `list(dict.keys())`... Despite the list was what it was used in py2 and maybe the consistency given by use exactly the same objects is enough, what is the benefit of converting here the dict_keys returned by default in py3, as it can be iterated exactly the same as a list?

pedrobaeza:
That's an automatic change from `2to3` tool, because it doesn't know if you need the element as an iterator or a list. In this exact case, `list()` can be avoided.

purchase_discount/__manifest__.py
pedrobaeza: Well, it was me also before (as freelance), so I prefer to simplify it.

jjscarafia (approved):
Tested on runbot. If possible add changes from this PR https://github.com/OCA/purchase-workflow/pull/504

pedrobaeza:
Added last commit and already approved, I merge this.

PR #471, opened 2017-10-22
[9.0][REF]purchase_request: call get_default name method inside copy

zakiuu:
Call `_get_default_name()` to get new name instead of  calling the sequence directly. This will give an entry point to override the copy method when generating a new name.


LoisRForgeFlow (approved):
:+1:

lmignon (approved):

JordiBForgeFlow (approved):

PR #474, opened 2017-11-04
[ADD] purchase_line_reference: Abstract reference column

lasley:
* New module to add an abstract reference column to `purchase.order.line` allowing for a direct back reference to the record it represents.


lasley:
Note I didn't add a view, because I don't think it would really be helpful. Not against it if someone wants it though 🤷‍♂️

gurneyalex:
I don't understand the use case. Can you explain?

rousseldenis:
@OCA/procurement-purchase-maintainers I think this one can be closed.

PR #479, opened 2017-11-21
[11.0][purchase_procurement_run_buy_hook]

JordiBForgeFlow:
Purchase Procurement Run Buy Hook
=================================

This module adds hooks to method '_run_buy' of model 'procurement.rule'. The
currently supported hooks are:

* Allow to add more data to an existing purchase order line, when a
  procurement is extending it.


rousseldenis:
@jbeficent You cannot do that as you don't know if other modules inherit the _run_buy function and you are not sure of the order modules are loaded in stack!

Make a PR on odoo/odoo, usually they fast merge such hooks.

JordiBForgeFlow:
@rousseldenis Thanks. I will try now. In the past I was quite unsuccessful.

JordiBForgeFlow:
@rousseldenis Done here: https://github.com/odoo/odoo/pull/21123. Let's pray now...

JordiBForgeFlow:
@rousseldenis Does not look that the PR that I opened to Odoo is going to be addressed soon. So I will continue with this PR. I have done a refactor to improve compatibility of the hook with multiple databases, and respecting inheritance chain.

rousseldenis:
@jbeficent I see, I was too optimistic about such PR's on Odoo...

JordiBForgeFlow:
@rgarnau @lreficent Can you review?

rgarnau (approved):
LGTM :+1:

purchase_procurement_run_buy_hook/hooks.py
```diff
@@ -0,0 +1,95 @@
...
+from odoo.addons.purchase.models.purchase import ProcurementRule
+
+
+def post_load_hook():
+
+    @api.multi
+    def _new_run_buy(self, product_id, product_qty, product_uom,
+                     location_id, name, origin, values):
+        if not hasattr(self, '_prepare_purchase_order_line_update'):
+            return self.action_invoice_create_original(
```
LoisRForgeFlow: Where is this declared? I am not able to find it

LoisRForgeFlow (commented):
Interesting way to add the hook! One comment:

purchase_procurement_run_buy_hook/hooks.py
JordiBForgeFlow: Ask @rgarnau

JordiBForgeFlow: @lreficent you're right. I was calling a wrong method. I now changed to call '_original_run_buy'.

JordiBForgeFlow:
@rousseldenis @lreficent can you update your review?

LoisRForgeFlow (approved):
LGTM now!

rousseldenis (approved):
It seems ok!

AaronHForgeFlow (approved):
:+1: This is compatible with modules that need the  hook and modules that don't need the hook. :cool:

PR #481, opened 2017-12-04
[9.0][ADD] purchase_tier_validation

LoisRForgeFlow:
Purchase Tier Validation
========================
This module extends the functionality of Purchase Orders to support a tier
validation process.

Configuration
=============

To configure this module, you need to:

#. Go to *Settings > Technical > Tier Validations > Tier Definition*.
#. Create as many tiers as you want for Purchase Order model.

Usage
=====

To use this module, you need to:

#. Create a Purchase Order triggering at least one "Tier Definition".
#. Click on *Request Validation* button.
#. Under the tab *Reviews* have a look to pending reviews and their statuses.
#. Once all reviews are validated click on *Confirm Order*.

------

* [x] : Depends on https://github.com/OCA/server-tools/pull/1085

### DIFF
--- a/purchase_tier_validation/__openerp__.py
+++ b/purchase_tier_validation/__openerp__.py
@@ -1,0 +1,22 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Eficent Business and IT Consulting Services S.L.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Purchase Tier Validation",
+    "summary": "Extends the functionality of Purchase Orders to "
+               "support a tier validation process.",
+    "version": "9.0.1.0.0",
+    "category": "Purchases",
+    "website": "https://github.com/OCA/purchase-workflow",
+    "author": "Eficent, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": [
+        "purchase",
+        "base_tier_validation",
+    ],
+    "data": [
+        "views/purchase_order_view.xml",
+    ],
+}
--- a/purchase_tier_validation/README.rst
+++ b/purchase_tier_validation/README.rst
@@ -1,0 +1,84 @@
+.. image:: https://img.shields.io/badge/license-AGPL--3-blue.png
+   :target: https://www.gnu.org/licenses/agpl
+   :alt: License: AGPL-3
+
+========================
+Purchase Tier Validation
+========================
+
+This module extends the functionality of Purchase Orders to support a tier
+validation process.
+
+Installation
+============
+
+This module depends on ``base_tier_validation``. You can find it at
+`OCA/server-tools <https://github.com/OCA/purchase-workflow>`_
+
+Configuration
+=============
+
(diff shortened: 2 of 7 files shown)


AaronHForgeFlow (approved):
:+1: This should be a standard

HviorForgeFlow (approved):
LGTM :+1:

purchase_tier_validation/__openerp__.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Eficent Business and IT Consulting Services S.L.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Purchase Tier Validation",
+    "summary": "WIP",
```
MiquelRForgeFlow: ?

andhit-r:
Thanks for the contribution. It will be a awsome module. I encounter exception when i tested it. Step to reproduce:

1. Create tier definition
selection_002

2. Create another tier definition
selection_003

After steps above i could not create or open purchase order data.

selection_001

HviorForgeFlow:
@andhit-r
You must provide a python evaluation of a field related to the model, for example, amount_total > 1000 in PO case

andhit-r:
@hveficent thanks for your info. I gave both my tier definition rec.id != 0. Exception no longer exist.

I encounter another issue. Both my tier definition use groups reviewer. User Administrator belong to both groups. I can approve first validation. But no succeed approve second validation. See this:
https://www.youtube.com/watch?v=Mp6-wwt1aRM&feature=youtu.be

LoisRForgeFlow:
@andhit-r By default Adminastrator only belong to "Billing" group on accounting app. Could you check that you upgraded the permissions of Administrator to "Accountant"?

andhit-r:
@lreficent ouch, i always assume that admin has all the superhuman access. My bad.

I have another question though. Validation process only workd if uset hit the request validation button, is that by design?

LoisRForgeFlow:
@andhit-r Yes, I couldn't find a way to properly auto-request validation with the current design. It's the trade-off to be able to have this much flexibility and extensibility. However, if you have access to validate all the reviews that are going to be needed, you don't need to request the validation it will be auto-requested and auto-approved.

andhit-r (commented):
Ok. Functional test 👍

andhit-r (approved):

andhit-r:
@lreficent What user should do if po or does not pass validation. Can user revise po and then request another validation? Or user has to make another po?

LoisRForgeFlow:
@andhit-r the user can cancel, reset back to draft, change as needed and request a validation again.

andhit-r:
@lreficent thanks for your explanation

purchase_tier_validation/__openerp__.py
LoisRForgeFlow: fixed

LoisRForgeFlow:
Rebased and squashed

LoisRForgeFlow:
Could anyone remove the WIP label?

JordiBForgeFlow (approved):

PR #484, opened 2017-12-15
[9.0][REF]purchase_request_to_rfq: use separate method for reception message

zakiuu:
Separate the logic of sending reception message from the `do_transfer` method to ease the inheritance and the override of that part of code

### DIFF
--- a/purchase_request_to_rfq/models/stock.py
+++ b/purchase_request_to_rfq/models/stock.py
@@ -31,8 +31,7 @@ class StockPicking(models.Model):
         return message
 
     @api.multi
-    def do_transfer(self):
-        super(StockPicking, self).do_transfer()
+    def _post_purchase_request_picking_confirm_message(self):
         request_obj = self.env['purchase.request']
         for picking in self:
             requests_dict = {}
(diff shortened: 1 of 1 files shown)


purchase_request_to_rfq/models/stock.py
```diff
@@ -58,3 +57,9 @@ def do_transfer(self):
                     self._purchase_request_picking_confirm_message_content(
                         picking, request, requests_dict[request_id])
                 request.message_post(body=message, subtype='mail.mt_comment')
+
+    @api.multi
+    def do_transfer(self):
+        res = super(StockPicking, self).do_transfer()
+        self._purchase_request_reception_message()
```
lmignon: @zakiuu I would prefer a more meaningful name for the method. May be `_post_purchase_request_picking_confirm_message`

lmignon (changes requested):

purchase_request_to_rfq/models/stock.py
zakiuu: @lmignon thanks for the review, done!

lmignon (approved):
@zakiuu Thank you! (code review)

ThomasBinsfeld (approved):

JordiBForgeFlow (approved):
+1 code review

JordiBForgeFlow:
@lreficent can you review?

andhit-r (approved):

LoisRForgeFlow (approved):
LGTM :+1:

zakiuu:
@jbeficent is it possible to merge this one?

gurneyalex (approved):

PR #488, opened 2018-01-12
[MIG] purchase_order_line_description: Migration to 10.0

cubells:
- Update Views
- Update README
- Update tests

cc @Tecnativa


purchase_order_line_description/__manifest__.py
```diff
@@ -1,10 +1,10 @@
...
-# (c) 2015 Alex Comba - Agile Business Group
+# Copyright 2015 Alex Comba - Agile Business Group
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
 
 {
     'name': "Purchase order line description",
-    'version': '8.0.1.0.0',
+    'version': '10.0.1.0.0',
     'category': 'Purchase Management',
     'author': "Agile Business Group, Odoo Community Association (OCA)",
```
pedrobaeza: Add Tecnativa

pedrobaeza (changes requested):

pedrobaeza (approved):
@carlosdauden please review

purchase_order_line_description/models/purchase_order_line.py
```diff
@@ -0,0 +1,21 @@
...
+
+    @api.onchange('product_id')
+    def onchange_product_id(self):
+        res = super(PurchaseOrderLine, self).onchange_product_id()
+        if self.product_id:
+            if self.user_has_groups(
+                    'purchase_order_line_description.'
+                    'group_use_product_description_per_po_line'):
+                if self.product_id.description_purchase:
+                    self.name = self.product_id.description_purchase
```
carlosdauden: Is posible simplify concat if?

carlosdauden (approved):
Minor comments

purchase_order_line_description/models/purchase_order_line.py
pedrobaeza: The idea is to replace it, not to concatenate it

carlosdauden: What about this other code?
```
if not self.product_id:
    return res
if (self.user_has_groups(
            'purchase_order_line_description.'
            'group_use_product_description_per_po_line') and
        self.product_id.description_purchase):
    self.name = self.product_id.description_purchase
return res
```

pedrobaeza: Yeah, good simplification. @carlosdauden please make the honors of modifying the PR squashing the commit.

gurneyalex (approved):

PR #489, opened 2018-01-16
[11.0] [ADD][purchase_update_seller_price] - Module to update the seller pri…

ougc27:
Hello

This module allows the update of the price of a seller on the product info when a purchase order is confirmed.

I hope this module is useful and can be merged.
Thanks & Regards


purchase_update_seller_price/models/purchase_order.py
```diff
@@ -0,0 +1,27 @@
...
+        for line in self.order_line:
+            partner = (
+                self.partner_id if not self.partner_id.parent_id
+                else self.partner_id.parent_id)
+            sellers = line.product_id.seller_ids
+            if partner in sellers.mapped('name'):
+                currency = (
+                    partner.property_purchase_currency_id or
+                    self.env.user.company_id.currency_id)
+                sellers.filtered(lambda x: x.name == partner).write({
```
alan196, line 24: What do you think about saving the price in the currency that you buy? Instead on converting to the currency of the company or the supplier? If you set multi-currency  you can add a price to every currency.

alan196 (changes requested):
Please add tests

purchase_update_seller_price/models/purchase_order.py
ougc27: This code was taken from the original Odoo method, do you need me to develop it in that way?

Regards

StefanRijnhart: It's the same as in the core, so the price computation looks alright to me. Do you also want to write the currency in case they are not the same? As in https://github.com/odoo/odoo/blob/11.0/addons/purchase/models/purchase.py#L454

StefanRijnhart: When selecting the supplierinfo to update, do you want to use _select_seller from https://github.com/odoo/odoo/blob/11.0/addons/product/models/product.py#L468, because it looks at minimum quantity and timesliced entries.

StefanRijnhart:
Thanks, this looks useful! Can you add a test? I just wrote a test for the exact same method so you should be able to reuse it: https://github.com/OCA/purchase-workflow/pull/407/files#diff-90fdbeaa6952a9a28515ba18e205ae17R118

purchase_update_seller_price/__init__.py
```diff
@@ -0,0 +1,4 @@
+# Copyright 2018 JARSA Sistemas S.A. de C.V.
```
nikul-serpentcs, line 1: @ougc27 IMO remove Copyright in all __init__.py file.

nikul-serpentcs (commented):
Improve Code

PR #494, opened 2018-01-24
[FIX] purchase_request_procurement: purchase request line created from a procurement should use the purchase unit of measure

benwillig:
When a PO is created from a procurement, quantities are computed based on the purchase unit of measure. Actually, ```purchase_request``` module simply uses the product's unit of measure. It should use the purchase unit of measure.


rousseldenis (approved):
Seems logical.

adrienpeiffer (approved):

JordiBForgeFlow (approved):

Issue #498, opened 2018-02-06
[purchase_location_by_line] : module is not allowing to add extra PO lines once PO is processed


bista-amitd:
To Reproduce:
1). Create one Purchase Order
2). Add one line on it
3). Confirm the purchase order, process the picking
4). Now try to add a new line and save, it's giving following warning:
'Quantities, Units of Measure, Products and Locations cannot be modified on stock moves that have already been processed (except by the Administrator).'

It's trying to write the location on processed stock move which is the processed product, it should write the location on newly created line only.

MiquelRForgeFlow:
This issue can be closed.

PR #499, opened 2018-02-06
[10.0] [FIX] location should be set to draft PO line only

bista-amitd:
Fixes: #498

### DIFF
--- a/purchase_location_by_line/models/purchase.py
+++ b/purchase_location_by_line/models/purchase.py
@@ -51,7 +51,7 @@ class PurchaseOrderLine(models.Model):
                 default_picking_location_id)
 
             location = line.location_dest_id or default_picking_location
-            if location:
+            if location and line.state != 'purchase':
                 line.move_ids.write(
                     {'location_dest_id': location.id})
         return res


rousseldenis (changes requested):
Seems good to me.

Please add a test case and change module version.

gurneyalex:
@bista-amitd do you need help with the CLA?

jarmokortetjarvi:
I Don't have a permission to do an actual review, but for what it's worth, this looks fine to me (except the missing test case).
Tested on a production-clone where this problem is present, and the fix works as intended.

AaronHForgeFlow (commented):
But if we allow this, is it consistent when changing the location in confirmed lines? I don't think so.

jarmokortetjarvi:
> But if we allow this, is it consistent when changing the location in confirmed lines? I don't think so.

This is a valid point @aheficent . Do you see a similar problem if this was done by filtering out done stock moves instead of checking the line status?

E.g.
```
if location:
	line.move_ids.filtered(lambda m: m.state != 'done').write(
	{'location_dest_id': location.id})
```

I feel like something like this would be more consistent when changing the location, but wouldn't try to modify validated moves.

AaronHForgeFlow:
At first sight that would be correct. We can try that.

jarmokortetjarvi:
@bista-amitd do you want to take this, or should I make a new PR?

PR #501, opened 2018-02-16
[FIX] purchase_discount: Handle case where taxes rounding is global

pedrobaeza:
In the case that taxes rounding is set to globally, Odoo requires again the line price unit, and currently ORM mixes values, so the only way to get a proper value is to overwrite that part, losing inheritability.

cc @Tecnativa

### DIFF
--- a/purchase_discount/README.rst
+++ b/purchase_discount/README.rst
@@ -21,7 +21,12 @@ Usage
 Known issues / Roadmap
 ======================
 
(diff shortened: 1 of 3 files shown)


jjscarafia (approved):
:+1: tested on runbut + code review

pedrobaeza:
Fast-tracking this one also and as the version has been changed in a previous PR that has been merged minutes ago, we avoid to change it 2 times.

PR #502, opened 2018-02-20
[9.0][FIX-MIG][product_by_supplier] Fix Migration

ivantodorovich:
This PR fixes some issues on 9.0:
- It was not possible to add supplier info from product form, because this module override the `default_product_tmpl_id` context (maybe in 8.0 odoo handled this differently)
- `product_tmpl_id` field was buggy on `supplierinfo.form`, because it was defining it when it was already there.

I suspect that this module had never been properly migrated from 8.0. It just had it's version changed.

---

ping @jjscarafia

### DIFF
--- a/product_by_supplier/views/product_supplierinfo_view.xml
+++ b/product_by_supplier/views/product_supplierinfo_view.xml
@@ -1,15 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <odoo>
-        <record id="product_template_form_view" model="ir.ui.view">
-            <field name="name">product.template.common.form</field>
-            <field name="model">product.template</field>
-            <field name="inherit_id" ref="product.product_template_form_view"/>
-            <field name="arch" type="xml">
-                <field name="seller_ids" position="attributes">
-                    <attribute name="context">{'from_product_form':True}</attribute>
-                </field>
-            </field>
-        </record>
 
         <record id="view_product_supplierinfo_search" model="ir.ui.view">
             <field name="name">product.supplierinfo.search</field>
@@ -56,11 +46,10 @@
             <field name="inherit_id" ref="product.product_supplierinfo_form_view"/>
             <field name="priority">99</field>
             <field name="arch" type="xml">
-                <xpath expr="//field[@name='product_name']" position="before">
-                    <field name="product_tmpl_id" string="Product"
-                    invisible="context.get('from_product_form')"
-                    required="not context.get('from_product_form')"/>
-                </xpath>
+                <field name="product_tmpl_id" position="replace"/>
+                <field name="product_id" position="before">
+                    <field name="product_tmpl_id" required="True" invisible="context.get('default_product_tmpl_id')"/>
+                </field>
             </field>
         </record>
 
--- a/product_by_supplier/views/product_supplierinfo_view.xml
+++ b/product_by_supplier/views/product_supplierinfo_view.xml
@@ -7,9 +7,9 @@
             <field name="arch" type="xml">
                 <search string="Product by Suppliers">
                     <field name="name" string="Supplier"/>
+                    <field name="product_tmpl_id" />
                     <field name="product_code" string="Supplier Product Code"/>
                     <field name="product_name" string="Supplier Product Name"/>
-                    <field name="product_tmpl_id" />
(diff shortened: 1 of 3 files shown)


ivantodorovich:
@pedrobaeza can you take a look at this? maybe add the needs review label?

ivantodorovich:
ping @leemannd , @Cedric-Pigeon

gurneyalex (approved):

leemannd (approved):

pedrobaeza (changes requested):
That's not the correct solution, as you are forcing to introduce again the template although coming from template itself via one2many. The solution is to have different views (you can do it with inheritance + primary view) for having the view in one2many without the field `product_tmpl_id`. In that case, Odoo will handle automatically that field (set by `inverse_name` attribute)

ivantodorovich:
Hi @pedrobaeza 

Thanks for your review!

I'm not sure I understand, though. Let me explain this changes.

On `v9` (as opposed to `v8`), the `product.template.common.form` already defines some context on the One2many field. So there's no need to hack into it. Besides, hacking it was removing the original context.

Also on `v9`, the `product.product_supplierinfo_form_view` already includes a `product_tmpl_id` field, so we can't add it anymore. Instead, I'm replacing it with our own.

That's all I did. Now, this module works almost like it did on `v8`.

--

I get that it's possible to achieve the same by defining a new view, inheriting the original but without extending it. If that where the case, instead of altering the one2many view, I'd define a new view for the action introduced by this module. It certainly is possible, but I don't get why is it needed or better than doing this.

product_by_supplier/views/product_supplierinfo_view.xml
```diff
@@ -56,11 +46,10 @@
...
             <field name="arch" type="xml">
-                <xpath expr="//field[@name='product_name']" position="before">
-                    <field name="product_tmpl_id" string="Product"
-                    invisible="context.get('from_product_form')"
-                    required="not context.get('from_product_form')"/>
-                </xpath>
+                <field name="product_tmpl_id" position="replace"/>
+                <field name="product_id" position="before">
+                    <field name="product_tmpl_id" required="True" invisible="context.get('default_product_tmpl_id')"/>
+                </field>
```
zaoral: @ivantodorovich 
I think that the field should not be replaced instead you could use a position="attributes" to add the new characteristic you need.

ivantodorovich: Hi @zaoral !

That was my initial intention. But the original location of the field on the view is not as good.
In fact, as in Odoo it's never shown, I think they never wasted a second thinking on the location, they've just placed it there.

zaoral:
> I get that it's possible to achieve the same by defining a new view, inheriting the original but without extending it. If that where the case, instead of altering the one2many view, I'd define a new view for the action introduced by this module. It certainly is possible, but I don't get why is it needed or better than doing this.

I agree, I think that if we need to show two different views one for o2m and another from the supplier.info form view then this should be added as a new issue and PR, but what is proposed here is to fix a problem with a buggy duplicated product_tmp_id field and a not needed context

product_by_supplier/views/product_supplierinfo_view.xml
zaoral: @ivantodorovich 

mmm.. ok I get it. The problem is that you want to move the location of the field.

First, I think that replace a field completly from a view specially if this one came from odoo addons directly is a delicate thing to do: I understand your claim but I think we should not replace the field IMHO. 

Sadly as far as I know we can't move a xml node (field) to another place in the view: if that is possible well would be perfect.

However, in this case, I have to agree with @pedrobaeza. The best option is create a new view (inherit from the original but explicitly set as primary in order to do not overwrite, only copy the complete view) will be the best approach, there you can replace the field and add what you like.

ivantodorovich:
@pedrobaeza @zaoral take a look now.

Instead of extending old views, we use new ones for the action introduced by this module.
Form view is actually inheriting original view, but it's set as primary so it doesn't affect original behaviour.

I have to agree it's 'cleaner' now. But still, this module was already inheriting views on `v8`. I've just wanted to fix a buggy behaviour from an incomplete migration.

pedrobaeza (approved):

PR #504, opened 2018-02-23
[10.0][IMP] Improove purchase_discount performance

### LINKED ISSUE #503: [10.0+] purchase_discount performance
Hi @pedrobaeza.

We've detected that aproving PO of many products takes to long and seams to be related to [this line](https://github.com/OCA/purchase-workflow/blob/10.0/purchase_discount/models/purchase_order.py#L87).

Do you remember why that is needed? We have comment it and it works ok. How can you get into an "infinite recursion"?

Just to let you know, we are not using this on v10, instead we have backported some improvements of v10 to v9, so perhups on v10 is not exactly the same but I guess it should be.

jjscarafia:
Close #503 
This state change is no longer needed because odoo has added this [line](https://github.com/odoo/odoo/blob/10.0/addons/purchase/models/purchase.py#L587)
Thanks @pedrobaeza for the quick explanation!

### DIFF
--- a/purchase_discount/__manifest__.py
+++ b/purchase_discount/__manifest__.py
@@ -10,7 +10,7 @@
               "Tecnativa, "
               "ACSONE SA/NV,"
               "Odoo Community Association (OCA)",
-    "version": "10.0.1.0.2",
+    "version": "10.0.1.0.3",
     "category": "Purchase Management",
     "depends": ["purchase"],
     "data": [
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Please change module version number according https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

jjscarafia:
@pedrobaeza . Done!

pedrobaeza (approved):
I fast-track this as it has been proved in the issue and to avoid conflicts with the other PR.

PR #508, opened 2018-03-02
8.0 backport purchase triple discount

legalsylvain:
backport of purchase_triple_discount from 10.0 to 8.0.

only the commit 5f44013 is to review.

~~Note : before merging, check if e2c3748 has been removed. (add temporary dependency to https://github.com/OCA/account-invoicing/pull/361)~~

### DIFF
--- a/purchase_triple_discount/models/purchase_order.py
+++ b/purchase_triple_discount/models/purchase_order.py
@@ -1,0 +1,42 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Tecnativa - David Vidal
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+from odoo.addons import decimal_precision as dp
+
+
+class PurchaseOrderLine(models.Model):
+    _inherit = "purchase.order.line"
+
+    @api.depends('discount2', 'discount3')
+    def _compute_amount(self):
+        super(PurchaseOrderLine, self)._compute_amount()
+
+    discount2 = fields.Float(
+        'Disc. 2 (%)',
+        digits=dp.get_precision('Discount'),
+        default=0.0,
+    )
+
+    discount3 = fields.Float(
+        'Disc. 3 (%)',
+        digits=dp.get_precision('Discount'),
+        default=0.0,
+    )
+
+    _sql_constraints = [
+        ('discount2_limit', 'CHECK (discount2 <= 100.0)',
+         'Discount 2 must be lower than 100%.'),
+        ('discount3_limit', 'CHECK (discount3 <= 100.0)',
+         'Discount 3 must be lower than 100%.'),
+    ]
+
+    def _get_discounted_price_unit(self):
+        price_unit = super(
+            PurchaseOrderLine, self)._get_discounted_price_unit()
+        if self.discount2:
+            price_unit *= (1 - self.discount2 / 100.0)
+        if self.discount3:
+            price_unit *= (1 - self.discount3 / 100.0)
+        return price_unit
--- a/purchase_triple_discount/models/purchase_order.py
+++ b/purchase_triple_discount/models/purchase_order.py
@@ -2,12 +2,18 @@
 # Copyright 2017 Tecnativa - David Vidal
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import api, fields, models
-from odoo.addons import decimal_precision as dp
+from openerp import api, fields, models
+from openerp.addons import decimal_precision as dp
 
 
 class PurchaseOrderLine(models.Model):
     _inherit = "purchase.order.line"
+
+    @api.model
+    def _calc_line_base_price(self, line):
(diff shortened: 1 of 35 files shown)


gurneyalex:
@legalsylvain  can you check the conflicts?

legalsylvain:
Hi @gurneyalex. 
done. I rebase. Should be green.
BTW weird thing : My PR was conflicting with po files present in an empty repository.  

See : https://github.com/OCA/purchase-workflow/tree/8.0/purchase_triple_discount

Note sure to understand why OCA transbot is creating PO files for unported module and so create new folders.

pedrobaeza:
In some moment, there was a branch under OCA organization that includes them. Do you push to OCA instead of grap in any moment?

legalsylvain:
> In some moment, there was a branch under OCA organization that includes them. Do you push to OCA instead of grap in any moment?

It's possible. I remember one time when I made an error on a git push. Not sure it was this repository. I deleted the incorrect branch on github but it could explain why transbot began to make its job.
Thanks.

legalsylvain:
Little up on that simple PR. CC : @chienandalu, @pedrobaeza 
(travis is red for unrelated reasons).

purchase_triple_discount/tests/test_purchase_discount.py
```diff
@@ -2,55 +2,19 @@
...
 
 class TestPurchaseOrder(common.SavepointCase):
 
     @classmethod
-    def setUpClass(cls):
-        super(TestPurchaseOrder, cls).setUpClass()
-        cls.partner = cls.env['res.partner'].create({
-            'name': 'Mr. Odoo',
-        })
-        cls.product1 = cls.env['product.product'].create({
```
pedrobaeza: Please don't remove these records by demo data, as demo data tend to change, and thus tests will fail with unexpected reasons.

purchase_triple_discount/models/purchase_order.py
```diff
@@ -32,11 +38,35 @@ def _compute_amount(self):
...
+            self, order, order_line, picking_id, group_id):
+        res = super(PurchaseOrder, self)._prepare_order_line_move(
+            order, order_line, picking_id, group_id)
+        for vals in res:
+            vals['price_unit'] = (vals.get('price_unit', 0.0) *
+                                  (1 - (order_line.discount2 / 100)) *
+                                  (1 - (order_line.discount3 / 100)))
+        return res
+
+#    def _get_discounted_price_unit(self):
```
pedrobaeza, line 65: Remove this

pedrobaeza (changes requested):

purchase_triple_discount/tests/test_purchase_discount.py
legalsylvain: Hi.

> as demo data tend to change

sorry, don't understand.

>  thus tests will fail with unexpected reasons

don't understand neither.

Well, a good pro for having demo data, is that the end user that test the module on runbot, on test env, can more easily test the module, without having to recreate all the data demo set.

pedrobaeza: Yes, keep both, but don't remove the creation of these records in these tests

purchase_triple_discount/models/purchase_order.py
legalsylvain: done.

purchase_triple_discount/tests/test_purchase_discount.py
legalsylvain: I understand what you ask @pedrobaeza but not why. Could you simply explain the reason or the use case that makes interesting keeping 40 lines of python code ?

pedrobaeza: Standard demo data changes over versions, and you are relying on that ones for the test. Keeping isolated base data for tests prevents you for having to change demo data + tests for making tests to work.

legalsylvain: @pedrobaeza . this is a port. this code will never be ported in V8> version. Thanks.

pedrobaeza: OK, but please take it into account for other modules

pedrobaeza:
Please check why Travis is red

legalsylvain:
Hi @pedrobaeza,

- These was conflict due to the merge of the recent PR : #560. (pot conflict) --> Solved.

- Travis was red because the module backported purchase_triple_discount breaks the existing module purchase_supplier_rounding_method (due to the fact that both inherit of same function and the order is not predictive). --> Solved, adding a glue module that fixes the problem.

Hope all is good now. #MakeTravisGreenAgain.

Thanks for the review !

purchase_triple_discount/models/purchase_order.py
legalsylvain: Note : should be done upstream, as it is a backport.

legalsylvain:
@pedrobaeza, I took into account your remarks. Could you update your review ? 
thanks !

pedrobaeza (approved):

gurneyalex (approved):

PR #516, opened 2018-03-13
[11.0][ADD] stock_landed_cost_company_percentage

bodedra:
https://github.com/OCA/purchase-workflow/issues/505

PR #518, opened 2018-03-14
[FIX] don't change the currency

gurneyalex:
in the test, changing the currency of the company
is not nice and can cause other modules' test to
fail for no good reason.

Instead, make sure that the rate of the company
currency is 1.

### DIFF
--- a/purchase_discount/tests/test_purchase_discount.py
+++ b/purchase_discount/tests/test_purchase_discount.py
@@ -18,8 +18,13 @@ class TestPurchaseOrder(common.SavepointCase):
             'name': 'Test product 2',
         })
         po_model = cls.env['purchase.order.line']
-        # Make sure currency is EUR for not having troubles with rates
-        cls.env.user.company_id.currency_id = cls.env.ref('base.EUR')
+        # Set the Exchange rate for the currency of the company to 1
+        # to avoid issues with rates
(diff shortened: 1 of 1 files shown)


gurneyalex:
I expect this fix to get a green build for e.g.  #516 which is failing with

2018-03-14 06:27:34,645 5034 ERROR openerp_test odoo.addons.purchase_discount.tests.test_purchase_discount: `     raise UserError(_('You cannot change the currency of the company since some journal items already exist'))

max3903 (approved):

bodedra (approved):

StefanRijnhart (approved):
I was confused because 'not nice' is not a compelling argument in test transactions which are rolled back and this must have been green once, but I see this is a very recent change in Odoo in https://github.com/odoo/odoo/commit/0e881fcd3a5

StefanRijnhart:
Short tracking this to fix test compatibility with Odoo.

StefanRijnhart:
Anyone who still needs to change the company's currency in their test can do what Odoo does and make the change in SQL ;-) https://github.com/odoo/odoo/commit/6deed45

gurneyalex:
@StefanRijnhart  sorry for the poor wording and thanks for the fast tracking :smile_cat:

PR #522, opened 2018-03-15
[ADD] purchase_picking_address

etobella:
Purchase Picking address
========================

This module allows to add a delivery address on picking types, so the purchase
order is more accurate.

Configuration
-----------------

#. Go to ’Inventory / Configuration / Settings’
#. Select the option ’Storage Locations’
#. Go to ‘Inventory / Warehouse Management / Operation Types’
#. Create a new operation type with code type vendors and select a 'Purchase
   delivery address'

Usage
-------

#. Go to 'Purchases / Requests for Quotation'
#. Create a Purchase order and select the created operation type
#. The delivery address of the purchase will be the one selected on the
   operation type


purchase_picking_address/models/purchase.py
```diff
@@ -0,0 +1,26 @@
...
+            self.dest_address_id = picking.purchase_delivery_address_id
+            return
+        super()._onchange_picking_type_id()
+
+    def _get_destination_location(self):
+        self.ensure_one()
+        picking_type = self.picking_type_id
+        if (
+            self.dest_address_id and
+            picking_type.purchase_delivery_address_id == self.dest_address_id
```
gurneyalex, line 23: why not use the address of the warehouse instead of creating a new field ?

etobella: Because sometimes you may have in the same warehouse multiple places that could receive materials. For example, if your warehouse has two accesses, you may want to receive one kind of products on street A and the other ones on street B. 
Transport cannot decide which one to use if you are using Warehouse address.

JordiBForgeFlow: @gurneyalex That is why you enter picking types and not warehouses in PO's. It allows you to receive in different places. But then correct documentation should follow!

rousseldenis: Ok, I had same remark as @gurneyalex.

And I suppose you don't have different reception locations (stock.location) in Odoo ?

JordiBForgeFlow: You must have different reception locations in Odoo. 
For example, a company receives sealable items from WH/Stock, whereas office supplies are purchased to be received diretly into WH/Office.

You would create two picking types: WH/Incoming and WH/Office, each with the corresponding default destination location.

Is that more clear now?

andhit-r: @etobella @jbeficent why not use purchase_transport_multi_address? Is there any scenario that can't be solved by purchase_transport_multi_address?

JordiBForgeFlow: purchase_transport_multi_address does not pick an address that is specific to a picking type. Assumes that the products are shipped to the warehouse's main address:
https://github.com/OCA/stock-logistics-transport/blob/8.0/purchase_transport_multi_address/model/purchase_order.py#L80

The issues stated in 'purchase_transport_multi_address' and 'stock_transport_multi_address' seem to be (partially) fixed in v11:
* In drop ship PO, the address of the customer appears in the PO and PO form
* In drop ship incoming shipments, the address of the customer appears in the picking form, but not in the screen.

What I found is that in case of transfers between warehouses (where Odoo creates a delivery order in WH1 and an incoming shipment in WH2) the picking form for the delivery order in WH1 is a mess - does not state where are the products really shipped to. 'stock_transport_multi_address'  would probably need to resolve the issue.

'stock_transport_multi_address' and 'purchase_transport_multi_address' are important  when you do drop shipments and cross warehouse moves. But they add a significant payload to the data model.

'purchase_picking_address' resolves a common issue from the purchase side with minimum impact.

Not sure...

andhit-r: > purchase_transport_multi_address does not pick an address that is specific to a picking type

True. But if we use different picking type from same warehouse to receipt, we need to configure additional route.

> 'stock_transport_multi_address' and 'purchase_transport_multi_address' are important when you do drop shipments and cross warehouse moves. But they add a significant payload to the data model.

> 'purchase_picking_address' resolves a common issue from the purchase side with minimum impact.

If that's the case i won't argue. Although it's better if we dont have two modules with simmilar goals.

etobella: Generally, odoo thinks that only one entrance must exist for a warehouse.  When I was checking the `purchase_transport_multi_address`, it had the same problems related to locations that I found on odoo. You can check it here:
https://github.com/etobella/stock-logistics-transport/blob/10.0/purchase_transport_multi_address/model/purchase_order.py#L80
Moreover, `stock_transport_multi_address` is discontinued since 8.0, so I am not sure if it is a good choice to migrate it as stock has changed a lot since 8.0.
So, I think there is three options:
- Maintain two unrelated modules
- Migrate and extend `stock_transport_multi_address`
- Create a module (`stock_location_address`) that allows to assign an address on locations and make `stock_transport_multi_address` dependant on it when it is migrated

I think that the third option is probably the simplest and more useful. What do you think?

etobella: I have made two small modules that could be used on the third option:
https://github.com/etobella/stock-logistics-transport/tree/11.0-add-stock_location_address

andhit-r: @etobella i'm still v8 user so i have a little knowledge regarding changes on later version. But i prefer option 3 since address field attach on location not picking type.

etobella:
I close this PR, because I have created a PR with the third option.
oca/stock-logistics-transport#46

PR #533, opened 2018-03-28
PORT 10 quick purchase

bealdav:
[ ] test to fix


quick_purchase/views/purchase_view.xml
```diff
@@ -7,13 +7,12 @@
         <field name="arch" type="xml">
             <field name="order_line" position="before">
                 <button name="add_product"
-                        string="Add product"
+                        string="Add"
```
legalsylvain: I was thinking to move the "Add" button in the header part. It will save space, and it's more Odoo spirit. what do you think ?

bealdav: you're right

mourad-ehm: I think from an ergonomic point of view it's better to have the button next to the command lines. Indeed, the user can immediately realize the utility of the button. If you put it in the head, it will be drowned in several buttons.

bealdav: @legalsylvain  What do you think of @mourad-ehm argument ? It makes sense

legalsylvain: Well, matter of taste. The "problem" I see is 
- that you loose space
- It's not in the Odoo UI Spirit. Buttons are very rarely in the form. Generally in the header.

But well, not a blocking point for me.

bealdav: finally I'm ok with mourad here

bealdav:
@mourad-ehm could you fix this test please 

https://travis-ci.org/OCA/purchase-workflow/jobs/469482909#L1428

PierrickBrun:
https://github.com/OCA/server-tools/pull/1510 needs to be merged for tests to pass

PR #463, opened 2017-10-05
[10.0] [MIG] product_supplier_code_purchase

nikul-serpentcs:
Product Supplier Code Purchase
==============================

This module adds to the purchase order line the supplier code defined in the product.

PR #535, opened 2018-04-04
[MIG][10.0]product_supplier_code_purchase

AaronHForgeFlow:
supersedes #463

### DIFF
--- a/product_supplier_code_purchase/models/purchase_order.py
+++ b/product_supplier_code_purchase/models/purchase_order.py
@@ -1,0 +1,29 @@
+# -*- coding: utf-8 -*-
+# Copyright 2015-17 Eficent Business and IT Consulting Services, S.L.
+#           (http://www.eficent.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from openerp import api, fields, models
+
+
+class PurchaseOrderLine(models.Model):
+    _inherit = 'purchase.order.line'
+
+    @api.multi
+    def _compute_product_supplier_code(self):
+        product_supplierinfo_obj = self.env['product.supplierinfo']
+        for line in self:
+            partner = line.order_id.partner_id
+            product = line.product_id
+            if product and partner:
+                supplier_info = product_supplierinfo_obj.search([
+                    '|', ('product_tmpl_id', '=', product.product_tmpl_id.id),
+                    ('product_id', '=', product.id),
+                    ('name', '=', partner.id)], limit=1)
+                if supplier_info:
+                    code = supplier_info.product_code or ''
+                    line.product_supplier_code = code
+        return True
+
+    product_supplier_code = fields.Char(string='Product Supplier Code',
+                                        compute=_compute_product_supplier_code)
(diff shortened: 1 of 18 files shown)


product_supplier_code_purchase/models/purchase_order.py
```diff
@@ -0,0 +1,27 @@
...
+    def _compute_product_supplier_code(self):
+        for line in self:
+            supplier_info = line.product_id.seller_ids.filtered(
+                lambda s: (s.product_id == line.product_id and
+                           s.name == line.partner_id))
+            if supplier_info:
+                code = supplier_info.product_code or ''
+                line.product_supplier_code = code
+
+    product_supplier_code = fields.Char(
```
rousseldenis: @aheficent Put fields definition on top. So, you can make
```python
compute='_compute_product_supplier_code'
```

product_supplier_code_purchase/models/purchase_order.py
```diff
@@ -0,0 +1,27 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+
+
+class PurchaseOrderLine(models.Model):
+    _inherit = 'purchase.order.line'
+
+    @api.multi
+    @api.depends('product_id.seller_ids.name',
```
rousseldenis: @aheficent Make it depends on partner_id too.

product_supplier_code_purchase/models/purchase_order.py
```diff
@@ -0,0 +1,27 @@
...
+                 'product_id.seller_ids.product_code')
+    def _compute_product_supplier_code(self):
+        for line in self:
+            supplier_info = line.product_id.seller_ids.filtered(
```
rousseldenis: @aheficent Careful with seller_ids as you can have several lines for a product with the same partner (supplierinfo is also supplier pricelist).
This code will crash if several.

You should maybe take supplierinfo[0].

rousseldenis (changes requested):
Code review. Minor changes

AaronHForgeFlow:
Pending tests still

rousseldenis (approved):

MiquelRForgeFlow (approved):

LoisRForgeFlow (approved):
:+1:

product_supplier_code_purchase/models/purchase_order.py
```diff
@@ -0,0 +1,27 @@
...
+
+from odoo import api, fields, models
+
+
+class PurchaseOrderLine(models.Model):
+    _inherit = 'purchase.order.line'
+
+    product_supplier_code = fields.Char(
+        string='Product Supplier Code',
+        compute='_compute_product_supplier_code'
```
legalsylvain, line 14: Should be great to have a function inv defined, to have the possibility, from the purchase order, to change the supplier code defined in the supplierinfo.

don't you think ? 

regards.

AaronHForgeFlow:
@legalsylvain that may be useful in some cases. I'm not used to inverse methods. When the user changes the supplier code in the PO line would that always change the code in the supplier.info?

legalsylvain:
It's just an idea !

But well : 

-> I think it's very usefull for the users, when the code is not defined, to set directly it in the current purchase order, instead of opening the product, and then the supplierinfo. (usability improved, reducing click quantity)

-> If the code is not set, (or wrong), changing it in the purchase order should update it in the product supplierinfo. I don't see any use case when the user doesn't want to update the code defined in the product.

AaronHForgeFlow:
Yes I think that would be a good improvement, avoiding many mouse clicks.

PR #539, opened 2018-04-13
[9.0] [FIX] Purchase Discount: use prepare method instead of new override

ThomasBinsfeld:
The override of the new() method produced some wrong behaviour in onchange methods
Use a clean prepare method override like in 10.0


pedrobaeza:
Please change module version number

ThomasBinsfeld:
@pedrobaeza Done, thanks

pedrobaeza (approved):

sbidoul (approved):

StefanRijnhart (approved):

PR #542, opened 2018-04-20
[ADD] New module to archive Purchase Orders and remove rights to dele…

BT-rmartin:
…te them.


purchase_archive/README.rst
```diff
@@ -0,0 +1,16 @@
+Purchase Archive
```
RoelAdriaans: Please use the OCA template for the readme: https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

purchase_archive/__manifest__.py
```diff
@@ -0,0 +1,37 @@
+# -*- encoding: utf-8 -*-
```
RoelAdriaans: Please use the OCA manifest, https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py as template

purchase_archive/models/purchase_order_ext.py
```diff
@@ -0,0 +1,11 @@
+# Copyright 2018 brain-tec AG - Raúl Martín
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+import odoo.addons.decimal_precision as dp
```
RoelAdriaans: Useless import

purchase_archive/models/purchase_order_ext.py
```diff
@@ -0,0 +1,11 @@
+# Copyright 2018 brain-tec AG - Raúl Martín
```
RoelAdriaans: Rename file to suit Odoo and OCA Guinelines , should be purchase_order.py

purchase_archive/security/ir.model.access.csv
```diff
@@ -0,0 +1,3 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
```
RoelAdriaans, line 1: Why redefine the security groups, as they are never used in this module?

purchase_archive/views/purchase_views_ext.xml
```diff
@@ -0,0 +1,33 @@
+<odoo>
```
RoelAdriaans: Rename file so suit oca/odoo standards, (purchase_order.xml)

RoelAdriaans (changes requested):
Code review, no functional test

purchase_archive/security/ir.model.access.csv
BT-rmartin: Because I'm removing the deletion right as now they are archived instead.

purchase_archive/README.rst
BT-rmartin: As far as I understand I just have to add the files into the readme folder and the README.rst is automatically generated?

BT-rmartin: Or do I have to execute somehow gen_addon_readme.py?

RoelAdriaans (approved):

bealdav (approved):

HviorForgeFlow (changes requested):
Code Review
Suggestion:
I wouldn't change that access right, I would use a ir.rule to define this restriction, and I would let user delete draft PO's.

PR #550, opened 2018-05-18
[11.0] Add purchase_line_procurement_group

grindtildeath:


jgrandguillaume (approved):
LGTM

max3903 (approved):

purchase_line_procurement_group/README.rst
```diff
@@ -0,0 +1,57 @@
+.. image:: https://img.shields.io/badge/license-AGPL--3-blue.png
+   :target: https://www.gnu.org/licenses/agpl
+   :alt: License: AGPL-3
+
+===============================
+Purchase Line Procurement Group
+===============================
+
+This module changes the way procurements generate purchase order lines.
+If some procurements are runned for the same products and locations, but have
```
JordiBForgeFlow: are run

purchase_line_procurement_group/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2018 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "Purchase Line Procurement Group",
+    "summary": "Group purchase order line according to procurement group",
+    "version": "11.0.1.0.0",
+    "category": "Purchase",
+    "website": "https://github.com/OCA/stock-logistics-workflow",
```
JordiBForgeFlow: change to purchase-workflow

purchase_line_procurement_group/tests/__init__.py
```diff
@@ -0,0 +1,3 @@
+# -*- coding: utf-8 -*-
```
JordiBForgeFlow: Remove all headers in init files.

JordiBForgeFlow (approved):
:+1: minor comments

grindtildeath:
It seems there's a different behaviour between odoo/odoo and OCA/OCB here :fearful:

grindtildeath:
It seems the travis issue should be fixed through rebase because of https://github.com/OCA/purchase-workflow/commit/659d4deaa6693a63f7923f03579e7885770bc8dc

grindtildeath:
I might need some help on this.

Explanation on last commit :
Raising the quantity fixes the test, but I'm not sure it's correct.
Let's consider the case functionnaly : 
- We have two Warehouses : WH and WH2
- WH2 is supplied through WH by creating a procurement (WH > WH2 route)
- There's an orderpoint on each warehouse stock location (WH/Stock and WH2/Stock) for the same product.

Test case : By running the scheduler, we don't know which orderpoint is going to be processed first so here are the two possibilities before the last change :

1a. As OP/WH2 is processed first, supply moves are generated for a quantity of 20 (max_qty) on WH2/Stock < WH/Stock < WH/QC. 
    Then when WH orderpoint is processed, its need (qty min 15) will be covered by the existing supply of 20, and no other supply is triggered. 

1b. As OP/WH is processed first, supply moves are generated for a quantity of 30 (max_qty) on WH/Stock < WH/QC.
    Then when WH2 orderpoint is processed, its need (qty min 10) will be covered by the supply of 30 and nothing is generated.
    That feels weird because if 20 (max qty OP/WH2) are removed from the 30 to supply WH2, the final qty on WH will be below the OP/WH and another supply will have to be triggered :/

Now with the last change here are the two possibilities :

2a. As OP/WH2 is processed first, supply moves are generated for a quantity of 20 (max_qty) on WH2/Stock < WH/Stock < WH/QC. 
    Then when WH orderpoint is processed, its need (min qty 40) will not be covered by the existing supply and another supply of 80 is triggered.

2b. As OP/WH is processed first, supply moves are generated for a quantity of 80 (max_qty) on WH/Stock < WH/QC.
    Then when WH2 orderpoint is processed, its need (qty max 20) will be covered by the supply of 30 but the supply of 20 is still generated.
    That feels weird because it seems to be more or less the same case as 1a where the supply is not triggered :/

I don't really know what's wrong or what's right, so some functional advice might help.

gurneyalex:
@grindtildeath  what you describe could be explained by http://portingguide.readthedocs.io/en/latest/dicts.html#changed-key-order and it's impact on this loop: https://github.com/odoo/odoo/blob/0b412c250afc60131188ce248087dae2c2a53574/addons/stock/models/procurement.py#L338

grindtildeath:
Thanks to @gurneyalex, we found the explanation about the test not working properly (https://github.com/OCA/purchase-workflow/pull/550#issuecomment-391285555) :

As the orderpoints are processed with a random order (https://github.com/OCA/purchase-workflow/pull/550#issuecomment-391328516), the test did fail when the OP of WH2/Stock was processed first, because the virtual available qty (and the outgoing_qty) on WH/Stock was then wrongly computed (20.0 instead of 0.0) at the time of processing the OP of WH/Stock, thus not generating its procurement.

However, when the OP of WH/Stock was processed first, the computation of virtual available qty on WH2/Stock was always 0.0 and the test did pass.

Finally, calling `_parent_store_compute()` on `stock.location` fixes the issue, as both OP are now computed correctly.

jgrandguillaume (approved):
LGTM thanks

pedrobaeza:
Now that I see this, IMO a better and more flexible solution (and more because I got some hooks in Odoo on the v11 code) is https://github.com/OCA/purchase-workflow/tree/10.0/procurement_purchase_no_grouping

PR #559, opened 2018-06-20
[10.0] purchase_commercial_partner: add compute_sudo=True and readonly=True

alexis-via:


pedrobaeza (approved):
Change version number

pedrobaeza:
Lint errors unrelated. I'll fix it after fast-tracking this.

PR #560, opened 2018-07-03
[FIX] remove translation for an non existing module

legalsylvain:
Trivial PR. remove from branch 8.0 a repository purchase_triple_discount that only contain translation.


pedrobaeza (approved):
Fast-tracking as obvious

legalsylvain:
thanks !

PR #564, opened 2018-07-18
[11.0] purchase_line_procurement_group : Fix stock move generation

grindtildeath:
TODO :
 - [x] Add tests


max3903 (approved):

purchase_line_procurement_group/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 {
     "name": "Purchase Line Procurement Group",
     "summary": "Group purchase order line according to procurement group",
-    "version": "11.0.1.0.0",
+    "version": "11.0.1.1.0",
```
simahawk, line 6: I see only "fix" commits: shall we go for 1.0.1?

purchase_line_procurement_group/tests/test_po_line_proc_group.py
```diff
@@ -87,3 +87,15 @@ def test_po_line_proc_group(self):
                 self.assertAlmostEqual(line.product_qty, 20)
             else:
                 self.assertAlmostEqual(line.product_qty, 30)
+        po.button_confirm()
```
simahawk:
```python
# ensure generated stock move are not merged together
# if they come from purchase order lines with different procurement groups
```
can be a proper comment?

simahawk (approved):
LG, can you add a comment on test tho?

purchase_line_procurement_group/__manifest__.py
grindtildeath: Actually the "generation of stock move" part is more an improvement, as it wasn't considered before. Moreover, it's already used in one of our projects with 11.0.1.1.0 so let's keep it like that.

PR #569, opened 2018-07-23

vinclar:
Add analytic field in the return to be copied from procurement


vinclar:
@benwillig and @rvalyi jbeficent

Is it posible to add this to the project to allow to be copied the analytic account from the procurement?

PR #571, opened 2018-07-29
[10.0] Allow copy analytic field from procurement on purchase request…

vinclar:
… lines


vinclar:
Add tests param

PR #572, opened 2018-07-29
[10.0] Allow copy analytic field from procurement on purchase request…

vinclar:
… lines

@benwillig @jbeficent 

I know that the test aren't passing, but I don't know how they work. If you think it's useful for the community please merge the change.
Thanks


rousseldenis:
@vinclar The problem here is that you have to depend on procurement_analytic (see Travis log) which is defined here https://github.com/OCA/account-analytic/tree/10.0.

So, maybe putting a new module ('purchase_request_procurement_analytic') that inherits the _prepare_purchase_request_line() function (you don't want necessarily to implement analytic)

AaronHForgeFlow:
agree with @rousseldenis this feature must be in another module that inherits from procurement_analytic and purchase_request_procurement

vinclar:
Thanks, will give a try to develop this simple new module

PR #574, opened 2018-08-08
[10.0][FIX] Subcontracted service: can't adapt type 'procurement.rule'

jvpascual:
when a sale order is confirmed, 
ProgrammingError: can't adapt type 'procurement.rule'

### DIFF
--- a/subcontracted_service/models/procurement.py
+++ b/subcontracted_service/models/procurement.py
@@ -21,7 +21,7 @@ class ProcurementOrder(models.Model):
         res = super(ProcurementOrder, self)._find_suitable_rule()
         if self._is_subcontracted_service():
             return (
-                self.warehouse_id.subcontracting_service_proc_rule_id.id
+                self.warehouse_id.subcontracting_service_proc_rule_id
             )
         return res
 
@@ -31,10 +31,6 @@ class ProcurementOrder(models.Model):
         if not res:
             rule_id = self._find_suitable_rule()
             if rule_id:
-                self.write({'rule_id': rule_id})
-                if isinstance(rule_id, int):
-                    self.write({'rule_id': rule_id})
-                else:
-                    self.write({'rule_id': rule_id.id})
+                self.write({'rule_id': rule_id.id})
                 return True
         return res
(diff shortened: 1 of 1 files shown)


JordiBForgeFlow:
@hveficent can you review?

HviorForgeFlow:
@jvpascual  I cannot reproduce error you've got. Any idea about which modulo or configuration has made it raise?

jvpascual:
Hi @hveficent,
in the attached error trace, can see the modules, the unique that is not in oca is vit_mrp_cost,
imagen

Kind regards,

HviorForgeFlow:
@jvpascual can you copy the full log in plain text??

jvpascual:
Hi @hveficent, the file is attached.
[subcontracted_service_error.txt](https://github.com/OCA/purchase-workflow/files/2274210/subcontracted_service_error.txt)

subcontracted_service/models/procurement.py
```diff
@@ -31,6 +31,9 @@ def _assign(self):
         if not res:
             rule_id = self._find_suitable_rule()
             if rule_id:
-                self.write({'rule_id': rule_id})
```
HviorForgeFlow, line 34: change to self.write({'rule_id': rule_id.id})

HviorForgeFlow: in _find_suitable_rule if subcontracted service remove last .id like:
self.warehouse_id.subcontracting_service_proc_rule_id

HviorForgeFlow: I think this way it should work, could you try in your instance?

HviorForgeFlow (changes requested):

subcontracted_service/models/procurement.py
jvpascual: Yes, it works.
i've just leave the old self.write({'rule_id': rule_id}) in case it come as object.

HviorForgeFlow:
Travis is failing. You must attend the other change, I think you have some module that is overwriting/extending function _find_suitable_rule(), could you find out which module it is..

jvpascual:
the function _find_suitable_rule(), only appears in the next modules: stock, procurement and subcontracted_service.
in some cases, can arrive as an object.

HviorForgeFlow:
Ok, I got it. But then fix _find_suitable_rule in this module to return an object and keep rule_id write as previous commit.

jvpascual:
Done.

HviorForgeFlow (approved):
Please squash commits into 1

jvpascual:
i've lost the origin.

PR #575, opened 2018-08-10
[10.0][FIX] Subcontracted service: can't adapt type 'procurement.rule'

jvpascual:
I've created a new thread, because something went wrong in the other one.


HviorForgeFlow (approved):
Open upon #574

LoisRForgeFlow (approved):
Thanks :+1:

PR #578, opened 2018-08-24
[10][ADD] purchase update date order

hparfr:
This module recalculate the purchase.date_order when needed..
(It's the last time to place an order and be delivered on time).


bealdav:
Just that to fix
https://travis-ci.org/OCA/purchase-workflow/jobs/420001085#L615

hieulucky111 (approved):

PR #585, opened 2018-09-17
[FIX] Do not round the unit price after discount is applied

StefanRijnhart:
To prevent differences with the invoice that is created from the purchase,
the unit price must not be rounded after the discount is applied.

E.g. 195PCE of a 14.49 product against 7% discount should be computed as
round(195 * 14.49 * .93, prec), not as 195 * round(14.49 * .93, prec)

### DIFF
--- a/purchase_discount/models/purchase_order.py
+++ b/purchase_discount/models/purchase_order.py
@@ -2,10 +2,12 @@
 # Copyright 2004-2009 Tiny SPRL (<http://tiny.be>).
 # Copyright 2016 ACSONE SA/NV (<http://acsone.eu>)
 # Copyright 2015-2017 Tecnativa - Pedro M. Baeza
+# Copyright 2018 Opener B.V. - Stefan Rijnhart
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import api, fields, models
 import odoo.addons.decimal_precision as dp
+from odoo.tools import frozendict
 
 
 class PurchaseOrder(models.Model):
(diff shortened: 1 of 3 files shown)


StefanRijnhart:
I believe the tests were disabled on this one so I set at_install to True as well

purchase_discount/tests/test_purchase_discount.py
```diff
@@ -7,8 +7,8 @@
 from odoo import fields
 
 
-class TestPurchaseOrder(common.HttpCase):
-    at_install = False
+class TestPurchaseOrder(common.TransactionCase):
+    at_install = True
```
pedrobaeza: Please enable `post_install` instead for avoiding conflicts when testing with other modules, but indeed both type of tests are disabled right now.

StefanRijnhart: Done!

purchase_discount/tests/test_purchase_discount.py
```diff
@@ -7,9 +7,9 @@
 from odoo import fields
 
 
-class TestPurchaseOrder(common.HttpCase):
+class TestPurchaseOrder(common.TransactionCase):
```
pedrobaeza: This was change to `HttpCase` for avoiding a block due to wkhtmltopdf and it's a trick for avoiding it. Are you changing this for any particular reason?

purchase_discount/models/purchase_order.py
```diff
@@ -47,18 +49,28 @@ class PurchaseOrderLine(models.Model):
...
+            discounted_price_unit = line._get_discounted_price_unit()
+            if discounted_price_unit != line.price_unit:
+                precision = line.order_id.currency_id.decimal_places
+                company = line.company_id or self.env.user.company_id
+                if company.tax_calculation_rounding_method == 'round_globally':
+                    precision += 5
+                orig_context = self.env.context
+                price = round(
+                    line.product_qty * discounted_price_unit, precision)
+                self.env.context = frozendict(
```
pedrobaeza, line 69: Changing context this way can be a little hackish and maybe affect cache mechanism. Are you sure there's no side effect?

StefanRijnhart: I have not noticed any side effect (I'm using it elsewhere too) but you may ask @Yajo https://github.com/odoo/odoo/issues/7472#issuecomment-247058243

(thanks for the trick, @yajo!)

purchase_discount/tests/test_purchase_discount.py
StefanRijnhart: I reviewed the tests because they were disabled and it did not look like a HttpCase to me. Reverted this.

purchase_discount/models/purchase_order.py
yajo: Indeed this is a hack that should be avoided if possible. AFAIK the only use case where it is justified is for onchanges.

Please change it for:

```python
self.with_context(base_values=(price, price, price))._compute_amount()
```

which is the correct way of altering context.

pedrobaeza: The problem with that, @Yajo, is that compute fields depend on that context, so we won't get the achieved effect. That's why we initially replace the field value instead of doing this.

StefanRijnhart: @yajo this is an onchange, I mean, it is a computed field that is computed when dependent fields are modified in the interface, and where `self` can be an NewId.

yajo: OK, I thought that computed fields did not have this problem. It's OK then.

MiquelRForgeFlow (approved):
LGTM

pedrobaeza (approved):

PR #586, opened 2018-09-19
[11.0][ADD] purchase_order_archive

astirpe:
On a system with a high volume of purchases, the number of purchase orders displayed in the list view can become huge. This module allows to archive Purchase Orders that are in status Locked or Cancelled.


pedrobaeza:
I think this will affect purchase analysis report, isn't it?

astirpe:
@CasVissers

astirpe:
@pedrobaeza The `purchase_report` view in postgres does not make use of the newly added field `active`. So this module does not affect the purchase analysis report, in the sense that the archived purchases will be taken into account in the report.

The module was mainly meant to hide the old inactive purchase orders from the purchase order list.

pedrobaeza:
OK, indeed. Thanks for the clarification.

bodedra (approved):
LGTM :+1:

pankk (approved):

tarteo (approved):
Simple but nice, looks good. The readme was probably most of the work ;)

PR #590, opened 2018-09-25
[ADD] purchase_order_line_stock_available module

cubells:
Simple migration of module from https://github.com/odoomrp/odoomrp-utils repository.

This module allows to view the virtual stock quantity of a product in the
purchase order line.

cc @Tecnativa


purchase_order_line_stock_available/models/purchase.py
```diff
@@ -0,0 +1,7 @@
+from odoo import fields, models
+
+
+class PurchaseOrderLine(models.Model):
+    _inherit = 'purchase.order.line'
+
+    virtual_available = fields.Float(related='product_id.virtual_available')
```
rousseldenis: Should be readonly

rousseldenis (changes requested):
Code review. Minor comment.

@cubells Do you have tested it in terms of performance with a purchase with lot of lines?

IMHO this is not a migration as the module was not part of OCA before

cubells:
@rousseldenis I've tested the module in a purchase with 237 lines and I have not noticed performance problems.

rousseldenis:
> @rousseldenis I've tested the module in a purchase with 237 lines and And I have not noticed performance problems.

Ok, with a lot of quants and existing moves too ?

cubells:
@rousseldenis yes

rousseldenis (approved):
Code review

pedrobaeza (approved):

PR #593, opened 2018-10-05
[FIX] purchase_landed_cost: error on deleting cost lines

cubells:
Before this pr:
If user try to delete some cost line and this cost line is referenced in any expense line into the field 'affected lines', a not _sensitive_ error is shown to user.

After this pr:
Error is catched and an _clear_ message is shown to user.

cc @Tecnativa

### DIFF
--- a/purchase_landed_cost/__manifest__.py
+++ b/purchase_landed_cost/__manifest__.py
@@ -5,14 +5,14 @@
 
 {
     'name': 'Purchase landed costs - Alternative option',
-    'version': '10.0.1.0.0',
-    "author": "OdooMRP team,"
-              "AvanzOSC,"
-              "Tecnativa,"
-              "Joaquín Gutierrez,"
-              "Odoo Community Association (OCA)",
+    'version': '10.0.2.0.0',
+    "author": u"OdooMRP team,"
+              u"AvanzOSC,"
+              u"Tecnativa,"
+              u"Joaquín Gutierrez,"
+              u"Odoo Community Association (OCA)",
     'category': 'Purchase Management',
-    'website': 'http://www.odoomrp.com',
+    'website': 'https://github.com/OCA/purchase-workflow',
     'summary': 'Purchase cost distribution',
     'depends': [
         'stock',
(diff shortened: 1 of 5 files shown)


pedrobaeza:
Can't this be handled through ondelete="cascade"?

cubells:
@pedrobaeza 
cost_lines has `ondelete` parameter: https://github.com/Tecnativa/purchase-workflow/blob/11.0-product_landed_cost_fix_error/purchase_landed_cost/models/purchase_cost_distribution.py#L120

and error is raised when deleting a cost line without remove reference in expense lines.

pedrobaeza:
That ondelete doesn't serve. It should be on the `many2one` part, not on the `one2many`

cubells:
I've tested `ondelete` on`affected_lines` field and it does not work, but `affected_lines` field is Many2many.

pedrobaeza (approved):

pedrobaeza:
@Roodin were you finally using this?

purchase_landed_cost/__manifest__.py
```diff
@@ -5,12 +5,12 @@
...
-              "Joaquín Gutierrez,"
-              "Odoo Community Association (OCA)",
+    'version': '10.0.2.0.0',
+    "author": u"OdooMRP team,"
+              u"AvanzOSC,"
+              u"Tecnativa,"
+              u"Joaquín Gutierrez,"
+              u"Odoo Community Association (OCA)",
     'category': 'Purchase Management',
     'website': 'http://www.odoomrp.com',
```
rafaelbn: Change to https://github.com/OCA/purchase-workflow
```suggestion
    'website': 'https://github.com/OCA/purchase-workflow',
```
As said in https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py#L11

rafaelbn (approved):

rafaelbn:
Are we going to improve  codecov/patch — 76.47% of diff hit (target 81.1%) ?

Can we merge ?

pedrobaeza:
Yes, merging, the coverage is good on the project and module, but as this is reducing code, that's why it's red.

PR #596, opened 2018-10-18
[11.0][MIG] product_by_supplier

LoisRForgeFlow:
Migration to v11.

PR #603, opened 2018-10-31
[DEL] product_by_supplier

MiquelRForgeFlow:
As @lreficent commented here https://github.com/OCA/purchase-workflow/pull/596#issuecomment-434623285, the migration of product_by_supplier from v10 to v11 is handled here https://github.com/OCA/OpenUpgrade/pull/1595.


pedrobaeza:
Fast-tracking this.

PR #606, opened 2018-11-12
[11.0][FIX] Purchase_discount set correct price on stock account entries when discount is set

### LINKED ISSUE #562: [11.0] purchase_discount: Stock moves created with unit_price before discount
The module is doing the right thing using `_get_stock_move_price_unit()` from the purchase order: https://github.com/odoo/odoo/blob/f6b3247c31ee176944a060afb911021d071d8a32/addons/purchase/models/purchase.py#L682

But there's also `_get_price_unit()` https://github.com/odoo/odoo/blob/ab5fcb29650349fa641c6130bf6dcbdc1ec28a07/addons/purchase/models/stock.py#L37 which makes things a bit confusing on which approach to take, because right now the stocks moves are being created with the price used by `_get_price_unit()`

A solution might be to apply the same logic from https://github.com/OCA/purchase-workflow/blob/11.0/purchase_discount/models/purchase_order.py#L51 to `_get_price_unit()` 

Right now, the bigger the discount the bigger the error on the cost when using Perpetual accounting.

Fixing this might close:
https://github.com/OCA/purchase-workflow/issues/413

CasVissers-360ERP:
Fixes #562


purchase_discount/models/__init__.py
```diff
@@ -2,3 +2,4 @@
 
 from . import purchase_order
 from . import account_invoice
+from . import stock_move
```
pedrobaeza: Please add final line break

pedrobaeza (commented):
Are you sure the price doesn't contain already the discount. Here we set it before calling super:

https://github.com/OCA/purchase-workflow/blob/a6a702f07f6927ded22402a9b3cd0ff36a43f57c/purchase_discount/models/purchase_order.py#L51

CasVissers-360ERP:
@pedrobaeza Yes I'm 100% sure. Will sign the CLA now.

pedrobaeza:
But then the other one should be removed?

pedrobaeza:
I have arrived to the bug this morning and add a better fix in 90ca5317705b711aa75932df99c0f581a40db218, as yours exclude some cases like taxes included in price. I have also included a unit test for checking this.

Thanks any way for your PR.

PR #608, opened 2018-11-22
[12.0][MIG] purchase_discount: Migrate to v12

sudhir-erpharbor:
Migrated purchase_discount module in v12.
Please review and give feedback.


purchase_discount/__manifest__.py
```diff
@@ -4,15 +4,17 @@
...
-    "name": "Purchase order lines with discounts",
+    "name": "Purchase Order Lines with Discounts",
     "author": "Tiny, "
               "Acysos S.L., "
               "Tecnativa, "
-              "ACSONE SA/NV,"
+              "ACSONE SA/NV, "
               "Odoo Community Association (OCA)",
-    "version": "11.0.1.0.1",
+    "version": "12.0.1.0.1",
```
pedrobaeza: 12.0.1.0.0

pedrobaeza (changes requested):
There are now some hook methods for not doing the discount in such hacky way. Can you please check them?

purchase_discount/__manifest__.py
sudhir-erpharbor: Thanks for your feedback.
Done. Would you please explain me what do you mean by hook? If you can please give any example of that, then I can sure work on that and check that.

pedrobaeza: I mean to overwrite this method instead of current hacking:

https://github.com/odoo/odoo/blob/02ce0022b2a5562a4b1c6f912a901d5b863df238/addons/purchase/models/purchase.py#L452

purchase_discount/models/purchase_order.py
```diff
@@ -0,0 +1,65 @@
...
+    discount = fields.Float(
+        string='Discount (%)', digits=dp.get_precision('Discount'),
+    )
+
+    _sql_constraints = [
+        ('discount_limit', 'CHECK (discount <= 100.0)',
+         'Discount must be lower than 100%.'),
+    ]
+
+    def _get_discounted_price_unit(self):
```
nikul-serpentcs, line 47: @sudhir-erpharbor IMO Add `@api.multi`

purchase_discount/tests/test_purchase_discount.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+class TestPurchaseOrder(common.SavepointCase):
+    at_install = False
+    post_install = True
+
+    @classmethod
+    def setUpClass(cls):
+        super(TestPurchaseOrder, cls).setUpClass()
+        cls.product_1 = cls.env['product.product'].create({
```
nikul-serpentcs: @sudhir-erpharbor IMO, here `cls.env['product.product']` and other objects & ref use multiple time, So add global variable and use it.

nikul-serpentcs (commented):
Improve Code

purchase_discount/models/purchase_order.py
sudhir-erpharbor: self.ensure_one() is already added in the method. I think it does not matter to have api.multi.

purchase_discount/tests/test_purchase_discount.py
sudhir-erpharbor: Done

purchase_discount/__manifest__.py
sudhir-erpharbor: Done. Please review again

purchase_discount/tests/test_purchase_discount.py
```diff
@@ -0,0 +1,121 @@
...
+        self.purchase_order.button_confirm()
+        moves = self.purchase_order.picking_ids.move_lines
+        move = moves.filtered(lambda x: x.purchase_line_id == self.po_line_1)
+        self.assertEqual(move.price_unit, 5,)
+        move = moves.filtered(lambda x: x.purchase_line_id == self.po_line_2)
+        self.assertEqual(move.price_unit, 161)
+        move = moves.filtered(lambda x: x.purchase_line_id == self.po_line_3)
+        self.assertEqual(move.price_unit, 10)
+        # Change price to launch a recalculation of totals
+        print ("\n\nPPPPPPPPP    ")
```
nikul-serpentcs: @sudhir-erpharbor Remove It

sudhir-erpharbor: Ooops! Thanks for pointing it out.

purchase_discount/models/purchase_order.py
```diff
@@ -0,0 +1,61 @@
...
+        value before calling super and restoring it later for assuring
+        maximum inheritability.
+        """
+        price_unit = False
+        price = self._get_discounted_price_unit()
+        if price != self.price_unit:
+            # Only change value if it's different
+            price_unit = self.price_unit
+            self.price_unit = price
+        price = super(PurchaseOrderLine, self)._get_stock_move_price_unit()
```
AaronHForgeFlow, line 71: Maybe it is possible to call super at the beginning and then check the discount and change the value is needed, don't you think?

AaronHForgeFlow (commented):
Functional tests good. Only one comment.

nikul-serpentcs (approved):
Code Review LGTM

purchase_discount/models/purchase_order.py
pedrobaeza: `@api.multi` is the default decorator since v10, so no need to explicitly add it.

purchase_discount/models/purchase_order.py
sudhir-erpharbor: @aheficent  Not sure. But looking at the code I am feeling that at first price_unit is changed with the discounted amount (unit_price - discount) and then calling super so that everything in the super method(s) will get calculated based on discounted unit_price and then changing the unit_price back to original.

AaronHForgeFlow: But the original method is only getting the price unit. Then you can change it or not. Anyway it is not blocking.

sudhir-erpharbor:
@pedrobaeza Any suggestions / feedback?

purchase_discount/models/purchase_order.py
```diff
@@ -13,18 +13,14 @@ class PurchaseOrderLine(models.Model):
 
     @api.depends('discount')
     def _compute_amount(self):
```
pedrobaeza, line 30: No need to keep this empty method.

purchase_discount/__manifest__.py
```diff
@@ -4,20 +4,22 @@
 # © 2016 ACSONE SA/NV (<http://acsone.eu>)
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
 {
-    "name": "Purchase order lines with discounts",
+    "name": "Purchase Order Lines with Discounts",
```
pedrobaeza: Why making this unneeded changes?

purchase_discount/__manifest__.py
```diff
@@ -4,20 +4,22 @@
...
     "author": "Tiny, "
               "Acysos S.L., "
               "Tecnativa, "
-              "ACSONE SA/NV,"
+              "ACSONE SA/NV, "
```
pedrobaeza: There's no need for extra spaces. They are ripped out when parsed by manifest reader.

purchase_discount/__manifest__.py
```diff
@@ -4,20 +4,22 @@
...
               "Odoo Community Association (OCA)",
-    "version": "11.0.1.0.1",
+    "version": "12.0.1.0.0",
+    "license": "AGPL-3",
```
pedrobaeza, line 14: Why do you change the order of the keys? It's better to just reduce the diff to minimum

purchase_discount/models/purchase_order.py
```diff
@@ -13,18 +13,14 @@ class PurchaseOrderLine(models.Model):
...
-            super(PurchaseOrderLine, line)._compute_amount()
-            if price_unit:
-                line.price_unit = price_unit
+        super(PurchaseOrderLine, self)._compute_amount()
+
+    def _prepare_compute_all_values(self):
+        vals = super(PurchaseOrderLine, self)._prepare_compute_all_values()
+        # This is always executed for allowing other modules to use this
+        # with different conditions than discount != 0
+        vals.update({'price_unit': self._get_discounted_price_unit()})
```
pedrobaeza, line 35: No need to keep the method `_get_discounted_price_unit`, as now you already have an inheritable method for consecutive modifications (for example, the triple discount).

purchase_discount/models/purchase_order.py
```diff
@@ -13,18 +13,14 @@ class PurchaseOrderLine(models.Model):
...
-                price_unit = line.price_unit
-                line.price_unit = price
-            super(PurchaseOrderLine, line)._compute_amount()
-            if price_unit:
-                line.price_unit = price_unit
+        super(PurchaseOrderLine, self)._compute_amount()
+
+    def _prepare_compute_all_values(self):
+        vals = super(PurchaseOrderLine, self)._prepare_compute_all_values()
+        # This is always executed for allowing other modules to use this
```
pedrobaeza: This explanation is from super. Don't copy this here again.

purchase_discount/views/report_purchaseorder.xml
```diff
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <odoo>
+
```
pedrobaeza, line 3: No need to add extra noise with these unneeded changes.

purchase_discount/views/purchase_discount_view.xml
```diff
@@ -1,39 +1,40 @@
...
-    <field name="name">purchase_discount.order.line.tree</field>
-    <field name="model">purchase.order.line</field>
-    <field name="inherit_id" ref="purchase.purchase_order_line_tree"/>
-    <field name="arch" type="xml">
-      <field name="price_unit" position="after">
-        <field name="discount"/>
-      </field>
-    </field>
-  </record>
+    <record model="ir.ui.view" id="purchase_order_line_form2">
```
pedrobaeza, line 4: Please don't change indentation for not adding extra noise, although this doesn't conform 4 spaces indentation.

pedrobaeza (changes requested):
@sudhir-erpharbor thanks for the migration. In general, you should keep the diff at minimum in a migration although some things don't fit your code style or some general style convention.

There's an extra thing I thing is not needed: the method `_get_stock_move_price_unit`. Isn't this this method also call the same `_prepare_compute_all_values` method for getting the price? If not, this is the patch to do on Odoo.

purchase_discount/models/purchase_order.py
sudhir-erpharbor: I think we need it. I had to override `_compute_amount` to add `discount` field in depends. So that change in discount field will trigger the `_compute_amount` method and will recalculate the price.
@pedrobaeza

purchase_discount/__manifest__.py
sudhir-erpharbor: I thought the module name should be in upper case letters.
@pedrobaeza

purchase_discount/models/purchase_order.py
sudhir-erpharbor: AFAIK the `_get_discounted_price_unit` method contains a common code to calculate the price based on the discount. This method is being called from 2 different methods (`_prepare_compute_all_values` and `_get_stock_move_price_unit`)

Let me know if you still think that `_get_discounted_price_unit` method should be removed.
@pedrobaeza

purchase_discount/views/report_purchaseorder.xml
sudhir-erpharbor: Ok. Next time I will keep this in mind before making such changes.
Thanks for letting me know on this.

sudhir-erpharbor:
Made changes as per review.
@pedrobaeza

purchase_discount/models/purchase_order.py
sudhir-erpharbor: Done

legalsylvain:
Hi @sudhir-erpharbor. Thanks for porting this module. could you take a look on that issue : 
https://github.com/OCA/purchase-workflow/issues/623
could you be agree to add in your current PR, the feature that are in "purchase_supplierinfo_discount" ? 
Otherwise, I can do it.

Regards.

sudhir-erpharbor:
> Hi @sudhir-erpharbor. Thanks for porting this module. could you take a look on that issue :
> #623
> could you be agree to add in your current PR, the feature that are in "purchase_supplierinfo_discount" ?
> Otherwise, I can do it.
> 
> Regards.

Done. Purchase triple discount module is merged in this PR. Please review.

pedrobaeza:
> Done. Purchase triple discount module is merged in this PR. Please review.

:scream: it was `purchase_supplierinfo_discount`, not `purchase_triple_discount`.

sudhir-erpharbor:
> > Done. Purchase triple discount module is merged in this PR. Please review.
> 
> scream it was `purchase_supplierinfo_discount`, not `purchase_triple_discount`.

Oops!! Let me revert the changes and will merge the purchase_supplierinfo_discount.
CC: @pedrobaeza

sudhir-erpharbor:
Done. I will merge the code of `purchase_supplierinfo_discount` module of v11 to this module asap.
Thanks for pointing out the silly mistake :)
CC: @pedrobaeza

pedrobaeza:
You can simply remove the commits instead reverting them. At the end, we generally squash migration commits to one (although this case we can leave 2 commits, one for the migration of the module, and another for the merge of the module).

sudhir-erpharbor:
> You can simply remove the commits instead reverting them. At the end, we generally squash migration commits to one (although this case we can leave 2 commits, one for the migration of the module, and another for the merge of the module).

Done. Removed the commits of the `purchase_triple_discount` module.
Sorry about the noise.

CC: @pedrobaeza

purchase_discount/__manifest__.py
pedrobaeza: I don't think so. Please respect it as it was.

purchase_discount/models/purchase_order.py
pedrobaeza: OK

purchase_discount/models/purchase_order.py
pedrobaeza: Yes, remove it and include here the logic.

purchase_discount/views/purchase_discount_view.xml
pedrobaeza: This is still pending

purchase_discount/views/report_purchaseorder.xml
pedrobaeza: This is still pending.

pedrobaeza:
Please squash all your migration commits together.

This won't fully work until https://github.com/odoo/odoo/pull/29983 is merged.

eLBati:
I took the the liberty of creating a new PR, with cleaner commit history: https://github.com/OCA/purchase-workflow/pull/668

@sudhir-erpharbor  let me know if it's OK for you to move there and if there are other improvements left to do
cc @pedrobaeza

sudhir-erpharbor:
> I took the the liberty of creating a new PR, with cleaner commit history: #668
> 
> @sudhir-erpharbor let me know if it's OK for you to move there and if there are other improvements left to do
> cc @pedrobaeza

Ok that is fine.

PR #610, opened 2018-11-24
[11.0] purchase_exception : Corrected runbot id

JayVora-SerpentCS:


purchase_exception/README.rst
```diff
@@ -39,6 +39,7 @@ Contributors
 ------------
 
 * Mourad EL HADJ MIMOUNE <mourad.elhadj.mimoune@akretion.com>
+* Serpent Consulting Services Pvt. Ltd. <support@serpentcs.com>
```
pedrobaeza: Please don't add you as contributor changing only the runbot ID.

anandkansagra (approved):

pedrobaeza (changes requested):
See my comment above

JayVora-SerpentCS:
Updated

purchase_exception/README.rst
JayVora-SerpentCS: Removed. I added contributor as a fact that we contributed to correct the wrong direction

bodedra (approved):
:+1:

rousseldenis:
@OCA/procurement-purchase-maintainers We can merge this. Link is redirecting to runboat

PR #612, opened 2018-11-26
[ADD] purchase_order_line_deep_sort: new module

cubells:
Allow user to define sort criteria for purchase order lines.

cc @Tecnativa


purchase_order_line_deep_sort/models/purchase_order.py
```diff
@@ -0,0 +1,52 @@
...
+        sequence = 0
+        key = eval("lambda p: p.%s" % order)
+        for line in self.order_line.sorted(key=key, reverse=reverse):
+            sequence += 10
+            line.sequence = sequence
+
+    @api.multi
+    def write(self, values):
+        res = super(PurchaseOrder, self).write(values)
+        if 'order_line' in values or 'line_order' in values or \
```
nikul-serpentcs, line 53: I think here direct check via a list
`if ['order_line' , 'line_order' , 'line_direction'] in values`

purchase_order_line_deep_sort/tests/test_purchase_order_line_deep_sort.py
```diff
@@ -0,0 +1,158 @@
...
+
+class TestPurchaseOrderLineDeepSort(common.SavepointCase):
+
+    @classmethod
+    def setUpClass(cls):
+        super(TestPurchaseOrderLineDeepSort, cls).setUpClass()
+
+        po_model = cls.env['purchase.order']
+        cls.po_line_model = cls.env['purchase.order.line']
+        cls.product_1 = cls.env['product.product'].create({
```
nikul-serpentcs: here `cls.env['product.product']` and other objects use multiple time, so use as a global variable.

nikul-serpentcs (commented):
Improve Code

purchase_order_line_deep_sort/models/purchase_order.py
```diff
@@ -0,0 +1,52 @@
...
+            self.line_direction = False
+
+    @api.multi
+    def _sort_purchase_line(self):
+        if not self.line_order and not self.line_direction:
+            return
+        order = self.line_order
+        reverse = self.line_direction == 'desc'
+        sequence = 0
+        key = eval("lambda p: p.%s" % order)
```
chienandalu: Wouldn't it be better to use Odoo's `safe_eval`?

pedrobaeza: This is not Python code, and I prefer to not have an opened door for any kind of expression.

purchase_order_line_deep_sort/models/purchase_order.py
pedrobaeza: This is not valid, as the order of the keys should be the same. It would be valid if you do sets, but not with lists.

purchase_order_line_deep_sort/tests/test_purchase_order_line_deep_sort.py
pedrobaeza: This is not adding too much value in terms of performance, but we'll change it if we have to touch again the module.

purchase_order_line_deep_sort/models/purchase_order.py
pedrobaeza: I mean: this is coming from some pre-defined values in selection field, and doing `safe_eval` is not going to work because lambda expressions are forbidden there, but the scope of this is totally limited.

pedrobaeza (approved):

purchase_order_line_deep_sort/models/purchase_order.py
pedrobaeza: I'm seeing any way that we should silence the linter. Let me check alternatives.

pedrobaeza: I have used another safer technique for this.

pedrobaeza:
@Nikul-Chaudhary @chienandalu you can review again.

purchase_order_line_deep_sort/views/purchase_order_views.xml
```diff
@@ -0,0 +1,22 @@
...
+        <field name="model">purchase.order</field>
+        <field name="inherit_id" ref="purchase.purchase_order_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//notebook" position="inside">
+                <page string="Other Info">
+                    <group name="line_order">
+                        <label for="line_order"/>
+                        <div>
+                            <field name="line_order" class="oe_inline"/>
+                            (<field name="line_direction" class="oe_inline" attrs="{'required': [('line_order', '!=', False)]}"/>)
```
Nikul-Chaudhary, line 14: @cubells content in ()

pedrobaeza: That's on purpose for better UI

Nikul-Chaudhary (approved):
Code Review LGTM :+1:

purchase_order_line_deep_sort/models/purchase_order.py
```diff
@@ -0,0 +1,62 @@
...
+            return
+        reverse = self.line_direction == 'desc'
+        sequence = 0
+        sorted_lines = self.order_line.sorted(
+            key=lambda p: resolve_subfields(p, self.line_order),
+            reverse=reverse,
+        )
+        for line in sorted_lines:
+            sequence += 10
+            line.sequence = sequence
```
rousseldenis, line 48: @cubells @pedrobaeza You cannot update a recordset like that in a write() or a create(). You are not sure of the behaviour !

pedrobaeza: Why do you always review after being months here and it's merged? Any way, this is something correct, and I don't see how can you improve this. The only optimization is to check first if the value is the same, but that's not usual to happen several times.

rousseldenis: Don't have antennas like you. Depends on activity. But I'd prefer letting a comment to make author warned than nothing even if merged.

pedrobaeza: Hehe, OK, no particular problem, but for avoiding something to be available in repo if you don't consider it correct, but I still don't see the problem you are pointing.

rousseldenis: It's not recommended to update a recordset like `recordset.property = value' in a write() or create() as you are not sure what could be re-triggered. You also are not sure that changes will be commited as all the changes are done in the super().

Better changing dict passed to create or write. Or to use compute or inverse.

pedrobaeza: That's a good general advice, but this time, everything is very constrained to the specific change of some values, and this has to be done this way for covering all cases from write/create in the purchase.order and purchase.order.line. Little optimizations can be done for passing the `sequence` value in the same dictionary, but they are not possible in other cases and it will complicate unnecessary the code.

pedrobaeza: So the summary is:

1. There's no possible problem with infinite loops.
2. Maybe in some cases you can optimize an SQL call.

My opinion is that 2 doesn't worth right now, but if you think it is, please propose a way to avoid that calls.

PR #613, opened 2018-11-26
[12.0][MIG] purchase_order_line_sequence

bjeficent:
Migration of module purchase_order_line_sequence to version 12


pedrobaeza:
Isn't this a native feature now?

AaronHForgeFlow:
@pedrobaeza I think you are right. However, this module does more things so I think it is also suitable to be ported:
- Propagates the same sequence to the stock move lines
- Shows the number in the PO line
- Shows the number in the report

pedrobaeza:
The sequence is also propagated (not explicitly, but through record generation), and the other 2 features are noise IMO, but then please change readme properly.

AaronHForgeFlow:
Yes better to update the readme. But the sequence is not propagated in standard, or at least, the sorting of the move lines is not the same as the sorting of the purchase order lines. I just tested on runbot: http://404354-12-0-96506d.runbot15.odoo.com/web?#id=37&action=747&model=purchase.order&view_type=form&menu_id=494

purchase_order_line_sequence/models/purchase.py
```diff
@@ -0,0 +1,88 @@
...
+                                   "displaying the purchase order.",
+                              default=9999)
+
+    sequence2 = fields.Integer('Sequence',
+                               help="Displays the sequence of the line in "
+                                    "the purchase order.",
+                               related='sequence', readonly=True)
+
+    @api.multi
+    def _create_stock_moves(self, picking):
```
pedrobaeza: Why not using `_prepare_stock_moves` method?

bjeficent: i didn't contemplate that solution. But you are right, this way is better

HviorForgeFlow:
@mreficent travis is :red_circle:

AaronHForgeFlow (approved):
Works like a charm :+1:

AaronHForgeFlow (approved):
:+1:

AaronHForgeFlow:
@hveficent mr trvis is green now. Thanks @mreficent

AdriaGForgeFlow (approved):
Could you use this PR to update also Copyright from Eficent to Forgeflow?

Tested on runbot and seems to work as expected :+1:

AaronHForgeFlow:
@pedrobaeza even if you will not use the module is it ok to you to merge it?

tonygalmiche:
Beginner's question:
Why this module has the label 'merged' but it does not appear here:
https://github.com/OCA/purchase-workflow/issues/591
Thank you in advance

HviorForgeFlow:
It is not automated, added btw

PR #614, opened 2018-11-30
[11.0][ADD] purchase_blanket_order

AdriaGForgeFlow:
This module adds purchase blanket orders in v11.0


pedrobaeza:
But isn't this already present on core with "Purchase Agreements" with the option "Blanket order"?

AdriaGForgeFlow:
It is indeed but in my opinion it is mixed with a quite different element such as Purchase Tender and it has limited capabilities. For example, the existing module allows you to order bigger quantities than specified in the blanket order which, in my opinion, shouldn't allow.
Our proposal goes hand to hand with the already existing module sale_blanket_order with the intention to enhance the purchase blanket order capabilities and to make the interface a little bit more intuitive.

pedrobaeza:
:-1: for me for this approach. Extend current features instead of creating a totally replacement.

JordiBForgeFlow:
@pedrobaeza The purchase_requisition module is a mess. It neither provides a good blanket order management process, nor provides a good tendering process.

I am really weary of making complicated extensions to a module that is poorly designed.

pedrobaeza:
Diverging from the main concept is always a problem, and I can tell you about that in several points of the past. We can discuss what are the problems in our CS if you want for getting to a proper extension system, but trust me, reinventing the wheel is never the best option.

purchase_blanket_order/models/blanket_orders.py
```diff
@@ -0,0 +1,359 @@
...
+        ('draft', 'Draft'),
+        ('open', 'Open'),
+        ('expired', 'Expired'),
+    ], compute='_compute_state', store=True, copy=False)
+    validity_date = fields.Date(
+        readonly=True,
+        states={'draft': [('readonly', False)]})
+    date_order = fields.Datetime(
+        readonly=True,
+        required=True,
```
Nikul-Chaudhary, line 81: readonly=True, & required=True, ??

purchase_blanket_order/models/blanket_orders.py
```diff
@@ -0,0 +1,359 @@
...
+        'res.users', string='Responsible', readonly=True,
+        states={'draft': [('readonly', False)]})
+    company_id = fields.Many2one(
+        'res.company', string='Company', default=_default_company,
+        readonly=True,
+        states={'draft': [('readonly', False)]})
+    purchase_count = fields.Integer(compute='_compute_purchase_count')
+
+    # Fields use to filter in tree view
+    original_qty = fields.Float(
```
Nikul-Chaudhary: @ageficent no need to default, Float alwayse default 0.0, 
Change everywhare

https://odoo-new-api-guide-line.readthedocs.io/en/latest/fields.html#float

Nikul-Chaudhary (changes requested):
Improved Code

purchase_blanket_order/models/blanket_orders.py
AdriaGForgeFlow: What do you mean? @Nikul-Chaudhary

JordiBForgeFlow:
@ageficent as discussed, can you add an option to view the lines? Like it's done in purchase requests, where you can navigate to the lines, and from the lines add an action to create a purchase order.

@pedrobaeza then we'll continue the work in our own repo. I spent months with purchase_requests and I know well enough what I'm doing.

purchase_blanket_order/models/blanket_orders.py
Nikul-Chaudhary: It's Two are needed?? because if readony field, How can put date here, and it's required also..

AdriaGForgeFlow: Because in draft mode the field is not readonly, the user must introduce the date order when in draft mode and then the field cannot be modified.

legalsylvain:
> Diverging from the main concept is always a problem,[...] reinventing the wheel is never the best option.

Well in fact it is sometimes the best option. AFAIK, the asset addons of the OCA does'nt depend on the odoo core features, and the reason was that it was a mess to use the base module.

> then we'll continue the work in our own repo.

I have no point of view about the both approach, since I do not use any. So I abstain about the code. But maybe it could be great to have more thoughts from OCA members that are using core feature or your module.

purchase_blanket_order/models/blanket_orders.py
```diff
@@ -0,0 +1,359 @@
...
+    payment_term_id = fields.Many2one(
+        'account.payment.term', string='Payment Terms', readonly=True,
+        states={'draft': [('readonly', False)]})
+    confirmed = fields.Boolean()
+    state = fields.Selection(selection=[
+        ('draft', 'Draft'),
+        ('open', 'Open'),
+        ('expired', 'Expired'),
+    ], compute='_compute_state', store=True, copy=False)
+    validity_date = fields.Date(
```
HviorForgeFlow, line 73: A help description would be nice

HviorForgeFlow (changes requested):
There could be a button to turn Blanket Order to draft, because when you confirm a Blanket Order there is no way back to edit any field. Blanket orders are used when there is a recurring need for expendable goods, so, if order items are similar a good way to fast create a blanket order could be to duplicate an existing one, move to draft and edit there.

purchase_blanket_order/models/blanket_orders.py
Nikul-Chaudhary: @ageficent Ok, Here Domain, I don't see :smile:

HviorForgeFlow (changes requested):
Functional review:
- When you duplicate a Blanket Order you keep its status, it should come back to draft.

I think solving this will fix some minor issues about Draft naming, no confirmation option on duplicates...

purchase_blanket_order/models/purchase_order.py
```diff
@@ -0,0 +1,84 @@
...
+
+    blanket_order_id = fields.Many2one(
+        'purchase.blanket.order', string='Origin blanket order',
+        related='order_line.blanket_order_line.order_id',
+        readonly=True)
+
+    @api.constrains('partner_id')
+    def check_partner_id(self):
+        if self.partner_id:
+            if self.order_line:
```
nikul-serpentcs: @ageficent no need to `if self.order_line:`

purchase_blanket_order/models/purchase_order.py
```diff
@@ -0,0 +1,84 @@
...
+    _inherit = 'purchase.order'
+
+    blanket_order_id = fields.Many2one(
+        'purchase.blanket.order', string='Origin blanket order',
+        related='order_line.blanket_order_line.order_id',
+        readonly=True)
+
+    @api.constrains('partner_id')
+    def check_partner_id(self):
+        if self.partner_id:
```
nikul-serpentcs: @ageficent No need to check `if self.partner_id:` because In Odoo base `partner_id` is required

nikul-serpentcs (commented):
Improve Code

purchase_blanket_order/models/blanket_orders.py
```diff
@@ -0,0 +1,521 @@
...
+        states={'draft': [('readonly', False)]})
+    line_ids = fields.One2many(
+        'purchase.blanket.order.line', 'order_id', string='Order lines',
+        track_visibility='always', copy=True)
+    line_count = fields.Integer(
+        string='Purchase Blanket Order Line count',
+        compute='_compute_line_count',
+        readonly=True
+    )
+    product_id = fields.Many2one('product.product',
```
HviorForgeFlow: I miss ```domain=[('purchase_ok', '=', True)]``` here

purchase_blanket_order/models/blanket_orders.py
```diff
@@ -0,0 +1,521 @@
...
+        ('expired', 'Expired'),
+    ], compute='_compute_state', store=True, copy=False,
+        track_visibility='always',)
+    validity_date = fields.Date(
+        readonly=True,
+        states={'draft': [('readonly', False)]},
+        track_visibility='always',
+        help="Date until which the blanket order will be valid, after this "
+             "date the blanket order will be marked as expired")
+    date_order = fields.Datetime(
```
HviorForgeFlow: Add some help description. What does this variable represents?

HviorForgeFlow (changes requested):

cristinamartinrod (approved):
Tested on runbot 👍🏻

LoisRForgeFlow (approved):
Tested on runbot, works like a charm :+1:

purchase_blanket_order/models/purchase_order.py
```diff
@@ -0,0 +1,156 @@
...
+            date_planned = fields.Date.to_string(
+                fields.Date.from_string(self.date_planned))
+            if line.blanket_order_line and \
+                    line.blanket_order_line.date_schedule and \
+                    line.blanket_order_line.date_schedule != date_planned:
+                    raise ValidationError(_(
+                        'Schedule dates defined on the Purchase Order Line '
+                        'and on the Blanket Order Line do not match.'))
+
+    @api.constrains('order_id.currency_id')
```
HviorForgeFlow: This should be:
```suggestion
    @api.constrains('currency_id')
```
because if not a warning is displayed:

```
WARNING:odoo.models:method purchase.order.line.check_currency: @constrains parameter 'order_id.currency_id' is not a field name
```

HviorForgeFlow (changes requested):

HviorForgeFlow (approved):

PR #616, opened 2018-12-03
[FIX] Update fr.po

quentinDupont:
Fix Typo


pedrobaeza:
Please use Weblate instead: https://odoo-community.org/page/translate

PR #619, opened 2018-12-07
[9.0] [REF] Purchase Request to RFQ: separate PO line unlink message sendin…

ThomasBinsfeld:
Separate PO line unlink message sending from unlink method.

Sending the messages in the unlink method makes it impossible to override/cancel.

### DIFF
--- a/purchase_request_to_rfq/models/purchase_order.py
+++ b/purchase_request_to_rfq/models/purchase_order.py
@@ -193,19 +193,14 @@ class PurchaseOrderLine(models.Model):
 
     @api.multi
     def _po_line_unlink_purchase_request_message_content(self):
-        self.ensure_one()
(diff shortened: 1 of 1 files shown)


purchase_request_to_rfq/models/purchase_order.py
```diff
@@ -193,19 +193,14 @@ def _po_line_purchase_request_cancel_message_content(self):
 
     @api.multi
     def _po_line_unlink_purchase_request_message_content(self):
```
RobinetDenisAcsone, line 195: The method should be renamed (`_po_line_unlink_purchase_request_message`?) or perhaps split the method in 2: keep the previous `_po_line_unlink_purchase_request_message_content` if someone want to overwrite only the message (also keep backward compatibility) and add another which do the `request.message_post`

PR #622, opened 2018-12-18
8.0 mig purchase order line sequence

mathiasfrancke:
Hello,
I needed a module from Odoo 10 in Odoo 8. Now i would like to share the back port with the OCA.

Best Regards
Mathias

**Depends**
- [ ] [8.0 mig stock picking line sequence](https://github.com/OCA/stock-logistics-workflow#485)


purchase_order_line_sequence/README.rst
```diff
@@ -0,0 +1,61 @@
...
+
+Usage
+=====
+
+-This module depends on the following module(s) :
+- * stock_picking_line_sequence (https://github.com/OCA/stock-logistics-workflow/)
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/142/10.0
```
bodedra: s/10.0/8.0

purchase_order_line_sequence/__openerp__.py
```diff
@@ -0,0 +1,31 @@
...
+    'name': 'Purchase order lines with sequence number',
+    'summary': 'Adds sequence to PO lines and propagates it to'
+               'Invoice lines and Stock Moves',
+    'version': '8.0.1.0.0',
+    'category': 'Purchase Management',
+    'author': "Camptocamp, "
+              "Eficent, "
+              "Serpent CS, "
+              "Odoo Community Association (OCA)",
+    'website': 'http://www.camptocamp.com',
```
bodedra: Change website with `https://github.com/OCA/purchase-workflow`

purchase_order_line_sequence/models/purchase.py
```diff
@@ -0,0 +1,87 @@
...
+    sequence = fields.Integer(help="Gives the sequence of the line when "
+                                   "displaying the purchase order.",
+                              default=9999)
+
+    sequence2 = fields.Integer(help="Displays the sequence of the line in "
+                                    "the purchase order.",
+                               related='sequence', readonly=True)
+
+    @api.multi
+    def _create_stock_moves(self, picking):
```
bodedra: In version 8.0, these method available at `purchase.order` object with more arguments. Here is reference https://github.com/odoo/odoo/blob/8.0/addons/purchase/purchase.py#L840

Possible to add these method in test case?

bodedra (changes requested):

bodedra:
Additionally, These module is depend on `stock_picking_line_sequence` so first we have to back port it.

screenshot from 2018-12-18 22-39-02

bodedra:
@mathiasfrancke I found your PR for that module at https://github.com/OCA/stock-logistics-workflow/pull/485

So can you update your PR description like

Depends:
-------------

- [ ] https://github.com/OCA/stock-logistics-workflow/pull/485

purchase_order_line_sequence/README.rst
mathiasfrancke: corrected. comes with next push.

purchase_order_line_sequence/__openerp__.py
mathiasfrancke: corrected. comes with next push.

purchase_order_line_sequence/README.rst
nikul-serpentcs: @mathiasfrancke @bodedra Use new README by fragment system, copying https://github.com/OCA/maintainer-tools/tree/master/template/module/readme and editing or removing the needed sections. You can then run locally oca-gen-addon-readme if you have OCA/maintainer-tools installed and you want to pregenerate the README.rst file (preferred), or keep previous README.rst file.

purchase_order_line_sequence/models/purchase.py
mathiasfrancke: Hello @bodedra ,
thanks for your note. I don't understand what you mentioned here. 
This is a method on the model `purchase.order.line`.
The method you linked is on the model `purchase.order`.
So both have the same name and different signatures but also the are on different objects.

bodedra: @mathiasfrancke In v8.0, we don't have same method in `purchase.order.line` compare to v10.0. But we have same method in `purchase.order` object level. 

Here is method references:

v8.0 - purchase.order - https://github.com/odoo/odoo/blob/8.0/addons/purchase/purchase.py#L840

v10.0 - purchase.order.line - https://github.com/odoo/odoo/blob/10.0/addons/purchase/models/purchase.py#L719

So we have make compatible code to work same in v8.0 & v.10.0

mathiasfrancke: Migrated missing method.

mathiasfrancke:
Since there is a problem merging https://github.com/OCA/stock-logistics-workflow/pull/485 this PR is on hold from my side. It is requirement for this PR.

mathiasfrancke:
I will close this PR because it is depenend on [stock_picking_line_sequence](https://github.com/OCA/stock-logistics-workflow/pull/485) where the PR is also closed because of duplicate with _stock_pciking_reorder_lines_ module

PR #624, opened 2018-12-28
[purchase_request] fix in procurement

JordiBForgeFlow:
Without this fix when you create a procurement from an orderpoint, that creates a purchase request, it will raise the error: 
`  File "/parts/purchase-workflow11/purchase_request/models/procurement_rule.py", line 56, in _make_pr_get_domain
    (gpo == 'propagate' and values['group_id'].id) or False
AttributeError: 'bool' object has no attribute 'id'`


LoisRForgeFlow (approved):

bodedra (approved):

JordiBForgeFlow:
Thanks for your reviews! Now merging.

PR #632, opened 2019-01-17
12.0 mig purchase landed cost

rvalyi:
standard v12 migration:

- product.uom -> uom.uom
- label with for
- now depends on new purchase_stock module to ensure stock.move has the required purchase_line_id field


pedrobaeza:
Please fix little runbot warning:
```
WARNING openerp_test odoo.models: The model import.landed.cost.pickings.wizard has no _description

```

purchase_landed_cost/wizard/import_invoice_line.py
```diff
@@ -0,0 +1,34 @@
...
+
+    @api.multi
+    def action_import_invoice_line(self):
+        self.ensure_one()
+        self.env['purchase.cost.distribution.expense'].create({
+            'distribution': self.env.context['active_id'],
+            'invoice_line': self.invoice_line.id,
+            'invoice_id': self.invoice_line.invoice_id.id,
+            'ref': self.invoice_line.name,
+            'expense_amount': self.invoice_line.price_subtotal,
```
novawish, line 32: Can we do a conversion (`compute`) form foreign currency for invoices that have different currency from company's currency (in multi-currency)?

purchase_landed_cost/models/purchase_cost_distribution.py
```diff
@@ -0,0 +1,575 @@
...
+
+    @api.onchange('type')
+    def onchange_type(self):
+        if self.type and self.type.default_amount:
+            self.expense_amount = self.type.default_amount
+
+    @api.onchange('invoice_line')
+    def onchange_invoice_line(self):
+        self.invoice_id = self.invoice_line.invoice_id.id
+        self.expense_amount = self.invoice_line.price_subtotal
```
novawish, line 570: Can we do a conversion (`compute`) form foreign currency for invoices that have different currency from company's currency (in multi-currency)?

novawish (changes requested):
We are facing an issue with multi-currency configuration: The Vendor Bills may have different currencies compared with company's currency. I added a suggestion, can you tell me what you think?

purchase_landed_cost/wizard/import_invoice_line.py
novawish: After thinking again, i think we can use field `price_subtotal_signed` instead

purchase_landed_cost/models/purchase_cost_distribution.py
novawish: After thinking again, i think we can use field `price_subtotal_signed` instead

purchase_landed_cost/wizard/import_landed_cost_pickings_wizard.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+class ImportLandedCostPickingsWizard(models.TransientModel):
+    _name = 'import.landed.cost.pickings.wizard'
+
+    possible_picking_ids = fields.Many2many(
+        comodel_name="stock.picking", string="Possible pickings")
+    picking_ids = fields.Many2many(
+        comodel_name="stock.picking", string="Pickings",
+        domain="[('id', 'in', possible_picking_ids[0][2])]")
```
novawish, line 14: I tested and apparently this does not work anymore.
I think we don't need `possible_picking_ids` and we can write a function that return directly the domain like this:
image

pedrobaeza:
Superseeded by #706

PR #642, opened 2019-01-28
[11.0][add][purchase_force_invoiced]

JordiBForgeFlow:
This module adds the possibility for users to force the invoice status of the
purchase orders to 'No Bill to Receive', even when not all the
quantities ordered or delivered have been invoiced.

This feature useful in the following scenario:

* The supplier disputes the quantities to be billed for, after the
  products have been delivered to her/him, and you agree to reduce the
  quantity to invoice (without expecting a refund).

* When migrating from a previous Odoo version, in some cases there is less
  quantity billed to what was delivered, and you don't want these old purchase
  orders to appear in your 'Waiting Bills' list

Usage
=====

#. Create a purchase order and confirm it.
#. Receive the products/services.
#. Create an vendor bill and reduce the invoiced quantity. The purchase order
   invoicing status is 'Waiting Bills'.
#. Lock the Purchase Order, to change the status of it to 'Done'.
#. Check the field 'Force Invoiced'. Now the quantities of the PO

This module is analogous to https://github.com/OCA/sale-workflow/blob/11.0/sale_force_invoiced/README.rst

The PR introduces a change to the module 'purchase_open_qty' to consider quantity to bill = 0 when the purchase has invoice status = 'invoiced'. This will bring compatibility with the module purchase_force_invoiced without requiring and explicit dependency.


purchase_open_qty/__manifest__.py
```diff
@@ -6,7 +6,7 @@
     "name": "Purchase Open Qty",
     "summary": "Allows to identify the purchase orders that have quantities "
                "pending to invoice or to receive.",
-    "version": "11.0.1.1.0",
+    "version": "11.0.1.1.1",
```
elicoidal, line 9: open a second PR

purchase_force_invoiced/view/purchase_view.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
```
elicoidal, line 5: copyright in XML files

purchase_force_invoiced/readme/USAGE.rst
```diff
@@ -0,0 +1,6 @@
+#. Create a purchase order and confirm it.
+#. Receive the products/services.
+#. Create an vendor bill and reduce the invoiced quantity. The purchase order
```
elicoidal:
```suggestion
#. Create a vendor bill and reduce the invoiced quantity. The purchase order
```

purchase_force_invoiced/readme/USAGE.rst
```diff
@@ -0,0 +1,6 @@
...
+   invoicing status is 'Waiting Bills'.
+#. Lock the Purchase Order, to change the status of it to 'Done'.
```
elicoidal:
```suggestion
#. Lock the Purchase Order and change its status to 'Done'.
```

purchase_force_invoiced/readme/USAGE.rst
```diff
@@ -0,0 +1,6 @@
...
+#. Check the field 'Force Invoiced'. Now the quantities of the PO
```
elicoidal: the sentence is incomplete

purchase_force_invoiced/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,13 @@
+This module adds the possibility for users to force the invoice status of the
+purchase orders to 'No Bill to Receive', even when not all the
+quantities ordered or delivered have been invoiced.
+
+This feature useful in the following scenario:
```
elicoidal:
```suggestion
This feature is useful in the following scenario:
```

purchase_force_invoiced/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,13 @@
+This module adds the possibility for users to force the invoice status of the
+purchase orders to 'No Bill to Receive', even when not all the
+quantities ordered or delivered have been invoiced.
```
elicoidal:
```suggestion
quantities, ordered or delivered, have been invoiced.
```

purchase_force_invoiced/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2019 Eficent Business and IT Consulting Services S.L.
+# Copyright 2019 Aleph Objects, Inc.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+{
+    'name': 'Purchase Force Invoiced',
+    'summary': 'Allows to force the billing status of the purchase order to '
+               'Invoiced',
```
elicoidal:
```suggestion
               '"Invoiced"',
```

elicoidal (approved):
LGTM

pedrobaeza:
The same as said in https://github.com/OCA/account-invoicing/pull/497#issuecomment-458327454

elicoidal:
@pedrobaeza I understand your point but tend to understand @jbeficent one's too.
There are many cases where despite the figures, you get to negotiate the VB and will never receive it.

What do you propose in these cases?

JordiBForgeFlow:
@pedrobaeza I understand your concern in https://github.com/OCA/account-invoicing/pull/497#issuecomment-458327454

We need a method to manage under/over tolerances. See https://wiki.scn.sap.com/wiki/display/ERPSCM/Under+and+Overdelivery+Tolerances+in+Purchasing

Because it is very usual for a company to agree with a vendor that a +-5% difference in quantity received vs qty invoiced is acceptable.

At this moment, when you order 100 and receive 101, the PO will be always set to status "Awaiting Bills", because you have received a bit more. So the PO is always in the list of 'To Bill'. And Purchasing knows that they are not expecting to receive a bill for this extra unit!

I am happy to refactor the module to something like 'purchase_delivery_tolerance'.

But in the end we need to say: This PO has no bills awaiting, even we have received more/less than planned.

JordiBForgeFlow:
See another approach here https://docs.oracle.com/cd/E41948_01/fscm92pbh1/eng/fscm/spog/concept_UnderstandingtheClosePurchaseOrdersProcess-9f3dcb.html

It is obvious that under/over tolerances are common.

pedrobaeza:
For me it's totally unacceptable that someone charges more than I have received. But any way, if it's the case, you simply have to do an extra incoming picking, and move to scrap the remaining quantity. And the reverse apply the same with a return.

I know that this seems more work to do at the beginning, but if not done that way, you will have inconsistent information in one moment or other (and more following a migration path). Look how Odoo unlink the lock condition to the invoice status for this reason.

I'm still totally reluctant to this method.

JordiBForgeFlow:
need your review @lreficent @ageficent

AdriaGForgeFlow (approved):
Functional review :+1:

PR #643, opened 2019-01-29
[12.0] MIG purchase_commercial_partner

alexis-via:


pedrobaeza:
Please squash together adjacent "OCA Transbot.." commits using https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests

purchase_commercial_partner/models/purchase.py
```diff
@@ -0,0 +1,13 @@
...
+
+from odoo import fields, models
+
+
+class PurchaseOrder(models.Model):
+    _inherit = 'purchase.order'
+
+    commercial_partner_id = fields.Many2one(
+        'res.partner', related='partner_id.commercial_partner_id', store=True,
+        index=True, string='Commercial Vendor', readonly=True)
```
nikul-serpentcs: @alexis-via no need to `readonly`

bealdav: I think yes, related fields with full access leads to weird and unwanted behavior, bad performance: crazy sql.

In master, related fields are now readonly by default, isn't it ?

pedrobaeza: In v12, related are already `readonly` by default.

bealdav: you're right ;-)
https://github.com/odoo/odoo/blob/12.0/odoo/fields.py#L422

pedrobaeza: We included in migration guide the reverse: if you rely on default `False` value (for example, for res.config.settings fields), you have to explicitly set the contrary: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0#tasks-to-do-in-the-migration

That's why the Python mantra "Explicit better than implicit" works. For this case, it's harmless to leave it explicit.

HviorForgeFlow (approved):
Code review LGTM

bealdav (approved):

PR #646, opened 2019-02-05
[11.0][IMP] purchase_request. Improve searching

AaronHForgeFlow:
Sometimes it is difficult to see if there are existing draft purchase request for specific product because:
- We cannot search by purchase requests by product
- Purchase request lines are filtered by state

This PR makes the changes to search purchase request by product and to see draft purchase request lines. There is already a constraint that prevents to create quotations from draft purchase request lines so this will not cause any break in the flow.

### DIFF
--- a/purchase_request/models/purchase_request.py
+++ b/purchase_request/models/purchase_request.py
@@ -86,6 +86,9 @@ class PurchaseRequest(models.Model):
                                readonly=False,
                                copy=True,
                                track_visibility='onchange')
+    product_id = fields.Many2one('product.product',
+                                 related='line_ids.product_id',
+                                 string='Product')
     state = fields.Selection(selection=_STATES,
                              string='Status',
                              index=True,
--- a/purchase_request/__manifest__.py
+++ b/purchase_request/__manifest__.py
@@ -5,7 +5,7 @@
     "name": "Purchase Request",
     "author": "Eficent, "
               "Odoo Community Association (OCA)",
-    "version": "11.0.1.0.0",
+    "version": "11.0.1.1.0",
     "summary": "Use this module to have notification of requirements of "
                "materials and/or external services and keep track of such "
                "requirements.",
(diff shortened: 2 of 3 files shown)


JordiBForgeFlow (approved):
:+1: code review and tested in runbot.

HviorForgeFlow (approved):
LGTM Functional and code review :+1:

AdriaGForgeFlow (approved):
LGTM :+1: Functional test on runbot and code review

MiquelRForgeFlow (approved):
LGTM

purchase_request/models/purchase_request.py
```diff
@@ -86,6 +86,9 @@ def _compute_is_editable(self):
                                readonly=False,
                                copy=True,
                                track_visibility='onchange')
+    product_id = fields.Many2one('product.product',
+                                 related='line_ids.product_id',
+                                 string='Product')
```
rousseldenis, line 91: @aheficent Careful here. When you will write on purchase request, you will generate a write also on product.
`readonly=True`

AaronHForgeFlow: Is it not readonly by deafult? Odoo does the same in purchases: https://github.com/odoo/odoo/blob/11.0/addons/purchase/models/purchase.py#L155

rousseldenis: Introduced in v12. Odoo has also wrong code :smile:

rousseldenis: v11 : https://github.com/odoo/odoo/blob/11.0/odoo/fields.py#L413

v12 : https://github.com/odoo/odoo/blob/12.0/odoo/fields.py#L422

AaronHForgeFlow: Fix proposed: https://github.com/OCA/purchase-workflow/pull/648

PR #649, opened 2019-02-06
[11.0][FIX] purchase_request

AdriaGForgeFlow:
When creating a RFQ from Purchase Request Lines, in the wizard, the quantity to purchase is not transfered to the RFQ and the quantity is set to the request line quantity. This PR fixes that issue.

### DIFF
--- a/purchase_request/models/purchase_request.py
+++ b/purchase_request/models/purchase_request.py
@@ -428,9 +428,9 @@ class PurchaseRequestLine(models.Model):
 
         rl_qty = 0.0
         # Recompute quantity by adding existing running procurements.
-        for rl in po_line.purchase_request_lines:
-            rl_qty += rl.product_uom_id._compute_quantity(
-                rl.product_qty, purchase_uom)
+        for rl in po_line:
+            rl_qty += rl.purchase_request_lines.product_uom_id\
+                ._compute_quantity(rl.product_qty, purchase_uom)
         qty = max(rl_qty, supplierinfo_min_qty)
         return qty
 
--- a/purchase_request/models/purchase_request.py
+++ b/purchase_request/models/purchase_request.py
@@ -429,8 +429,9 @@ class PurchaseRequestLine(models.Model):
         rl_qty = 0.0
         # Recompute quantity by adding existing running procurements.
         for rl in po_line:
-            rl_qty += rl.purchase_request_lines.product_uom_id\
-                ._compute_quantity(rl.product_qty, purchase_uom)
+            for purchase_request_line in rl.purchase_request_lines:
+                rl_qty += purchase_request_line.product_uom_id\
+                    ._compute_quantity(rl.product_qty, purchase_uom)
         qty = max(rl_qty, supplierinfo_min_qty)
         return qty
 


HviorForgeFlow:
@ageficent in that wizard you should be able to modify quantity as has already been approved. Fix should be to use xml readonly tag on quantity field.

AdriaGForgeFlow:
> @ageficent in that wizard you should be able to modify quantity as has already been approved. Fix should be to use xml readonly tag on quantity field.

Thanks. I was not sure if the user was allowed to modify the quanities or not from the wizard. Updated the code. :+1:

HviorForgeFlow (changes requested):
I would also add readonly here.

AaronHForgeFlow (approved):
:+1: LGTM 

I understand that if the manager wants to buy more/less quantity the purchase request must pass again the approval process.

purchase_request/wizard/purchase_request_line_make_purchase_order_view.xml
```diff
@@ -33,7 +33,7 @@
                               <field name="request_id" readonly="1"/>
                               <field name="product_id" readonly="1"/>
                               <field name="name"/>
-                              <field name="product_qty"/>
+                              <field name="product_qty" readonly="1"/>
```
JordiBForgeFlow: @hveficent that is not correct. The person that initiates the purchase could decide to create two partial purchase orders, each with a separate quantity. The purchaser should be free to decide in the wizard.

JordiBForgeFlow: Considering my comment, then this PR is not required, I think

AaronHForgeFlow: @jbeficent It is needed then. I tested with @ageficent and if the user changes the quantity in the wizard the quotation will reflect the quantity in the purchase request line not the one in the wizard. It ignores the quantity in the wizard.

JordiBForgeFlow: But seems that @ageficent reversed the change, because the change now is only going to add readonly=1

AaronHForgeFlow: Exactly. Commit e221bfc3db749406e327a69d34ad5d2b6ea5f2ce does the fix. It is the first one in this PR.

HviorForgeFlow: You are right @jbeficent, but it's still a way to limit quantity has been approved to purchase. For example, you can buy 30 when your request was on 20. I think there is need to perform some limitation like stock_quant_manual_assign module does.

HviorForgeFlow (approved):

HviorForgeFlow (changes requested):
Travis red tests failing

AdriaGForgeFlow:
[WIP] Pending to backport allocations from v12 purchase_request

JordiBForgeFlow:
Perhaps we should not backport allocations.

HviorForgeFlow:
Sure @jbeficent, it needs to be fixed on V12 instead

HviorForgeFlow:
@ageficent can you check #791 in order to see if this is fixed there? Thanks!

HviorForgeFlow:
@AdriaGForgeFlow is this required anymore?

PR #650, opened 2019-02-12
[11.0][MIG] purchase_order_type: Migration to 11.0

oihane:


purchase_order_type/__init__.py
```diff
@@ -0,0 +1,3 @@
+# -*- coding: utf-8 -*-
```
nikul-serpentcs: @oihane no need to here

purchase_order_type/__manifest__.py
```diff
@@ -0,0 +1,40 @@
+# -*- coding: utf-8 -*-
```
nikul-serpentcs: @oihane Copyright should be 

```
# Copyright (C) 2015 Camptocamp SA
```

purchase_order_type/models/purchase_order.py
```diff
@@ -0,0 +1,64 @@
+# -*- coding: utf-8 -*-
```
nikul-serpentcs: same here or Other files

nikul-serpentcs (changes requested):
Improve Code

purchase_order_type/__init__.py
oihane: Done

purchase_order_type/__manifest__.py
oihane: I didn't have any idea if I could change it.

purchase_order_type/models/purchase_order.py
oihane: done, in every py file

anajuaristi (approved):

anajuaristi (approved):

tbaden:
@nikul-serpentcs your changes seem resolved. anything else or good to go?
@oihane please squash your commits into one migration commit

nikul-serpentcs (approved):
Only code Review LGTM :+1:

HviorForgeFlow:
Code Review
Why is defined a relation between Partner an Purchase Type? What is the use case?

oihane:
@hveficent is similar to `sale_order_type`, you can define a default purchase type in a supplier if you select that supplier the purchase order will have the type defined in the supplier instead of the default one.

HviorForgeFlow (approved):

PR #652, opened 2019-02-13
[12.0][MIG] purchase_tier_validation: port to v12

naglis:
IIUC, the tests currently fail because the dependency `base_tier_validation` is not yet available (see OCA/server-ux/pull/52).

### DIFF
--- a/oca_dependencies.txt
+++ b/oca_dependencies.txt
@@ -1,0 +1,1 @@
+server-ux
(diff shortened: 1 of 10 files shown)


LoisRForgeFlow:
This one can be closed as it is already merge in the other PR.

naglis:
Thanks @lreficent :)

PR #655, opened 2019-02-15
[ADD] purchase_invoice_no_zero_lines: avoid zero quantity lines in ac…

jonerikceberio:
…count invoice lines from purchase
To answer https://github.com/OCA/OCB/pull/814
While coding I found nearly the same in account-invoicing/purchase_stock_picking_return_invoicing


pedrobaeza:
Please avoid plurals in module name. You can put `purchase_invoicing_no_zero_line`. Other thing: this should go to OCA/account-invoicing as a better host, the same as the other module you mention.

jonerikceberio:
Hm, funny, I was just complaining to myself about the hr_holidayS and still failed to see the plural in mine. Ok, let's go to account-invoicing.

pedrobaeza:
Hehe, we can't do anything about that one as it's core, but we can do it on ours! Check guidelines about this: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#modules

PR #660, opened 2019-02-26
[11.0][ADD] purchase_order_recommendation: New Module

chienandalu:
This module adds a recommended products wizard to current purchase order.

 It is based on delivered products to customer locations in a given range of
dates, and allows the purchase manager to quickly know the most sold products
for the current supplier, which results in an easy to use hint to improve
the purchase workflow.

cc @Tecnativa


purchase_order_product_recommendation/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2019 David Vidal <david.vidal@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "Purchase Order Product Recommendation",
+    "summary": "Recommend products to buy to supplier based on history",
+    "version": "11.0.1.0.1",
```
nikul-serpentcs: @chienandalu `11.0.1.0.0`

purchase_order_product_recommendation/wizards/purchase_order_recommendation.py
```diff
@@ -0,0 +1,290 @@
...
+class PurchaseOrderRecommendation(models.TransientModel):
+    _name = 'purchase.order.recommendation'
+    _description = 'Recommended products for current purchase order'
+
+    order_id = fields.Many2one(
+        'purchase.order',
+        'Purchase Order',
+        default=lambda self: self._default_order_id(),
+        required=True,
+        readonly=True,
```
nikul-serpentcs, line 19: @chienandalu ??
```
        required=True,
        readonly=True
```

purchase_order_product_recommendation/wizards/purchase_order_recommendation.py
```diff
@@ -0,0 +1,290 @@
...
+    )
+    units_virtual_available = fields.Float(
+        readonly=True,
+    )
+    units_included = fields.Float()
+    wizard_id = fields.Many2one(
+        'purchase.order.recommendation',
+        'Wizard',
+        ondelete='cascade',
+        required=True,
```
nikul-serpentcs, line 242: same here

nikul-serpentcs (changes requested):
Improved Code

purchase_order_product_recommendation/wizards/purchase_order_recommendation.py
chienandalu: Its an OCA convention (https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#pep8-options):

> When making a comma separated list, dict, tuple, ... with one element per line, append a comma to the last element. This makes it so the next element added only changes one line in the changeset instead of changing the last element to simply add a comma.

purchase_order_product_recommendation/wizards/purchase_order_recommendation.py
chienandalu: same as https://github.com/OCA/purchase-workflow/pull/660#discussion_r260253622

rousseldenis:
@chienandalu Are you aware of this module ? https://github.com/OCA/stock-logistics-warehouse/tree/11.0/stock_demand_estimate

Could this one maybe help your process ?

purchase_order_product_recommendation/wizards/purchase_order_recommendation.py
nikul-serpentcs: @chienandalu Wrong Topic

My question is `required & readonly` are True then how user put a value, in fields not any domain right now?

chienandalu: The wizard is invoked from the order itself and that's where the id comes from. When a user calls the wizard from a PO is setting the value already.

chienandalu:
@rousseldenis Yes I know. We discarded it as we just wanted  a simple assistant based on a very specific criteria (in the way `sale_order_product_recommendation` does)

purchase_order_product_recommendation/wizards/purchase_order_recommendation.py
nikul-serpentcs: @chienandalu Sounds Good

purchase_order_product_recommendation/wizards/purchase_order_recommendation.py
nikul-serpentcs: Also, same as here ??

chienandalu: Sure, a wizard line without it's parent model would be useless :wink:

purchase_order_product_recommendation/wizards/purchase_order_recommendation.py
```diff
@@ -0,0 +1,297 @@
...
+from odoo.tools import DEFAULT_SERVER_DATETIME_FORMAT
+from datetime import datetime, timedelta
+
+
+class PurchaseOrderRecommendation(models.TransientModel):
+    _name = 'purchase.order.recommendation'
+    _description = 'Recommended products for current purchase order'
+
+    order_id = fields.Many2one(
+        'purchase.order',
```
pedrobaeza: Use keyword arguments (`comodel_name`, `string`...)

purchase_order_product_recommendation/wizards/purchase_order_recommendation.py
```diff
@@ -0,0 +1,297 @@
...
+        help='Final date to compute recommendations.',
+    )
+    line_ids = fields.One2many(
+        'purchase.order.recommendation.line',
+        'wizard_id',
+        'Products',
+    )
+    line_amount = fields.Integer(
+        'Number of recommendations',
+        default=0,
```
pedrobaeza, line 39: Why default=0? It should be a sane value like 10 or 15

purchase_order_product_recommendation/wizards/purchase_order_recommendation.py
```diff
@@ -0,0 +1,297 @@
...
+        required=True,
+        help='Stablish a limit on how many recommendations you want to get.'
+             'Leave it as 0 to set no limit',
+    )
+    last_compute = fields.Char()
+
+    @api.model
+    def _default_order_id(self):
+        return self.env.context.get('active_id', False)
```
pedrobaeza, line 49: Check if the active_model is `purchase.order`

purchase_order_product_recommendation/wizards/purchase_order_recommendation.py
```diff
@@ -0,0 +1,297 @@
...
+
+    @api.multi
+    def _get_total_days(self):
+        """Compute days between the initial and the end date"""
+        day = (datetime.strptime('{} 00:00:00'.format(self.date_end),
+               DEFAULT_SERVER_DATETIME_FORMAT) + timedelta(days=1) -
```
pedrobaeza: Use `fields.Datetime.from_string`

purchase_order_product_recommendation/wizards/purchase_order_recommendation.py
```diff
@@ -0,0 +1,297 @@
...
+               datetime.strptime('{} 00:00:00'.format(self.date_begin),
+               DEFAULT_SERVER_DATETIME_FORMAT)).days
+        return day
+
+    @api.multi
+    def _find_move_line(self, src='internal', dst='customer'):
+        """"Returns a dictionary from the move lines in a range of dates
+            from and to given location types"""
+        supplierinfo_obj = self.env['product.supplierinfo']
+        supplierinfo = supplierinfo_obj.search([
```
pedrobaeza: You should explain this in the README, that only products declared here are searched.

pedrobaeza: And name variables according its content. This should be plural as it contains several records.

purchase_order_product_recommendation/wizards/purchase_order_recommendation.py
```diff
@@ -0,0 +1,297 @@
...
+               DEFAULT_SERVER_DATETIME_FORMAT) + timedelta(days=1) -
+               datetime.strptime('{} 00:00:00'.format(self.date_begin),
+               DEFAULT_SERVER_DATETIME_FORMAT)).days
+        return day
+
+    @api.multi
+    def _find_move_line(self, src='internal', dst='customer'):
+        """"Returns a dictionary from the move lines in a range of dates
+            from and to given location types"""
+        supplierinfo_obj = self.env['product.supplierinfo']
```
pedrobaeza: For performance reasons, I would say here to pass the context for avoiding the prefetch of all fields, and this will be extended to the rest of objects through `mapped`.

purchase_order_product_recommendation/wizards/purchase_order_recommendation.py
```diff
@@ -0,0 +1,297 @@
...
+        supplierinfo = supplierinfo_obj.search([
+            ('name', '=', self.order_id.partner_id.id)])
+        product_tmpl = supplierinfo.mapped('product_tmpl_id')
+        product = supplierinfo.mapped('product_id')
+        product |= self.env['product.product'].search(
```
pedrobaeza: Use `products |= product_tmpls.mapped('product_variant_ids')` directly for achieving the same in a simpler way, as a search with a list of IDs can be less efficient if the list in long.

purchase_order_product_recommendation/wizards/purchase_order_recommendation.py
```diff
@@ -0,0 +1,297 @@
...
+
+    @api.multi
+    @api.onchange('order_id', 'date_begin', 'date_end', 'line_amount')
+    def _generate_recommendations(self):
+        """Generate lines according to received and delivered items"""
+        last_compute = '{}-{}-{}-{}'.format(
+            self.id, self.date_begin, self.date_end, self.line_amount)
+        # Avoid execute onchange as times as fields in api.onchange
+        # ORM must control this?
+        if self.last_compute == last_compute:
```
pedrobaeza: Why is this needed?

pedrobaeza (changes requested):
Some little things

purchase_order_product_recommendation/wizards/purchase_order_recommendation.py
chienandalu: In `sale_order_product_recommendation` is suposed to avoid running the onchange for every field declared, but indeed a made a debug and that's not happening. Can bring some light to this, @Yajo ?

yajo: No idea, but indeed it seems redundant with the onchange decorator. Is it possible that this comes from v7, where we didn't have such decorator?

pedrobaeza: Well, this module was first developed for v9, so I don't think so. Let's remove it.

bjeficent (approved):
Functional test in local :+1:

pedrobaeza (approved):

HviorForgeFlow:
Is there any situation where unit_delivered is different from zero??

chienandalu:
Yes, if there were units received the wizard adds the product to the lines as well.

HviorForgeFlow:
Also for deliver ones?

chienandalu:
Yes :slightly_smiling_face:

PR #663, opened 2019-03-04
[IMP] product_supplier_info_discount: update supplier info if it changes

chienandalu:
- If the supplier already has a supplier info for a product, it will be updated
  if the discount or the price changes.

cc @Tecnativa


pedrobaeza:
After conversations, this feature shouldn't be contained in this module, so closing.

PR #664, opened 2019-03-04
12.0 isabel currency

IsabelEsparza:
@alan196 could you check it please? 

Thank You!


purchase_supplierinfo_currency/models/product.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
+
+    def _prepare_sellers(self, params):
+        res = super()._prepare_sellers(params)
+        if not params:
+            return res
+        if params.get('supplier'):
```
nikul-serpentcs, line 15: @IsabelEsparza Here use `Else`

```
if params.get('supplier'):
    currency_id = params.get('supplier').currency_id.id
else:
    currency_id = params.get('order_id').partner_id.currency_id.id
```

nikul-serpentcs (changes requested):
Improve Code

purchase_supplierinfo_currency/models/product.py
IsabelEsparza: Thank you, 
I change it right now.

alan196 (approved):
👍 Tested functionally and code review

PR #668, opened 2019-03-30
12.0 mig purchase_discount

eLBati:
From https://github.com/OCA/purchase-workflow/pull/608


mrcast (approved):
Functional test OK!

purchase_discount/__init__.py
```diff
@@ -0,0 +1,3 @@
+# -*- coding: utf-8 -*-
```
nikul-serpentcs: @eLBati No need to `-*- coding: utf-8 -*-`

purchase_discount/views/product_supplierinfo_view.xml
```diff
@@ -0,0 +1,23 @@
+<odoo>
+
+    <record model="ir.ui.view" id="product_supplierinfo_form_view">
+         <field name="model">product.supplierinfo</field>
+         <field name="inherit_id" ref="product.product_supplierinfo_form_view"/>
+         <field name="arch" type="xml">
+              <xpath expr="//div[field[@name='price']]" position="after">
```
nikul-serpentcs: IMO here use `<field name="price" position="after">`

nikul-serpentcs (changes requested):
netpicking

pedrobaeza:
We still need https://github.com/odoo/odoo/pull/29983 for completing the module

eLBati:
@nikul-serpentcs thanks, done

eLBati:
@pedrobaeza what happens using `purchase_discount` without that fix?

pedrobaeza:
I'm not sure if stock move price is correct. And with the hack in previous versions, sometimes there are side effects.

pedrobaeza:
@legalsylvain we talked about merging `product_supplierinfo_discount` in this module, isn't it?

eLBati:
@pedrobaeza done with 82536a40eb1064da39f3900f8b4e538c39156bed

pedrobaeza:
Good to know.

Tardo:
please, get Travis happy

purchase_discount/views/product_supplierinfo_view.xml
geluboros: I think this xpath was good as it was before this change, since it considered the price label div.
With this change, the discount label is no longer visible. I think this should be reverted.

eLBati:
In the end, I will not use this module, so I will not work on it. If someone can take care of it, please do. Otherwise the PR can be closed

lmiguens-solvos:
@pedrobaeza I'd finish this PR.

eLBati:
@lmiguens-solvos you can create a new PR starting from my branch `eLBati:12.0-mig-purchase_discount`

lmiguens-solvos:
@eLBati Thanks

pedrobaeza:
Superseeded by #731

PR #674, opened 2019-04-24
[12.0][MIGR] purchase order line deep sort

renatonlima:


pedrobaeza:
I have come to this after reviewing and ordering a merge in #773. Sorry for not seeing this. But I think it's still good to merge the other one as there's a problem with the name of the fields in res.config.settings that can't start with `default_`. I hope you don't mind. For the rest, if you are using this PR, it's compatible with the other one without doing anything except switching.

Closing then.

Issue #551, opened 2018-05-31
[11]Subcontracted Service install error


Moto017:
The operation cannot be completed, probably due to the following:
- deletion: you may be trying to delete a record while other records still reference it
- creation/update: a mandatory field is not correctly set

[object with reference: picking_type_id - picking.type.id]

pedrobaeza:
Please put the detailed steps to reproduce the error.

Moto017:
Using 11 CE. Attempting app install gives the above error.

pedrobaeza:
Which module?

Moto017:
subcontracted_service

pedrobaeza:
@hveficent can you check?

HviorForgeFlow:
Yes, I could reproduce your error. Main problem is on init hook that creates a procurement rule in main company warehouse. If there is no buy stock_picking_type associated to this warehouse, installation raises this error.
I'm not so sure how to solve this problem aiming in standarization. Should I remove init_hook, or just prompt a message if stock_pciking_type is not found?
cc. @jbeficent @lreficent

HviorForgeFlow:
@Moto017 I would like you to explain a little bit better your use case of this module, it would help to improve it, thanks.

Moto017:
@hveficent O course.  
We subcontract portions of manufacturing with a mix of material supplied by us and supplied by subcontractor.  Our best use to handle inventory levels is with BOM's, therefore, there is no direct link to our vendor bills.  Creation of the PO by service will then verify and match production quantities and amounts.  Hope this is helpful!

HviorForgeFlow:
I was asking about routes, if you are using multiple warehouses, among other customization, because I don't understand how you don't provide Odoo inventory with an incoming route in main company.

Moto017:
We are using multi company as well as multiple warehouses.  We do have an incoming route in main company under Buy.  Perhaps it is a multi-company error as I noticed these are not defined in procurements.  Your commit has allowed installation and will check use.

PR #677, opened 2019-04-30
[MIG] subcontracted_service: Migration to 12.0

skukered:


HviorForgeFlow (approved):
Functional + Code review :+1:

subcontracted_service/models/warehouse.py
```diff
@@ -43,11 +43,10 @@ def _get_vals_for_proc_rule_subcontracting(self):
     @api.multi
     def _set_subcontracting_service_proc_rule(self):
         for rec in self:
-            if rec.subcontracting_service_proc_rule_id:
-                continue
-            vals = rec._get_vals_for_proc_rule_subcontracting()
-            rule = self.env['stock.rule'].create(vals)
-            rec.subcontracting_service_proc_rule_id = rule.id
+            if not rec.subcontracting_service_proc_rule_id:
```
MiquelRForgeFlow, line 47: I don't understand this change. Unnecessary diff. Please, revert it.

subcontracted_service/models/warehouse.py
```diff
@@ -22,16 +22,7 @@ def _get_buy_route(self):
     def _get_vals_for_proc_rule_subcontracting(self):
         self.ensure_one()
         picking_type = self.in_type_id
-        if not picking_type:
```
MiquelRForgeFlow: Please, could you explain why there is no need to validate picking_type? 🤔

HviorForgeFlow: It was added due to this issue: https://github.com/OCA/purchase-workflow/issues/551

subcontracted_service/models/warehouse.py
HviorForgeFlow: It does the same without a continue.

HviorForgeFlow (commented):
Are you planning to finish it?

subcontracted_service/models/warehouse.py
MiquelRForgeFlow: Yeah, I know. But it's unnecessary to add this code diff.

skukered:
I'll check it out this week. Sorry

> Are you planning to finish it?

skukered:
> Are you planning to finish it?

I'll check it out this week. Sorry

subcontracted_service/models/warehouse.py
HviorForgeFlow: 4 lines against code diff from README is minimalistic

skukered:
"It does the same without a continue." @hveficent 
with the continue, I must add more code in the test to go through there

LoisRForgeFlow:
@skukered  I think 97.95 codecov is enough. Could you squash all your migration commits into one?

JordiBForgeFlow (approved):

PR #679, opened 2019-05-07
[11.0][FIX] purchase_order_line_deep_sort: Fix compute order when order lines are created directly

sergio-teruel:
cc @Tecnativa

### DIFF
--- a/purchase_order_line_deep_sort/models/purchase_order.py
+++ b/purchase_order_line_deep_sort/models/purchase_order.py
@@ -1,5 +1,6 @@
 # Copyright 2018 Tecnativa - Vicent Cubells <vicent.cubells@tecnativa.com>
 # Copyright 2019 Tecnativa - Pedro M. Baeza
+# Copyright 2019 Tecnativa - Sergio Teruel
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3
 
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

purchase_order_line_deep_sort/models/purchase_order.py
```diff
@@ -45,6 +46,8 @@ def resolve_subfields(obj, line_order):
         )
         for line in sorted_lines:
             sequence += 10
```
Tardo, line 48: Magic number...

Tardo (approved):

purchase_order_line_deep_sort/models/purchase_order.py
pedrobaeza: Yes, as it's a suitable gap between lines. This is used in Odoo in other places. Any better suggestion?

Tardo: It's recommended to use a constant

pedrobaeza: Ah, well, I get you. OK, as this is out of this PR, let's merge it and we can improve it later with that constant for being in header.

PR #685, opened 2019-05-23
price_discount: Add Tests

hugosantosred:
Add tests for check if price_subtotal is equal between invoice and purchase


hugosantosred:
Related to #686 and only to confirm that the new tests are failing

Issue #558, opened 2018-06-12
[9.0] purchase_discount Discounts are not calculated correctly


mtbochaca:
When I make a purchase order with a discount, it is not calculated correctly. 
Ex: Units: 30 Price: 19.75 Discount: 50% Odoo total line: 296,40 Real total line: 296,25
Tested in runbot. 
http://3328685-9-0-20c98b.runbot2.odoo-community.org/web?debug=1#id=10&view_type=form&model=purchase.order&menu_id=232&action=303

pedrobaeza:
This version might miss some of the patches on 10.0. Please check if it's the case.

mtbochaca:
Not. The problem it's the same in version 10!!!
http://3331579-10-0-9cc0f2.runbot2.odoo-community.org/web?debug=1#id=8&view_type=form&model=purchase.order&menu_id=201&action=302
The problem it's in a rounding method:
Price_unit: 19.75 * discount: 0.5 = 9,875 Rounding (????) = 9,88 * Units = 30 = 296.40
The correct calculation:
Price_unit: 19.75 * discount: 0.5 = 9,875  * Units:30 =  296.25 Rounding here!!!!

I'm working in a code for solve the problem, but I'm not a programing. Sorry

legalsylvain:
Do you use purchase_rounding_method ? Because what you describe is exactly how works this module.

regards.

hugosantosred:
We are experiencing problems with purchase_discount module also in v11.

For example a Purchase Order with one line of a product with a price unit of 23.10 * 50 units and a discount of 1%, the subtotal amount computed for that line is 1143.50

Then validate and invoice the order, the line in the invoice is still the same with 50 units at 23.10 each and a discount of 1% but the computed subtotal amount in this case is 1143.45

So this behaviour is not correct because the amount in the purchase order and in the invoice must be the same.

This seems to be an error with the method that computes the discount in each line, because is using the line.price_unit for store the value of the price_unit with discount, so the amount is rounded to the same decimal positions as the price_unit in the database, losing some precission here

pedrobaeza:
which tax round policy do you have? Can you design a test to be added to the module that proves the problem?

hugosantosred:
@pedrobaeza I've made a PR with the tests for check if the price_subtotal is equal between invoice and purchase.

https://github.com/OCA/purchase-workflow/pull/685

PR #686, opened 2019-05-23
[11.0] [fix] purchase_discount

hugosantosred:
Use all the decimal positions from _get_discounted_price_unit to compute
the purchase line amount, for not lose decimal precision.
This can lead to unbalanced amounts between the purchase and the related invoice

https://github.com/OCA/purchase-workflow/issues/558

### DIFF
--- a/purchase_discount/tests/test_purchase_discount.py
+++ b/purchase_discount/tests/test_purchase_discount.py
@@ -28,7 +28,7 @@ class TestPurchaseOrder(common.SavepointCase):
             'currency_id': cls.env.user.company_id.currency_id.id,
             'rate': 1.00,
             'name': fields.Date.today()
-            })
+        })
         cls.purchase_order = cls.env['purchase.order'].create({
             'partner_id': cls.env.ref('base.res_partner_3').id,
         })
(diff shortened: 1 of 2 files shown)


pedrobaeza:
@legalsylvain what do you think about this?

purchase_discount/tests/test_purchase_discount.py
```diff
@@ -126,3 +141,17 @@ def test_invoice(self):
...
+            'purchase_id': self.purchase_order_check_amount.id,
+        })
+        invoice.purchase_order_change()
+        line = invoice.invoice_line_ids.filtered(
+            lambda x: x.purchase_line_id == self.po_line_check_amount
+        )
+        line.quantity = self.po_line_check_amount.product_qty
+        self.assertEqual(line.discount, 1)
+        self.assertAlmostEqual(line.price_subtotal,
+                               self.po_line_check_amount.price_subtotal)
```
pedrobaeza, line 157: You must use fixed data in tests (for example, `15.03`), not references to other fields.

legalsylvain:
Hi.

About the original bug, mentionned here #558. 

> Price_unit: 19.75 * discount: 0.5 = 9,875 Rounding (????) = 9,88 * Units = 30 = 296.40
The correct calculation:
Price_unit: 19.75 * discount: 0.5 = 9,875 * Units:30 = 296.25 Rounding here!!!!

I don't see as a bug, and for me, should be an option on the supplier.

I'm using a lot discount (1 & 2) on purchase module. In fact, some of my suppliers are using the first method of rounding, other are using the second one.

Indeed the default method should be the second one, as it is a more accurate.

My proposal : include the code of purchase_rounding_method in the module purchase_triple_discount.

https://github.com/OCA/purchase-workflow/tree/8.0/purchase_supplier_rounding_method

What do you think ?

vincent-hatakeyama:
The sale module has a discount on sales order lines, and the price on the line would be 296.25 in the example provided. The purchase module discount is expected to work the same way, but without this fix it does not.

vincent-hatakeyama:
I still think this should be merged to fix issue #850. It seems that some work need to be done to include it thought.

PR #690, opened 2019-06-03
[11.0][IMP] purchase_request: Allow change product on make PO

HviorForgeFlow:
Allow change product on make purchase order wizard allows this workflow case:
- Purchase request users create a PR without a specific product
- Once it is validated, the responsible, who can be the same user if it also is in purchase user or the responsible of create PO's, can modify product before create the PO.

CC @Eficent


AdriaGForgeFlow (approved):
:+1:

JordiBForgeFlow (approved):

PR #691, opened 2019-06-06
[11.0][FIX] purchase_order_line_deep_sort: Fix sorted with string and False values

sergio-teruel:
cc @Tecnativa

### DIFF
--- a/purchase_order_line_deep_sort/models/purchase_order.py
+++ b/purchase_order_line_deep_sort/models/purchase_order.py
@@ -32,8 +32,12 @@ class PurchaseOrder(models.Model):
         def resolve_subfields(obj, line_order):
             subfields = line_order.split('.')
             res = obj
+            str_fields = ('text', 'varchar', 'timestamp', 'date')
             for subfield in subfields:
-                res = getattr(res, subfield)
+                if res._fields[subfield].column_type[0] in str_fields:
+                    res = getattr(res, subfield) or ''
+                else:
+                    res = getattr(res, subfield)
             return res
 
         if not self.line_order and not self.line_direction:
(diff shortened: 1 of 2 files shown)


purchase_order_line_deep_sort/models/purchase_order.py
```diff
@@ -32,8 +32,12 @@ def _sort_purchase_line(self):
         def resolve_subfields(obj, line_order):
             subfields = line_order.split('.')
             res = obj
+            str_fields = (fields.Char, fields.Text)
             for subfield in subfields:
-                res = getattr(res, subfield)
+                if isinstance(res._fields[subfield], str_fields):
+                    res = getattr(res, subfield) or ''
+                else:
+                    res = getattr(res, subfield)
```
JayVora-SerpentCS, line 40: IMHO, only one line should work,
res = getattr(res, subfield) or ''
We would prefer to have the test case or example where you need If condition (may be a good comment on code)

sergio-teruel: Mhuuu!! But or '' os not always, if you order by a boolean field you do not want an empty string.
I will do tests for this use case... :smile:

pedrobaeza:
Waiting a test case

sergio-teruel:
Hi, I will try resume this PR:
The python function sorted crash when you try make a order string values with boolean values:
my_list = ['a', 'b', False, '2019-06-15'] --> TypeError: '<' not supported between instances of 'str' and 'int'
But NOT crash when the type list elements are numeric and boolean values
my_list = [1, 2, False] --> [False, 1, 2]
So when we make purchase order lines sorted method check column types form fields to set empty string for ('text', 'varchar', 'timestamp', 'date') column types.

Test case done.

purchase_order_line_deep_sort/models/purchase_order.py
sergio-teruel: @JayVora-SerpentCS I change isinstance function by fields.column_type method.

pedrobaeza (approved):

pedrobaeza:
@JayVora-SerpentCS please review

JayVora-SerpentCS (approved):

PR #693, opened 2019-06-11
[FIX] purchase_discount: Use sudo in purchase lines

nicomacr:
We do it with sudo to avoid permission errors in case the user validates the order and does not have permission to modify purchase lines.

For ex: User with not allow to purchase, and try to validate an order and the purchase order line has discount applied.

### DIFF
--- a/purchase_discount/models/stock_move.py
+++ b/purchase_discount/models/stock_move.py
@@ -13,7 +13,10 @@ class StockMove(models.Model):
         maximum inheritability.
         """
         price_unit = False
-        po_line = self.purchase_line_id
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
This is due to the ugly hack we need to do. Hopefully in v12 this will be totally avoided (thanks to the introduced hook, but not totally working: https://github.com/odoo/odoo/pull/29983)

nicomacr:
@pedrobaeza thanks, yes i notice that was for that.

ValentinoDefelice (approved):

zaoral (approved):

PR #702, opened 2019-06-17
[12.0][MIG] procurement_purchase_no_grouping: Migration to 12.0

Tardo:
Cc @Tecnativa


pedrobaeza (changes requested):
Please take advantage of a new method that is present now in v12 for not doing the context + search hack:

https://github.com/odoo/odoo/blob/9e9e9acef1a65483a4a3a318bde5c247b681942a/addons/purchase_stock/models/stock_rule.py#L70

pedrobaeza:
It seems the method is used already, but the context part is still being passed. AFAIK, that's not needed anymore.

pedrobaeza:
Sorry for the noise... It seems all OK after checking deeply.

pedrobaeza (approved):

tbaden (approved):
thanks for the work 👍 
I did a code review: LGTM

PR #704, opened 2019-06-18
[IMP] purchase_order_product_recommendation: warehouse filter

chienandalu:
- If you have multiple warehouses, you can also constrain the recommendations to
the deliveries of specific ones.

cc @Tecnativa

### DIFF
--- a/purchase_order_product_recommendation/wizards/purchase_order_recommendation.py
+++ b/purchase_order_product_recommendation/wizards/purchase_order_recommendation.py
@@ -46,6 +46,14 @@ class PurchaseOrderRecommendation(models.TransientModel):
         default=False,
         help='Show all products with supplier infos for this supplier',
     )
+    warehouse_ids = fields.Many2many(
+        comodel_name='stock.warehouse',
+        string='Warehouse',
+        help='Constrain search to an specific warehouse',
+    )
+    warehouse_count = fields.Integer(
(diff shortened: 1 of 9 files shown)


rafaelbn (approved):
Functionally tested :+1:

purchase_order_product_recommendation/wizards/purchase_order_recommendation.py
```diff
@@ -71,18 +79,19 @@ def _find_move_line(self, src='internal', dst='customer'):
...
-                ('date', '<=', '{} 23:59:59'.format(self.date_end)),
-                ('location_id.usage', '=', src),
-                ('location_dest_id.usage', '=', dst),
-                ('state', '=', 'done'),
-            ],
-            ['product_id', 'qty_done'],
-            ['product_id'],
-        )
+            domain, ['product_id', 'qty_done'], ['product_id'])
         # Manual ordering that circumvents ORM limitations
```
victor-ditay, line 95: I would like to know more about this ordering. Maybe deserve to be a tool for any read_group as an option

pedrobaeza: The problem is that you can't limit the `read_group` operation, because you don't know totals before performing the query, so we sort according that totals and limit to the top n ones.

carlosdauden (approved):

sergio-teruel (approved):

PR #705, opened 2019-06-19
[10.0][ENH]purchase_request allocations

AaronHForgeFlow:
Followed the same approach as [stock_request module](https://github.com/OCA/stock-logistics-warehouse/tree/11.0/stock_request)

This allows to track closely the status of the purchase request:
* What lines have been satisfied
* How much qty is pending
* How much qty was cancelled

This also sets the purchase request done when all lines are satisfied.

If the idea is welcomed it will be ported to newer versions


AaronHForgeFlow:
Closing for now.

PR #706, opened 2019-06-26
12.0 mig purchase_landed_cost

ernestotejeda:
Supersedes https://github.com/OCA/purchase-workflow/pull/632

@novawish , about your review in https://github.com/OCA/purchase-workflow/pull/632:

https://github.com/OCA/purchase-workflow/pull/632#pullrequestreview-207155769
> We are facing an issue with multi-currency configuration: The Vendor Bills may have different currencies compared with company's currency. I added a suggestion, can you tell me what you think?

I agree with you that we should take into account the multi-currency, but I do not think that substituting price_subtotal for price_subtotal_signed will solve the problem, because price_subtotal_signed is in the currency of the company of the invoice, not in the currency of the company of the "Purchase cost distributions ". What i did was get the price_subtotal_signed from the invoice and convert it to the currency of the "Purchase cost distributions".
I also converted the currency of the expense type (purchase.expense.type) to the currency of the "Purchase cost distributions" in the Onchange of Expense type. 
what do you think?

About https://github.com/OCA/purchase-workflow/pull/632#discussion_r259899190

> I tested and apparently this does not work anymore.
> I think we don't need `possible_picking_ids` and we can write a function that return directly the domain like this:

I think it is a good idea to write a function that directly returns the domain, but as it is now it works according to the tests that I have done. Could you check again that this does not work?

Cc @Tecnativa


purchase_landed_cost/models/purchase_order.py
```diff
@@ -0,0 +1,31 @@
...
+        self.ensure_one()
+        line_obj = self.env['purchase.cost.distribution.line']
+        lines = line_obj.search([('purchase_id', '=', self.id)])
+        if lines:
+            mod_obj = self.env['ir.model.data']
+            model, action_id = tuple(
+                mod_obj.get_object_reference(
+                    'purchase_landed_cost',
+                    'action_purchase_cost_distribution'))
+            action = self.env[model].browse(action_id).read()[0]
```
carlosdauden: Why don't use direcly
```suggestion
            action = self.env.ref('purchase_landed_cost.action_purchase_cost_distribution').read()[0]
```
and remove previous code.

purchase_landed_cost/models/purchase_order.py
```diff
@@ -0,0 +1,31 @@
...
+            ids = set([x.distribution.id for x in lines])
```
carlosdauden: More cool :wink: 
```suggestion
            distributions = lines.mapped('distribution')
            if len(distributions) == 1:            
                form = self.env.ref('purchase_landed_cost.purchase_cost_distribution_form')
                action['views'] = [(form.id, 'form')]
                action['res_id'] = distributions.id
            else:
                action['domain'] = [('id', 'in', distributions.ids)]
```

purchase_landed_cost/models/stock_picking.py
```diff
@@ -0,0 +1,30 @@
...
+
+class StockPicking(models.Model):
+    _inherit = 'stock.picking'
+
+    @api.multi
+    def action_open_landed_cost(self):
+        self.ensure_one()
+        line_obj = self.env['purchase.cost.distribution.line']
+        lines = line_obj.search([('picking_id', '=', self.id)])
+        if lines:
```
carlosdauden, line 15: Create new common method to pass lines and use in both models or change same commented above.

carlosdauden (approved):
Code revision

purchase_landed_cost/models/purchase_cost_distribution.py
```diff
@@ -562,12 +562,26 @@ def _compute_display_name(self):
     @api.onchange('type')
     def onchange_type(self):
         if self.type and self.type.default_amount:
-            self.expense_amount = self.type.default_amount
+            # set expense_amount in the currency of the distribution
```
victor-ditay: Good text self-explanatory but might be better as the docstring of the function?

purchase_landed_cost/models/purchase_cost_distribution.py
```diff
@@ -562,12 +562,26 @@ def _compute_display_name(self):
...
+            company = self.company_id or self.env.user.company_id
+            cost_date = self.distribution.date or fields.Date.today()
+            self.expense_amount = currency_from._convert(amount, currency_to,
+                                                         company, cost_date)
 
     @api.onchange('invoice_line')
     def onchange_invoice_line(self):
         self.invoice_id = self.invoice_line.invoice_id.id
-        self.expense_amount = self.invoice_line.price_subtotal
+        # set expense_amount in the currency of the distribution
```
victor-ditay: Good text self explanatory but might be better as the docstring of the function?

purchase_landed_cost/wizard/import_invoice_line.py
```diff
@@ -24,11 +24,20 @@ class ImportInvoiceLine(models.TransientModel):
     @api.multi
     def action_import_invoice_line(self):
         self.ensure_one()
+        dist_id = self.env.context['active_id']
```
victor-ditay, line 27: Could you add docstring to explain whole funcion?

victor-ditay (changes requested):

purchase_landed_cost/models/purchase_order.py
ernestotejeda: Done

purchase_landed_cost/models/purchase_order.py
ernestotejeda: Done

purchase_landed_cost/models/stock_picking.py
ernestotejeda: Done

purchase_landed_cost/models/purchase_cost_distribution.py
ernestotejeda: Done

purchase_landed_cost/models/purchase_cost_distribution.py
```diff
@@ -0,0 +1,603 @@
...
+        help="The line volume in m3.")
+    company_id = fields.Many2one(
+        comodel_name="res.company", related="distribution.company_id",
+        store=True,
+    )
+
+    @api.model
+    def get_action_purchase_cost_distribution(self):
+        xml_id = 'purchase_landed_cost.action_purchase_cost_distribution'
+        action = self.env.ref(xml_id, raise_if_not_found=True).read()[0]
```
carlosdauden: True is the default value for raise_if_not_found:
https://github.com/odoo/odoo/blob/36551615b7265ad83dbf9d1207ff9ee59b7069f3/odoo/api.py#L792
```suggestion
        action = self.env.ref(xml_id).read()[0]
```

ernestotejeda: Right, change done

ernestotejeda:
Changes done

pedrobaeza:
@victormmtorres can you re review this?

rafaelbn (approved):
:+1:

PR #707, opened 2019-06-27
[11.0][IMP] purchase_date_planned_manual: Avoid merging lines with different date_planned

MiquelRForgeFlow:
Solves https://github.com/OCA/purchase-workflow/pull/422#issuecomment-311287468.

### DIFF
--- a/purchase_date_planned_manual/models/purchase_order.py
+++ b/purchase_date_planned_manual/models/purchase_order.py
@@ -52,3 +52,23 @@ class PurchaseOrderLine(models.Model):
             'This line is scheduled for: %s. \n However it is now planned to '
             'arrive late.') % Dt.to_string(Dt.context_timestamp(
                 self, Dt.from_string(self.date_planned))))
+
+    def _merge_in_existing_line(self, product_id, product_qty, product_uom,
+                                location_id, name, origin, values):
+        if self.date_planned == values.get('date_planned'):
+            return super(PurchaseOrderLine, self)._merge_in_existing_line(
+                product_id, product_qty, product_uom,
+                location_id, name, origin, values)
+        return False
+
+
+class ProcurementRule(models.Model):
+    _inherit = 'procurement.rule'
+
+    def _prepare_purchase_order_line(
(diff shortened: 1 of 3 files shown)


MiquelRForgeFlow:
Please, could you check @lreficent ?

LoisRForgeFlow (changes requested):
It could work, but please can you add a test case to be sure?

MiquelRForgeFlow:
Tests added :green_apple:

LoisRForgeFlow (changes requested):
Tested locally, it is not merging prcourements into the same line but planned date is not respected in PO
image

MiquelRForgeFlow:
Fixed

MiquelRForgeFlow:
ping: @lreficent

purchase_date_planned_manual/models/purchase_order.py
```diff
@@ -52,3 +52,24 @@ def action_delayed_line(self):
...
             'arrive late.') % Dt.to_string(Dt.context_timestamp(
                 self, Dt.from_string(self.date_planned))))
+
+    def _merge_in_existing_line(self, product_id, product_qty, product_uom,
+                                location_id, name, origin, values):
+        if self.date_planned == values.get('date_planned'):
+            return super(PurchaseOrderLine, self)._merge_in_existing_line(
+                product_id, product_qty, product_uom,
+                location_id, name, origin, values)
+        else:
```
nikul-serpentcs: @mreficent here `else` is needed ??

I think It Should be a better :smiley: 
```
if self.date_planned == values.get('date_planned'):
    return super(PurchaseOrderLine, self)._merge_in_existing_line(
        product_id, product_qty, product_uom,
        location_id, name, origin, values)
return False
```

MiquelRForgeFlow: done

HviorForgeFlow (approved):

JordiBForgeFlow (approved):

JordiBForgeFlow:
@mreficent can you forward port to 12.0?

PR #709, opened 2019-07-02
[10.0][ENH]better extensibility for messages in purchase requests

AaronHForgeFlow:
This PR does two things:

-  Pass the objects to the methods that makes the messages instead of the values directly. This way we can easily change the message in a another module.

- Displays the display_name of the product, instead of the description in the purchase request (This is the only functional change here. It makes sense as we have stock moves and products involved)

### DIFF
--- a/purchase_request_to_rfq/models/purchase_order.py
+++ b/purchase_request_to_rfq/models/purchase_order.py
@@ -22,12 +22,17 @@ class PurchaseOrder(models.Model):
             request.name, self.name)
 
         for line in request_dict.values():
+            if line['request_line'].product_id:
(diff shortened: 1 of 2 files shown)


MiquelRForgeFlow (approved):
LGTM

hparfr (approved):
lgtm

HviorForgeFlow (approved):

rousseldenis (approved):

MiquelRForgeFlow:
@aheficent shouldn't this be forward ported?

AaronHForgeFlow:
@mreficent yes good point

PR #540, opened 2018-04-13
[11.0][MIG] purchase last price info

celm1990:

PR #713, opened 2019-07-09
[11.0][MIG] purchase_last_price_info

AdriaGForgeFlow:
Standard migration of module purchase_last_price_info with some modifications.

* Remove computed fields to be able to put field in tree view without affecting scalability
* Add methods on Purchase Order model that will compute last price info upon PO confirmation or cancellation

Superseeds #540 

cc ~ @Eficent


rousseldenis:
Indeed, that could be better, but can slow great purchase orders confirmation.

You should at least add migration script to initialize existing data.

HviorForgeFlow (approved):
Functional review :+1:

AaronHForgeFlow (approved):
:+1: Tested on runbot

purchase_last_price_info/models/product.py
```diff
@@ -1,49 +1,51 @@
 
-from odoo import api, fields, models
+from odoo import fields, api, models
```
MiquelRForgeFlow: it's recommended to have them in alphabetical order

purchase_last_price_info/migrations/11.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2019 Eficent Business and IT Consulting Services S.L.
+#   (http://www.eficent.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import api, SUPERUSER_ID
+
+
+def migrate(cr, version):
+    if not version:
+        return
+    with api.Environment.manage():
```
MiquelRForgeFlow: Don't! It's better to reuse the post_init_hook to avoid duplicated code.

I think you should do:

```python
from openupgradelib import openupgrade
from odoo.addons.purchase_last_price_info import set_last_price_info


@openupgrade.migrate()
def migrate(env, version):
    set_last_price_info(env.cr, env.registry)
```

MiquelRForgeFlow:
please, squash same author adjacent not relevant commits.

rafaelbn:
Please @ageficent add usage section, is short like:

- Go To Products -> Tab Purchase (see screenshot)

You can add this screenshot:




Is there any other please to check this info? Only in the product form?

Thanks

cc @chienandalu

rafaelbn (commented):
Improve readme for new odoo users could test it easily please :smile:

purchase_last_price_info/models/product.py
```diff
@@ -0,0 +1,71 @@
...
+                    _compute_quantity(last_line.price_unit,
+                                      last_line.product_uom)
+                last_supplier = last_line.order_id.partner_id
+
+            # Assign values to record
+            product.last_purchase_date = date_order
+            product.last_purchase_price = price_unit_uom
+            product.last_supplier_id = last_supplier
+            # Set related product template values
+            product.product_tmpl_id.set_product_template_last_purchase(
```
chienandalu, line 53: What if there are variants with different prices?

AdriaGForgeFlow: In the product template form view you will only see the information if there are one or less variants. When there are multiple variants even if they have different prices it does not matter since you will not be able to see the information in the product template.

chienandalu: Ok :)

purchase_last_price_info/models/product.py
```diff
@@ -0,0 +1,71 @@
...
+            # Check if Order ID was passed, to speed up the search
+            if order_id:
+                lines = PurchaseOrderLine.search([
+                    ('order_id', '=', order_id),
+                    ('product_id', '=', product.id)], limit=1)
+            else:
+                lines = PurchaseOrderLine.search(
+                    [('product_id', '=', product.id),
+                     ('state', 'in', ['purchase', 'done'])]).sorted(
+                    key=lambda l: l.order_id.date_order, reverse=True)
```
chienandalu, line 35: why not limit to one here?

purchase_last_price_info/models/product.py
```diff
@@ -0,0 +1,71 @@
...
+
+                date_order = last_line.order_id.date_order
+                # Compute Price Unit in the Product base UoM
+                price_unit_uom = product.product_tmpl_id.uom_id.\
+                    _compute_quantity(last_line.price_unit,
+                                      last_line.product_uom)
+                last_supplier = last_line.order_id.partner_id
+
+            # Assign values to record
+            product.last_purchase_date = date_order
```
chienandalu, line 49: Do you need to assign values if no lines are found

AdriaGForgeFlow: Yes, you need this in case you cancel all Purchase Orders related to that product. In this case you need to set the values of the product to False and 0, otherwise they would stay set to the last cancelled PO values.

purchase_last_price_info/models/product.py
AdriaGForgeFlow: I'm not sure it can be done since you are doing the search first, then sorting with the method sorted(). If you limit to 1 the search you may not get the most recent PO.

purchase_last_price_info/models/product.py
chienandalu: Good point :+1:

purchase_last_price_info/models/product.py
HviorForgeFlow: You could try to use order='date_order desc' on search, but not sure as you are taking order id

chienandalu: Well, at the end these are parameters that translate into an sql query (something like `SELECT * FROM purchase_order_line WHERE product_id = '3' AND state IN ('purchase', 'done') SORT BY date DESC`), so I'd say it's going to do it well at once

AdriaGForgeFlow: The problem here is that you are searching in the purchase order line table but using the purchase order date to sort. It would require a more complex sql query with a `JOIN` I think.

PR #715, opened 2019-07-10
[12.0][ADD] product_form_purchase_link

rousseldenis:
As Odoo removed links in product forms and replaced it by a purchase.report model, this reintroduces the link to purchases.


bealdav (approved):

AaronHForgeFlow:
What is the difference with the standard button?

image

rousseldenis:
> What is the difference with the standard button?
> 
> image

If you click, you go to a new model 'purchase.report' with graph/pivot views and you don't have access to purchases themselves...

AaronHForgeFlow (approved):
:+1: It works and it makes sense

rousseldenis:
Soon the little brother for Sales as Odoo did the same :smile:

sbidoul:
`product_purchase` might be too generic as a module name?

rousseldenis:
> `product_purchase` might be too generic as a module name?

product_purchase_link ?

bealdav:
I ask myself if we don't need to switch from smart button to action button.
Isn't it confusing to have so many blocks in UI and so many computed fields ?

In small project no problem but in big ones ...

rousseldenis:
> I ask myself if we don't need to switch from smart button to action button.
> Isn't it confusing to have so many blocks in UI and so many computed fields ?
> 
> In small project no problem but in big ones ...

The computed field is done once (as it appears yet in first button).

I agree that it can be bulky to use but I don't like Actions when it should be accessible quickly.

If you have a more suitable solution (In Purchase tab maybe, but it requires two clicks).

rousseldenis:
Any idea for the name ? Suggested one is the good one ?

bodedra:
`product_purchase_report_link` or these LGTM `product_purchase_link`

LoisRForgeFlow:
I think product_purchase_link is good, but to be extra-specific about the functionality added you could go for `product_form_purchase_link`.

rousseldenis:
> I think product_purchase_link is good, but to be extra-specific about the functionality added you could go for `product_form_purchase_link`.

I like that last one !

LoisRForgeFlow:
@rousseldenis can you do the rename then? :smile:

pedrobaeza:
I prefer `product_purchase_info`

AdriaGForgeFlow (approved):
> I think product_purchase_link is good, but to be extra-specific about the functionality added you could go for `product_form_purchase_link`.

Agreed :+1:

rafaelbn:
@rousseldenis please could you please rename? Thanks!

LoisRForgeFlow:
@rousseldenis ping, are you going to rename this?

rousseldenis:
@rafaelbn @lreficent Done !

legalsylvain:
Thanks @rousseldenis for this usefull module.

PR #718, opened 2019-07-17
[WIP][10.0][FIX] purchase_request_procurement update procurement state

hparfr:
Before this PR:
When a PR is done, the state of its procurements are not updated.
After this PR:
When a PR is done, its procurements states are set to done.

I think the procurement should done if the purchase request is done.
Note: having the procurement in exception if the purchase request is rejected is already in place.

It's related to the issue described in this PR: #719


hparfr:
It's not generic enough.

PR #719, opened 2019-07-17
[10.0][FIX] purchase_request_to_rfq no link to procurement

hparfr:
Before this PR:
purchase order lines created from the wizard adds a reference to the procurement
If the order line is removed, the procurement goes in exception

After this PR
purchase order lines created from the wizard have not procurement id
If the order line is removed, there is no impact the on procurement


AaronHForgeFlow:
IMHO it is correct that the procurement goes to exception. Otherwise it will be always "Running", won't it?

hparfr:
You can use purchase_request modules to make bids. Like create 3 different rfq to 3 different suppliers and only confirm one.

The procurement should stop running when all the quantities have been received.

AaronHForgeFlow:
So in case I create RFQ instead of bids, and I cancel the purchase requests, the procurement order will be always running? Is it possible to keep the link at least in those cases?

hparfr:
I don't get what you mean : 
- a procurement creates an PRL,
- a RFQ is created from the PRL
- the PR is rejected ? then the procurement is cancelled ?

HviorForgeFlow:
ping @aheficent

AaronHForgeFlow:
> I don't get what you mean :
> 
> * a procurement creates an PRL,
> * a RFQ is created from the PRL
> * the PR is rejected ? then the procurement is cancelled ?

Yes, as long as there is no way to relaunch he procurement for non technical user, is it?

PR #726, opened 2019-07-25
[ADD] purchase_exception: migration script to v12

MiquelRForgeFlow:
Add migration script.

### DIFF
--- a/purchase_exception/migrations/12.0.1.0.1/noupdate_changes.xml
+++ b/purchase_exception/migrations/12.0.1.0.1/noupdate_changes.xml
@@ -1,0 +1,16 @@
+<?xml version='1.0' encoding='utf-8'?>
+<odoo>
+  <record id="ir_cron_test_po_order_except" model="ir.cron" forcecreate="True">
+    <field name="code">model.test_all_draft_orders()</field>
+  </record>
+  <record id="po_excep_no_email" model="exception.rule">
+<!--    <field name="rule_group"/>-->
+    <field name="code">if not self.partner_id.email:
+            failed=True</field>
+  </record>
+  <record id="pol_excep_qty_check" model="exception.rule">
+<!--    <field name="rule_group"/>-->
+    <field name="code">if self.product_qty &lt; 0:
+            failed=True</field>
+  </record>
+</odoo>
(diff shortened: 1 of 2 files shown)


MiquelRForgeFlow:
@sudhir-erpharbor

purchase_exception/migrations/12.0.1.0.1/noupdate_changes.xml
```diff
@@ -0,0 +1,16 @@
...
+  <record id="ir_cron_test_po_order_except" model="ir.cron" forcecreate="True">
+    <field name="code">model.test_all_draft_orders()</field>
+  </record>
+  <record id="po_excep_no_email" model="exception.rule">
+<!--    <field name="rule_group"/>-->
+    <field name="code">if not self.partner_id.email:
+            failed=True</field>
+  </record>
+  <record id="pol_excep_qty_check" model="exception.rule">
+<!--    <field name="rule_group"/>-->
```
HviorForgeFlow, line 12: Cleanup unused code

HviorForgeFlow (commented):
Can you provide a little explanation about what this migration script is fixing? Thanks

purchase_exception/migrations/12.0.1.0.1/noupdate_changes.xml
MiquelRForgeFlow: It's better to not delete in this case. If in the future we check again the noupdate changes from v11 and v12 versions, then this way we will not be tempted to add the commented code.

MiquelRForgeFlow:
@hveficent In v12, respect to v11, some changes have been done in xml records that are with "noupdate = 1". So, in order to make effective this changes during migration, this code is needed.

HviorForgeFlow (approved):

AaronHForgeFlow (approved):
:+1: LGTM

HviorForgeFlow:
@pedrobaeza I'm not sure if ocabot can merge this fix to migration script from 12.0.1.0.1 because it will change version. What should we do?

pedrobaeza:
we can merge without version bump:

/ocabot merge

et voilá

HviorForgeFlow:
I didin't realize about that way :smile: 
Thanks you so much!

PR #734, opened 2019-08-09
[11.0][MIG] purchase_fiscal_position_update

MiquelRForgeFlow:
Standard migration.

Fractioned Readme will be implemented in v12.


alexis-via:
This feature seems native in v11, cf https://github.com/odoo/odoo/blob/11.0/addons/purchase/models/purchase.py#L242 (untested, I don't use v11)
So we don't need this module any more.

MiquelRForgeFlow:
I am not sure. Are you sure it's the same? What you point is also in v10 and v11, as you can see: https://github.com/odoo/odoo/blob/10.0/addons/purchase/models/purchase.py#L239. So if it's the same, I don't understand why the module in v10 exists.

PR #735, opened 2019-08-09
[12.0][MIG] purchase_fiscal_position_update

MiquelRForgeFlow:
Standard migration of https://github.com/OCA/purchase-workflow/pull/734.


legalsylvain (approved):
Diff review & test on runbot.
Works like a charm.
thanks @mreficent

legalsylvain:
CC : @alexis-via

alexis-via:
This feature is native in v12, so I don't see any reason to port that module to v12.
cf https://github.com/odoo/odoo/blob/12.0/addons/purchase/models/purchase.py#L202

MiquelRForgeFlow:
I am not sure. Are you sure it's the same? What you point is also in v10 and v11, as you can see: https://github.com/odoo/odoo/blob/10.0/addons/purchase/models/purchase.py#L239. So if it's the same, I don't understand why the module in v10 exists.

alexis-via:
Porting this module to v10 was a mistake (my mistake !). I just made a test on a v10 DB without this module ; I confirm taxes are updated on lines when you change the fiscal position.

MiquelRForgeFlow:
hahahaha ok!

PR #739, opened 2019-08-21
[12.0][ADD] purchase_invoice_plan

kongrattapong:
## Ready to review.

### DIFF
--- a/purchase_invoice_plan/__init__.py
+++ b/purchase_invoice_plan/__init__.py
@@ -1,0 +1,5 @@
+# Copyright 2019 Ecosoft Co., Ltd (http://ecosoft.co.th/)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html)
+
+from . import models
+from . import wizard
--- a/purchase_invoice_plan/__manifest__.py
+++ b/purchase_invoice_plan/__manifest__.py
@@ -1,0 +1,23 @@
+# Copyright 2019 Ecosoft Co., Ltd (http://ecosoft.co.th/)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html)
+
+{
+    'name': 'Purchase Invoice Plan',
+    'summary': 'Add to purchases order, ability to manage future invoice plan',
+    'version': '12.0.1.0.0',
+    'author': 'Ecosoft,Odoo Community Association (OCA)',
+    'license': 'AGPL-3',
+    'website': 'https://github.com/OCA/purchase-workflow/',
+    'category': 'Purchase',
+    'depends': ['account',
+                'purchase_open_qty',
+                'purchase_stock',
+                ],
+    'data': ['security/ir.model.access.csv',
+             'wizard/purchase_create_invoice_plan_view.xml',
+             'wizard/purchase_make_planned_invoice_view.xml',
+             'views/purchase_view.xml',
+             ],
+    'installable': True,
+    'development_status': 'alpha',
+}
--- a/purchase_invoice_plan/models/__init__.py
+++ b/purchase_invoice_plan/models/__init__.py
@@ -1,0 +1,4 @@
+# Copyright 2019 Ecosoft Co., Ltd (http://ecosoft.co.th/)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html)
+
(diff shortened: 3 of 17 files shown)


kittiu:
@kongrattapong is this ready for review? If so, please remove [WIP] from the PR name.

kongrattapong:
> @kongrattapong is this ready for review? If so, please remove [WIP] from the PR name.

@kittiu ready to review krub

kittiu (approved):

kittiu:
@Saran440 would you help review and test this

HviorForgeFlow (changes requested):
Functional review it is a good feature.

Little comments:
- I miss access to purchase invoice plan view model, to list them, not just Invoice Plan tab on PO.
- Plan date is not propagated to Invoice when using creating all remaining bills. Not sure if in Bill Date or Due Date.
- Invoice plan can be only created on a draft PO, once is confirmed PO I can't manipulate it. I would like to have the ability to decide if I use Invoice Plan or not until my next step is create bill.

Do this comments make sense?

kittiu:
Thank you @hveficent for your comment, let's address one by one,

> I miss access to purchase invoice plan view model, to list them, not just Invoice Plan tab on PO.

You mean, you want to have a new menu "Purchase Invoice Plan" as new menu? Just wonder what use case is it for?
Note: The intention of invoice plan is just a helper. Even there is a plan, you don't need to follow them all.

> Plan date is not propagated to Invoice when using creating all remaining bills. Not sure if in Bill Date or Due Date

It should be Bill Date. Is it not propagated? I will check.

> Invoice plan can be only created on a draft PO, once is confirmed PO I can't manipulate it. I would like to have the ability to decide if I use Invoice Plan or not until my next step is create bill.

We don't allow because, I afraid that, says, there is 10 installment and user used it for 3 installments. But then, he decided to edit the plan, I just afraid it won't be consistent anymore.

What I could do is, once PO created 1st invoice, we will disallow creating again, will this works?

HviorForgeFlow:
> > I miss access to purchase invoice plan view model, to list them, not just Invoice Plan tab on PO.
> 
> You mean, you want to have a new menu "Purchase Invoice Plan" as new menu? Just wonder what use case is it for?
> Note: The intention of invoice plan is just a helper. Even there is a plan, you don't need to follow them all.

Maybe it this is not the use case, but: If I want to take an overview about all Invoice plans with my vendors I need to go 1 per 1 on their related purchase to see the plan and the status.

A Entry, in vendors like Vendor Invoice Plans calling a tree view with all Invoice Plans created by the system can help me to speed up my periodically revision. Not mandatory, nice to have.

> > Invoice plan can be only created on a draft PO, once is confirmed PO I can't manipulate it. I would like to have the ability to decide if I use Invoice Plan or not until my next step is create bill.
> 
> We don't allow because, I afraid that, says, there is 10 installment and user used it for 3 installments. But then, he decided to edit the plan, I just afraid it won't be consistent anymore.

It depends on the workflow of the users. I mean, I can't activate/deactivate invoice plan once I confirmed the order, if I forget to add my invoice plan or I reach a deal with vendor later than order confirmation to allocate payments in different installments.

A good first approach is:
> What I could do is, once PO created 1st invoice, we will disallow creating again, will this works?

kittiu:
@hveficent let me enhance it first. Thanks for the constructive comment :+1:

kittiu:
@hveficent Fixed all 3 points, you can make review again.
Thank you very much, they all good points.

HviorForgeFlow (approved):
@kittiu  I've been testing the module again, maybe Invoice Plan Tab could be located on Invoice App, not sure, it depends which users are interested on see that, nevertheless they also can get this information filetring Purchase Orders where Invoice Plan is set. So I think I was mistaken about point 1.

Other points are covered with new changes. 

LGTM :+1:

Let's see how perform this module on customers.

Thanks!

kittiu:
@hveficent also, there is another module `purchase_deposit` which is already merged.
As both of them will create invoice, as separated module they are not good together (as they not aware of each other).
I am creating a new bridge module `purchase_invoice_plan_deposit` which will auto install if both are installed. I will let you know when it is ready. :)

Note: Featured that we just did for purchase_invoice_plan, I will port them also to sale_invoice_plan.

Thank you!

PR #740, opened 2019-08-23
[10][FIX]purchase_request_to_rfq Always use Product Unit Of Measure precision

hparfr:
Be consistent with : https://github.com/OCA/purchase-workflow/blob/10.0/purchase_request/models/purchase_request.py#L250

### DIFF
--- a/purchase_request_to_rfq/models/purchase_request.py
+++ b/purchase_request_to_rfq/models/purchase_request.py
@@ -8,6 +8,7 @@ from datetime import datetime
 from dateutil.relativedelta import relativedelta
 
 from odoo import _, api, exceptions, fields, models
+import odoo.addons.decimal_precision as dp
 
 
 class PurchaseRequestLine(models.Model):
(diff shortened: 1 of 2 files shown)


purchase_request_to_rfq/models/purchase_request.py
```diff
@@ -60,8 +61,10 @@ def _compute_purchase_state(self):
                     temp_purchase_state = 'draft'
             rec.purchase_state = temp_purchase_state
 
-    purchased_qty = fields.Float(string='Quantity in RFQ or PO',
-                                 compute="_compute_purchased_qty")
+    purchased_qty = fields.Float(
+        string='Quantity in RFQ or PO',
+        digits=dp.get_precision('Product Unit of Measure'),
```
LoisRForgeFlow, line 66: Are you sure this one is needed? as it is inherited if should already have digits defined.

LoisRForgeFlow: Nevermind, they are different fields... my bad.

LoisRForgeFlow (approved):

HviorForgeFlow (approved):

JordiBForgeFlow (approved):

PR #744, opened 2019-09-09
[12.0][ADD] purchase_supplier_approve: supplier approval funcionality

dsidorenkovas-rgbconsulting:
This module adds an approval state for suppliers. There are 3 available states:

  * Pending to approval (by default)
  * Approved
  * Not approved

Suppliers can be approved only by a specific group of users: **Suppliers approval**.

The supplier approval state is used in purchase order confirmation process.
Only orders with approved suppliers can be confirmed.


kittiu:
Thanks contribution :+1: 
To start with, can you change name of this PR to [12.0][ADD] purchase_supplier_approve
Also, please greenify the travis first. :)

purchase_supplier_approve/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2019 RGB Consulting - Domantas Sidorenkovas
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    'name': "Purchase Supplier Approve",
+    'version': '12.0.1.0.0',
+    'category': 'Purchase',
+    'depends': ['purchase'],
+    'author': "RGB Consulting",
```
MeetKD: Add OCA as author.

purchase_supplier_approve/models/res_partner.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    supplier_approval = fields.Selection([
+        ('pda', 'Pending approval'),
+        ('apd', 'Approved'),
+        ('nad', 'Not approved'),
+    ], default='pda', readonly=True, track_visibility='onchange')
```
MeetKD: It would be great if you can add help string.

MeetKD (changes requested):

purchase_supplier_approve/__manifest__.py
dsidorenkovas-rgbconsulting: Hi,
As you can see I made this change in the following commit:
https://github.com/OCA/purchase-workflow/pull/744/commits/b6f823d3cfa9d4a71fd8cc4b6128cda08aeb6d33#diff-8e65f9724534767cee0c8449ebf1c8dcR9

purchase_supplier_approve/models/res_partner.py
dsidorenkovas-rgbconsulting: Hi again,
As you asked, I made this change also in the following commit:
https://github.com/OCA/purchase-workflow/pull/744/commits/b6f823d3cfa9d4a71fd8cc4b6128cda08aeb6d33#diff-d44cb370023ae01225035b3421763d4eR15

dsidorenkovas-rgbconsulting:
> Thanks contribution 
> To start with, can you change name of this PR to [12.0][ADD] purchase_supplier_approve
> Also, please greenify the travis first. :)

Hi @kittiu,

I renamed the PR as you asked and also checked some travis errors. The current errors are caused by tests from other modules, since in this module we are blocking purchase order confirmation.

purchase_supplier_approve/models/res_partner.py
MeetKD: Looks good :+1:

purchase_supplier_approve/__manifest__.py
MeetKD: :+1:

purchase_supplier_approve/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2019 RGB Consulting - Domantas Sidorenkovas
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
```
kittiu, line 3: This is a minor comment and non-blocking, but would you use single quote to all of your code? It is helpful when we do searhing for code later on.

purchase_supplier_approve/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+    'depends': ['purchase'],
+    'author': "RGB Consulting,"
+              "Odoo Community Association (OCA)",
+    'license': 'AGPL-3',
+    'website': "https://github.com/OCA/purchase-workflow",
+    'data': [
+        'security/purchase_security.xml',
+        'views/res_partner_view.xml',
+        'wizard/supplier_approval_wizard.xml',
+    ],
```
kittiu, line 18: Do you plan to be maintainer of this module? (It will helps for you to easily merge it later).
'maintainers': ['your_git_login'],

purchase_supplier_approve/wizard/supplier_approval_wizard.py
```diff
@@ -0,0 +1,30 @@
...
+        return self.env['res.partner'].browse(self._context.get('active_id'))
+
+    partner_id = fields.Many2one(comodel_name='res.partner',
+                                 default=_default_partner, string='Supplier')
+
+    supplier_approval = fields.Selection([
+        ('pda', 'Pending approval'),
+        ('apd', 'Approved'),
+        ('nad', 'Not approved'),
+    ], default='pda', track_visibility='onchange')
```
kittiu: Track visibility is not needed I think, as this is a wizard

purchase_supplier_approve/wizard/supplier_approval_wizard.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+class SupplierValidationWizard(models.TransientModel):
+    _name = "supplier.approval.wizard"
+    _description = "Supplier Validation Wizard"
+
+    def _default_partner(self):
+        return self.env['res.partner'].browse(self._context.get('active_id'))
+
+    partner_id = fields.Many2one(comodel_name='res.partner',
```
kittiu: Looks like, partner_id is not really needed, right? (you need it only because you want to set its default).
On open wizard, I would prefer to use `def default_get(...)` to set default supplier approval state (you can use _default_partner or _get_partner for partner object)
On save, again, you can still use _get_partner to get the partner and write it.

Note:
* This way, we can avoid using _onchange_partner_id, just to get the default approve state. To use method for it right purpose.
* Code can be shorten a bit

WDYT?

purchase_supplier_approve/models/res_partner.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2019 RGB Consulting - Domantas Sidorenkovas
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    supplier_approval = fields.Selection([
```
kittiu: I would prefer field name like, supplier_approve_status, or supplier_status, which easier to understand at first glance.
For selection, I would prefer,
[('pending_approve', 'Pending Approval'), ('approved', 'Approved'), ('not_approved', 'Not approved')]
My point is about readability. As such, when you see this column from database,
`supplier_approve_status = 'approved'`, would be easier to understand than, `supplier_approve = 'apd'`

purchase_supplier_approve/models/purchase.py
```diff
@@ -0,0 +1,15 @@
...
+from odoo.exceptions import ValidationError
+
+
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+
+    @api.multi
+    def button_confirm(self):
+        if any(rec.partner_id.supplier_approval != 'apd' for rec in self):
+            raise ValidationError(_('You cannot confirm a purchase order if the supplier is not approved.'))
```
kittiu: I seem many lines with length > 79 chars, it should be in new line.
I also wonder why it can pass travis test... my editor whould show error like this,
image

dsidorenkovas-rgbconsulting:
Hi,

I made the requested changes. Let me know if there is anything else to be done.

kittiu:
@dsidorenkovas-rgbconsulting this module does make test in other module failed with this,
```
You cannot confirm a purchase order if the supplier is not approved.
```
To coexists this module with others, is it ok to default state = 'approved' ? Will it still make sense?
Otherwise, I am not sure how this module can be merged...

geluboros:
Hi @kittiu,

You are right, this module cannot coexist in this repository because it's making all the tests fail.

I think it's important to maintain "Pending Approval" as a default state, so maybe a solution could be to not block purchase confirmations for suppliers in this state.
We could add a system parameter flag that allows to configure if the pending state is a blocking state or not (False by default). Ofc, this should be well documented in the module's Readme to avoid any confusion.

What do you think about this approach? If you think it could work, let us know so we can apply the changes, otherwise this PR can be closed.

Regards,
Gelu

kittiu:
Yes it is good opinion IMHO. Make it configurable :)

dsidorenkovas-rgbconsulting:
Hi @kittiu,
I made the proposed changes. Also, I added another commit for child contacts managment.

PR #746, opened 2019-09-11
Add module to facilitate purchasing product on weight but managed on other u.m.

sergiocorato:
This module aim to avoid creating many uom conversion measure using a field already present in product (weight) when conversion is only needed for purchase reason (e.g. iron stuff ordered in meters and purchased in kgs).
This is done computing unit price of purchase on price of purchase in sellers * unit weight of product in product (is added a boolean in product to enable the computation, and shown a 'kg' near the price on sellers to clarify to the user).


sergiocorato:
strange error on travis https://travis-ci.org/OCA/purchase-workflow/jobs/583629584 :confused:

pedrobaeza:
Seems a Travis glitch. Restarted.

sergiocorato:
> Seems a Travis glitch. Restarted.

thanks, runbot too sounds strange (time execution 8ms)

pedrobaeza:
It has a warning, but logs are not accessible. I have launched a rebuild.

sergiocorato:
> It has a warning, but logs are not accessible. I have launched a rebuild.

Runbot was normal after rebuild, but red, even if no errors was in log. I tried a commit to force another build, but now logs are not accessible and red anyway.

pedrobaeza:
We finally have the log, and the warning is:
```
2019-09-12 07:24:49,430 187 WARNING openerp_test odoo.models: @onchange('product_id', 'product_id.weight', 'product_qty') parameters must be field names
```

sergiocorato:
> We finally have the log, and the warning is:
> 
> ```
> 2019-09-12 07:24:49,430 187 WARNING openerp_test odoo.models: @onchange('product_id', 'product_id.weight', 'product_qty') parameters must be field names
> ```

thanks, so i need to create a related field to this field?

purchase_product_weight/models/purchase.py
```diff
@@ -0,0 +1,58 @@
...
+
+class PurchaseOrderLine(models.Model):
+    _inherit = 'purchase.order.line'
+
+    weight_total = fields.Float(string='Total weight')
+    product_weight = fields.Float(
+        related='product_id.weight',
+        string='Product Weight')
+
+    @api.onchange('product_id', 'product_weight', 'product_qty')
```
kittiu: I am just wondering why do we need to have product_weight here at all?
At the time of onchange, there shouldn't be change of changing the weight, or is it?
(I think normally, changing of weight can be for computed field)

purchase_product_weight/models/purchase.py
```diff
@@ -0,0 +1,58 @@
...
+    def _compute_total_weight(self):
```
kittiu: For onchange, the method should named "_onchange_xxx" ? (as coding guideline)

purchase_product_weight/models/purchase.py
```diff
@@ -0,0 +1,58 @@
...
+        for line in self:
+            line.weight_total = line.product_id.weight * \
+                line.product_qty
+
+    @api.onchange('product_qty', 'product_uom', 'weight_total')
+    def _onchange_quantity(self):
```
kittiu, line 18: Overall, I think this part of code is quite complicate, I can't understand it easily.
Can make as sub methods call with some description on it for easy reading?

kittiu: I..e, a routing for date_planned, and a routing for price_unit

purchase_product_weight/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,17 @@
...
+* add a Seller with a price.
+
+Then create a purchase order with the product created, unit price will be
+computed with: 'seller price' * 'product weight'.
+
+As weight can variate, a computed field 'total weight' will be added to
+purchase order line, and if this value is changed the unit price will be
+re-computed on 'total weight' / 'product qty'.
+
+In this way, all other logic remains untouched, but the user can have a simply
```
kittiu:
```suggestion
In this way, all other logic remains untouched, but the user can have a simple
```

purchase_product_weight/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,17 @@
...
+solution to manage product with price on weight but with a different u.m. on
```
kittiu:
```suggestion
solution to manage product with price on weight but with a different unit of measure on
```
Note: IMO, u.m. abbrev is not common, at least on my country. May be we can use the full word instead?

kittiu:
AFAIK, I think the name of this PR should be, [10.0][ADD] purchase_product_weight
and the branch could name 10.0-add-purchase_product_weight, though these are not strict, but can be good for readability.

By the way, this is actually ADD module or IMP? If this is an ADD, the version in manifest should be 10.0.1.0.0

After all, thank you for your contribution. :+1:

purchase_product_weight/models/purchase.py
sergiocorato: right, done!

purchase_product_weight/models/purchase.py
sergiocorato: Ok, I removed date_planned, which is already computed in original onchange called with super() (the function is copied-pasted and modified from the original one)

purchase_product_weight/models/purchase.py
sergiocorato: I removed it, thanks!
(I would like to have a computed field which user can modify by hand and if user goes on product from PO and alter the weight, the field was recomputed, but I don't know if it is possible)

sergiocorato:
Superseeded by #747 
Thanks @kittiu !

PR #748, opened 2019-09-17
12.0 product supplierinfo discount

krunal267:


pedrobaeza:
This is included now in `purchase_discount` module.

PR #747, opened 2019-09-16
[10.0][ADD] purchase product weight new module

sergiocorato:

PR #749, opened 2019-09-18
[12.0][ADD] New module purchase product weight

sergiocorato:
This module ease the use of product managed on any unit of measure but purchased on weight, without adding multiple conversion, but simply using weight on product. A future extension can be use of volume too.


sergiocorato:
migrated from #747

purchase_product_weight/models/product.py
```diff
@@ -0,0 +1,11 @@
+# Copyright 2019 Sergio Corato <https://github.com/sergiocorato>
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+from odoo import models, fields
```
Trivedi-Vacha-SerpentCS: Import should be in alphabetical order

purchase_product_weight/models/product_supplierinfo.py
```diff
@@ -0,0 +1,11 @@
+# Copyright 2019 Sergio Corato <https://github.com/sergiocorato>
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+from odoo import models, fields
```
Trivedi-Vacha-SerpentCS: Same as above

purchase_product_weight/models/purchase.py
```diff
@@ -0,0 +1,87 @@
+# Copyright 2019 Sergio Corato <https://github.com/sergiocorato>
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+from odoo import models, fields, api
```
Trivedi-Vacha-SerpentCS: Same as above

purchase_product_weight/models/purchase.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class PurchaseOrderLine(models.Model):
+    _inherit = 'purchase.order.line'
+
+    weight_total = fields.Float(string='Total weight')
```
Trivedi-Vacha-SerpentCS: Good to define help on the field.

purchase_product_weight/models/product.py
sergiocorato: done

purchase_product_weight/models/purchase.py
sergiocorato: done

purchase_product_weight/models/product_supplierinfo.py
sergiocorato: done, thanks

purchase_product_weight/models/purchase.py
sergiocorato: done

purchase_product_weight/models/purchase.py
```diff
@@ -0,0 +1,92 @@
...
+        help='Total weight computed from product weight * product quantity. '
+             'It can be written manually, in this case unit price will be '
+             'recomputed on Total weight / product quantity * seller price '
+             'if field "Compute price on weight" on product is flagged.')
+
+    @api.onchange('product_id', 'product_qty')
+    def _onchange_total_weight(self):
+        for line in self:
+            line.weight_total = line.product_id.weight * \
+                line.product_qty
```
ps-tubtim: I think this function's name is `_onchange_weight_total` better than`_onchange_total_weight`. Do you think so?

sergiocorato: Right, done

PR #757, opened 2019-10-03
[12.0][FIX]allocation creation when min_qty is greater than qty requested

AaronHForgeFlow:
Basically, allocation system fails when the minimum ordered qty is greater than the qty requested. With the current code we are creating allocations for the minimum order quantity, which is incorrect.

Besides that, I have also give creation and writing permissions to stock user to the allocations model and the purchase request line model. Other wise a basic warehouse user would not be able to receive the purchase orders. The new access control list will not give them menu access to purchase request, so I think security is still consistent.

Ping: @hveficent

### DIFF
--- a/purchase_request/security/ir.model.access.csv
+++ b/purchase_request/security/ir.model.access.csv
@@ -9,6 +9,6 @@ access_purchase_order_line_user,access.purchase.order.line.user,model_purchase_o
 access_purchase_order_manager,access.purchase.order.manager,model_purchase_order,purchase_request.group_purchase_request_manager,1,0,0,0
 access_purchase_order_line_manager,access.purchase.order.line.manager,model_purchase_order_line,purchase_request.group_purchase_request_manager,1,0,0,0
 access_purchase_request_allocation_manager,purchase.request.allocation,model_purchase_request_allocation,group_purchase_request_manager,1,1,1,1
-access_purchase_request_allocation_user,purchase.request.allocation,model_purchase_request_allocation,base.group_user,1,0,0,0
-access_purchase_request_line_stock_user,purchase.request.line.stock,model_purchase_request_line,stock.group_stock_user,1,0,0,0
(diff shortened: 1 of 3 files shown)


AaronHForgeFlow:
Tests for the failing use case added.

purchase_request/security/ir.model.access.csv
```diff
@@ -9,6 +9,6 @@ access_purchase_order_line_user,access.purchase.order.line.user,model_purchase_o
 access_purchase_order_manager,access.purchase.order.manager,model_purchase_order,purchase_request.group_purchase_request_manager,1,0,0,0
 access_purchase_order_line_manager,access.purchase.order.line.manager,model_purchase_order_line,purchase_request.group_purchase_request_manager,1,0,0,0
 access_purchase_request_allocation_manager,purchase.request.allocation,model_purchase_request_allocation,group_purchase_request_manager,1,1,1,1
-access_purchase_request_allocation_user,purchase.request.allocation,model_purchase_request_allocation,base.group_user,1,0,0,0
-access_purchase_request_line_stock_user,purchase.request.line.stock,model_purchase_request_line,stock.group_stock_user,1,0,0,0
+access_purchase_request_allocation_user,purchase.request.allocation,model_purchase_request_allocation,base.group_user,1,1,1,0
+access_purchase_request_line_stock_user,purchase.request.line.stock,model_purchase_request_line,stock.group_stock_user,1,1,1,0
```
MiquelRForgeFlow, line 13: to revert again in v13

MiquelRForgeFlow (approved):

HviorForgeFlow (approved):

purchase_request/security/ir.model.access.csv
AaronHForgeFlow: For the bless of the new compute fields calculation :hugs:

PR #764, opened 2019-10-15
[12.0][IMP] purchase_invoice_plan, add hook function to get plan_qty

kittiu:
Make hook function (prepare for new addon purchase_invoice_plan_deposit)
@hveficent


HviorForgeFlow (approved):
FF

PR #767, opened 2019-10-17
[12.0][ADD] purchase_requisition_tier_validation

ps-tubtim:


HviorForgeFlow:
Functional Review:
I'm only block to move forward from states when I go from Bid Selection to Closed.
I'm not so sure if this is the purpose of the module. How many validations you want to have?

Tier Validation Engine searchs for next state on state_to, and then check if it comes from a state_from to trigger the validation block.

ps-tubtim:
@hveficent Thank you for review. I go along with your comment.

> I'm only block to move forward from states when I go from Bid Selection to Closed.

purchase_requisition_tier_validation/models/purchase_requisition.py
```diff
@@ -0,0 +1,11 @@
+# Copyright 2019 Ecosoft Co., Ltd. (http://ecosoft.co.th)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class PurchaseRequisition(models.Model):
+    _name = 'purchase.requisition'
+    _inherit = ['purchase.requisition', 'tier.validation']
+    _state_from = ['open']
```
HviorForgeFlow:
```suggestion
    _state_from = ['draft']
```

purchase_requisition_tier_validation/models/purchase_requisition.py
```diff
@@ -0,0 +1,11 @@
...
+    _state_to = ['done']
```
HviorForgeFlow:
```suggestion
    _state_to = ['ongoing', 'in_progress']
```

HviorForgeFlow (changes requested):
I think you misunderstand me. 
First of all, Where would be required to place the validation?
I guess it should be done like this:
- In case of Blanket Order Agreement: from Draft to Ongoing
- In case of Call for Tenders: from draft to Confirmed

kittiu:
@ps-tubtim is final? You need to do as per change request first?

ps-tubtim:
@hveficent Sorry, I misunderstand. Thank you for review.
@kittiu Thank you, I don't receive notification this pr.

HviorForgeFlow (approved):
LGTM :+1: 

Not sure about Call for Tenders workflow, not familiarized with it.

purchase_requisition_tier_validation/tests/test_purchase_requisition_tier_validation.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2019 Ecosoft Co., Ltd. (http://ecosoft.co.th)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.tests import common
+
+
+class TestHrExpenseTierValidation(common.TransactionCase):
+    def setUp(self):
+        super(TestHrExpenseTierValidation, self).setUp()
```
LoisRForgeFlow, line 9: Incorrect name

LoisRForgeFlow (approved):
LGTM, very minor comment below.

PR #768, opened 2019-10-18
[12.0][FIX] Purchase Request - Fix singleton error

CasVissers-360ERP:
Fix for singleton error: 

 File "/opt/odoo/customer/community/purchase_request/models/purchase_order.py", line 197, in write
    prev_qty_received = self.qty_received
  File "/opt/odoo/odoo-server/odoo/fields.py", line 987, in __get__
    record.ensure_one()
  File "/opt/odoo/odoo-server/odoo/models.py", line 4724, in ensure_one
    raise ValueError("Expected singleton: %s" % self)
ValueError: Expected singleton: purchase.order.line(89, 90, 91, 92, 93)


JordiBForgeFlow:
@aheficent

HviorForgeFlow:
Hello @CasVissers, can you provide the use case when this error is raised? It would help to add a test and verify that workflow keeps correct after this changes. Thanks!

purchase_request/models/purchase_order.py
```diff
@@ -194,8 +194,9 @@ def _prepare_request_message_data(
     def write(self, vals):
         #  it is done here instead of method _update_received_qty
         #  to make sure this work for services
-        prev_qty_received = self.qty_received
-        res = super(PurchaseOrderLine, self).write(vals)
-        if vals.get('qty_received', False):
-            self.update_service_allocations(prev_qty_received)
+        for line in self:
+            prev_qty_received = line.qty_received
+            res = super(PurchaseOrderLine, line).write(vals)
```
astirpe, line 199: I think the super() should be moved outside the loop and can be called only once, over self instead of each line.
As @hveficent  said, if you provide a use case or the steps to reproduce the issue, we can have a better look at this issue.

astirpe (changes requested):

purchase_request/models/purchase_order.py
kittiu: I got the same error while writing other addon that some how calling purchase_lines.write(...)
Will this works?
```
    @api.multi
    def write(self, vals):
        #  it is done here instead of method _update_received_qty
        #  to make sure this work for services
        res = super(PurchaseOrderLine, self).write(vals)
        if vals.get('qty_received', False):
            for line in self:
                prev_qty_received = line.qty_received
                line.update_service_allocations(prev_qty_received)
        return res
```

HviorForgeFlow: Much better, @CasVissers apply this.

HviorForgeFlow:
ping @CasVissers

PR #774, opened 2019-10-24
[12.0][ADD] purchase_work_acceptance

ps-tubtim:
This module work about work acceptance creation on Purchase Order.

Process:
1. Create Purchase Order normally
2. After Confirm Order (Status = 'Purchase'), Click button 'Create WA'
3. Accept order


purchase_work_acceptance/models/work_acceptance.py
```diff
@@ -0,0 +1,166 @@
...
+    )
+    origin = fields.Char(
+        string='Source Document',
+        copy=False,
+        help="Reference of the document that generated this work acceptance."
+    )
+    date_due = fields.Datetime(
+        string='Due Date',
+        required=True,
+        index=True,
```
kittiu: Normally, we will need index for the joining field, ie.., Many2one from other, which expect to have SQL Join, for better performance.
For normal filed, mostly we don't need index. (too many index is also not good, it make system slow own create new record).

purchase_work_acceptance/models/work_acceptance.py
```diff
@@ -0,0 +1,166 @@
...
+        copy=False,
+    )
+    partner_id = fields.Many2one(
+        comodel_name='res.partner',
+        string='Vendor',
+        required=True,
+        change_default=True,
+        track_visibility='always',
+    )
+    acceptor_id = fields.Many2one(
```
kittiu:
```suggestion
    responsible_id = fields.Many2one(
```

purchase_work_acceptance/models/work_acceptance.py
```diff
@@ -0,0 +1,166 @@
...
+        comodel_name='res.partner',
```
kittiu:
```suggestion
        comodel_name='hr.employee',
```

kittiu: Normally, it should be employee, except you really need partner for some reason.

kittiu:
Some design questions:
- Why do you need to split file into work_acceptance.py and work_acceptance_line.py, also those xml files also split. Work acceptance line seem to have separated view, but I don't see any menu for "Work Acceptance Line". (and I don't think we need it).
- WA lines seem to have too many fields. I think we can only focus on quantity. Because WA objective is to control the acceptance quantity (and so to make sure IN do not accept more than WA granted). (also, it will be much easier if we do not touch on price/taxes)

ps-tubtim:
@kittiu Ready for review.

purchase_work_acceptance/models/stock_picking.py
```diff
@@ -0,0 +1,38 @@
...
+
+    @api.multi
+    def button_validate(self):
+        self.ensure_one()
+        if self.wa_id:
+            wa_lines = self.env['work.acceptance.line']
+            for move in self.move_ids_without_package:
+                id = move.purchase_line_id.id
+                wa_line = wa_lines.search([('purchase_line_id', '=', id)])
+                if move.quantity_done > wa_line.product_qty:
```
kittiu: Let's compare for equal.
```suggestion
                if move.quantity_done != wa_line.product_qty:
```
But please use float_compare

purchase_work_acceptance/models/stock_picking.py
```diff
@@ -0,0 +1,38 @@
...
+                    raise ValidationError(_('You cannot validate a transfer if'
+                                          ' Done more than accepted quantity'))
```
kittiu:
```suggestion
                                          ' done quantity is more than accepted quantity'))
```

kittiu:
## Key points

* I think we should have purchase configuration for Work Acceptance now.
  * - [ ] Enable work acceptance process
    * Shows WA Reference in Vendor Bill / Incoming Shipment (IN only) and in Wizard...but not required.
  * - [ ] Enforce work acceptance on every purchase (show only if above is selected)
     * Make required WA fields on Wizard, and Invoice and Picking, when they are created from purchase order (this we if invoice/picking is created directly, required=False, invisible=False)

Note: If not enable, system will work just like there is no WA at all (no menu, no field, no check)

---
* I think it is a good idea now, to use WA to control exactly both Quantity and Price Unit (we was checking  only quantity). (I see that, you are passing both quantity and price unit, which is good).
  * Let's throw Error when,
    * Product's Quantity and Price is not match with WA
    * Picking / Invoice has some product lines that is not matched in WA (ie.., manually created line)
---
* You seem to use WA's line  to create Invoice Lines (this is incorrect)
  * In fact, this WA should not overwrite how invoice line is created by PO (if no WA selected, system should works as normal)
  * WA Line should just help changing the quantity and price unit and nothing else.

In other words,
* Create Bill without selecting "WA", the Billing has no line !!!
  * If WA is not selected, it should bypass WA and work as normal.
---
* A lot of code can be written differently (I have more detailed comment), but generally,
  * Many field store=False or removed.
  * Many code can be removed or shorten.

purchase_work_acceptance/__manifest__.py
```diff
@@ -0,0 +1,29 @@
...
+{
+    'name': 'Work Acceptance',
+    'version': '12.0.1.0.0',
+    'category': 'Purchase Management',
+    'author': 'Ecosoft, '
+              'Odoo Community Association (OCA)',
+    'license': 'AGPL-3',
+    'website': 'https://github.com/OCA/purchase-workflow',
+    'depends': [
+        'hr',
```
kittiu: Why do we need hr?

purchase_work_acceptance/models/account_invoice.py
```diff
@@ -0,0 +1,169 @@
...
+from odoo.tools.float_utils import float_compare
+
+
+class AccountInvoice(models.Model):
+    _inherit = 'account.invoice'
+
+    wa_id = fields.Many2one(
+        comodel_name='work.acceptance',
+        string='WA Reference',
+        domain=[('state', '=', 'accept')],
```
kittiu: List only WA of only PO we used to create this invoice. Not all PO's WA.

purchase_work_acceptance/models/account_invoice.py
```diff
@@ -0,0 +1,169 @@
...
+        store=True,
```
kittiu: Why store=True? Is this computed?

kittiu: Add some help as following (you can use this help to all wa_id field, it will help user knows what it is used for).
```
To control quantity and unit price of the vendor bill, to be according to the quantity and unit price of the work acceptance.
```

purchase_work_acceptance/models/account_invoice.py
```diff
@@ -0,0 +1,169 @@
...
+    )
+
+    @api.multi
+    def action_invoice_open(self):
+        self.ensure_one()
```
kittiu: Can we not use self.ensure_one() ?
If you use ensure_one(), it will break the existing action_invoice_open(), there must be some place that been using this as multi.

purchase_work_acceptance/models/account_invoice.py
```diff
@@ -0,0 +1,169 @@
...
+        if self.wa_id:
+            wa_lines = self.env['work.acceptance.line']
+            for line in self.invoice_line_ids:
+                id = line.purchase_line_id.id
+                wa_line = wa_lines.search([('purchase_line_id', '=', id)])
+                if line.quantity > wa_line.product_qty:
```
kittiu: Let's check for "=", not ">".
And let's check both quantity and price_unit, shall we?

purchase_work_acceptance/models/account_invoice.py
```diff
@@ -0,0 +1,169 @@
...
+                    raise ValidationError(_('You cannot validate a bill if '
+                                          'Quantity more than accepted quantity'))
+        return super(AccountInvoice, self).action_invoice_open()
+
+    def _prepare_invoice_line_from_wa_line(self, line):
```
kittiu: From this point on, I think it need to be redo in different way. From what I see (correct if I am wrong), you are trying to use WA's line to create invoice line.

**WA line should not be used to create invoice line**

Invoice line should be created as as normal. But during it creation, just try to snap in and change the quantity and price_unit as in WA.

WA Line has purchase_line_id, and invoice_line also has purchase_line_id, so it should be easy to do so. In the method _prepare_invoice_line_from_po_line() which is now used to prepare invoice line, inherit it, it is easy to change the quantity and price_unit.

**Note:** it is super important that, the addon we create, do not **replace** the existing functionality, but only **extend** it.

purchase_work_acceptance/models/account_invoice.py
```diff
@@ -0,0 +1,169 @@
...
+        return data
+
+    def _onchange_product_id(self):
+        domain = super(AccountInvoice, self)._onchange_product_id()
+        if self.wa_id:
+            self.uom_id = self.product_id.uom_po_id
+        return domain
+
+    @api.onchange('wa_id')
+    def wa_id_change(self):
```
kittiu: I think this method can be removed, given the above comment.

purchase_work_acceptance/models/account_invoice.py
```diff
@@ -0,0 +1,169 @@
...
+
+        self.invoice_line_ids += new_lines
+        self.payment_term_id = self.wa_id.purchase_id.payment_term_id
+        self.purchase_id = self.wa_id.purchase_id.id
+        self.origin = self.wa_id.purchase_id.name
+        self.env.context = dict(self.env.context, from_wa_id_change=True)
+        return {}
+
+    @api.onchange('currency_id')
+    def _onchange_currency_id(self):
```
kittiu: I think this method can be removed, given the above comment.

purchase_work_acceptance/models/account_invoice.py
```diff
@@ -0,0 +1,169 @@
...
+        if self.currency_id:
+            for line in self.invoice_line_ids.filtered(lambda r: r.wa_line_id):
+                date = self.date or self.date_invoice or fields.Date.today()
+                company = self.company_id
+                line.price_unit = line.wa_id.currency_id._convert(
+                    line.wa_line_id.price_unit, self.currency_id, company,
+                    date, round=False)
+
+    @api.model
+    def create(self, vals):
```
kittiu: I think this method can be removed, given the above comment.
Note: as we are not creating this invoice from WA. We still create invoice from purchase, just to adjust it a bit.

purchase_work_acceptance/models/account_invoice.py
```diff
@@ -0,0 +1,169 @@
...
+            message = _(
+                "This vendor bill has been created from: %s"
+                ) % (",".join(
+                    ["<a href=# data-oe-model=work.acceptance data-oe-id=" +
+                     str(line.id) + ">" + line.name + "</a>" for line in wa]))
+            invoice.message_post(body=message)
+        return invoice
+
+    @api.multi
+    def write(self, vals):
```
kittiu: I think we also don't need this method.

Note: if we still use exiting `def purchase_order_change()` to create invoice line as Odoo standard, I think we can just add onchange to `wa_id` just to trigger `purchase_order_change()` again. Almost no code we have to write.

purchase_work_acceptance/models/account_invoice.py
```diff
@@ -0,0 +1,169 @@
...
+    _inherit = 'account.invoice.line'
+
+    wa_line_id = fields.Many2one(
+        comodel_name='work.acceptance.line',
+        string='Work Acceptance Line',
+        ondelete='set null',
+        index=True,
+        readonly=True
+    )
+    wa_id = fields.Many2one(
```
kittiu: Do we need to create this field? 
Note: I see that there are many fields that also not needed (store=True fields) also in other model. Make sure that it is really necessary before we have it, then have it.

purchase_work_acceptance/models/purchase.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+class PurchaseOrder(models.Model):
+    _inherit = 'purchase.order'
+
+    wa_count = fields.Integer(
+        compute='_compute_wa',
+        string='WA count',
+        default=0,
+        store=True
```
kittiu: Do we need store=True? Is non-stored field enough?
Note: again, try not to store field if not necessary.

purchase_work_acceptance/models/purchase.py
```diff
@@ -0,0 +1,121 @@
...
+    )
+    wa_ids = fields.Many2many(
+        comodel_name='work.acceptance',
+        compute='_compute_wa',
```
kittiu: Coding standard
```suggestion
        compute='_compute_wa_ids',
```

BUT, why do you need to compute at all? There is field 'purchase_id' in work.acceptance already. Can we just create field like this?
```
wa_ids = fields.One2many(
  comodel_name='work.acceptance',
  inverse_name='purchase_id',
```

purchase_work_acceptance/models/purchase.py
```diff
@@ -0,0 +1,121 @@
...
+        string='Work Acceptances',
+        copy=False,
```
kittiu: I think normally, compute field do not need `copy`
Also, do we need to make a store=True?

purchase_work_acceptance/models/purchase.py
```diff
@@ -0,0 +1,121 @@
...
+        store=True,
+    )
+    wa_lines = fields.One2many(
```
kittiu: Coding standard
```suggestion
    wa_line_ids = fields.One2many(
```

purchase_work_acceptance/models/purchase.py
```diff
@@ -0,0 +1,121 @@
...
+        comodel_name='work.acceptance.line',
+        inverse_name='purchase_line_id',
+        string="WA Lines",
+        readonly=True,
+        copy=False
+    )
+
+    @api.depends('order_line.wa_lines.wa_id')
```
kittiu: We we don't need to store wa_count, we don't need depends.

purchase_work_acceptance/models/purchase.py
```diff
@@ -0,0 +1,121 @@
...
+    def _compute_wa(self):
+        for order in self:
+            wa = self.env['work.acceptance']
```
kittiu: You don't need to declare wa object everytime. You can just use something like,
```
order.wa_ids |= line.wa_lines.mapped('wa_id')
```

BUT, to find wa_count, why don't we just,
```
WA = self.env['work.acceptance']
for purchase in self:
  order.wa_count = WA.search_count([('purchase_id', '=', order.id)])
```

**AND ALSO, as in the above comment, we already have wa_ids, so just `len(order.wa_ids)` should do the job.**

Note: Try to write as short code as possible to do the same thing. Short code normally less buggy.

purchase_work_acceptance/models/purchase.py
```diff
@@ -0,0 +1,121 @@
...
+            'default_company_id': self.company_id.id,
+            'company_id': self.company_id.id,
+            'default_date_due': self.date_planned,
+            'default_wa_line_ids': [(0, 0, {
+                'purchase_line_id': line.id,
+                'name': line.name,
+                'product_uom': line.product_uom.id,
+                'product_id': line.product_id.id,
+                'price_unit': line.price_unit,
+                'product_qty': line.product_qty - line.qty_accepted,
```
kittiu: WA's suggested quantity should be, (please ready from redmine)
* Service case: Purchase Qty - Billed Qty
* Product case: Purchase Qty - Received Qty

And so, that's why I don't think `qty_accepted` is necessary.


**Note:** We only want to use WA as a helper to existing feature. In my opinion, WA's quantity is not that important. If you don't agree, please discuss.

purchase_work_acceptance/models/purchase.py
```diff
@@ -0,0 +1,121 @@
...
+            'default_partner_id': self.partner_id.id,
+            'default_partner_ref': self.partner_ref,
+            'default_currency_id': self.currency_id.id,
+            'default_company_id': self.company_id.id,
+            'company_id': self.company_id.id,
+            'default_date_due': self.date_planned,
+            'default_wa_line_ids': [(0, 0, {
+                'purchase_line_id': line.id,
+                'name': line.name,
+                'product_uom': line.product_uom.id,
```
kittiu, line 48: Do we need this?
```suggestion
```

purchase_work_acceptance/models/purchase.py
```diff
@@ -0,0 +1,121 @@
...
+                'product_id': line.product_id.id,
+                'price_unit': line.price_unit,
```
kittiu, line 50: It is ok to have price_unit now, as it can be used to enforce invoice price too.

purchase_work_acceptance/models/work_acceptance.py
```diff
@@ -0,0 +1,241 @@
...
+        readonly=True
+    )
+    price_unit = fields.Float(
+        string='Unit Price',
+        required=True,
+    )
+    price_subtotal = fields.Monetary(
+        compute='_compute_amount',
+        string='Subtotal',
+        store=True
```
kittiu: Do we need to store it?

purchase_work_acceptance/models/purchase.py
```diff
@@ -0,0 +1,121 @@
...
+                'product_qty': line.product_qty - line.qty_accepted,
+                'price_subtotal': line.price_subtotal,
```
kittiu: It is a compute field, we don't need to assign value.
```suggestion
```
Overall, I think this method can be shorter than too.

purchase_work_acceptance/models/work_acceptance.py
```diff
@@ -0,0 +1,241 @@
...
+    _description = 'Work Acceptance'
+    _order = 'id desc'
+
+    name = fields.Char(
+        required=True,
+        index=True,
+        copy=False,
+        default='New'
+    )
+    origin = fields.Char(
```
kittiu: We don't need this. We already have `purchase_id` to as source document.
```suggestion
```

purchase_work_acceptance/models/purchase.py
```diff
@@ -0,0 +1,121 @@
...
+
+    @api.multi
+    def action_view_wa(self):
+        act = self.env.ref('purchase_work_acceptance.action_work_acceptance')
+        result = act.read()[0]
+        create_wa = self.env.context.get('create_wa', False)
+        result['context'] = {
+            'default_purchase_id': self.id,
+            'default_partner_id': self.partner_id.id,
+            'default_partner_ref': self.partner_ref,
```
kittiu: I think we don't need this too.
```suggestion
```

purchase_work_acceptance/models/purchase.py
```diff
@@ -0,0 +1,121 @@
...
+            'default_currency_id': self.currency_id.id,
```
kittiu: I saw in field currency_id, you already have default value for it. So, this wouldn't need.
```suggestion
```

purchase_work_acceptance/models/purchase.py
```diff
@@ -0,0 +1,121 @@
...
+            'default_company_id': self.company_id.id,
```
kittiu, line 42: I saw in field company_id, you already have default value for it. So, this wouldn't need.
```suggestion
            'default_company_id': self.company_id.id,
```
Note: and also, why we have both `default_company_id` and `company_id` here?

purchase_work_acceptance/models/purchase.py
```diff
@@ -0,0 +1,121 @@
...
+            for line in order.order_line:
+                wa |= line.wa_lines.mapped('wa_id')
+            order.wa_ids = wa
+            order.wa_count = len(wa)
+
+    @api.multi
+    def action_view_wa(self):
+        act = self.env.ref('purchase_work_acceptance.action_work_acceptance')
+        result = act.read()[0]
+        create_wa = self.env.context.get('create_wa', False)
```
kittiu, line 38: Only if create_wa == True, that you will need default values in result['context'], right? Should it be in the if clause some how?

purchase_work_acceptance/models/work_acceptance.py
```diff
@@ -0,0 +1,241 @@
...
+        copy=False,
+        default=fields.Datetime.now,
+    )
+    date_accept = fields.Datetime(
+        string='Accepted Date',
+        required=True,
+        copy=False,
+        default=fields.Datetime.now,
+    )
+    partner_ref = fields.Char(
```
kittiu: Not sure why we need this.
```suggestion
```

purchase_work_acceptance/models/purchase.py
```diff
@@ -0,0 +1,121 @@
...
+        }
+        if len(self.wa_ids) > 1 and not create_wa:
+            result['domain'] = "[('id', 'in', " + str(self.wa_ids.ids) + ")]"
+        else:
+            res = self.env.ref(
+                'purchase_work_acceptance.view_work_acceptance_form', False)
+            result['views'] = [(res and res.id or False, 'form')]
+            if not create_wa:
+                result['res_id'] = self.wa_ids.id or False
+        result['context']['default_origin'] = self.name
```
kittiu: We need this field?
```suggestion
```

purchase_work_acceptance/models/purchase.py
```diff
@@ -0,0 +1,121 @@
...
+        result['context']['default_reference'] = self.partner_ref
```
kittiu: We need this field?
```suggestion
```

purchase_work_acceptance/models/purchase.py
```diff
@@ -0,0 +1,121 @@
...
+                'view_id': wizard.id,
+                'target': 'new',
+            }
+        return super(PurchaseOrder, self).action_view_invoice()
+
+
+class PurchaseOrderLine(models.Model):
+    _inherit = 'purchase.order.line'
+
+    qty_accepted = fields.Float(
```
kittiu: After some though, I think we can consider removing this field. Please discuss if you see otherwise.
My reasons,
* WA is used only to control the exact quantity in invoice and picking.
* WA can be created and accepted, but never used by invoice or picking. (we intend to make it flexible).
* Removing this field, don't make any problem in our use case (do we?)

purchase_work_acceptance/models/purchase.py
```diff
@@ -0,0 +1,121 @@
...
+        compute='_compute_qty_accepted',
+        string="Accepted Qty",
+        default=0.0,
+        digits=dp.get_precision('Product Unit of Measure'),
+        store=True,
+    )
+    wa_lines = fields.One2many(
```
kittiu: Coding standard
```suggestion
    wa_line_ids = fields.One2many(
```

purchase_work_acceptance/models/purchase.py
```diff
@@ -0,0 +1,121 @@
...
+        comodel_name='work.acceptance.line',
+        inverse_name='purchase_line_id',
+        string="WA Lines",
+        readonly=True,
+        copy=False
+    )
+
+    @api.depends('wa_lines.wa_id.state', 'wa_lines.product_qty')
+    def _compute_qty_accepted(self):
```
kittiu: We may don't need this, if we agree not to have this `qty_accepted`

purchase_work_acceptance/models/stock_picking.py
```diff
@@ -0,0 +1,38 @@
...
+        comodel_name='work.acceptance',
+        string='WA Reference',
+        domain=[('state', '=', 'accept')],
+    )
+
+    @api.multi
+    def button_validate(self):
+        self.ensure_one()
+        if self.wa_id:
+            wa_lines = self.env['work.acceptance.line']
```
kittiu: Coding standard, always use camel case (capital letter), when declare Class.
```suggestion
            WALine = self.env['work.acceptance.line']
```

purchase_work_acceptance/models/stock_picking.py
```diff
@@ -0,0 +1,38 @@
...
+            for move in self.move_ids_without_package:
+                id = move.purchase_line_id.id
+                wa_line = wa_lines.search([('purchase_line_id', '=', id)])
```
kittiu: Can we have more than 1 wa_line with same `purchase_line_id` ??? This can cause a bug on back orders as it start to refer to > 1 wa_lines.

NOTE: now I can see that, there is no direct link between stock move and wa line, since stock move is created before WA. Lets do this,
* Don't worry about the link. 
* Just sum quantity by product_id, and just ensure that both WA and Picking has equal product's quantities.
* This will affect _onchange_wa_id() too. You need to rewrite the done quantity using this concept.
  * Just smartly fill in the done quantity by products
  * I.e., even if purchase order have 1 product line with 10 qty, picking can have 10 of the same product, and 1 qty each and be valid.

purchase_work_acceptance/models/stock_picking.py
```diff
@@ -0,0 +1,38 @@
...
+        domain=[('state', '=', 'accept')],
+    )
+
+    @api.multi
+    def button_validate(self):
+        self.ensure_one()
+        if self.wa_id:
+            wa_lines = self.env['work.acceptance.line']
+            for move in self.move_ids_without_package:
+                id = move.purchase_line_id.id
```
kittiu: Use meaningful variable.
```suggestion
                po_line_id = move.purchase_line_id.id
```

purchase_work_acceptance/models/stock_picking.py
kittiu: Note for general practice.
* Beware of number of records can be on the result set, and make the variable name explicit. Normally, search will result in multi.
* If you want to use search(), and you expect the result to have len() = 1. Always use .ensure_one() after it. Or use search with limit = 1.

purchase_work_acceptance/models/stock_picking.py
```diff
@@ -0,0 +1,38 @@
...
+                    raise ValidationError(_('You cannot validate a transfer if'
+                                          ' Done more than accepted quantity'))
+        return super(Picking, self).button_validate()
+
+    @api.onchange('wa_id')
+    def wa_id_change(self):
+        if self.wa_id:
+            po_lines = self.env['purchase.order.line']
+            for move in self.move_ids_without_package:
+                id = move.purchase_line_id.id
```
kittiu:
```suggestion
                po_line_id = move.purchase_line_id.id
```

purchase_work_acceptance/models/stock_picking.py
```diff
@@ -0,0 +1,38 @@
...
+                wa_line = wa_lines.search([('purchase_line_id', '=', id)])
+                if move.quantity_done > wa_line.product_qty:
+                    raise ValidationError(_('You cannot validate a transfer if'
+                                          ' Done more than accepted quantity'))
+        return super(Picking, self).button_validate()
+
+    @api.onchange('wa_id')
+    def wa_id_change(self):
+        if self.wa_id:
+            po_lines = self.env['purchase.order.line']
```
kittiu:
```suggestion
            PurchaseLine = self.env['purchase.order.line']
```

purchase_work_acceptance/models/stock_picking.py
```diff
@@ -0,0 +1,38 @@
...
+            for move in self.move_ids_without_package:
+                id = move.purchase_line_id.id
+                po_line = po_lines.search([('id', '=', id)])
+                qty_limit = po_line.qty_accepted - po_line.qty_received
+                move.quantity_done = qty_limit
```
kittiu: NOTE: base on above comment about no link between po line and stock move, please rewrite this onchange.

purchase_work_acceptance/models/work_acceptance.py
```diff
@@ -0,0 +1,241 @@
...
+    )
+    date_due = fields.Datetime(
+        string='Due Date',
+        required=True,
+        copy=False,
+    )
+    date_receive = fields.Datetime(
+        string='Received Date',
+        required=True,
+        copy=False,
```
kittiu: If you have both required=True, and copy=False, you will get error immediatelly when copy.
Please fix the same for other fields.
```suggestion
```

purchase_work_acceptance/models/work_acceptance.py
```diff
@@ -0,0 +1,241 @@
...
+    def create(self, vals):
+        if vals.get('name', 'New') == 'New':
+            vals['name'] = self.env['ir.sequence'].next_by_code(
+                'work.acceptance') or '/'
+        return super(WorkAcceptance, self).create(vals)
+
+    @api.multi
+    def button_accept(self, force=False):
+        self.write({'state': 'accept'})
+        return {}
```
kittiu:
```suggestion
```

purchase_work_acceptance/models/work_acceptance.py
```diff
@@ -0,0 +1,241 @@
...
+
+    @api.multi
+    def button_draft(self):
+        self.write({'state': 'draft'})
+        return {}
```
kittiu:
```suggestion
```

purchase_work_acceptance/models/work_acceptance.py
```diff
@@ -0,0 +1,241 @@
...
+
+    @api.multi
+    def button_cancel(self):
+        self.write({'state': 'cancel'})
+        return {}
```
kittiu:
```suggestion
```

purchase_work_acceptance/models/work_acceptance.py
```diff
@@ -0,0 +1,241 @@
...
+    @api.model
+    def create(self, vals):
+        if vals.get('name', 'New') == 'New':
+            vals['name'] = self.env['ir.sequence'].next_by_code(
+                'work.acceptance') or '/'
+        return super(WorkAcceptance, self).create(vals)
+
+    @api.multi
+    def button_accept(self, force=False):
+        self.write({'state': 'accept'})
```
kittiu:
```suggestion
        self.write({'state': 'accept',
                         'date_accept': fields.Datetime.now()})
```

purchase_work_acceptance/models/work_acceptance.py
```diff
@@ -0,0 +1,241 @@
...
+        string='Received Date',
+        required=True,
+        copy=False,
+        default=fields.Datetime.now,
+    )
+    date_accept = fields.Datetime(
+        string='Accepted Date',
+        required=True,
+        copy=False,
+        default=fields.Datetime.now,
```
kittiu: Let's don't default this, and have it set when WA is accepted, ok?
```suggestion
```

purchase_work_acceptance/models/work_acceptance.py
```diff
@@ -0,0 +1,241 @@
...
+    )
+    product_uom = fields.Many2one(
+        comodel_name='uom.uom',
+        string='Product Unit of Measure',
+    )
+    product_id = fields.Many2one(
+        comodel_name='product.product',
+        string='Product',
+    )
+    product_type = fields.Selection(
```
kittiu: Do we need this?

purchase_work_acceptance/models/work_acceptance.py
```diff
@@ -0,0 +1,241 @@
...
+
+    name = fields.Text(
+        string='Description',
+        required=True
+    )
+    product_qty = fields.Float(
+        string='Quantity',
+        required=True,
+    )
+    product_uom = fields.Many2one(
```
kittiu: Do we need this?

purchase_work_acceptance/models/work_acceptance.py
```diff
@@ -0,0 +1,241 @@
...
+        store=True
+    )
+    wa_id = fields.Many2one(
+        comodel_name='work.acceptance',
+        string='WA Reference',
+        index=True,
+        required=True,
+        ondelete='cascade'
+    )
+    company_id = fields.Many2one(
```
kittiu: Do we need this?

purchase_work_acceptance/models/work_acceptance.py
```diff
@@ -0,0 +1,241 @@
...
+        comodel_name='res.company',
+        related='wa_id.company_id',
+        string='Company',
+        store=True,
+        readonly=True
+    )
+    state = fields.Selection(
```
kittiu: Do we need this?

purchase_work_acceptance/models/work_acceptance.py
```diff
@@ -0,0 +1,241 @@
...
+        related='wa_id.state',
+        store=True,
+        readonly=False
+    )
+
```
kittiu:
```suggestion
```

purchase_work_acceptance/models/work_acceptance.py
```diff
@@ -0,0 +1,241 @@
...
+        readonly=True,
+    )
+    purchase_line_id = fields.Many2one(
+        comodel_name='purchase.order.line',
+        string='Purchase Order Line',
+        ondelete='set null',
+        index=True,
+        readonly=True,
+    )
+    purchase_id = fields.Many2one(
```
kittiu: Need this?

purchase_work_acceptance/models/work_acceptance.py
```diff
@@ -0,0 +1,241 @@
...
+        comodel_name='purchase.order',
+        related='purchase_line_id.order_id',
+        string='Purchase Order',
+        store=False,
+        readonly=True,
+        related_sudo=False,
+    )
+
+    @api.depends('product_qty', 'price_unit')
```
kittiu: We we don't store, we won't need this,
```suggestion
```

purchase_work_acceptance/readme/USAGE.rst
```diff
@@ -0,0 +1,5 @@
+**Work Acceptance**
```
kittiu: Please add more detail about how to use this. We can also talk about invoice and picking.

purchase_work_acceptance/views/work_acceptance_views.xml
```diff
@@ -0,0 +1,175 @@
...
+                <field name="responsible_id"/>
+                <field name="company_id" groups="base.group_multi_company" options="{'no_create': True}"/>
+                <field name="origin"/>
+                <field name="currency_id" invisible="1"/>
+                <field name="state"/>
+            </tree>
+        </field>
+    </record>
+
+    <record id="view_work_acceptance_kanban" model="ir.ui.view">
```
kittiu: Let's remove kanban view for now. I think the tree view will do just fine now.
In future, we could have kanban, report, calendar and activity view also :)
Let's keep things simple, for this first version.

purchase_work_acceptance/views/work_acceptance_views.xml
```diff
@@ -0,0 +1,175 @@
...
+                                               <field name="price_unit"/>
+                                           </group>
+                                           <group>
+                                               <field name="company_id" groups="base.group_multi_company" options="{'no_create': True}"/>
+                                           </group>
+                                       </group>
+                                   </sheet>
+                               </form>
+                           </field>
+                       </page>
```
kittiu, line 61: Please add Chartter. I think it will be very useful. When people do acceptance, they can have conversation back and forth.
Please also learn how PO is writing messages in the chatter. You can learn from it and make a nice history.

purchase_work_acceptance/wizard/select_work_acceptance_wizard.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class SelectWorkAcceptanceWizard(models.TransientModel):
+    _name = 'select.work.acceptance.wizard'
+    _description = 'Select Work Acceptance Wizard'
+
+    wa_id = fields.Many2one(
+        comodel_name='work.acceptance',
+        string='Work Acceptance',
+        domain=[('state', '=', 'accept')],
```
kittiu: Must filter only WA of this PO
```suggestion
        domain=lambda self: [('state', '=', 'accept'), ('purchase_id', '=', self._context.get('active_id')],
```

purchase_work_acceptance/wizard/select_work_acceptance_wizard_views.xml
```diff
@@ -0,0 +1,23 @@
...
+        <field name="name">select_work_acceptance_wizard</field>
+        <field name="model">select.work.acceptance.wizard</field>
+        <field name="type">form</field>
+        <field name="arch" type="xml">
+             <form string="Select Work Acceptance">
+                 <group>
+                     <field name="wa_id" options="{'no_create': True}"/>
+                 </group>
+                 <footer>
+                     <button name="button_ok"
```
kittiu:
```suggestion
                     <button name="button_create_vendor_bill"
```

purchase_work_acceptance/wizard/select_work_acceptance_wizard_views.xml
```diff
@@ -0,0 +1,23 @@
...
+                             string="Ok" type="object"
```
kittiu:
```suggestion
                             string="Create Vendor Bill" type="object"
```

kittiu:
After fixing all this, I think it is good time **to do Test Script** now. From the redmine, I split into 2 cases. But I think you can also do it as one case, to make it shorter by having PO with both service and product. I would suggest, i.e.,
* 2 service line with different product.
* 3 product lines with 2 line same product, and one line different.

Then, do 2 WA. 1st to accept partial 1 service and 1 product (2 lines in total). 2nd to accept the rest. This should cover most cases.

kittiu:
Even I don't enable WA at all, I still see this button.
image

kittiu:
image

purchase_work_acceptance/models/stock_picking.py
```diff
@@ -0,0 +1,54 @@
...
+                move_product = Moves.search([
+                    ('picking_id', '=', self.id), ('product_id', '=', product.id)])
+                if sum(wa_product.mapped('product_qty')) != \
+                        sum(move_product.mapped('quantity_done')):
+                    raise ValidationError(_('You cannot validate a transfer if done'
+                                          ' quantity not equal accepted quantity'))
+        return super(Picking, self).button_validate()
+
+    @api.onchange('wa_id')
+    def wa_id_change(self):
```
kittiu: Not serious, but naming convention
```suggestion
    def _onchange_wa_id(self):
```

purchase_work_acceptance/tests/test_purchase_work_acceptance.py
```diff
@@ -0,0 +1,129 @@
...
+                        'product_qty': 12.0})]
+        })
+        work_acceptance.button_accept()
+        self.assertEqual(work_acceptance.state, 'accept')
+        self.assertEqual(purchase_order.wa_count, 1)
+        # Received Products
+        picking = purchase_order.picking_ids[0]
+        self.assertEqual(len(picking.move_ids_without_package), 1)
+        picking.wa_id = work_acceptance
+        picking.wa_id_change()
```
kittiu:
```suggestion
        picking._onchange_wa_id()
```

purchase_work_acceptance/models/stock_picking.py
```diff
@@ -0,0 +1,54 @@
...
+                [('picking_id', '=', self.id)]).mapped('product_id')
+            if len(move_products) != len(wa_products):
+                raise ValidationError(_('You cannot validate a transfer if done'
+                                      ' quantity not equal accepted quantity'))
+            for product in move_products:
+                wa_product = WALines.search([
+                    ('wa_id', '=', self.wa_id.id), ('product_id', '=', product.id)])
+                move_product = Moves.search([
+                    ('picking_id', '=', self.id), ('product_id', '=', product.id)])
+                if sum(wa_product.mapped('product_qty')) != \
```
kittiu: Sorry, I don't say earlier. But lets also check the UoM too.
For example, 
* Choose WA will auto populate both quantity and uom -- **1 Dozen**.
* If user change it to **12 Units**, it should be OK

purchase_work_acceptance/models/account_invoice.py
```diff
@@ -0,0 +1,53 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models, _
+from odoo.exceptions import ValidationError
+
+
+class AccountInvoice(models.Model):
+    _inherit = 'account.invoice'
+
+    is_required_wa_id = fields.Boolean(
```
kittiu: As boolean, I think no need "_id" nor "is_"
```suggestion
    require_wa = fields.Boolean(
```

purchase_work_acceptance/models/stock_picking.py
```diff
@@ -0,0 +1,54 @@
...
+                        sum(move_product.mapped('quantity_done')):
+                    raise ValidationError(_('You cannot validate a transfer if done'
+                                          ' quantity not equal accepted quantity'))
+        return super(Picking, self).button_validate()
+
+    @api.onchange('wa_id')
+    def wa_id_change(self):
+        if self.wa_id:
+            for move in self.move_ids_without_package:
```
kittiu: I don't see any logic here this is using WA's quantity to apply to Done.
Note: as discussed previously, please make sure sum of product quantity (with multi uom) is correct too, i.e., when multi line with same products, we need to sum quantity to check.
image

purchase_work_acceptance/models/account_invoice.py
```diff
@@ -0,0 +1,53 @@
...
+    )
+
+    def _get_is_required_wa_id(self):
+        return self.env.user.has_group(
+            'purchase_work_acceptance.group_work_acceptance_enforce')
+
+    @api.multi
+    def action_invoice_open(self):
+        if self.wa_id:
+            WALines = self.env['work.acceptance.line']
```
kittiu:
```suggestion
            WALine = self.env['work.acceptance.line']
```

purchase_work_acceptance/models/account_invoice.py
```diff
@@ -0,0 +1,53 @@
...
+            for line in self.invoice_line_ids:
+                po_line_id = line.purchase_line_id.id
+                WALine = WALines.search([
```
kittiu: Use lower case for variable, and with 's' to signify multi
```suggestion
                walines = WALines.search([
```

purchase_work_acceptance/models/account_invoice.py
```diff
@@ -0,0 +1,53 @@
...
+             'according to the quantity and unit price of the work acceptance.',
+    )
+
+    def _get_is_required_wa_id(self):
+        return self.env.user.has_group(
+            'purchase_work_acceptance.group_work_acceptance_enforce')
+
+    @api.multi
+    def action_invoice_open(self):
+        if self.wa_id:
```
kittiu: This **break @api.multi**, why can't we loop?
Please contact me if you are not sure what I mean.
```
        if self.wa_id:   # <--------- 
```

purchase_work_acceptance/models/account_invoice.py
kittiu: Are there other way to work? I think search in every loop is not OK.
Note: When write the code, always imagine that the possibility of having 1,000 lines, and it can be very slow. This way helps you write code better.

kittiu:
Can we use WA quantity to help fill the quantity. (or is it too difficult?, let me know).

image

image

purchase_work_acceptance/models/account_invoice.py
kittiu: Again, you can always keep WA data in dict, tuple, list and use it to compare.

kittiu: Don't forget to make it work for multi UoM too.

purchase_work_acceptance/models/account_invoice.py
```diff
@@ -0,0 +1,53 @@
...
+                        line.price_unit != WALine.price_unit:
+                    raise ValidationError(_('You cannot validate a bill if '
+                                          'Quantity more than accepted quantity'))
+        return super(AccountInvoice, self).action_invoice_open()
+
+
+class AccountInvoiceLine(models.Model):
+    _inherit = 'account.invoice.line'
+
+    wa_line_id = fields.Many2one(
```
kittiu: Is this field used? I don't see it anywhere.
```suggestion
```

purchase_work_acceptance/models/purchase.py
```diff
@@ -0,0 +1,91 @@
...
+    _inherit = 'purchase.order'
+
+    wa_count = fields.Integer(
+        compute='_compute_wa_ids',
+        string='WA count',
+        default=0,
+    )
+    wa_ids = fields.Many2many(
+        comodel_name='work.acceptance',
+        inverse_name='purchase_id',
```
kittiu: Do you want this to be Many2one or Many2many?
For Many2many, we don't need inverse_name.
```suggestion
```

purchase_work_acceptance/models/purchase.py
```diff
@@ -0,0 +1,91 @@
...
+        comodel_name='work.acceptance.line',
+        inverse_name='purchase_line_id',
+        string="WA Lines",
+        readonly=True,
+    )
+
+    def _compute_wa_ids(self):
+        for order in self:
+            for line in order.order_line:
+                order.wa_ids |= line.wa_line_ids.mapped('wa_id')
```
kittiu:
```suggestion
            order.wa_ids = order.mapped('order_line').mapped('wa_line_ids').mapped('wa_id')
```

purchase_work_acceptance/models/purchase.py
```diff
@@ -0,0 +1,91 @@
...
+            order.wa_count = len(order.wa_ids)
+
+    @api.multi
+    def action_view_wa(self):
+        act = self.env.ref('purchase_work_acceptance.action_work_acceptance')
```
kittiu, line 36: Please add
`self.ensure_one()`

purchase_work_acceptance/models/purchase.py
```diff
@@ -0,0 +1,91 @@
...
+                'purchase_work_acceptance.view_work_acceptance_form', False)
+            result['views'] = [(res and res.id or False, 'form')]
+            if not create_wa:
+                result['res_id'] = self.wa_ids.id or False
+        return result
+
+    @api.multi
+    def action_view_invoice(self):
+        create_bill = self.env.context.get('create_bill', False)
+        if create_bill:
```
kittiu: Should we make it shorter like this?
```suggestion
        if self._context.get('create_bill', False):
```

purchase_work_acceptance/models/work_acceptance.py
```diff
@@ -0,0 +1,189 @@
...
+        string='Responsible Person',
+        required=True,
+        change_default=True,
+        track_visibility='always',
+    )
+    currency_id = fields.Many2one(
+        comodel_name='res.currency',
+        string='Currency',
+        required=True,
+        default=lambda self: self.env.user.company_id.currency_id.id
```
kittiu: What will happen if PO is in "USD" but company currency is "THB"?
Will we have problem?

purchase_work_acceptance/views/work_acceptance_views.xml
```diff
@@ -0,0 +1,132 @@
...
+                   <button name="button_cancel" type="object" states="draft,accept" string="Cancel"/>
+                   <field name="state" widget="statusbar" statusbar_visible="draft,accept" readonly="1"/>
+               </header>
+               <sheet>
+                   <div class="oe_title">
+                       <h1><field name="name" readonly="1"/></h1>
+                   </div>
+                   <group>
+                       <group>
+                           <field name="partner_id" attrs="{'readonly': [('state', '!=', 'draft')]}"/>
```
kittiu: On all the readonly condition on every fields, if you can do it in Model, please do.
Except there are reason not to do.

kittiu:
Please check Travis.

kittiu:
Following comment, seem not yet fixed. Feel free to contact me if I confused you.

image

Note: typo, "Don't worry about the lines"

kittiu (changes requested):

purchase_work_acceptance/models/account_invoice.py
ps-tubtim: @kittiu I should change this code to 

```
 for rec in self:
       if rec.wa_id:
       ...
```

right?

purchase_work_acceptance/models/account_invoice.py
ps-tubtim: Thanks, this field isn't used.

purchase_work_acceptance/models/purchase.py
ps-tubtim: Thank you, I want this field to be Many2many. I will delete inverse_name.

purchase_work_acceptance/models/purchase.py
ps-tubtim: I will changed.

purchase_work_acceptance/models/work_acceptance.py
ps-tubtim: I think I should change default to PO's cerrency because WA is focus quantity, not price.

ps-tubtim: I think if I pass PO's currency,  we can't found problem.

purchase_work_acceptance/views/work_acceptance_views.xml
ps-tubtim: I can do it in Model.

purchase_work_acceptance/models/stock_picking.py
ps-tubtim: Done

purchase_work_acceptance/models/stock_picking.py
ps-tubtim: Fixed

purchase_work_acceptance/models/account_invoice.py
ps-tubtim: Done

purchase_work_acceptance/__manifest__.py
```diff
@@ -0,0 +1,31 @@
...
+        'security/ir.model.access.csv',
+        'security/security.xml',
+        'views/account_invoice_views.xml',
+        'views/purchase_views.xml',
+        'views/res_config_settings_views.xml',
+        'views/stock_picking_views.xml',
+        'views/work_acceptance_views.xml',
+        'wizard/select_work_acceptance_wizard_views.xml',
+    ],
+    'maintainer': [
```
kittiu: Please add development status = alpha

purchase_work_acceptance/__manifest__.py
```diff
@@ -0,0 +1,31 @@
...
+        'ps-tubtim',
+    ],
+    'installable': True,
```
kittiu, line 29: Add development status = alpha

purchase_work_acceptance/__manifest__.py
```diff
@@ -0,0 +1,31 @@
+# Copyright 2019 Ecosoft Co., Ltd. (http://ecosoft.co.th)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Work Acceptance',
```
kittiu:
```suggestion
    'name': 'Purchase Work Acceptance',
```

purchase_work_acceptance/models/account_invoice.py
```diff
@@ -0,0 +1,73 @@
...
+
+from odoo import api, fields, models, _
+from odoo.exceptions import ValidationError
+
+
+class AccountInvoice(models.Model):
+    _inherit = 'account.invoice'
+
+    require_wa = fields.Boolean(
+        default=lambda self: self._get_require_wa(),
```
kittiu: Can this be compute field (store=False), so we don't need to create database field?

purchase_work_acceptance/security/ir.model.access.csv
```diff
@@ -0,0 +1,3 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_work_acceptance,access_work_acceptance,model_work_acceptance,purchase.group_purchase_manager,1,1,1,1
```
kittiu: Do we need to be purchase manager to access this model?
I think only purchase user should be enough, WDYT?

purchase_work_acceptance/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,11 @@
+** Show WA Reference fields on Purchase Order, Receipt, Vendor Bill **
```
kittiu: Please review all readme's files again

purchase_work_acceptance/views/work_acceptance_views.xml
```diff
@@ -0,0 +1,137 @@
...
+                           <field name="date_due"/>
+                           <field name="date_receive"/>
+                           <field name="date_accept" attrs="{'invisible': [('state', '=', 'draft')]}"/>
+                           <field name="company_id" groups="base.group_multi_company" options="{'no_create': True}"/>
+                       </group>
+                   </group>
+                   <notebook>
+                       <page string="Products">
+                           <field name="wa_line_ids" attrs="{'readonly': [('state', '!=', 'draft')]}">
+                                <tree string="Work Acceptance Lines" editable="bottom">
```
kittiu, line 36: Do not allow adding more WA Lines (create="0"). Currently, we can, but when save, it throw error.

kittiu: Even WA Line in WA is deleted, create Vendor Bill still not remove that line.
BUT...for IN it is ok not to remove.

kittiu: When create another WA, the WA for service should default with Quantity - Billed Qty.
Now, it still default with full amount.
Note: this will be different for storable product.

kittiu: Fix about Invoice reference too,
https://github.com/ecosoft-odoo/kanban/issues/11#issuecomment-550725825

kittiu (changes requested):

purchase_work_acceptance/__manifest__.py
ps-tubtim: Added

purchase_work_acceptance/__manifest__.py
ps-tubtim: Edited

purchase_work_acceptance/models/account_invoice.py
ps-tubtim: I will change to compute field.

purchase_work_acceptance/security/ir.model.access.csv
ps-tubtim: I agree. I will change it.

purchase_work_acceptance/readme/CONFIGURE.rst
ps-tubtim: Reviewed

purchase_work_acceptance/views/work_acceptance_views.xml
ps-tubtim: > Do not allow adding more WA Lines (create="0"). Currently, we can, but when save, it throw error.

I think it should allow adding more WA Lines for flexible.

> Even WA Line in WA is deleted, create Vendor Bill still not remove that line.
BUT...for IN it is ok not to remove.

Done

> When create another WA, the WA for service should default with Quantity - Billed Qty.
Now, it still default with full amount.
Note: this will be different for storable product.

Done

etobella:
@kittiu @ps-tubtim Really interesting, but I have a few comments:

- I think that the need of WA should be defined by product, this way, it is really flexible. For example, it will be required for a design process, but not necessary for delivery costs.
- Invoices should not be related to WA, as an invoice could be used to pay two or more Work Acceptances. In order to solve this, you should relate the Received items with the WA, now it will be automatic. This is similar with picking, you can have 4 pickings for 2 POs and just one invoice.

If you think this is interesting, we can make a PR for your branch. What do you think?

Also, why do you need WA on Pickings? it is really necessary there? IMO, the picking is the WA and no WA is necessary for products and consumables.

@jarroyomorales @jbeficent

kittiu:
@etobella our use case is for public sector (government), for their strict procurement process.
The main idea is for an acceptance committee to do the Work Acceptance when the goods arrive (qty and quality grading). What ever is accepted, IN and Invoice must be exact (to refrain from fraud). This is the whole point of having WA.

From your comments, seem you don't want to be so strict. What is your motivation / use case of using WA?

**Note:** In the project, we would need committee tab, quality grading tab. But we want to leave that for project specific thing.

etobella:
Well, for example this could be useful for equipments. Equipments should not be managed as products or consumables, as they are not stored on warehouses. So, it is interesting to include them as a new type of product, that will generate equipments. However, we want to check the receival of the product from the PO, so WA is a good approach.

Also, it can be interesting for services. For example, if asked a contractor to make repair on my building, I want to accept the work, so it is great to use WA.

So, IMO, the WA can be used for more companies if we enable a expected boolean on product. Also, you can ask the providers to send an invoice for each Work Acceptance, but sometimes they will make it wrong, with your solution this will fail.

On the other hand, if you want to enable WA for Pickings I think this should be another module (too much relation).

One last thing, you made the module related to hr (responsible). Should be it made related to `res.users`?

etobella:
I made a small Branch on 11.0 following your commit and some changes with my comments:
https://github.com/etobella/purchase-workflow/tree/11.0-mig-pwa

kittiu:
@etobella I see. You wanted to use WA as another document, not too strict.
Currently we have 2 configuration like this,

- [ ] Enable WA
  - [ ] Enforce WA

What if I change it to,

- [ ] Enable WA
  - [ ] Enforce WA on Goods Receipt
  - [ ] Enforce WA on Vendor Bill

Or even more fine grain like this,

- [ ] Enable WA on Purchase Order
  - [ ] Enable WA on Goods Receipt 
    - [ ] Enforce WA on Goods Receipt
  - [ ] Enable WA on Vendor Bill 
    - [ ] Enforce WA on Vendor Bill

Will this be enough for your use case? 

Note:
- I saw from your 11.0 commit, you just delete the WA from IN and Invoice. In the above configuration, if you just enable and not enforce, it will already work for you, is it?
- For configuration to be in product level, I think it will be too complex. WDYT?

etobella:
Good for me your opinion (the last option), but I will ask for a small hook in order to set the enforcement on hook level :smile: But I will make it on a custom module :wink:

Do you need that we make something?

kittiu:
We will do it on v12, then would need you to back port to v11 (you use it there?)
Oh, you can help review this PR and push it forward. Thanks!

etobella:
We will do both, obviously 😉

purchase_work_acceptance/models/account_invoice.py
```diff
@@ -0,0 +1,80 @@
...
+    def _compute_require_wa(self):
+        self.require_wa = self.env.user.has_group(
+            'purchase_work_acceptance.group_work_acceptance_enforce')
+
+    def _prepare_invoice_line_from_po_line(self, line):
+        res = super()._prepare_invoice_line_from_po_line(line)
+        wa_line = self.wa_id.wa_line_ids.filtered(
+            lambda l: l.purchase_line_id == line)
+        if wa_line:
+            res['quantity'] = wa_line['product_qty']
```
etobella: I think you should not change the quantity as easily. Maybe you should use the minimum between `wa_line.product_qty` and `res['quantity']`

You should change the price (or at least check it here)

Also, you should use wa_line.product_qty.

purchase_work_acceptance/models/account_invoice.py
```diff
@@ -0,0 +1,80 @@
...
+                            invoice_line[line.product_id.id] = qty_old + qty
+                        else:
+                            invoice_line[line.product_id.id] = qty
+                if wa_line != invoice_line:
+                    raise ValidationError(_('You cannot validate a bill if '
+                                          'Quantity not equal accepted quantity'))
+            rec._unlink_zero_quantity()
+        return super().action_invoice_open()
+
+    def _unlink_zero_quantity(self):
```
etobella: I think this should not be here, as the default behavior of purchase is not this, maybe this should need to be on another module. This could be interesting outside of work acceptance.

purchase_work_acceptance/models/purchase.py
```diff
@@ -0,0 +1,96 @@
...
+
+class PurchaseOrder(models.Model):
+    _inherit = 'purchase.order'
+
+    wa_count = fields.Integer(
+        compute='_compute_wa_ids',
+        string='WA count',
+        default=0,
+    )
+    wa_ids = fields.Many2many(
```
etobella: Maybe you can use a One2many as you already have purchase_id on the WA.

purchase_work_acceptance/models/purchase.py
```diff
@@ -0,0 +1,96 @@
...
+                'purchase_work_acceptance.view_work_acceptance_form', False)
+            result['views'] = [(res and res.id or False, 'form')]
+            if not create_wa:
+                result['res_id'] = self.wa_ids.id or False
+        return result
+
+    @api.multi
+    def action_view_invoice(self):
+        if self.env.context.get('create_bill', False) and self.env.user.has_group(
+                'purchase_work_acceptance.group_work_acceptance'):
```
etobella: Maybe we could use a new group: `force_work_acceptance_invoice group`

purchase_work_acceptance/models/purchase.py
```diff
@@ -0,0 +1,96 @@
...
+        act = self.env.ref('purchase_work_acceptance.action_work_acceptance')
+        result = act.read()[0]
+        create_wa = self.env.context.get('create_wa', False)
+        result['context'] = {
+            'default_purchase_id': self.id,
+            'default_partner_id': self.partner_id.id,
+            'default_company_id': self.company_id.id,
+            'default_currency_id': self.currency_id.id,
+            'default_date_due': self.date_planned,
+            'default_wa_line_ids': [(0, 0, {
```
etobella, line 45: Maybe you should only pass the lines with `qty != 0`

purchase_work_acceptance/models/purchase.py
```diff
@@ -0,0 +1,96 @@
...
+        comodel_name='work.acceptance.line',
+        inverse_name='purchase_line_id',
+        string="WA Lines",
+        readonly=True,
+    )
+
+    def _get_product_qty(self):
+        if self.product_id.type == 'service':
+            return self.product_qty - self.qty_invoiced
+        return self.product_qty - self.qty_received
```
etobella: Why? I don't understand this.

purchase_work_acceptance/models/purchase.py
```diff
@@ -0,0 +1,96 @@
...
+    wa_line_ids = fields.One2many(
+        comodel_name='work.acceptance.line',
+        inverse_name='purchase_line_id',
+        string="WA Lines",
+        readonly=True,
+    )
+
+    def _get_product_qty(self):
+        if self.product_id.type == 'service':
+            return self.product_qty - self.qty_invoiced
```
etobella: Now, you can create two WAs before invoicing and both with 1 unit. Now you can invoice 2 units and this is rare...
Maybe we should need to use `self.product_qty - sum(wal.product_qty for wal in self.wa_line_ids if wal.state != 'cancel')`

purchase_work_acceptance/models/stock_picking.py
```diff
@@ -0,0 +1,75 @@
...
+            'purchase_work_acceptance.group_work_acceptance_enforce')
+
+    @api.multi
+    def button_validate(self):
+        if self.wa_id:
+            wa_line = {}
+            for line in self.wa_id.wa_line_ids:
+                qty = line.product_uom._compute_quantity(
+                    line.product_qty, line.product_id.uom_id)
+                if qty > 0.0 and line.product_id.type != 'service':
```
etobella: you should make `type in ['product', 'consu']` as this is the standard logic of picking.

purchase_work_acceptance/models/work_acceptance.py
```diff
@@ -0,0 +1,218 @@
...
+        comodel_name='res.partner',
+        string='Vendor',
+        required=True,
+        change_default=True,
+        track_visibility='always',
+        readonly=True,
+        states={'draft': [('readonly', False)]},
+    )
+    responsible_id = fields.Many2one(
+        comodel_name='hr.employee',
```
etobella: Why employee? You should use res.users or res.partner in order to not make necessary hr

purchase_work_acceptance/models/work_acceptance.py
```diff
@@ -0,0 +1,218 @@
...
+    def create(self, vals):
+        if vals.get('name', 'New') == 'New':
+            vals['name'] = self.env['ir.sequence'].next_by_code(
+                'work.acceptance') or '/'
+        return super(WorkAcceptance, self).create(vals)
+
+    @api.multi
+    def button_accept(self, force=False):
+        self._unlink_zero_quantity()
+        self.write({'state': 'accept', 'date_accept': fields.Datetime.now()})
```
etobella, line 128: Here, it would be great to set the received_qty of the lines of services :smile:

etobella (changes requested):
I made some comments on code and on behavior. I have tested it on runbot, and works as expected by README, but I think some change should be needed.

Also, I made a PO with three products with 3 different scheduled dates. The PO created 3 pickings. So, I need to create 3 WAs (it is currently not working)? or just one should be necessary?

purchase_work_acceptance/models/work_acceptance.py
etobella: or `not in ['consu', 'product']`

purchase_work_acceptance/models/purchase.py
ps-tubtim: Edited: https://github.com/OCA/purchase-workflow/pull/774/commits/aa231bbdd8d25ed7e48a94eeacb137b06599d07d

purchase_work_acceptance/models/work_acceptance.py
ps-tubtim: Edited: https://github.com/OCA/purchase-workflow/pull/774/commits/aa231bbdd8d25ed7e48a94eeacb137b06599d07d

purchase_work_acceptance/models/purchase.py
ps-tubtim: Because product type = service don't have  qty_received.
But I will add qty_received follow https://github.com/OCA/purchase-workflow/pull/774#discussion_r345090165, this code will change.

purchase_work_acceptance/models/account_invoice.py
ps-tubtim: I will change `wa_line['product_qty']` to `wa_line.product_qty`.

But I don't understand why I should use the minimum between `wa_line.product_qty` and `res['quantity']`. I need to quantity of invoice = quantity of wa when choose wa.

purchase_work_acceptance/models/account_invoice.py
ps-tubtim: I agree. I will remove this code.

purchase_work_acceptance/models/purchase.py
ps-tubtim: Edited : https://github.com/OCA/purchase-workflow/pull/774/commits/74b8c5f27338d714143dfee5372105bf91e269cc

purchase_work_acceptance/models/purchase.py
ps-tubtim: Edited : https://github.com/OCA/purchase-workflow/pull/774/commits/74b8c5f27338d714143dfee5372105bf91e269cc

purchase_work_acceptance/models/purchase.py
ps-tubtim: Edited : https://github.com/OCA/purchase-workflow/pull/774/commits/74b8c5f27338d714143dfee5372105bf91e269cc

purchase_work_acceptance/models/purchase.py
ps-tubtim: Edited : https://github.com/OCA/purchase-workflow/pull/774/commits/74b8c5f27338d714143dfee5372105bf91e269cc

purchase_work_acceptance/models/stock_picking.py
ps-tubtim: Edited : https://github.com/OCA/purchase-workflow/pull/774/commits/74b8c5f27338d714143dfee5372105bf91e269cc

purchase_work_acceptance/models/work_acceptance.py
ps-tubtim: Edited : https://github.com/OCA/purchase-workflow/pull/774/commits/74b8c5f27338d714143dfee5372105bf91e269cc

ps-tubtim:
@etobella Thank you for review. :)
You can see file changed in https://github.com/OCA/purchase-workflow/pull/774/commits/74b8c5f27338d714143dfee5372105bf91e269cc

purchase_work_acceptance/models/purchase.py
```diff
@@ -12,8 +12,9 @@ class PurchaseOrder(models.Model):
         string='WA count',
         default=0,
     )
-    wa_ids = fields.Many2many(
+    wa_ids = fields.One2many(
         comodel_name='work.acceptance',
+        inverse_name='purchase_id',
         compute='_compute_wa_ids',
         string='Work Acceptances',
```
etobella, line 18: No need of compute here

purchase_work_acceptance/models/work_acceptance.py
```diff
@@ -120,6 +120,11 @@ def create(self, vals):
     @api.multi
     def button_accept(self, force=False):
         self._unlink_zero_quantity()
+        po_lines = self.purchase_id.order_line
+        for po_line in po_lines:
+            if po_line.product_id.type == 'service':
+                po_line.qty_received = self.wa_line_ids.filtered(
```
etobella, line 126: Type not in product consu, just for following the standard

purchase_work_acceptance/models/purchase.py
ps-tubtim: Edited : https://github.com/OCA/purchase-workflow/pull/774/commits/d48b95022bd9ec1b1bfd9832102f755e5cc709d8

purchase_work_acceptance/models/work_acceptance.py
ps-tubtim: Edited : https://github.com/OCA/purchase-workflow/pull/774/commits/d48b95022bd9ec1b1bfd9832102f755e5cc709d8

ps-tubtim:
@etobella Thanks, you can see files changed in https://github.com/OCA/purchase-workflow/pull/774/commits/d48b95022bd9ec1b1bfd9832102f755e5cc709d8

purchase_work_acceptance/models/account_invoice.py
etobella: If you create a PO with one line and create a WA with this line you can create two invoices for this product. This will give problems, obviously, as you will create two invoices paying the same WA/PO.
I have already tested on the runbot:

- Enable WA (forcing WA for invoice)
- Create a PO with one line with the product as a service
- Confirm PO
- Create a WA and accept WA
- Create invoice with this WA and confirm it
- Create another invoice with this WA and confirm it

You can check the PO00010 of the current runbot to check it

etobella:
Also, with this changes, maybe we should need to make qty received readonly when the line has WA.

ps-tubtim:
@etobella Thank you.

> Also, with this changes, maybe we should need to make qty received readonly when the line has WA.

Fixed.

purchase_work_acceptance/models/account_invoice.py
ps-tubtim: Now, I don't allow create many invoices same WA. It should 1 WA : 1 Invoice

etobella (approved):
LGTM :+1: 
Great work and thanks for attending my comments :smile: 
@jarroyomorales @jbeficent

kittiu:
@ps-tubtim thanks! Good work!

jarroyomorales (approved):
Tested on runbot and works as expected. LGTM! :+1:

etobella:
@kittiu, I will start working on the backport to 11.0 :wink:

HviorForgeFlow:
Functional review on runbot:
Forcing WA
I created a PO with 3 different product and create one WA for each product.
When I Confirm and Receipt Products I have a picking of 3 products but I can only select one WA.
Some module on runbot blocked me partial Picking and I'm blocked on that stage.

ps-tubtim:
@hveficent Thank you for review.
In this module, I allow use 1 WA : 1 picking : 1 invoice.
If you need receipt 3 products in  a picking, you should create WA with 3 products.

JordiBForgeFlow (approved):
:+1 technical and functional review. I made small changes in the README seccions here. https://github.com/Eficent/purchase-workflow/tree/12.0-add-purchase_work_acceptance.

For some reason it does not let me create a PR to your repo.

purchase_work_acceptance/models/stock_picking.py
```diff
@@ -0,0 +1,80 @@
...
+                else:
+                    wa_line[line.product_id.id] = qty
+            for move in self.move_ids_without_package:
+                if move.product_id.id in wa_line.keys():
+                    qty = wa_line[move.product_id.id]
+                    if move.product_uom_qty < qty:
+                        move.quantity_done = move.product_uom_qty
+                        wa_line[line.product_id.id] = qty - move.product_uom_qty
+                    else:
+                        move.quantity_done = qty
```
JordiBForgeFlow: @ps-tubtim this is wrong. You need to update the stock.move.line with quantities completed, not the stock move.

ps-tubtim:
@jbeficent Thanks, README  edited.

JordiBForgeFlow (changes requested):

purchase_work_acceptance/models/stock_picking.py
JordiBForgeFlow: @ps-tubtim this is not working for me. Perhaps it is working for you? Somehow the inverse method associated to the field 'quantity_done'. is not being called.

ps-tubtim: @jbeficent Thanks for review. I changed this code to
```      
for move_line in self.move_line_ids:
     if move_line.product_id.id in wa_line.keys():
          qty = wa_line[move_line.product_id.id]
          if move_line.product_uom_qty < qty:
                move_line.qty_done = move_line.product_uom_qty
                wa_line[line.product_id.id] = qty - move_line.product_uom_qty
          else:
                move_line.qty_done = qty
```

etobella: @jbeficent Could you recheck?

JordiBForgeFlow: @ps-tubtim does not yet work to me. Can you test from your end?

image

ps-tubtim: @jbeficent Sorry, I misunderstand. Thank you. Edited.

JordiBForgeFlow (approved):

PR #780, opened 2019-10-29
[12.0][MIG] purchase_order_product_recommendation

ernestotejeda:
Cc @Tecnativa TT20177


purchase_order_product_recommendation/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 {
     "name": "Purchase Order Product Recommendation",
     "summary": "Recommend products to buy to supplier based on history",
-    "version": "11.0.3.0.1",
+    "version": "12.0.3.0.1",
```
pedrobaeza: 12.0.1.0.0

purchase_order_product_recommendation/__manifest__.py
```diff
@@ -12,6 +12,7 @@
     "installable": True,
     "depends": [
         "purchase",
+        "stock",
```
pedrobaeza: Why this extra dependency?

pedrobaeza (changes requested):
Please improve code coverage in tests

purchase_order_product_recommendation/__manifest__.py
ernestotejeda: This module in the prevous version only depended on purchase and purchase depended on stock_account (https://github.com/odoo/odoo/blob/40db7a8cf25ef0b5dd5b9d89e3b92c19372b09f6/addons/purchase/__manifest__.py#L12), but in v12 purchase only depends on account, so we have to force the stock module dependency here

ernestotejeda:
Changes done

purchase_order_product_recommendation/__manifest__.py
pedrobaeza: Then depend on `purchase_stock`.

ernestotejeda: You're right. It's done, thank you

purchase_order_product_recommendation/wizards/purchase_order_recommendation.py
```diff
@@ -0,0 +1,316 @@
...
+    purchase_line_id = fields.Many2one(
+        comodel_name='purchase.order.line',
+    )
+    is_modified = fields.Boolean()
+
+    @api.multi
+    @api.depends('partner_id', 'product_id', 'units_included')
+    def _compute_price_unit(self):
+        for one in self:
+            one.price_unit = one.product_id._select_seller(
```
chienandalu, line 291: I'm having some performance issues with this compute in v11, but maybe in v12 it's benefited by the ORM improvements

chienandalu (approved):
Code and functional testing :+1:

pedrobaeza (approved):

PR #786, opened 2019-11-06
[11.0][FIX] purchase_order_product_recommendation: performance

chienandalu:
- When the wizard had a lot of lines, the price_unit computation had a
very poor performance whenenever we changed the qty to order (maybe due
to some ORM issue). Changing to an initial computation and then to a
computation on the onchange method, gets rid of the problem.

cc @Tecnativa TT19188

### DIFF
--- a/purchase_order_product_recommendation/wizards/purchase_order_recommendation.py
+++ b/purchase_order_product_recommendation/wizards/purchase_order_recommendation.py
@@ -137,6 +137,13 @@ class PurchaseOrderRecommendation(models.TransientModel):
         qty_to_order = abs(
             min(0, units_virtual_available - vals.get('qty_delivered', 0)))
         vals['is_modified'] = bool(qty_to_order)
+        units_included = order_line and order_line.product_qty or qty_to_order
+        price_unit = product_id._select_seller(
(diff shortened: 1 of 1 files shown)


purchase_order_product_recommendation/wizards/purchase_order_recommendation.py
```diff
@@ -152,6 +152,13 @@ def _prepare_wizard_line(self, vals, order_line=False):
                                qty_to_order),
             'is_modified': vals.get('is_modified', False),
         }
+        vals['price_unit'] = product_id._select_seller(
```
pedrobaeza: This is only syntactic sugar, but you can still return the dictionary, but adding the key with the call to `_select_seller`.

pedrobaeza (approved):

purchase_order_product_recommendation/wizards/purchase_order_recommendation.py
chienandalu: Yes, it can be made in a prettier way :) Change

ernestotejeda (approved):

pedrobaeza:
@ernestotejeda please apply this on v12 too

PR #789, opened 2019-11-06
[12.0][FIX] purchase_order_product_recommendation: performance

ernestotejeda:
- When the wizard had a lot of lines, the price_unit computation had a
very poor performance whenenever we changed the qty to order (maybe due
to some ORM issue). Changing to an initial computation and then to a
computation on the onchange method, gets rid of the problem.

Cc @Tecnativa


pedrobaeza (approved):
Let's fast-track this as it has been reviewed in 11.0

PR #793, opened 2019-11-21
[ADD][11.0] purchase_stock_return_request: New module

chienandalu:
This module extends stock_return_request to be able to link the returns to
their corresponding purchase orders in the case of returns to supplier as well
as to be able to get the corresponding refunds.

cc @Tecnativa TT20130


rafaelbn (approved):
Running in production for a long time! :+1:

PR #792, opened 2019-11-20
[12.0][FIX] Purchase request incorrect mapping in compute

CasVissers-360ERP:
Fixes traceback:
`ValueError: <class 'KeyError'>: "purchase_request_id" while evaluating`

purchase_request_allocation_ids doesn't have purchase_request_id field.

PR #798, opened 2019-11-25
[FIX] Incorrect mapping in compute

CasVissers-360ERP:
Fixes traceback:
ValueError: <class 'KeyError'>: "purchase_request_id" while evaluating

purchase_request_allocation_ids doesn't have purchase_request_id field.

Replaces #792


HviorForgeFlow:
https://github.com/OCA/purchase-workflow/pull/836

PR #799, opened 2019-11-25
[12.0][FIX] Purchase request incorrect mapping in compute

HviorForgeFlow:
@CasVissers


CasVissers-360ERP:
@HviorForgeFlow Incorrect mapping is still in current version?

PR #805, opened 2019-11-27
[IMP] purchase_order_product_recommendation: extra filters

chienandalu:
- Extract domain build to an independent method for easy extension by
third modules
- Extract product filter to an independent method for easy extension.
- New product internal category filter.
- New "Show all products" filter that allows to easily show all
purchable products.

cc @Tecnativa TT20722

PR #806, opened 2019-11-27
[ADD] purchase_order_product_recommendation_brand: New module

chienandalu:
Allows to filter product recommendations by brand.

Depends on 

- [x] https://github.com/OCA/purchase-workflow/pull/805 purchase_order_product_recommendation domain  method

cc @Tecnativa TT20723


purchase_order_product_recommendation_brand/wizards/purchase_order_recommendation.py
```diff
@@ -0,0 +1,47 @@
...
+        comodel_name='product.brand',
+        string="Brands",
+    )
+
+    def _get_products(self):
+        """Filter products of the given brands"""
+        products = super()._get_products()
+        # Filter products by brand if set.
+        # It will apply to show_all_partner_products as well
+        if self.product_brand_ids:
```
pedrobaeza, line 19: Why do you need this if you are already filtering in the domain?

chienandalu: In this method , I filter the supplier products (given by the proper supplierinfos). In `_get_all_products_domain` we obtain all the products independently of the supplier.

pedrobaeza: OK, I'm not sure about the performance of this.

pedrobaeza (approved):

rafaelbn (approved):

pedrobaeza:
I'm thinking this module should go to OCA/brand on 12.0

chienandalu:
> I'm thinking this module should go to OCA/brand on 12.0

I didn't know the repo, but anyway `product-attribute/product_brand` and `brand/brand` doesn't seem to have any relation, don't they?

chienandalu:
ok, now I see product_brand in the repo, but in the v13 branch

pedrobaeza:
Yes, it was moved later. We can wait for v13 if you think so.

PR #810, opened 2019-11-29
[IMP] purchase_order_product_recommendation: more filters

chienandalu:
Bring changes from https://github.com/OCA/purchase-workflow/pull/805 as well as split tests for better inherability

- Extract domain build to an independent method for easy extension by
third modules
- Extract product filter to an independent method for easy extension.
- New product internal category filter.
- New "Show all products" filter that allows to easily show all
purchable products.
- Make tests extensible by other modules

cc @Tecnativa TT20722

### DIFF
--- a/purchase_order_product_recommendation/tests/test_recommendation.py
+++ b/purchase_order_product_recommendation/tests/test_recommendation.py
@@ -1,6 +1,5 @@
 # Copyright 2019 David Vidal <david.vidal@tecnativa.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
-
 from datetime import date, datetime
 from odoo.tests.common import SavepointCase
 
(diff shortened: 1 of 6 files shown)


pedrobaeza (commented):
Although I didn't tell you before in the other PR, instead of creating such big diff, it's better to leave both classes in the same file (as they are very related), and previous conventions on the names are for putting `Base` for the base class and the same without base for the normal one.

purchase_order_product_recommendation/tests/test_recommendation.py
```diff
@@ -142,6 +151,12 @@ def wizard(self):
         wizard._generate_recommendations()
         return wizard
 
+
+class RecommendationCaseTests(RecommendationCase):
+    @classmethod
+    def setUpClass(cls):
```
pedrobaeza: No need of this

chienandalu: Yes, sorry

pedrobaeza (approved):

purchase_order_product_recommendation/wizards/purchase_order_recommendation.py
```diff
@@ -43,10 +43,20 @@ class PurchaseOrderRecommendation(models.TransientModel):
...
     )
     show_all_partner_products = fields.Boolean(
-        string='Show all products',
+        string='Show all supplier products',
         default=False,
         help='Show all products with supplier infos for this supplier',
     )
+    show_all_products = fields.Boolean(
+        string='Show all purchasable products',
+        default=False,
```
nikul-serpentcs, line 52: @chienandalu FYI no need to add `default=False`, Boolean always False

rafaelbn (approved):

PR #811, opened 2019-12-02
[FIX] purchase_order_general_discount: reset to default

chienandalu:
- A related field shoudn't have a default value. This was making the
setting to be reset from time to time.

cc @Tecnativa TT20861


pedrobaeza (approved):
Fast-tracking this fix

chienandalu:
Looks like this didn't merge properly

Issue #794, opened 2019-11-21
[11.0] Dealing with purchase cancellation with MTO


jonerikceberio:
Situation: a salesman makes a sale of a product configured as MTO, buy. A purchase is created, and the purchase manager confirms it. But someone thinks it would be a good idea to have some stock of the product, or changing the price, or something like that -> cancels the PO, modifies it and re-confirms it
Expected result: the picking is re-generated with everything linked to SO
Problem: the purchase move has been created with propagate=True (by default, as it is not informed), and therefore the sale move is cancelled when the purchase is cancelled. NO activity is sent to the salesman, so the SO is left with a cancelled picking (or partial one), and no one knows about it.

Solution: There are several options:

- The re-generation of the PO could be done, but it would be difficult
- We could send and activity to the salesman, but he has no way to re-create the picking
- Intercepting the PO cancellation and asking/warning the user

What I did in [https://github.com/Digital5-Odoo/purchase-workflow/commit/0724f0cb5ce54465952199c14986edcc7dd6f864](https://github.com/Digital5-Odoo/purchase-workflow/commit/0724f0cb5ce54465952199c14986edcc7dd6f864):

- Inform the propagate field in the purchase move with the value of the procurement rule that created it. That way we can configure it on UI as False, and therefore the sale move won´t be cancelled - module stock will just change it to MTS
- Upon cancelling the purchase: recompute the state of the sale move, so it will change from "Waiting for other operation" to "Waiting availability"

Most possibly there is a better solution, so I did not make a PR of it. Thought, proposals?

pedrobaeza:
Why do you cancel it instead of simply change the quantity in the PO line?

jonerikceberio:
Sorry, got that wrong: an example would be making the purchase to supplier A, and having the confirmation. But a couple of days later the provider calls you to inform you of a delay of the delivery, which you cannot accept. So you want to change the supplier.

pedrobaeza:
Indeed it requires to cancel the PO for changing the supplier.

I suggest that if you have that case frequently, remove the propagate=True in MTO stock rule and avoid the problem.

jonerikceberio:
[...]

PR #813, opened 2019-12-02
11.0 purchase make to order cancel

jonerikceberio:
Explained in #794, summary:
Situation: a salesman makes a sale of a product configured as MTO, buy. A purchase is created, and the purchase manager confirms it. But someone has to change the supplier -> cancels the PO, modifies it and re-confirms it.
Expected result: the picking is re-generated with everything linked to SO
Problem: the purchase move has been created with propagate=True (by default, as it is not informed), and therefore the sale move is cancelled when the purchase is cancelled. NO activity is sent to the salesman, so the SO is left with a cancelled picking (or partial one), and no one knows about it.
This module does the following actions:

- Inform the propagate field in the purchase move with the value of the procurement rule that created it. That way we can configure it on UI as False, and therefore the sale move won´t be cancelled - module stock will just change it to MTS
- Upon cancelling the purchase: recompute the state of the sale move, so it will change from "Waiting for other operation" to "Waiting availability"


purchase_make_to_order_cancel/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Purchase MTO Cancel',
+    'summary': 'Avoid the cancellation of sale pickings in case of a purchase '
+               'cancellation with a MTO product',
+    'version': '11.0.1.0.0',
+    'category': 'Purchase Management',
+    'author': 'Digital5, S.L., Odoo Community Association (OCA)',
+    'website': 'http://www.digital5.es',
```
MeetKD: Website -> https://github.com/OCA/purchase-workflow/

purchase_make_to_order_cancel/models/purchase.py
```diff
@@ -0,0 +1,44 @@
+# Copyright 2019 Digital5 S.L.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from openerp import models, api
```
MeetKD: openerp -> odoo

PR #819, opened 2019-12-11
[12.0][IMP]purchase_request

AaronHForgeFlow:
`requested_by` should not be copied when duplicating a purchase request

cc @forgeflow

PR #820, opened 2019-12-11
[10.0][FIX] purchase_request followers

AaronHForgeFlow:
backport #819 

Basically, when duplicating purchase_request the requested by should be the guy that duplicates the record. That way it will be also follower.

cc @HviorForgeFlow


AaronHForgeFlow:
Apologies, my previous solution was quite clumsy. Adding the current user as requested by works, because it makes sense, and because when saving the purchase request the current user will be assigned to the requested by field.

Regards.

HviorForgeFlow (approved):

PR #821, opened 2019-12-11
[FIX] purchase_order_product_recommendation: show recommendations for all products

chienandalu:
- Recommendations were only shown for the supplier products

cc @Tecnativa TT20722


pedrobaeza (approved):

yajo (approved):
A test would be good here.

PR #823, opened 2019-12-16
[12.0][ADD] purchase_manual_delivery

AdriaGForgeFlow:
This PR adds the module purchase_manual_delivery in v12.

The goal of this module is to allow the manual creation of incoming shipments. When installed,
a purchase order won't impact directly the stock. It will not create the pickings upon order confirmation.
It allows the reception and the impact on stock to be done manually when needed.
The goal is to be used when a supplier does not confirm the full order at once. As the supplier
confirms the different purchase order lines, the user can manually create the incoming
shipments for those specific lines.

cc ~ @forgeflow


purchase_manual_delivery/models/purchase_order.py
```diff
@@ -0,0 +1,88 @@
...
+    _inherit = "purchase.order.line"
+
+    existing_qty = fields.Float(
+        compute="_compute_existing_qty",
+        string="Existing Qty",
+        digits=dp.get_precision("Product Unit of Measure"),
+        help="Quantity already planned or shipped (stock movements "
+        "already created)",
+    )
+    pending_qty_to_receive = fields.Boolean(
```
HviorForgeFlow: I would use:
```suggestion
    pending_to_receive = fields.Boolean(
```
I doesn't not make sense to me to use quantity on a boolean, could be confusing.

AdriaGForgeFlow: Done :+1:

AaronHForgeFlow (approved):
:+1: Functional tests. Works like a charm. And the feature to select existing pickings is awesome :medal_sports:

purchase_manual_delivery/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+{
+    'name': 'Purchase Manual Delivery',
+    'summary': """
+        Prevents pickings to be auto generated upon Purchase Order confirmation
+        and adds the ability to manually generate them as the supplier confirms
+        the different purchase order lines.
+    """,
+    'version': '12.0.1.0.0',
+    'license': 'AGPL-3',
+    'author': 'Eficent Business and IT Consulting Services S.L.,'
```
AaronHForgeFlow: Please, ForgeFlow everywhere :grin:

AdriaGForgeFlow: Thanks for the heads up :+1:

HviorForgeFlow (approved):
Functional review
:+1:

HviorForgeFlow:
@AdriaGForgeFlow squash commits

AdriaGForgeFlow:
> @AdriaGForgeFlow squash commits

Done :+1:

HviorForgeFlow (approved):

rvalyi:
A nice use case I think is when you have electronic invoicing and you may want to create the incoming shipments based on the supplier electronic invoice. I'll review it, thanks!

PR #827, opened 2019-12-17
[12.0][FIX] purchase_open_qty

AdriaGForgeFlow:
Quick fix to trigger computation on PO state changed, otherwise it is not computed when the module is installed together with purchase_manual_delivery.

cc ~ @ForgeFlow @JordiBForgeFlow


LoisRForgeFlow (approved):
:+1:

HviorForgeFlow (approved):
FF

PR #832, opened 2019-12-18
[10.0][FIX]purchase_order_line_sequence. 

AaronHForgeFlow:
Do not update move sequence in move if not updateable move.

_create_picking method is called on write when decreasing the quantity. If all incoming shipments are done then you get an error on this module.

### DIFF
--- a/purchase_order_line_sequence/models/purchase.py
+++ b/purchase_order_line_sequence/models/purchase.py
@@ -31,12 +31,14 @@ class PurchaseOrder(models.Model):
         for order in self:
             if any([ptype in ['product', 'consu'] for ptype in
                     order.order_line.mapped('product_id.type')]):
-                picking = order.picking_ids.filtered(
-                    lambda x: x.state not in ('done', 'cancel'))[0]
-                for move, line in zip(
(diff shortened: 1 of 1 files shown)


purchase_order_line_sequence/models/purchase.py
```diff
@@ -31,12 +31,14 @@ def _create_picking(self):
...
                     order.order_line.mapped('product_id.type')]):
-                picking = order.picking_ids.filtered(
-                    lambda x: x.state not in ('done', 'cancel'))[0]
-                for move, line in zip(
-                        sorted(picking.move_lines,
-                               key=lambda m: m.id), order.order_line):
-                    move.write({'sequence': line.sequence})
+                pickings = order.picking_ids.filtered(
+                    lambda x: x.state not in ('done', 'cancel'))
+                if pickings:
```
AaronHForgeFlow, line 36: This is the only different line :smile:

HviorForgeFlow (approved):

HviorForgeFlow:
I noticed this problem also in v12

bodedra (approved):
LGTM :+1:

PR #835, opened 2019-12-30
[12.0] product_supplier_code_purchase

AaronHForgeFlow:
Standard migration + readme refactor + add coverage

cc @HviorForgeFlow  @ForgeFlow


legalsylvain (changes requested):
Hi @AaronHForgeFlow. Thanks for porting this module I'm just discovering.
I see it usefull. 
General design questions :
 
- Why the field is not stored to True ? It will avoid to recompute each time the purchase order is displayed, that can be long for big sized purchase order. Also, the value can change. (code can be updated, and it could be great to keep the original value Otherwise, it will be not the same text as for the field name of the purchase order line).
- I know that in recent version of odoo, when confirming a purchase order, it will create supplierinfo, if not exist for the current product. Should be great to use this value to have up to date supplierinfo. Don't you think ? 

kind regards.

AaronHForgeFlow:
Hi @legalsylvain , thanks for reviewing.

> * Why the field is not stored to True ? It will avoid to recompute each time the purchase order is displayed, that can be long for big sized purchase order. Also, the value can change. (code can be updated, and it could be great to keep the original value Otherwise, it will be not the same text as for the field name of the purchase order line).

Good point. I agree with you. I will change that.

> * I know that in recent version of odoo, when confirming a purchase order, it will create supplierinfo, if not exist for the current product. Should be great to use this value to have up to date supplierinfo. Don't you think ?

I agree, in case of creating new supplierinfo. In case of existing supplierinfo I prefer not to update it. It is common for customers to make mistakes and writing wrong codes here. That could affect then other parts.

Kind Regards.

legalsylvain:
> Good point. I agree with you. I will change that.

Thanks !

> I agree, in case of creating new supplierinfo. In case of existing supplierinfo I prefer not to update it. It is common for customers to make mistakes and writing wrong codes here. That could affect then other parts.

Thanks for your answer.

I just tested, and in Odoo purchase module, a supplierinfo is created using price, supplier, etc..., if it doens't exist. (As said, I propose to use the code field to create an up to date supplierinfo).
But if the supplierinfo exists, the purchase price (for exemple) is not updated (if changed). I think that your module should follow the same logic and let the user update manually the code, if needed (so we are agree with the implementation !)

Other question : did you thought to add the code also in the supplier invoice line model ? It should be great and it could allow a little glue module ``account_invoice_supplierinfo_update_product_code`` 
with [account_invoice_supplierinfo_update](https://github.com/OCA/account-invoicing/tree/12.0/account_invoice_supplierinfo_update) to check and update values, if incorrect.
Don't you think ? 
Maybe it should be in another module ``account_invoice_supplier_code``. Not a clear point of view regarding that question.

AaronHForgeFlow:
Thanks for your answer.

> I just tested, and in Odoo purchase module, a supplierinfo is created using price, supplier, etc..., if it doens't exist. (As said, I propose to use the code field to create an up to date supplierinfo). But if the supplierinfo exists, the purchase price (for exemple) is not updated (if changed). I think that your module should follow the same logic and let the user update manually the code, if needed (so we are agree with the implementation !)

Yes, I was thinking on the same think.

> Other question : did you thought to add the code also in the supplier invoice line model ? It should be great and it could allow a little glue module `account_invoice_supplierinfo_update_product_code`
> with [account_invoice_supplierinfo_update](https://github.com/OCA/account-invoicing/tree/12.0/account_invoice_supplierinfo_update) to check and update values, if incorrect.
> Don't you think ?
> Maybe it should be in another module `account_invoice_supplier_code`. Not a clear point of view regarding that question.

Perhaps  `account_invoice_supplier_code` should depends on this one and take always the code  from the PO line. Not sure.

legalsylvain:
Well, in fact, the field (on the account.invoice.line model, will be usefull only in a supplierinfo update workflow. Maybe we could consider to add only a new module account_invoice_supplierinfo_update_product_code that add this field.

AaronHForgeFlow:
In favor of just one single module. Are you planning to propose something?

legalsylvain:
> Are you planning to propose something?

Yes, I'll do a PoC. but not before T2 2020. I'll ping you if you want.

AaronHForgeFlow:
> > Are you planning to propose something?
> 
> Yes, I'll do a PoC. but not before T2 2020. I'll ping you if you want.

All right. That's good. For now I don't need that feature but perhaps in the feature I will need it. I will be glad to review that.

Regards.

AaronHForgeFlow:
I am wondering if it is best to not let the user to change the supplier code in the PO line, and take it always from he supplierinfo table. I don't really like this kind of duplicity. Perhaps it is better to leave the module as it is. Knowing that the field is stored, the PO line will always contain the product code at the date the PO line was created.

legalsylvain:
This duplicity is like the purchase price. I see more clicks to force the user to open the product form then supplierinfo form and then to set a product code and then to save (2 times).
Dont' you think ?

AaronHForgeFlow:
Yes, perhaps adding a security group so only people belonging to that group could alter the code in purchase_order_line could do the thing. Would that be fine?

legalsylvain:
Well, why not, even if I dont' understand exactly why. But in that case, I think it should not be a group, but an option in the "preference" form of the user, to let the possibility to the user to enable or not this feature. like https://github.com/OCA/web/blob/12.0/web_responsive/models/res_users.py#L10

A user that creates a purchase order has the right to update supplierinfo so it has no sense to block him if he has not the possibility to change the feature. Don't you think ?

But in reality, I don't understand exactly what you fear. That the user corrupt data ? or that he makes a mistake ? 

kind regards.

AaronHForgeFlow:
All right. I will not implement that security layer in this module. It seems it is not generic feature so better not to put more noise in the module.

supplierinfo codes now will be updated when confirming PO.

Thanks for the feedback.

product_supplier_code_purchase/__manifest__.py
```diff
@@ -1,13 +1,11 @@
...
+# Copyright 2015-17 ForgeFlow, S.L.
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
 {
     "name": "Product Supplier Code in Purchase",
     "summary": """This module adds to the purchase order line the supplier
                 code defined in the product.""",
-    "version": "10.0.1.0.0",
+    "version": "12.0.1.0.0",
     "author": "Eficent, Odoo Community Association (OCA)",
```
AdriaGForgeFlow: Could you update this also to ForgeFlow?

product_supplier_code_purchase/models/purchase_order.py
```diff
@@ -1,8 +1,5 @@
-# -*- coding: utf-8 -*-
-# Copyright 2015-17 Eficent Business and IT Consulting Services, S.L.
-#           (http://www.eficent.com)
+# Copyright 2015-17 ForgeFlow, S.L.
```
AdriaGForgeFlow, line 1: Could you update the year of this and all other Copyright comments to 2020

HviorForgeFlow:
@legalsylvain what should we do to move this forward?

AdriaGForgeFlow (commented):
Some comments below:

product_supplier_code_purchase/models/purchase_order.py
```diff
@@ -1,27 +1,64 @@
...
+                    partner_id=line.partner_id,
+                    quantity=line.product_qty,
+                    date=line.order_id.date_order
+                    and line.order_id.date_order.date(),
+                    uom_id=line.product_uom,
+                )
+                if seller:
+                    try:
+                        seller['product_code'] = line.product_supplier_code
+                    except AccessError:
```
legalsylvain, line 63: I don't understand such try / catch AccessError. Could you explain this point ?

legalsylvain (approved):
Code review . No test. thanks for : 
- adding tests
- changing compute to onchange. (better design)

Just one question regarding AccessError.

code review / no test.

product_supplier_code_purchase/models/purchase_order.py
AaronHForgeFlow: Odoo does the same thing when updating supplierinfo: https://github.com/odoo/odoo/blob/12.0/addons/purchase/models/purchase.py#L396

I guess this is when you have permissions in PO lines but not in supplierinfo. I did not get any error there, but just in case, I prefer to follow what Odoo does.

HviorForgeFlow (approved):

PR #839, opened 2020-01-11
[ADD] purchase_request_secondary_unit: New module 

alan196:
This new module allows the use of secondary units in a purchase request and transfers it to the Purchase order when is created from a purchase request.


ThomasBinsfeld (approved):

rousseldenis:
@alan196 Unfortunately you can't depends on purchase_order_secondary_unit keeping LGPL license as it's an AGPL module.

alan196:
The license is now AGPL 👍 Could you help me to review @rousseldenis

PR #852, opened 2020-01-29
[12.0][FIX]purchase_request

AaronHForgeFlow:
Fails when there is no group id in values, this happens on automatic reordering like reordering rules scenarios, no manual purchase requests.


HviorForgeFlow (approved):
LGTM. Can you forward port to v13 when merged? if required.

AdriaGForgeFlow (approved):
Code review :+1:

PR #856, opened 2020-02-06
[12.0][IMP] purchase_location_by_line

tafaRU:
This improvement allows to define a default destination location.

The default destination location can be defined in the product or in the product category.
The first one has high priority.


HviorForgeFlow (commented):
Functional review:
- Module is incompatible with purchase_manual_delivery, need to be uninstalled on runbot to test.
- Trigger for destination location on PO Line is only launched on changing product. Could be nice to have a depends also. And maybe a copy=false on location_dest_id variable. Rare corner case when default_location_destination is modified and a PO is copied from  an old one.

PR #870, opened 2020-02-29
[MIG] Migrated purchase_order_line_stock_available to v13.

sudhir-erpharbor:


tonygalmiche:
With runbot : 
image

image

Why -45 and not 0?

PR #637, opened 2019-01-26
[12.0] [MIG] purchase_minimum_amount

Nikul-Chaudhary:
Purchase Minimum Amount
=======================

This module allows you to establish and automate a specific Purchase Order Approval Block Reason in the system: the 'Minimum Purchase Order Amount per Vendor'.

PR #873, opened 2020-03-04
[12.0][FIX] purchase_request: allow Purchase users to see purchase requests

etobella:
Purchase Users should need to see purchase requests
#637

### DIFF
--- a/purchase_request/security/ir.model.access.csv
+++ b/purchase_request/security/ir.model.access.csv
@@ -12,3 +12,5 @@ access_purchase_request_allocation_manager,purchase.request.allocation,model_pur
 access_purchase_request_allocation_user,purchase.request.allocation,model_purchase_request_allocation,base.group_user,1,1,1,0
 access_purchase_request_line_stock_user,purchase.request.line.stock,model_purchase_request_line,stock.group_stock_user,1,1,1,0
 access_purchase_request_stock_user,purchase.request.stock,model_purchase_request,stock.group_stock_user,1,0,0,0
+access_purchase_request_purchase_user,purchase.request,model_purchase_request,purchase.group_purchase_user,1,0,0,0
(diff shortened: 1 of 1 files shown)


LoisRForgeFlow (approved):

LoisRForgeFlow:
cc @AaronHForgeFlow

jarroyomorales (approved):

AaronHForgeFlow (approved):
It makes sense to me. I think  in previous versions happens the same.

HviorForgeFlow:
Allowing Purchase User belong to Purchase Request User don't allow that user to see PRs from other users, just generate their own ones. It can be done by manual configuration.

I can't see benefits of this fix.

@etobella

HviorForgeFlow (changes requested):
ping @etobella 
So this is a workaround to allow testing, isn't it?
I could not see previous tests on related PR.

etobella:
The problem is, the order of execution. On previous repos the tests were not colliding, but now it is. The problem was that the user was trying to access stock requests but he was not allowed.

etobella:
@HviorForgeFlow maybe this is a workaround for a test, But purcase users should be able to see who asked for something. Isn't it?

HviorForgeFlow:
Then they should be Manager, or new group read only all, because Purchase Request Users can only see their own Purchase Requests.

@etobella

etobella:
Good for me any of both options, @LoisRForgeFlow which option do you prefer?

AaronHForgeFlow:
I vote for new group readonly all, it is less restrictive

etobella:
If we want to add this kind of permissions, the simplest would be to add the permission directly to the Purchase user

etobella:
I think now is enough @HviorForgeFlow

HviorForgeFlow (approved):
Ok! Better this way, thanks!

PR #876, opened 2020-03-12
[10.0][ADD] purchase_request - unlink pr and prl only when pr is draft

baimont:
When unlinking a purchase request or a purchase request line and purchase request is not in draft, this raises an error.

### DIFF
--- a/purchase_request/models/purchase_request.py
+++ b/purchase_request/models/purchase_request.py
@@ -181,6 +181,20 @@ class PurchaseRequest(models.Model):
         return res
 
     @api.multi
+    def _can_be_deleted(self):
+        self.ensure_one()
+        return self.state == 'draft'
+
+    @api.multi
+    def unlink(self):
+        for request in self:
+            if not request._can_be_deleted():
+                raise UserError(_(
+                    'You cannot delete a purchase request which is not draft.'
+                ))
+        return super(PurchaseRequest, self).unlink()
+
+    @api.multi
     def button_draft(self):
         self.mapped('line_ids').do_uncancel()
         return self.write({'state': 'draft'})
@@ -331,3 +345,18 @@ class PurchaseRequestLine(models.Model):
             requests = self.mapped('request_id')
             requests.check_auto_reject()
         return res
+
+    @api.multi
+    def _can_be_deleted(self):
+        self.ensure_one()
+        return self.request_state == 'draft'
+
+    @api.multi
+    def unlink(self):
+        for line in self:
+            if not line._can_be_deleted():
+                raise UserError(_(
+                    'You can only delete a purchase request line '
+                    'if the purchase request is in draft state.'
+                ))
+        return super(PurchaseRequestLine, self).unlink()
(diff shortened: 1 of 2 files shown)


AaronHForgeFlow:
Is it possible to delete confirmed Purchase request now?

AaronHForgeFlow (approved):
:+1: Just tested and you can delete purchase request in confirmed state in v10. That is not correct, so this fix it is ok to me. 

ping @HviorForgeFlow

benwillig (approved):

purchase_request/models/purchase_request.py
```diff
@@ -331,3 +340,14 @@ def write(self, vals):
...
+
+    @api.multi
+    def unlink(self):
+        if self.filtered(
+            lambda r: r.request_state != 'draft'
+        ):
+            raise UserError(_(
+                'You can only delete a purchase request line '
+                'if the purchase request is in draft state.'
+            ))
```
apineux: Why not use a for loop with raise such as you did in purchase_request? No need to filter everything if the first one is already in confirmed state.

purchase_request/models/purchase_request.py
```diff
@@ -331,3 +340,14 @@ def write(self, vals):
             requests = self.mapped('request_id')
             requests.check_auto_reject()
         return res
+
+    @api.multi
+    def unlink(self):
+        if self.filtered(
+            lambda r: r.request_state != 'draft'
+        ):
```
ThomasBinsfeld:
```suggestion
    @api.multi
    def _can_be_deleted()
        self.ensure_one()
        return self.request_state == 'draft'
    
    @api.multi
    def unlink(self):
        for rec in self:
            if not rec._can_be_deleted():
```

apineux:
@baimont Why not use a record rule to achieve the same result? Is this operation executed by the admin?

AaronHForgeFlow (changes requested):
Agree with the record rules proposal

baimont:
Thanks for your reviews @AaronHForgeFlow, @ThomasBinsfeld, @apineux and @benwillig !
 
About this comment by @apineux :
> @baimont Why not use a record rule to achieve the same result? Is this operation executed by the admin?

I would say I have no idea which one is better. Record rule would be nicer concerning admin, but searching for the same functionality across odoo and oca code, I encountered it a lot through python code. As doing the same is not a valid argument, may I ask you to comment this post to enlighten me about that?
In favor of python code: I feel this way is more maintainable. Record rules can quickly become hard to read.

apineux:
@baimont 
The record rule is a better choice in my opinion as it is the purpose of record rules. Overwriting the unlink method is stronger as it will block the deletion all the time for any user. If you just want to block users to delete request lines then use the record rules.
If you need a stronger block (even for admin user) because some 'cron' or else can also delete request lines then use the unlink method as you did.

baimont:
> @baimont
> The record rule is a better choice in my opinion as it is the purpose of record rules. Overwriting the unlink method is stronger as it will block the deletion all the time for any user. If you just want to block users to delete request lines then use the record rules.
> If you need a stronger block (even for admin user) because some 'cron' or else can also delete request lines then use the unlink method as you did.

@apineux, as I wanted something stronger, I went with the second option. But I modified the commit following @ThomasBinsfeld suggestion, making the feature easily inheritable.
I believe this is a good compromise.

apineux (approved):

ThomasBinsfeld (approved):

AaronHForgeFlow (approved):
:+1: It is fine to me

PR #888, opened 2020-04-07
[13.0][ADD] purchase_warn_message

HviorForgeFlow:
### Purchase Warn Message
This module add a warning popup on POs to ensure warning is populated
no only when partner is changed.

CC @ForgeFlow @MiquelRForgeFlow @JordiBForgeFlow


purchase_warn_message/__manifest__.py
```diff
@@ -0,0 +1,14 @@
...
+
+{
+    "name": "Purchase Warn Message",
+    "summary": """
+        Add a popup warning on purchase to ensure warning is populated""",
+    "version": "13.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "ForgeFlow, Odoo Community Association (OCA)",
+    "website": "www.github.com/OCA/stock-logistics-workflow.git",
+    "depends": ["purchase"],
```
MiquelRForgeFlow, line 12:
```suggestion
    "depends": ["purchase", "account_invoice_warn_message"],
```

HviorForgeFlow: It is not required to be dependency of account_invoice_warn_message. Modules are independant, aren't they?

MiquelRForgeFlow: Yes, but... When installing `sale` your are installing `account`, then when you are installing this module, it may be implicit you want the warning in account (account invoice), because invoicing your sales are part of the sale workflow.

MiquelRForgeFlow: sorry, I was thinking in sale, but the same argument applies for purchase.

HviorForgeFlow: Maybe they are related on functionalities, but not on dependencies nor requirements, so I would like to skip account_invoice_warn_message dependency.

AdriaGForgeFlow (approved):
Code review :+1:

purchase_warn_message/models/purchase_order.py
```diff
@@ -0,0 +1,23 @@
...
+
+    purchase_warn_msg = fields.Text(compute="_compute_purchase_warn_msg")
+
+    def _compute_purchase_warn_msg(self):
+        for rec in self:
+            if rec.partner_id:
+                if rec.partner_id.parent_id:
+                    p = rec.partner_id.parent_id
+                else:
+                    p = rec.partner_id
```
ValentinVinagre: you can change that for:
`p = rec.partner_id.commercial_partner_id`

purchase_warn_message/models/purchase_order.py
```diff
@@ -0,0 +1,23 @@
...
+                if rec.state not in ["done", "cancel"] and p.purchase_warn == "warning":
+                    rec.purchase_warn_msg = p.purchase_warn_msg
+                    continue
```
ValentinVinagre, line 23: why do you use continue?
you can add a variable after the `if rec.partner_id:` and only change the value in this if, and only use one write in the function. `rec.purchase_warn_msg = new_variable`.

purchase_warn_message/README.rst
```diff
@@ -0,0 +1,74 @@
+=====================
```
ValentinVinagre, line 1: the readme don't be separate in different files?

purchase_warn_message/models/purchase_order.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2020 ForgeFlow S.L.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
```
ValentinVinagre, line 2: add https please in all documents.

ValentinVinagre (commented):
Some comments.
Functional test -> nice work 👍

purchase_warn_message/models/purchase_order.py
```diff
@@ -0,0 +1,23 @@
...
+    purchase_warn_msg = fields.Text(compute="_compute_purchase_warn_msg")
+
+    def _compute_purchase_warn_msg(self):
+        for rec in self:
+            if rec.partner_id:
+                if rec.partner_id.parent_id:
+                    p = rec.partner_id.parent_id
+                else:
+                    p = rec.partner_id
+                if rec.state not in ["done", "cancel"] and p.purchase_warn == "warning":
```
ValentinVinagre: why the sate 'purchase' is not included?

ValentinVinagre:
Hi @HviorForgeFlow ,
When I checked the https://github.com/OCA/sale-workflow/pull/1113, I see that when you calculates the field '*_warn_msg' use one logic in sale and other logic in the purchase.

I think that these modules should follow the same logic, don't you think?

HaraldPanten:
Hi @HviorForgeFlow are you still working on this?

THX!

HviorForgeFlow:
@HaraldPanten change behavior to look like OCA/sale-workflow#1113
CC @MiquelRForgeFlow @ValentinVinagre

MiquelRForgeFlow (approved):

PR #890, opened 2020-04-10
[IMP] 12.0 purchase_delivery_split_date

gurneyalex:
This changeset implements a missing part in the module description. If
you have a purchase order with several lines, confirmed, then you change
the dates on some of the line, this would be reflected in the related
stock moves but not in the stock pickings which would not be split or
merged by date.

### DIFF
--- a/purchase_delivery_split_date/models/purchase.py
+++ b/purchase_delivery_split_date/models/purchase.py
@@ -70,6 +70,49 @@ class PurchaseOrderLine(models.Model):
                 picking)
         return moves
 
+    @api.multi
+    def write(self, values):
+        res = super().write(values)
+        if 'date_planned' in values:
+            self.mapped('order_id')._check_split_pickings()
+        return res
+
+
+class PurchaseOrder(models.Model):
+    _inherit = 'purchase.order'
+
+    def _check_split_pickings(self):
+        for order in self:
+            moves = self.env['stock.move'].search([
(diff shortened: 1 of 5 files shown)


purchase_delivery_split_date/models/purchase.py
```diff
@@ -70,6 +70,49 @@ def _create_stock_moves(self, picking):
...
+    def _check_split_pickings(self):
+        for order in self:
+            moves = self.env['stock.move'].search([
+                ('purchase_line_id', 'in', self.order_line.ids),
+                ('state', 'not in', ('cancel', 'done')),
+            ])
+            pickings = moves.mapped('picking_id')
+            pickings_by_date = {}
+            for pick in pickings:
+                pickings_by_date[pick.scheduled_date.date()] = pick
```
mmequignon, line 93: Could you get multiple pickings with the same scheduled_date here ?

gurneyalex: we could with manual date changes. In that case the moves would be grouped in one of them. I think.

mmequignon: So yes, no risk to accidentally skip some pickings here.

mmequignon (approved):

i-vyshnevska (approved):

Tonow-c2c (approved):

PR #901, opened 2020-04-29
[12.0][REF] purchase_landed_cost: Rename variables

clementmbr:
This PR is linked with #897 where I opened suggestions to improve purchase_landed_cost.

Here, the objective is to :
1. Avoid confusion between product's `standard_price` (which will be modified by clicking on 'Update Cost') and the sought-after "product's Landed Cost"
2. Avoid confusion between the words 'price', 'expenses' and 'cost'. Following the idea that :
    Cost = Price + Expenses

Thus, the following changes were made :

- delete `product_price_unit` (a **related** field to stock move's
`price_unit`)
- rename `standard_price_old` into **`product_price_unit`** (a **calculated**
 field which give stock move's `price_unit`)
- rename `standard_price_new` into **`landed_cost_unit`** (and reset it to
zero when the Cost Distribution calculation is canceled)
- rename `total_amount` into **`product_price_amount`**
- rename `cost_ratio` into **`expense_unit`**

----

The question I ask myself right now is if it will be useful to add two more fields :

- `standard_price_old` which would be the real _"old product's standard_price before the stock_move"_ 
- and `standard_price_new` which would be the real _"new updated product's standard_price after clicking on 'Update Cost' "_

These two fields may be interesting, but it might be difficult to display them on the same o2m tree view which already have 9 columns :

image

### DIFF
--- a/purchase_landed_cost/__manifest__.py
+++ b/purchase_landed_cost/__manifest__.py
@@ -4,7 +4,7 @@
 
 {
     'name': 'Purchase landed costs - Alternative option',
-    'version': '12.0.1.1.0',
+    'version': '12.0.2.0.0',
     "author": "AvanzOSC,"
               "Tecnativa,"
               "Joaquín Gutierrez,"
--- a/purchase_landed_cost/i18n/es.po
+++ b/purchase_landed_cost/i18n/es.po
@@ -22,8 +22,8 @@ msgid "Affected lines"
 msgstr "Líneas afectadas"
 
 #. module: purchase_landed_cost
-#: model:ir.model.fields,field_description:purchase_landed_cost.field_purchase_cost_distribution_line__total_amount
-msgid "Amount line"
+#: model:ir.model.fields,field_description:purchase_landed_cost.field_purchase_cost_distribution_line__product_price_amount
+msgid "Price amount"
 msgstr "Importe de línea"
 
 #. module: purchase_landed_cost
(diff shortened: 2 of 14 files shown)


pedrobaeza:
If you do this, you need to include migration scripts using `openupgradelib` and calling `rename_fields` method.

clementmbr:
@pedrobaeza Yes it will be done if these changes are approved.
Do you think they make sense ?
Should I also add these 2 informative fields referring to the _"old standard_price before the stock move"_ and the _"new standard_price after the stock_move and after the added landed_cost expenses"_ ?

clementmbr:
@pedrobaeza I've just added the migration scripts.
Travis is still red here because of small flake8 problems that were corrected on #897 .

However I hope the rest of the PR is good to be reviewed now.

legalsylvain (approved):
make sense. Code review / no test.

pedrobaeza:
I see 2 renamings incorrect, putting `expenses` in plural, which is very misleading and tend to typos coding. Also, Travis should be green. If there's the need of cleaning the branch, please do it in other PR which will be merged quicker.

clementmbr:
@pedrobaeza ok i've done a very small PR for the flake8 problems : #907 

About this name for the Cost Distribution Expenses i thought `expenses_unit` and `expenses_amount` would be interesting to help differentiate them from fields related to a single expense... But I agree it can lead to typos errors coding.

I've changed them to `expense_unit` and `expense_amount` then. I'll do the rebase to make Travis green as soon as #907 is merged.

Thank you very much!

clementmbr:
@pedrobaeza now Travis is green on this PR too and I corrected the plural in `expenses`.

Hope it looks good for you now ! Let me know if there is anything more I can do to be able to merge this PR.
cc @rvalyi @legalsylvain

PR #902, opened 2020-04-30
[12.0][FIX] purchase_open_qty

AdriaGForgeFlow:
This module introduces two filters, in the Purchase Order Line search view, which use the fields:
* pending_qty_to_invoice
* pending_qty_to_receive

However, those fields are not defined in the model so every time we try to use them it gives an error. This PR fixes that issue by adding those two fields to the purchase.order.line model.

cc ~ @ForgeFlow

### DIFF
--- a/purchase_open_qty/readme/CONTRIBUTORS.rst
+++ b/purchase_open_qty/readme/CONTRIBUTORS.rst
@@ -1,2 +1,3 @@
 * Miquel Raïch <miquel.raich@eficent.com>
 * Andreas Dian Sukarno Putro <andreasdian777@gmail.com>
+* Eric Antones <eantones@nuobit.com>
--- a/purchase_open_qty/views/purchase_view.xml
+++ b/purchase_open_qty/views/purchase_view.xml
@@ -36,10 +36,10 @@
             <filter name="hide_cancelled" position="after">
                 <filter name="pending_qty_to_invoice"
                         string="Pending Qty to Bill"
-                        domain="[('pending_qty_to_invoice','=',True)]"/>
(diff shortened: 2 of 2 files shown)


AaronHForgeFlow (approved):
:+1: Code Review. LGTM.

AaronHForgeFlow:
Maybe the new fields are not necessary, but changing the field names :thinking: https://github.com/OCA/purchase-workflow/pull/910

eantones:
@AdriaGForgeFlow  you are duplicating code and introducing an unnecessary computation, you don't really need a computed field at line level because you already have the data stored on qty_to_invoice and qty_to_receive. In my opinion the approach on https://github.com/OCA/purchase-workflow/pull/910 is more efficient, simpler and has much less code so much less risk to introduce bugs.
cc @AaronHForgeFlow 
Thanks! :-)

AdriaGForgeFlow:
> @AdriaGForgeFlow you are duplicating code and introducing an unnecessary computation, you don't really need a computed field at line level because you already have the data stored on qty_to_invoice and qty_to_receive. In my opinion the approach on #910 is more efficient, simpler and has much less code so much less risk to introduce bugs.
> cc @AaronHForgeFlow
> Thanks! :-)

Will do, thanks for the suggestion

AdriaGForgeFlow:
@AaronHForgeFlow @eantones  Can you review again?

eantones (approved):

MiquelRForgeFlow (approved):

AaronHForgeFlow (approved):
Still LGTM

PR #903, opened 2020-04-30
[13.0][ADD] purchase_order_line_packaging_qty

MiquelRForgeFlow:
Inspired in https://github.com/OCA/sale-workflow/pull/1114.

### DIFF
--- a/purchase_order_line_packaging_qty/__init__.py
+++ b/purchase_order_line_packaging_qty/__init__.py
@@ -1,0 +1,1 @@
+from . import models
--- a/purchase_order_line_packaging_qty/__manifest__.py
+++ b/purchase_order_line_packaging_qty/__manifest__.py
@@ -1,0 +1,18 @@
+# Copyright 2020 Camptocamp SA
+# Copyright 2020 ForgeFlow, S.L.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+{
+    "name": "Purchase Order Line Packaging Quantity",
+    "summary": "Define quantities according to product packaging"
+    " on purchase order lines",
+    "version": "13.0.1.0.0",
+    "development_status": "Alpha",
+    "category": "Warehouse Management",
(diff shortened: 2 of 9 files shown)


JordiBForgeFlow:
cc @jgrandguillaume @didierdonze

JordiBForgeFlow:
@MiquelRForgeFlow When you create a PO and enter an initial quantity and package you cannot change neither the quantity in normal units or the package quantity. It stays in the original value. Can you review?

MiquelRForgeFlow:
@JordiBForgeFlow could you check again, please?

jgrandguillaume:
Hi @MiquelRForgeFlow ,

Thanks for the PR. Functional test, I see that the order of fields is different from the last version we have on the sale here https://github.com/OCA/sale-workflow/pull/1114 (see attached image).

Screenshot from 2020-06-02 11-12-30

Naturally, you always prefer to 1st chose the packaging, then its quantity which update the unit quantity.

Could you adapt this ?

jgrandguillaume (changes requested):
See my last comment

MiquelRForgeFlow:
@jgrandguillaume done

grindtildeath (commented):
Looks ok technically.
However, since most of the code is the same that I wrote in https://github.com/OCA/sale-workflow/pull/1114 , I feel like dropping the Camptocamp copyright and relicensing to LGPL is not right, and that authorship must be respected. :slightly_smiling_face:

MiquelRForgeFlow:
@grindtildeath is it ok now?

jgrandguillaume:
Hi @MiquelRForgeFlow 

Still need to move it there:

Screenshot from 2020-06-02 14-22-06

otherwie good

MiquelRForgeFlow:
ready

grindtildeath (approved):

jgrandguillaume (approved):
LGTM now, thanks @MiquelRForgeFlow

PR #904, opened 2020-04-30
[ADD] purchase_propagate_qty

gurneyalex:
This module propagates qty decrease on purchase lines to the stock.moves


purchase_propagate_qty/models/purchase.py
```diff
@@ -0,0 +1,32 @@
...
+
+_logger = logging.getLogger(__name__)
+
+
+class PurchaseOrderLine(models.Model):
+    _inherit = "purchase.order.line"
+
+    @api.multi
+    def write(self, values):
+        res = super().write(values)
```
mmequignon, line 14: Couldn't it be possible to get rid of that local variable since it's not used in the method ?

mmequignon (commented):
In the module description, you talk about the default behaviour when qty is increased on a purchase line.
Is this behaviour still active ? Correct me if I'm wrong, but I think your module handle both cases already.

purchase_propagate_qty/models/purchase.py
gurneyalex: which one? `res` ? I return it at the end of the method (on line 32)

mmequignon: Yes, I mean, why not directly return super().write(values) ?

gurneyalex:
> In the module description, you talk about the default behaviour when qty is increased on a purchase line.
> Is this behaviour still active ? Correct me if I'm wrong, but I think your module handle both cases already.

I updated the description. The behavior in case of an increase is slightly the same as in normal Odoo (except that the change is propagated on chained moves)

purchase_propagate_qty/models/purchase.py
```diff
@@ -0,0 +1,28 @@
...
+        for line in self:
+            if line.state != 'purchase':
+                continue
```
legalsylvain, line 22: if you replace by ``for line in self.filtered(lambda x: x.state != 'purchase'):`` it could be faster and more synthetic, don't you think ?

purchase_propagate_qty/models/purchase.py
```diff
@@ -0,0 +1,28 @@
...
+            for move in line.move_dest_ids | line.move_ids:
+                if move.state in ('cancel', 'done'):
+                    continue
+                if move.product_id != line.product_id:
```
i-vyshnevska, line 26: do you need also check for uom? there can be 2 same products with different uom's

purchase_propagate_qty/models/purchase.py
gurneyalex: Actually I don't think so: `filtered()` does a loop, calling a function each time (see https://github.com/odoo/odoo/blob/13.0/odoo/models.py#L5243): This means 2 loops instead of 1 and N additional function calls. I have not benchmarked function calls in Python3 but this used to be much slower.

purchase_propagate_qty/models/purchase.py
gurneyalex: @i-vyshnevska I don't think so: I'm checking the stock moves related to the purchase line -> 1. they should have the same uom and 2. I'm working with the fields which are converted in the product's default unit, so I believe I'm safe.

purchase_propagate_qty/models/purchase.py
legalsylvain: Ok ! Thanks. I didn't know !

i-vyshnevska (approved):

purchase_propagate_qty/models/purchase.py
legalsylvain: It's not the same algorithm. (Not the same order)

houssine78 (approved):

HaraldPanten (approved):

dreispt:
Travis red. Rebuilding to recheck.

HviorForgeFlow:
@legalsylvain @mmequignon can merge be proceed?

legalsylvain:
@HviorForgeFlow 

sorry, I was on holidays. yes you can. My question was answered. I think it's ok now.

kind regards.

PR #905, opened 2020-04-30
[IMP] [12.0] purchase_delivery_split_date

gurneyalex:
* [FIX] when adding a new line on a confirmed PO, split the delivery (this was
  done only if a date was changed on an existing line)

* [IMP] when the quantity on a line is changed, the onchange would reset the
  planned date -> change this to prevent setting a date earlier than the one on
  the line, since if we are using this module the user probably has manually
  set the date first

### DIFF
--- a/purchase_delivery_split_date/__manifest__.py
+++ b/purchase_delivery_split_date/__manifest__.py
@@ -4,7 +4,7 @@
 
 {
     "name": "Purchase Delivery Split Date",
-    "version": "12.0.2.0.0",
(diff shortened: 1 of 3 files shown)


gurneyalex:
See also #904 : new module to propagate qty decrease on PO lines, which comes in handy if you want to split a PO line

guewen (approved):

HaraldPanten (approved):

hparfr (approved):

HviorForgeFlow (approved):

PR #908, opened 2020-05-05
[12.0] Add module purchase_reception_status

alexis-via:
I may have missed something, but I didn't find a simple module that would add the reception status of the purchase order. The OCA module purchase_picking_state by @chafique-delli  could do the job, but it depends on purchase_stock. On the long term, I think this module could replace the module purchase_picking_state.

### DIFF
--- a/purchase_reception_status/__manifest__.py
+++ b/purchase_reception_status/__manifest__.py
@@ -1,0 +1,19 @@
+# Copyright 2020 Akretion France (http://www.akretion.com/)
+# @author: Alexis de Lattre <alexis.delattre@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Purchase Reception Status',
+    'version': '12.0.1.0.0',
+    'category': 'Purchases',
+    'license': 'AGPL-3',
+    'summary': 'Add reception status on purchase orders',
+    'author': 'Akretion,Odoo Community Association (OCA)',
+    'maintainers': ['alexis-via'],
+    'website': 'https://github.com/OCA/purchase-workflow',
+    'depends': ['purchase'],
+    'data': [
+        'views/purchase_order.xml',
+    ],
+    'installable': True,
+}
--- a/purchase_reception_status/models/purchase_order.py
+++ b/purchase_reception_status/models/purchase_order.py
@@ -1,0 +1,50 @@
+# Copyright 2020 Akretion France (http://www.akretion.com/)
+# @author: Alexis de Lattre <alexis.delattre@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+from odoo.tools import float_compare
+
+
+class PurchaseOrder(models.Model):
+    _inherit = 'purchase.order'
+
+    reception_status = fields.Selection([
+        ('no', 'Nothing Received'),
+        ('partial', 'Partially Received'),
+        ('received', 'Fully Received'),
+        ], compute='_compute_reception_status',
+        string='Reception Status', store=True)
+    force_received = fields.Boolean(
+        string='Force Received',
+        readonly=True, states={'done': [('readonly', False)]}, copy=False,
+        help="If true, the reception status will be forced to Fully Received, "
+        "even if some lines are not fully received. "
+        "To be able to modify this field, you must first lock the order.")
+
+    @api.depends(
+        'state', 'force_received',
+        'order_line.qty_received', 'order_line.product_qty')
+    def _compute_reception_status(self):
+        prec = self.env['decimal.precision'].precision_get(
+            'Product Unit of Measure')
+        for order in self:
+            status = 'no'
+            if order.state in ('purchase', 'done'):
+                if order.force_received:
+                    status = 'received'
+                elif all([
+                        float_compare(
+                            line.qty_received,
+                            line.product_qty,
+                            precision_digits=prec) >= 0
+                        for line in order.order_line]):
+                    status = 'received'
+                elif any([
+                        float_compare(
+                            line.qty_received,
+                            0,
+                            precision_digits=prec) > 0
+                        for line in order.order_line]):
+                    status = 'partial'
+            order.reception_status = status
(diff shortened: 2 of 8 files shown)


rvalyi (approved):

alexis-via:
My only concern is when you have a PO line for shipping costs. This is a service but you certainly don't want to manually write the received qty manually on the PO line (as you would do with other services). Any ideas welcomed.

quentinDupont:
Works fine :white_check_mark: 
I have a question when you use stock. Maybe it's out of your scope but it's a bit confusing :+1: 

- Confirme purcharse order
- On PO : lock and check Fully Received
- Go to picking
- On Picking : write partial received quantity and validate
- On PO : Now you have your PO as "Fully Received" but with partial quantity :man_shrugging:

purchase_reception_status/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+{
+    'name': 'Purchase Reception Status',
+    'version': '12.0.1.0.0',
+    'category': 'Purchases',
+    'license': 'AGPL-3',
+    'summary': 'Add reception status on purchase orders',
+    'author': 'Akretion,Odoo Community Association (OCA)',
+    'maintainers': ['alexis-via'],
+    'website': 'https://github.com/OCA/purchase-workflow',
+    'depends': ['purchase'],
```
ValentinVinagre, line 14: Add 'stock' too, because if not, it doesn't make sense

purchase_reception_status/models/purchase_order.py
```diff
@@ -0,0 +1,50 @@
...
+
+from odoo import api, fields, models
+from odoo.tools import float_compare
+
+
+class PurchaseOrder(models.Model):
+    _inherit = 'purchase.order'
+
+    reception_status = fields.Selection([
+        ('no', 'Nothing Received'),
```
ValentinVinagre, line 13: add a new option in case the purchase order only has services:
 ('no_rec', 'Nothing to Received'),

purchase_reception_status/models/purchase_order.py
```diff
@@ -0,0 +1,50 @@
...
+        'order_line.qty_received', 'order_line.product_qty')
+    def _compute_reception_status(self):
+        prec = self.env['decimal.precision'].precision_get(
+            'Product Unit of Measure')
+        for order in self:
+            status = 'no'
+            if order.state in ('purchase', 'done'):
+                if order.force_received:
+                    status = 'received'
+                elif all([
```
ValentinVinagre, line 36: Add the new elif. This indicate the case only services, we don't have "Nothing to Received"
`elif all(prod.type == 'service' for prod in order.order_line.product_id):
                    status = 'no_rec'`

purchase_reception_status/models/purchase_order.py
```diff
@@ -0,0 +1,50 @@
...
+                        float_compare(
+                            line.qty_received,
+                            line.product_qty,
+                            precision_digits=prec) >= 0
+                        for line in order.order_line]):
```
ValentinVinagre, line 41: filter services:
`for line in order.order_line.filtered(lambda x: x.product_id.type != 'service')]):`

purchase_reception_status/models/purchase_order.py
```diff
@@ -0,0 +1,50 @@
...
+                    status = 'received'
+                elif any([
+                        float_compare(
+                            line.qty_received,
+                            0,
+                            precision_digits=prec) > 0
+                        for line in order.order_line]):
```
ValentinVinagre, line 48: filter services:
`for line in order.order_line.filtered(lambda x: x.product_id.type != 'service')]):`

purchase_reception_status/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,5 @@
+This module adds a field *Receiption Status* on purchase orders. On a confirmed purchase order, it can have 3 different values:
+
+* Nothing Received
```
ValentinVinagre, line 3: add the new option

ValentinVinagre (commented):
Hi @alexis-via ,
I propose some changes and issue, what do you think?

purchase_reception_status/__manifest__.py
alexis-via: I disagree. It makes sense event without stock. When you purchase a service, you may want to follow the "reception" of this service by manually entering the "received qty", then the field "reception_status" is still usefull.

purchase_reception_status/models/purchase_order.py
alexis-via: I disagree, for the same reason as above.

purchase_reception_status/models/purchase_order.py
alexis-via: Same answer

purchase_reception_status/models/purchase_order.py
alexis-via: same answer

purchase_reception_status/models/purchase_order.py
alexis-via: same answer

purchase_reception_status/readme/DESCRIPTION.rst
alexis-via: same answer

alexis-via:
@quentinDupont If you check the option "Force fully received", it means that you want to force it to that status, so you can't compain if it displays Partially received !!! I'm afraid you misundertood the purpose of this option "Force received".

quentinDupont:
> @quentinDupont If you check the option "Force fully received", it means that you want to force it to that status, so you can't compain if it displays Partially received !!! I'm afraid you misundertood the purpose of this option "Force received".

I understand the purpose of this module, I was just pointing out a possible user error. Sometimes I prefer not to have only the possibility to answer the user "you misunderstood and can't complain" and I prefer to improve the design of the software.
Anyway, this is not a blocking point for me, and this module does what it was designed to do :+1:

quentinDupont (approved):

purchase_reception_status/__manifest__.py
ValentinVinagre: humm okay 👍

alexis-via:
@quentinDupont FYI, the option "Force received" is only available to Purchase Manager, not to regular purchase users.

PR #912, opened 2020-05-11
[13.0][MIG] purchase_triple_discount

ps-tubtim:
Standard Migration.

This PR is waiting for OCA/account-invoicing#661, you can always use this for readability.

- [ ] [13.0][MIG] account_invoice_triple_discount


ps-tubtim:
As most people agree to move to this approach #930, so I would like to close this for now.

PR #923, opened 2020-06-06
[12.0] [FIX] purchase_request: disable creation from form view

sswapnesh:
New purchase request line should not be created from `form` view as well.

Create is already disable on list view at https://github.com/OCA/purchase-workflow/blob/12.0/purchase_request/views/purchase_request_line_view.xml#L9

PR #924, opened 2020-06-06
[13.0] [FIX] purchase_request: disable creation from form view

sswapnesh:
New purchase request line should not be created from form view as well.

Create is already disabled on list view at https://github.com/OCA/purchase-workflow/blob/13.0/purchase_request/views/purchase_request_line_view.xml#L11


AaronHForgeFlow (approved):
:+1: Makes sense to me

HviorForgeFlow (approved):

sswapnesh:
Could also look at 12's PR https://github.com/OCA/purchase-workflow/pull/923?
(As we need the fix for v12 originally)

PR #925, opened 2020-06-06
[FIX] purchase_discount: set currency rate correctly

sswapnesh:
As mentioned on https://github.com/OCA/purchase-workflow/pull/923#issuecomment-640065252

Before this commit, we was creating new currency rate for same date which was triggering https://github.com/odoo/odoo/commit/01142701cc566b3084326af9649fd85db02b4220 constraints so the tests were failing.

No we search for currency rate of the same day and set rate to `1.0` or create new rate.


sswapnesh:
@pedrobaeza

sswapnesh:
@pedrobaeza

pedrobaeza (approved):

PR #926, opened 2020-06-06
[13.0] [FIX] purchase_discount: set currency rate correctly

sswapnesh:
As mentioned on #923 (comment)

Before this commit, we was creating new currency rate for same date which was triggering odoo/odoo@0114270 constraints so the tests were failing.

No we search for currency rate of the same day and set rate to `1.0` or create new rate.

### DIFF
--- a/purchase_discount/tests/test_purchase_discount.py
+++ b/purchase_discount/tests/test_purchase_discount.py
@@ -19,15 +19,26 @@ class TestPurchaseOrder(common.SavepointCase):
         )
         cls.product_2 = product_obj.create({"name": "Test product 2"})
         po_model = cls.env["purchase.order.line"]
+        currency_rate_model = cls.env["res.currency.rate"]
         # Set the Exchange rate for the currency of the company to 1
         # to avoid issues with rates
-        cls.env["res.currency.rate"].create(
-            {
-                "currency_id": cls.env.user.company_id.currency_id.id,
-                "rate": 1.00,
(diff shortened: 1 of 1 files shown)


purchase_discount/tests/test_purchase_discount.py
```diff
@@ -21,13 +21,17 @@ def setUpClass(cls):
...
         # Set the Exchange rate for the currency of the company to 1
         # to avoid issues with rates
-        cls.env["res.currency.rate"].create(
-            {
-                "currency_id": cls.env.user.company_id.currency_id.id,
-                "rate": 1.00,
-                "name": fields.Date.today(),
-            }
-        )
+        latest_currency_rate_line = currency_rate_model.search([
```
pedrobaeza: No `currency_rate_model` defined before

sswapnesh: Bad copy-paste, Fixing now

sswapnesh: Fixed, 
(PS: Wondering why it was not fixed earlier :thinking: )

pedrobaeza:
Please check Travis as it seems there's still a problem.

sswapnesh:
~Not sure what could be the reason~
~(PS: I have not setp up black/pre-commit yet, so I can't go in-depth for now)~

sswapnesh:
Can be merged now

PR #928, opened 2020-06-15
[ADD] purchase_delays

mmequignon:
This modules adds a transport lead time to the origin supplier delay.


purchase_delays/readme/HISTORY.rst
```diff
@@ -0,0 +1,4 @@
+13.0.1.0.0 (2020-06-15)
```
i-vyshnevska: nobody uses History file, mostly we are droping it

purchase_delays/models/product_supplierinfo.py
```diff
@@ -0,0 +1,32 @@
...
+    transport_delay = fields.Integer(
+        string="Transport Lead Time", default=0, required=True
+    )
+
+    @api.depends("supplier_delay", "transport_delay")
+    def _compute_delay(self):
+        for record in self:
+            record.delay = record.supplier_delay + record.transport_delay
+
+    def _inverse_delay(self):
```
i-vyshnevska: this inverse shouldn't be here, you shouldn't update fields on which you are computing this field

mmequignon: Yeah, but the build is failing since `supplier_calendar` is trying to set `delay`

mmequignon: What I can do is to create a new field instead.

pedrobaeza:
Please don't use plurals in module name.

purchase_delays/models/product_supplierinfo.py
i-vyshnevska: then you need to add to the description it's not compatible with supplier_calendar, and AFAIS the problem in the test not in the model, you may need to adapt the test

Tonow-c2c (commented):
You can easily add tests :-)

purchase_delays/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,2 @@
+This module overrides the `product.supplierinfo` model.
+It splits the delay into two new delay fields, `Supplier Lead Time` and `Transport Lead Time`, to sum them into the actual `Lead Time`
```
jgrandguillaume: I would precise:
"This allow users to get the information about the transportation lead time in order to plan the transport properly when relevant."

purchase_delays/models/product_supplierinfo.py
jgrandguillaume: @mmacross Not sure how to solve this technically, but on a functional side it is correct that if you use the supplier_calendar it should set the supplier_delay (not the delay) if you use this module. Because calendar is related to when your supplier works, not transport.

cc @i-vyshnevska @LoisRForgeFlow

jgrandguillaume (changes requested):
What about naming the module "vendor_transport_lead_time" to be in-line with convention raised by @pedrobaeza

jgrandguillaume:
@mmequignon This module don't need to be compatible with supplier-calendar ones. So just metnion in the Readme. cc @i-vyshnevska

purchase_delays/models/product_supplierinfo.py
mmequignon: I have set that inverse function because of the tests that was crashing because they was trying to set delay.
I'm not sure it's necessary to keep it if I declare those two modules as incompatible.

i-vyshnevska: we don't need to be compatible with supplier_calendar but as it installed after the test will fail

jgrandguillaume: @i-vyshnevska Can you detail why it's bad to get an inver on delay? I personally think it's a good thing. What to I miss ? 
My proposal @mmequignon would be setting the delay would set the supplier delay and keep transport to 0.

jgrandguillaume: Real math here ;) If delay i set then:
 Supplier delay = delay - transport delay
 transport delay is kept

That's how I would do

i-vyshnevska (approved):

purchase_delays/models/product_supplierinfo.py
i-vyshnevska: @jgrandguillaume  because delay is computed field based on two other delays, if inverse should update transport_delay, than an update of transport_delivery should update delay and goes to eternity, odoo framework prevent this on some level but this looks wrong for me
@mmequignon you need to add a flag on test on supplier_calendar -post_install

jgrandguillaume: Understood @i-vyshnevska  thanks. @mmequignon Will call you to discuss it.

mmequignon:
@pedrobaeza @jgrandguillaume @Tonow-c2c 
Does it looks better to you ?

jgrandguillaume (approved):
LGTM thanks a lot

marcelsavegnago (approved):

PR #917, opened 2020-05-23
[13.0] [MIG] purchase_reception_status

Rad0van:
This is simple port of purchase_receptions_status to 13.0

PR #934, opened 2020-07-07
[13.0] Migrated Purchase Reception Status to v13

kayue:


kayue:
Sorry duplicated with #917.

pedrobaeza:
And with no commit history, so closing.

PR #939, opened 2020-07-10
[13.0][FIX] purchase_work_acceptance, no purchase_line_id in wa_line coz of readonly flag

kittiu:
When PO -> create WA. The readonly nature of reference purchase_line_id cause no link to WA because purchase_line_id was not written when saved.

Note: I surprise why it works before, but this should be very simple fix to fast forward.


etobella (approved):
LGTM

marcelsavegnago (approved):

ps-tubtim (approved):

PR #944, opened 2020-07-16
[FIX] purchase_isolated_rfq, add uninstall_hook to restore original action domain/context

kittiu:


HviorForgeFlow:
functional review: Works as expected

What I noticed RFQs are moved to Purchase Order status, this can lead to duplicated information, not sure how to deal with it. 

Maybe moving RFQ to cancel on uninstall as the link between RFQ and PO is lost when uninstalling module

To keep data consistency if module is uninstalled and installed maybe a custom column creating like OU ones on uninstall to keep the link and moving the RFQ to cancel and reading if found on installation. But then it is also required to match the RFQ sequence... Anyway just thinking on weirdest use case. It can be kept like this

I think move RFQs to Done status is not a good approach as they are counted as Purchase Orders on Purchase Reporting.
What about move them to achieved once they are validated? 

Just a few thoughts on that :smile: 

@kittiu

kittiu:
@HviorForgeFlow thank you for a thoughtful comments. ;)
Conceptual, I just think that whatever RFQ has been converted to PO, should be cancelled (as it already has new life).
Will try and see.

Saran440 (approved):
Functional Test :+1:

Saran440:
@HviorForgeFlow Could you review, please? :)

HviorForgeFlow (approved):

PR #948, opened 2020-07-21
[12.0][FIX] purchase_order_line_price_history: get line product

ernestotejeda:
cc @Tecnativa TT24733


pedrobaeza:
Fast-tracking this patch, as the module is useless without it: /ocabot merge patch

PR #949, opened 2020-07-23
[12.0][FIX] purchase_work_acceptance: name repeat in core

ps-tubtim:
Edit name in `res_config_settings_views.xml ` because of repeat in core. (https://github.com/odoo/odoo/blob/909f2aaaaae34c7ef7b69d62614bd7acfde74591/addons/purchase/views/res_config_settings_views.xml#L16)


kittiu (approved):
Thanks

HviorForgeFlow (approved):

PR #953, opened 2020-07-28
[12.0][ADD] purchase_order_line_price_history_discount: new module

ernestotejeda:
cc @Tecnativa TT24924


CarlosRoca13:
When I update the product price of a product with the history, the discount is not added. I don't know if that should happen, but could be a good upgrade.

ernestotejeda:
Thanks @CarlosRoca13 , I have talked with @pedrobaeza and we think it may be an update, so I did it. Take a look at this, please.
I also add a unit test.

purchase_order_line_price_history_discount/wizards/purchase_order_line_price_history.py
```diff
@@ -0,0 +1,14 @@
...
+from odoo import fields, models
+
+
+class PurchaseOrderLinePriceHistoryline(models.TransientModel):
+    _inherit = "purchase.order.line.price.history.line"
+
+    discount = fields.Float(related="purchase_order_line_id.discount")
+
+    def _prepare_purchase_order_line_vals(self):
+        vals = super()._prepare_purchase_order_line_vals()
```
Tardo, line 12:
```suggestion
    def _prepare_purchase_order_line_vals(self):
        self.ensure_one()
        vals = super()._prepare_purchase_order_line_vals()
```

Tardo (changes requested):

purchase_order_line_price_history_discount/wizards/purchase_order_line_price_history.py
ernestotejeda: I think in this case it isn't necessary. When the method is called from super(), `self.ensure_one()` will be executed.
https://github.com/OCA/purchase-workflow/pull/953/files#diff-b595047d950f62cc98957ba740167de9R110

pedrobaeza: I agree

CarlosRoca13 (approved):
LGTM :+1:

pedrobaeza (approved):
Code review

PR #954, opened 2020-07-28
[13.0][MIG] purchase_order_line_price_history: Migration to 13.0

ernestotejeda:
cc @Tecnativa TT24923


HviorForgeFlow (commented):
Functional Review: Works as expected

Could be feasible to move the set price button to the Purchase history tree view? Or at least drop the default Save ... Close buttons on the History Lines popup.

CarlosRoca13 (approved):
Functional and code review :+1:

ernestotejeda:
Thanks @HviorForgeFlow, I have dropped the default Save buttons on the History Lines popup. 
And I also added the hook method added here in v12 https://github.com/OCA/purchase-workflow/pull/953

purchase_order_line_price_history/wizards/purchase_order_line_price_history.py
```diff
@@ -0,0 +1,95 @@
...
+            if self.include_commercial_partner:
+                commercial_ids = self.partner_id.commercial_partner_id.ids
+                domain += [("partner_id", "child_of", commercial_ids)]
+            else:
+                domain += [("partner_id", "child_of", self.partner_id.ids)]
+        vals = []
+        order_lines = self.env["purchase.order.line"].search(domain, limit=20)
+        order_lines -= self.purchase_order_line_id
+        for order_line in order_lines:
+            vals.append((0, False, {"purchase_order_line_id": order_line.id}))
```
Tardo:
```suggestion
        vals = [(0, False, {"purchase_order_line_id": order_line.id}) for order_line in order_lines]
```

Tardo (approved):
:+1: LGTM. Code review... only non-blocking comments.

ernestotejeda:
@Tardo Changes done
Sometimes I think that in some cases it is more readable if we don't use list comprehension, but it is certainly a guideline to use them, so let use them. Thanks.

HviorForgeFlow (approved):

PR #960, opened 2020-07-31
[12.0][REF+FIX] purchase_order_line_price_history*: Refactor tests for using same base

pedrobaeza:
This way, we use DRY principle and avoid difficulties on migrations/maintenance.

Revisiting tests for *_discount module, some asserts are not needed, as they are from base module, and added a relevant one about wizard line.

@Tecnativa


CarlosRoca13 (approved):
It looks better :+1:

yajo (approved):

ernestotejeda (approved):

marcelsavegnago (approved):

PR #962, opened 2020-08-09
[IMP] procurement_purchase_no_grouping: Add company level policy

Rad0van:
This is my attempt to create a slightly enhanced version of procurement_purchase_no_grouping. The idea here is to provide also System default setting that will cause the logic fall back to system-wide setting set up in System settings. The reason for this is we have more than 250k products in more than 300 categories so setting up each category would be rather laborious. This should also work for existing users - they only need to setup the system-wide default.

### DIFF
--- a/procurement_purchase_no_grouping/__manifest__.py
+++ b/procurement_purchase_no_grouping/__manifest__.py
@@ -10,7 +10,7 @@
     "website": "https://github.com/OCA/purchase-workflow",
     "category": "Procurements",
     "depends": ["purchase_stock"],
-    "data": ["views/product_category_view.xml"],
+    "data": ["views/product_category_view.xml", "views/res_config_settings_views.xml"],
     "installable": True,
     "license": "AGPL-3",
 }
--- a/procurement_purchase_no_grouping/models/res_company.py
+++ b/procurement_purchase_no_grouping/models/res_company.py
@@ -1,12 +1,11 @@
-# Copyright 2015 AvanzOsc (http://www.avanzosc.es)
-# Copyright 2015-2016 - Pedro M. Baeza <pedro.baeza@tecnativa.com>
+# Copyright 2020 - Radovan Skolnik <radovan@skolnik.info>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
 
 from odoo import fields, models
 
 
-class ProductCategory(models.Model):
-    _inherit = "product.category"
+class ResCompany(models.Model):
+    _inherit = "res.company"
 
     procured_purchase_grouping = fields.Selection(
         [
@@ -15,15 +14,15 @@ class ProductCategory(models.Model):
             ("order", "No order grouping"),
         ],
         string="Procured purchase grouping",
+        default="standard",
         help="Select the behaviour for grouping procured purchases for the "
         "the products of this category:\n"
-        "* Standard grouping (default): Procurements will generate "
+        "* Standard grouping: Procurements will generate "
         "purchase orders as always, grouping lines and orders when "
         "possible.\n"
         "* No line grouping: If there are any open purchase order for "
         "the same supplier, it will be reused, but lines won't be "
         "merged.\n"
         "* No order grouping: This option will prevent any kind of "
-        "grouping.\n"
-        "* <empty>: If no value is selected, system-wide default will be used.",
+        "grouping.",
     )
--- a/procurement_purchase_no_grouping/models/res_config_settings.py
+++ b/procurement_purchase_no_grouping/models/res_config_settings.py
@@ -1,0 +1,15 @@
+# Copyright 2020 - Radovan Skolnik <radovan@skolnik.info>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    """Configuration of default value for procurement purchase grouping."""
+
+    _inherit = "res.config.settings"
+    _description = "Procurement purchase grouping settings"
+
+    procured_purchase_grouping = fields.Selection(
+        related="company_id.procured_purchase_grouping", readonly=False,
+    )
--- a/procurement_purchase_no_grouping/tests/test_procurement_purchase_no_grouping.py
+++ b/procurement_purchase_no_grouping/tests/test_procurement_purchase_no_grouping.py
@@ -48,6 +48,9 @@ class TestProcurementPurchaseNoGrouping(common.SavepointCase):
         )
         self.stock_rule._run_buy([(procurement, rule)])
 
+    def _set_system_grouping(self, grouping):
+        self.env.company.procured_purchase_grouping = grouping
+
     def test_procurement_grouped_purchase(self):
         self.category.procured_purchase_grouping = "standard"
         self._run_procurement()
@@ -70,6 +73,41 @@ class TestProcurementPurchaseNoGrouping(common.SavepointCase):
 
     def test_procurement_no_grouping_order_purchase(self):
         self.category.procured_purchase_grouping = "order"
+        self._run_procurement()
+        self._run_procurement()
+        orders = self.env["purchase.order"].search([("origin", "=", self.origin)])
+        self.assertEqual(len(orders), 2, "Procured purchase orders are the same")
+        self.assertEqual(
+            len(orders.mapped("order_line")),
+            2,
+            "Procured purchase orders lines are the same",
+        )
+
+    def test_procurement_system_grouped_purchase(self):
+        self.category.procured_purchase_grouping = None
+        self._set_system_grouping("standard")
+        self._run_procurement()
+        self._run_procurement()
+        orders = self.env["purchase.order"].search([("origin", "=", self.origin)])
+        self.assertEqual(len(orders), 1, "Procured purchase orders are not the same")
+        self.assertEqual(
+            len(orders.order_line), 1, "Procured purchase orders lines are not the same"
+        )
+
+    def test_procurement_system_no_grouping_line_purchase(self):
+        self.category.procured_purchase_grouping = None
+        self._set_system_grouping("line")
+        self._run_procurement()
+        self._run_procurement()
+        orders = self.env["purchase.order"].search([("origin", "=", self.origin)])
+        self.assertEqual(len(orders), 1, "Procured purchase orders are not the same")
+        self.assertEqual(
+            len(orders.order_line), 2, "Procured purchase orders lines are the same"
+        )
+
+    def test_procurement_system_no_grouping_order_purchase(self):
+        self.category.procured_purchase_grouping = None
+        self._set_system_grouping("order")
         self._run_procurement()
         self._run_procurement()
         orders = self.env["purchase.order"].search([("origin", "=", self.origin)])
--- a/procurement_purchase_no_grouping/README.rst
+++ b/procurement_purchase_no_grouping/README.rst
@@ -26,7 +26,8 @@ Procurement Purchase No Grouping
 |badge1| |badge2| |badge3| |badge4| |badge5| 
 
 This module allows to not group generated purchase orders from procurements.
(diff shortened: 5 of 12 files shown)


Rad0van:
@pedrobaeza, @sergio-teruel  and @carlosdauden could maybe have a look and leave some comments? Thank you.

procurement_purchase_no_grouping/__manifest__.py
```diff
@@ -5,12 +5,12 @@
 
 {
     "name": "Procurement Purchase No Grouping",
-    "version": "13.0.1.0.0",
+    "version": "13.0.1.1.0",
```
pedrobaeza: Don't bump version yourself, but let the ocabot to do it for you on merge.

procurement_purchase_no_grouping/models/product_category.py
```diff
@@ -10,15 +10,18 @@ class ProductCategory(models.Model):
...
     procured_purchase_grouping = fields.Selection(
         [
+            ("system", "System default"),
             ("standard", "Standard grouping"),
             ("line", "No line grouping"),
             ("order", "No order grouping"),
         ],
         string="Procured purchase grouping",
-        default="standard",
+        required=True,
```
pedrobaeza: Please don't change this: not putting anything is for using the upper configuration. The criteria order should be: check the product, if no policy, check the category, and if not, go to system parameter.

pedrobaeza (changes requested):

procurement_purchase_no_grouping/models/product_category.py
Rad0van: @pedrobaeza not sure I follow this one. My findings looking into database is that when you install the module it will fill in the default value to all product categories. You can then change it to None but that kind of defeats my use case (which might be specific I admit). So if None is used to indicate to go one level higher then I guess there should be no default value. Or am I missing something?

pedrobaeza: Yes, both things: I don't want any default value nor being required.

Rad0van: > Yes, both things: I don't want any default value nor being required.

OK, that makes sense. So None will fall back to system default. Will do right away.

Rad0van:
@pedrobaeza can you please help me here? Tests in some other modules fail as if there was no system setting - looks like this usually:

`File "/home/travis/build/OCA/purchase-workflow/purchase_request/models/stock_rule.py", line 92, in _run_buy
return super(StockRule, self)._run_buy(procurements)
File "/home/travis/build/OCA/purchase-workflow/procurement_purchase_no_grouping/models/stock_rule.py", line 19, in _run_buy
)["procured_purchase_grouping"]
KeyError: 'procured_purchase_grouping'`

I am scratching my head here - not understanding why this works in my module... Any help would be appreciated. Thank you.

HviorForgeFlow:
I think the problem is that product categories that are loaded before you install the module do not fetch default value correctly.
Maybe you need some init hook to initialize that information... not sure, maybe @pedrobaeza can confirm this issue.
@Rad0van 

I would store that config on company level, but maybe there is another way to do it...

Rad0van:
> I think the problem is that product categories that are loaded before you install the module do not fetch default value correctly.
> Maybe you need some init hook to initialize that information... not sure, maybe @pedrobaeza can confirm this issue.
> @Rad0van
> 
> I would store that config on company level, but maybe there is another way to do it...

@HviorForgeFlow hmm, but it seems to me it is failing on retrieving the system setting. Something like that happened once to me when I was testing - that initially the value wasn't there for some reason even it had default. But this happens randomly through the tests.

procurement_purchase_no_grouping/models/res_config_settings.py
```diff
@@ -0,0 +1,34 @@
...
+
+    default_procured_purchase_grouping = fields.Selection(
+        [
+            ("standard", "Standard grouping"),
+            ("line", "No line grouping"),
+            ("order", "No order grouping"),
+        ],
+        string="Procured purchase grouping",
+        default="standard",
+        default_model="product.category",
```
pedrobaeza: You should store this other way, as this means that new categories will take this value instead of empty one, which can't be the desired behavior, as it's not the same having "system wide grouping" than the last default that it was. I think a better granularity is to have this selection at res.company, at put here a related field.

Rad0van: > You should store this other way, as this means that new categories will take this value instead of empty one, which can't be the desired behavior, as it's not the same having "system wide grouping" than the last default that it was. I think a better granularity is to have this selection at res.company, at put here a related field.

@pedrobaeza will (try to) do. But does this really solve the failed tests? In them it seems as if the system settings in some of the tests wasn't there at all.

pedrobaeza: Well, other approach, other tests, so it's worth trying. Anyway, when you have it, I'll see specifically possible problems

Rad0van:
@pedrobaeza well it seems to work now - you were right ;-) Now, should I also rework it as you suggested - i.e. put the setting into res.company?  So that it would be specific per res.company? Looking at my installations there aren't that many such things at res.company levels. Can you please be more specific? Will gladly do in  away you suggest.

pedrobaeza:
In a multi-company environment, it can make sense to have a different grouping policy per company, and thus, the value is stored at res.company level instead of a system wide level (ir.config_parameter). There are several fields in res.company in core that follows this way.

procurement_purchase_no_grouping/models/res_config_settings.py
```diff
@@ -0,0 +1,51 @@
...
+        "* No line grouping: If there are any open purchase order for "
+        "the same supplier, it will be reused, but lines won't be "
+        "merged.\n"
+        "* No order grouping: This option will prevent any kind of "
+        "grouping.",
+    )
+
+    @api.model
+    def get_values(self):
+        res = super(ResConfigSettings, self).get_values()
```
bealdav: no need Class name and self if it's the same than current class in python3

Rad0van: Good point for a Python newbie ;-) Thanx. Will do.

Rad0van:
> In a multi-company environment, it can make sense to have a different grouping policy per company, and thus, the value is stored at res.company level instead of a system wide level (ir.config_parameter). There are several fields in res.company in core that follows this way.

@pedrobaeza will look into that. Now I squashed all the commits to one.

Rad0van:
@pedrobaeza done. I hope at least ;-)

procurement_purchase_no_grouping/models/res_config_settings.py
```diff
@@ -0,0 +1,27 @@
...
+    """Configuration of default value for procurement purchase grouping."""
+
+    _inherit = "res.config.settings"
+    _description = "Procurement purchase grouping settings"
+
+    procured_purchase_grouping = fields.Selection(
+        related="company_id.procured_purchase_grouping",
+        string="Procured purchase grouping",
+        readonly=False,
+        help="Select the behaviour for grouping procured purchases for the "
```
pedrobaeza: Don't add the help nor the string, as it's taken from the related field.

procurement_purchase_no_grouping/models/res_company.py
```diff
@@ -0,0 +1,29 @@
...
+
+    procured_purchase_grouping = fields.Selection(
+        [
+            ("standard", "Standard grouping"),
+            ("line", "No line grouping"),
+            ("order", "No order grouping"),
+        ],
+        string="Procured purchase grouping",
+        default="standard",
+        required=True,
```
pedrobaeza: It's better to avoid this `required`, as in tests, this will be a nuisance, as this constraint is put on DB, but default is not applying until next run, so it provokes failures. Instead, put `required="1"` in the res_config_settings view for UI.

procurement_purchase_no_grouping/tests/test_procurement_purchase_no_grouping.py
```diff
@@ -79,3 +82,38 @@ def test_procurement_no_grouping_order_purchase(self):
             2,
             "Procured purchase orders lines are the same",
         )
+
+    def test_procurement_system_grouped_purchase(self):
+        self.category.procured_purchase_grouping = None
+        self._set_system_grouping("standard")
+        self._run_procurement()
+        self._run_procurement()
```
pedrobaeza, line 90: Why 2 times?

procurement_purchase_no_grouping/models/res_config_settings.py
Rad0van: OK, will do.

procurement_purchase_no_grouping/models/res_company.py
Rad0van: But that would mean that unless someone sets it, it will be None? In combination with that Product Category does not have a default value we can run into situation where there will be no value at all - i.e. directly after installation.

procurement_purchase_no_grouping/tests/test_procurement_purchase_no_grouping.py
Rad0van: Honestly - I wouldn't know. That's the way it is in original code. Was wondering when debugging why but let it go. Should that be limited to one?

procurement_purchase_no_grouping/models/res_company.py
pedrobaeza: No, required != default. The default will still be there. This is only a technical question for not setting the NOT NULL constraint on the DB, so you don't have problems in tests.

procurement_purchase_no_grouping/tests/test_procurement_purchase_no_grouping.py
pedrobaeza: Yes, try to put only one and see if tests still pass

procurement_purchase_no_grouping/models/res_company.py
Rad0van: Still learning new thing. Thanx a lot. Will do.

procurement_purchase_no_grouping/tests/test_procurement_purchase_no_grouping.py
Rad0van: OK.

Rad0van: So it seems the double run was required.

Rad0van:
@pedrobaeza All done except for the double procurement run. Without it even the original tests that didn't rely on default value failed.

pedrobaeza (approved):
OK, it's curious the double run thing.

Please change the commit message to:
```
[IMP] procurement_purchase_no_grouping: Add company level policy
```

Rad0van:
@pedrobaeza thank you far valuable input.

Rad0van:
@bealdav can you please re-review? Or maybe @sergio-teruel or @carlosdauden could have a look? Thank you.

bealdav:
Yes I have to, but before I need to know more on the module that you amended, I don't use it up to now.

Rad0van:
> Yes I have to, but before I need to know more on the module that you amended, I don't use it up to now.

Oh I thought you knew it. Before you could set that grouping type per Product Category I just added company level policy for Product Categories where no value is set.

pedrobaeza:
Bad timing for this. My colleagues are on holidays. When they return, they can eventually review the PR.

Rad0van:
> Bad timing for this. My colleagues are on holidays. When they return, they can eventually review the PR.

:-) Sure, no problem.

HviorForgeFlow:
@Rad0van @pedrobaeza @sergio-teruel @carlosdauden any news on this?

Rad0van:
> @Rad0van @pedrobaeza @sergio-teruel @carlosdauden any news on this?

Waiting for approval from others. Not much I can do now.

bealdav (approved):

sergio-teruel (approved):

PR #969, opened 2020-08-31
[FIX] vendor_transport_lead_time: fixing 'delay' field + PO integration

sebalix:
About the fix:

* move the constraint in a `@api.constrains` method
* able to create a seller (with 'create' method) without filling the `delay` field
* if we set `delay` when calling 'create', it takes precedence over other delay fields (inverse method is called)
* to make thing working, `delay` should not be required anymore
* do not override the `delay` user's label

### DIFF
--- a/vendor_transport_lead_time/models/product_supplierinfo.py
+++ b/vendor_transport_lead_time/models/product_supplierinfo.py
@@ -10,16 +10,19 @@ class ProductSupplierinfo(models.Model):
     _inherit = "product.supplierinfo"
 
     delay = fields.Integer(
-        string="Lead Time",
-        compute="_compute_delay",
-        inverse="_inverse_delay",
-        readonly=True,
+        compute="_compute_delay", inverse="_inverse_delay", store=True, required=False
     )
     supplier_delay = fields.Integer(
-        string="Supplier Lead Time", default=0, required=True
+        string="Supplier Lead Time",
+        default=0,
+        required=True,
+        help="Supplier lead time in days.",
     )
     transport_delay = fields.Integer(
-        string="Transport Lead Time", default=0, required=True
+        string="Transport Lead Time",
+        default=0,
+        required=True,
+        help="Transport lead time in days.",
     )
 
     @api.depends("supplier_delay", "transport_delay")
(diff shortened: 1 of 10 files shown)


vendor_transport_lead_time/models/product_supplierinfo.py
```diff
@@ -29,10 +32,19 @@ def _compute_delay(self):
...
                 )
-            else:
-                record.supplier_delay = diff
+
+    @api.model
+    def _setup_fields(self):
+        # "remove" the default lambda on "delay" field (from 'product' module)
+        # to not let Odoo put a value in this field when 'create' is called
+        self._fields["delay"].default = False
+        super()._setup_fields()
```
mmequignon, line 50: Not sure to understand. Why don't you set the default value directly on the field?

vendor_transport_lead_time/models/purchase_order_line.py
```diff
@@ -0,0 +1,47 @@
...
+            return False
+        seller = self.product_id._select_seller(
+            partner_id=self.partner_id,
+            quantity=self.product_qty,
+            date=self.order_id.date_order and self.order_id.date_order.date(),
+            uom_id=self.product_uom,
+        )
+        if not seller:
+            return False
+        return fields.Datetime.subtract(self.date_planned, days=seller.transport_delay)
```
mmequignon, line 40:
```suggestion
        return self._get_date_planned(seller)
```
Delay on the seller should include transport_delay, right?
_get_date_planned should be fine, no?
https://github.com/odoo/odoo/blob/13.0/addons/purchase/models/purchase.py#L627

vendor_transport_lead_time/models/product_supplierinfo.py
sebalix: Because we have no default value to set on this field (it is computed now), and anyway Odoo will try to set the default value defined in the standard `product`, what we do not want.
I don't define the default to `False` here, I override the generated `lambda` of the field as Odoo is checking this `default` attribute with some `if field.default` here and there in the ORM.

vendor_transport_lead_time/models/purchase_order_line.py
sebalix: We are subtracting the transport delay here, the `date_planned` is already computed as you said, but we also want to compute the `supplier_date_planned` without taking account the transport delay.

mmequignon: Hmm, if you say so. I'm not sure to understand why the supplier date planned should not include the delivery delay, though.

mmequignon (approved):

guewen (approved):

PR #978, opened 2020-09-23
[IMP] purchase_request: Speed up stock_move company constraint

moylop260:
Stock move is a table that could increase a lot.
So, it's important avoid slow methods in the middle

The constraint method `stock_move._check_company_purchase_request`
compares `purchase_request_allocation.company_id` vs `stock_move.company_id`
to be sure they have the same value

`pra.company_id` was a related field to
`purchase_request_line_id.request_id.company_id`
without `store=True`
Then this constraint consume a lot of time getting all these iteration of records.
For that it was changed to `store=True` and an index was added to get it faster.

And the method was using a loop getting company record by record.
It is so slow too.
It was changed directly using a query since that the ORM doesn't support `JOIN`
And was added an index to `pra.stock_move_id` field to get `JOIN` faster

### DIFF
--- a/purchase_request/models/purchase_request_allocation.py
+++ b/purchase_request/models/purchase_request_allocation.py
@@ -18,11 +18,13 @@ class PurchaseRequestAllocation(models.Model):
                                  comodel_name='res.company',
                                  readonly=True,
                                  related='purchase_request_line_id.request_id.'
-                                         'company_id'
+                                         'company_id',
+                                 store=True, index=True,
                                  )
     stock_move_id = fields.Many2one(string='Stock Move',
                                     comodel_name='stock.move',
                                     ondelete='cascade',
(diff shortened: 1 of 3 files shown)


moylop260:
Hello @AaronHForgeFlow  @HviorForgeFlow @LoisRForgeFlow @MiquelRForgeFlow

What do you think about this change to `purchase_request` module?

Regards!

luisg123v (approved):
Great improvement.

Is red related?

hugho-ad (approved):

HviorForgeFlow (approved):
It looks great IMO

moylop260:
@hugho-ad 

Could you help me to backport/fowardport it, please?

hugho-ad:
sure

PR #979, opened 2020-09-29
[12.0] [MIG] purchase_stock_return_request: Migration to 12.0

CarlosRoca13:
cc @Tecnativa TT25247

This module extend the fuctionality of https://github.com/OCA/stock-logistics-workflow/tree/12.0/stock_return_request to link the returns to the corresponding purchase order. In the next image you can see the error that appear when we try to return one unit without a picking.
image

please @chienandalu @joao-p-marques can you review this?


purchase_stock_return_request/models/stock_return_request.py
```diff
@@ -0,0 +1,85 @@
...
+    def _action_confirm(self):
+        res = super()._action_confirm()
+        if self.state == 'confirmed':
+            self.purchase_order_ids = self.returned_picking_ids.mapped(
+                'purchase_id')
+        return res
+
+    def action_view_purchases(self):
+        """Display returned purchases"""
+        action = action = self.env.ref('purchase.purchase_form_action')
```
joao-p-marques: Why this chained assignment? Did you mean this?
```suggestion
        action = self.env.ref('purchase.purchase_form_action')
```

purchase_stock_return_request/models/stock_return_request.py
```diff
@@ -0,0 +1,85 @@
...
+        result = action.read()[0]
+        result['context'] = {}
+        purchases = self.mapped('purchase_order_ids')
+        if not purchases or len(purchases) > 1:
```
joao-p-marques: You could replace this by:
```suggestion
        if len(purchases) != 1:
```

purchase_stock_return_request/models/stock_return_request.py
```diff
@@ -0,0 +1,85 @@
...
+            result['domain'] = "[('id', 'in', %s)]" % (purchases.ids)
+        elif len(purchases) == 1:
```
joao-p-marques: And this:
```suggestion
        else:
```

CarlosRoca13:
@joao-p-marques changes done!

joao-p-marques (approved):
Everything LGTM!

chienandalu (approved):

PR #991, opened 2020-10-06
[FIX] Convert invoice amount to expense currency

TheCloneMaster:
Invoice currency is NOT being used to convert to landed cost currency.
Currently it takes company currency and converts to landed cost currency, so if invoice is in different currency, amount is wrong.


pedrobaeza (approved):

HviorForgeFlow (approved):
Makes sense

Issue #971, opened 2020-09-02
[13.0] purchase_isolated_rfq cancel RFQ raise error


youring:
Try on runbot, create a new RFQ, save, and click the Cancel button, raise error,
`Error:
Odoo Server Error

Traceback (most recent call last):
  File "/.repo_requirements/odoo/odoo/http.py", line 624, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/.repo_requirements/odoo/odoo/http.py", line 310, in _handle_exception
    raise pycompat.reraise(type(exception), exception, sys.exc_info()[2])
  File "/.repo_requirements/odoo/odoo/tools/pycompat.py", line 14, in reraise
    raise value
  File "/.repo_requirements/odoo/odoo/http.py", line 669, in dispatch
    result = self._call_function(**self.params)
  File "/.repo_requirements/odoo/odoo/http.py", line 350, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/.repo_requirements/odoo/odoo/service/model.py", line 94, in wrapper
    return f(dbname, *args, **kwargs)
  File "/.repo_requirements/odoo/odoo/http.py", line 339, in checked_call
    result = self.endpoint(*a, **kw)
  File "/.repo_requirements/odoo/odoo/http.py", line 915, in __call__
    return self.method(*args, **kw)
  File "/.repo_requirements/odoo/odoo/http.py", line 515, in response_wrap
    response = f(*args, **kw)
  File "/.repo_requirements/odoo/addons/web/controllers/main.py", line 1326, in call_button
    action = self._call_kw(model, method, args, kwargs)
  File "/.repo_requirements/odoo/addons/web/controllers/main.py", line 1314, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/.repo_requirements/odoo/odoo/api.py", line 380, in call_kw
    method = getattr(type(model), name)
AttributeError: type object 'purchase.order' has no attribute 'action_cancel'`

ping @kongrattapong @kittiu

PR #992, opened 2020-10-07
[FIX][13.0] purchase_isolated_rfq cancel RFQ raise error 

Nikul-Chaudhary:
issue #971


Saran440 (changes requested):
Functional Test
should it change button name to `button_cancel` ?

HviorForgeFlow:
> Functional Test
> should it change button name to `button_cancel` ?

I think yes! But it should be tested

HviorForgeFlow:
> Functional Test
> should it change button name to `button_cancel` ?

ping @Nikul-Chaudhary

kittiu:
Let's go on, it is a small fix :)
https://github.com/OCA/purchase-workflow/pull/992

PR #996, opened 2020-10-16
[ADD] purchase_request: Make product.template#purchase_request a property

hbrunn:
this way we can turn on the request workflow for a product per company

### DIFF
--- a/purchase_request/__manifest__.py
+++ b/purchase_request/__manifest__.py
@@ -4,7 +4,7 @@
 {
     "name": "Purchase Request",
     "author": "ForgeFlow, Odoo Community Association (OCA)",
-    "version": "13.0.3.0.1",
+    "version": "13.0.4.0.0",
     "summary": "Use this module to have notification of requirements of "
     "materials and/or external services and keep track of such "
     "requirements.",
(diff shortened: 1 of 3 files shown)


AaronHForgeFlow (approved):
What happens if you already migrated to properties using https://github.com/OCA/purchase-workflow/pull/997?

HviorForgeFlow:
> What happens if you already migrated to properties using #997?

Migrate will fail when quering this:
12.0-purchase_request-company-dep-route=# select id, coalesce(purchase_request, False) from product_template
12.0-purchase_request-company-dep-route-# ;
ERROR:  column "purchase_request" does not exist
LINE 1: select id, coalesce(purchase_request, False) from product_te...

@hbrunn fix conflicts as #999 has already been applied, Thanks!

AaronHForgeFlow:
so then, either we don't touch version 12 or we adapt the scripts here.

AaronHForgeFlow (changes requested):
so then we adapt the scripts here

hbrunn:
rebased. The migration will be fine here if done already in v12 because it only kicks in when the [the column exists](https://github.com/OCA/purchase-workflow/pull/996/files#diff-78044130d89e82e899e67380e28ca732f2ba580a4d48732a8095fbc58b00cce8R7)

HviorForgeFlow (approved):

AaronHForgeFlow (approved):

PR #997, opened 2020-10-16
[ADD] purchase_request: Make product.template#purchase_request a property

hbrunn:
this way we can turn on the request workflow for a product per company

v13 PR in #996

### DIFF
--- a/purchase_request/__manifest__.py
+++ b/purchase_request/__manifest__.py
@@ -5,7 +5,7 @@
     "name": "Purchase Request",
     "author": "Eficent, "
               "Odoo Community Association (OCA)",
-    "version": "12.0.4.0.0",
+    "version": "12.0.5.0.0",
     "summary": "Use this module to have notification of requirements of "
                "materials and/or external services and keep track of such "
                "requirements.",
--- a/purchase_request/migrations/12.0.5.0.0/post-migration.py
+++ b/purchase_request/migrations/12.0.5.0.0/post-migration.py
@@ -1,0 +1,22 @@
+from odoo import api, SUPERUSER_ID
+from odoo.tools.sql import column_exists
+
+
+def migrate(cr, version=None):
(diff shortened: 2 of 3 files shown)


pedrobaeza:
You can't use the same field name for the new company dependent field, as the SQL column is dropped. If I don't remember bad, you can't even use pre-migration for renaming the column, as it's immediately dropped when Python is loaded.

pedrobaeza:
And another thing: this is not `[ADD]`, but `[IMP]`, as you are not adding a module.

hbrunn:
the migration script worked just fine when I developed this, can you point to some code that's supposed to drop the column?

StefanRijnhart:
@hbrunn It's probably https://github.com/odoo/odoo/blob/12.0/odoo/addons/base/models/ir_model.py#L691 that Pedro is referring to. When the field's xmlid is found to be missing after the upgrade, the column is dropped. But it might be that that only happens after the migration script has already run. Did you test on openupgrade? We disabled the dropping of columns there.

hbrunn:
this happens in https://github.com/odoo/odoo/blob/12.0/odoo/addons/base/models/ir_model.py#L1673 which is called after all migration scripts have run: https://github.com/odoo/odoo/blob/12.0/odoo/modules/loading.py#L473

I think the sentiment is from previous versions where property fields were actually function fields, there indeed the column was dropped immediately.

StefanRijnhart:
Ah yes, the fields record stays on so that would not drop the column.

HviorForgeFlow (approved):
migration process tested
LGTM :+1:

pedrobaeza:
Good that the process has improved. Indeed the problem was detected in 10.0.

AdriaGForgeFlow (approved):

AaronHForgeFlow (approved):

PR #1000, opened 2020-10-21
[14.0][MIG] purchase_work_ acceptance

ps-tubtim:
Standard Migration.


ps-tubtim:
Move to https://github.com/OCA/purchase-workflow/pull/1006.

PR #1009, opened 2020-11-04
 [FIX] purchase_quick : do not raise an error, if the key is not in the context

legalsylvain:
trivial patch

### DIFF
--- a/purchase_quick/models/product_product.py
+++ b/purchase_quick/models/product_product.py
@@ -24,7 +24,7 @@ class ProductProduct(models.Model):
     @api.depends("po_line_ids")
     def _compute_process_qty(self):
         res = super(ProductProduct, self)._compute_process_qty()
-        if self.env.context["parent_model"] == "purchase.order":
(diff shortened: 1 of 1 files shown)


rvalyi (approved):
ok but isn't self.env.context.get("parent_model") enough?

purchase_quick/models/product_product.py
```diff
@@ -24,7 +24,7 @@ class ProductProduct(models.Model):
     @api.depends("po_line_ids")
     def _compute_process_qty(self):
         res = super(ProductProduct, self)._compute_process_qty()
-        if self.env.context["parent_model"] == "purchase.order":
+        if self.env.context.get("parent_model", False) == "purchase.order":
             po_lines = self.env["purchase.order.line"].search(
```
bealdav, line 28: Yes it's works without , False

Thanks for the fix @legalsylvain , good catch

PierrickBrun (approved):

PR #1013, opened 2020-11-16
[13.0][MIG] purchase_allowed_product

Dariosi:


gaikaz:
@manuelcalerosolis @pedrobaeza 
Could we ask for a review? :)

purchase_allowed_product/models/product.py
```diff
@@ -19,18 +19,20 @@ def _search(
...
+                self.env["res.partner"]
+                .browse(restrict_supplier_id)
+                .commercial_partner_id
+            )
             supplierinfos = self.env["product.supplierinfo"].search(
                 [("name", "=", seller.id)]
             )
             args += [
                 "|",
-                ("product_tmpl_id", "in", supplierinfos.mapped("product_tmpl_id").ids),
```
pedrobaeza: This change is not correct. This will give `singleton error` in case of multiple supplierinfos found.

purchase_allowed_product/__manifest__.py
```diff
@@ -4,18 +4,18 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 {
     "name": "Purchase and Invoice Allowed Product",
-    "summary": "This module allows to select only products that can be "
-    "supplied by the vendor",
-    "version": "12.0.1.0.0",
+    "summary": """This module allows to select only products that can be """
```
pedrobaeza: There's no need of this change

purchase_allowed_product/tests/test_purchase_allowed_product.py
```diff
@@ -25,76 +25,52 @@ def setUp(self):
...
-            self.assertEqual(self.purchase.use_only_supplied_product, True)
-        context = {
-            "restrict_supplier_id": self.purchase.partner_id.id,
-            "use_only_supplied_product": self.purchase.use_only_supplied_product,
-        }
+        with Form(
+            self.env["purchase.order"], view="purchase.purchase_order_form"
+        ) as purchase_form:
+            purchase_form.partner_id = self.partner_4
+            purchase_form.save()
```
pedrobaeza: No need of this. The `with` automatically saves the form for you. Anyway, you are querying the form itself, not the resultant record.

pedrobaeza (changes requested):

manuelcalerosolis:
I have done several tests with the module and it works correctly

purchase_allowed_product/__manifest__.py
Dariosi: Ok, I'll change it back.

purchase_allowed_product/models/product.py
Dariosi: Thank you for review. I agree with you that the code `supplierinfos.product_tmpl_id.ids` would be incorrect in v12 but since v13, multi-relational field access is supported and works like a mapped https://www.odoo.com/documentation/13.0/reference/orm.html#map. This change should not cause an error even supplierinfos is a recordset.

purchase_allowed_product/tests/test_purchase_allowed_product.py
Dariosi: You are rigt. Thanks for the explanation.

purchase_allowed_product/models/product.py
pedrobaeza: I'm afraid not:

Selección_005

It seems their initial intention was to allow it and then go back again to previous behavior, but they don't update documentation.

@Feyensv can you please confirm and rectify documentation if so?

Feyensv: Since 13, mapped is implicit, but only for relational fields, the ORM will raise when accessing another type of field on multiple records at once:
```python
partners = users.partner_id
user_names = users.mapped('name') # necessary because len(users) > 1
# implicit mapped also works for long relational walkthrough:
currencies = users.partner_id.company_id.currency_id
currency_names = users.partner_id.company_id.currency_id.mapped('name')
currency_rates = users.partner_id.company_id.currency_id.rate_ids
```

Thus for the ids logic, it will work fine, but not in your example Pedro ;)

pedrobaeza: Thanks for the clarification, Victor. I think the documentation can be clarified for avoiding this misunderstanding, as it's a bit tricky and for me was not clear, or is it only me?

pedrobaeza (approved):

purchase_allowed_product/models/product.py
Feyensv: It is documented, but maybe not highlighted enough in the big ORM page ^^.
image

pedrobaeza: Yeah, I mean something more explicit. What about?

```
Since v13, you can chain relational fields (many2one, one2many, many2many) without using mapped when having multi recordsets, but not for other field types.

records.partner_id  # == records.mapped('partner_id')
records.partner_id.bank_ids  # == records.mapped('partner_id.bank_ids')
records.partner_id.mapped('name')  # name field is not relational, so it needs mapped
```

Dariosi:
Thanks to all reviewers for their time.
@pedrobaeza  @manuelcalerosolis  I am new in a contributor role. I guess that's why i have a questions:
should I do anything else about this code before merge? Can it be merged now if everythig is ok?

purchase_allowed_product/models/product.py
pedrobaeza: @Feyensv do you see correct my proposal for changing the documentation? Should I put it in a PR?

pedrobaeza:
@Dariosi see https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#review for the rules. You can review other PRs and ask in exchange yours to be reviewed.

purchase_allowed_product/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,8 @@
+* Chafique Delli <chafique.delli@akretion.com>
+* Mourad EL HADJ MIMOUNE <mourad.elhadj.mimoune@akretion.com>
+* Hieu, Vo Minh Bao <hieu.vmb@komit-consulting.com>
+* `Tecnativa <https://www.tecnativa.com>`__:
+
+  * Manuel Calero
+  * Pedro M. Baeza
+  * Darius Žižys <darius@vialaurea.lt>
```
HviorForgeFlow: You should add a line break before your name and the untab to get out of Tecnativa's company contributors

HviorForgeFlow (commented):
Code review

purchase_allowed_product/readme/CONTRIBUTORS.rst
Dariosi: Sorry. I'll fix that... How hard is to get a job at Tecnativa? :smile:

HviorForgeFlow:
Travis is red. I think a rebase would fix it

Dariosi:
@HviorForgeFlow Rebase didn’t help.
There seems to be a problem with the purchase_discount module:
image

Not sure what I could do in this case.

kittiu:
Please try rebase again. There is a latest one. It fixed mine.

kittiu (approved):

PR #1014, opened 2020-11-16
[13.0][MIG] purchase_security: Migration to 13.0

joao-p-marques:
Migration and some small improvements. 
Changed the group name from "Purchase / Purchase (own orders)" to "Purchase / User (own orders)", as suggested @pedrobaeza 
Also improved README

@Tecnativa 
TT26750


pedrobaeza (approved):

victoralmau:
Please, check Travis error

joao-p-marques:
Thanks @victoralmau 

I have changed the tests. In v12 I got an `AccessError` exception when trying to access the POs when I hadn't permissions, and I was asserting I got it. Now, I wasn't getting it in the test, so I changed the assertion to compare the size of the received record set. 

However, I don't know if this is right. Should I be getting the exception anyway? @pedrobaeza

purchase_security/tests/test_access_rights.py
```diff
@@ -0,0 +1,129 @@
+# Copyright 2020 Tecnativa - Víctor Martínez
```
victoralmau, line 1:
```suggestion
```

victoralmau (approved):

PR #1036, opened 2020-12-03
[13.0][FIX] purchase_isolated_rfq, fix cancel error

kittiu:
Based on this issue https://github.com/OCA/purchase-workflow/issues/971
and to replace this PR https://github.com/OCA/purchase-workflow/pull/992 (as it is pending)


HviorForgeFlow:
FF
/ocabot merge patch

PR #1041, opened 2020-12-12
[IMP] purchase_order_product_recommendation: min with for numeric step cells

ernestotejeda:
cc @Tecnativa TT25581

A min width is set for the wizard list field labeled 'Qty'. If the wizard list has many columns and the field has a very small width, editing it with the numeric_step widget is very difficult.

image

### DIFF
--- a/purchase_order_product_recommendation/__manifest__.py
+++ b/purchase_order_product_recommendation/__manifest__.py
@@ -15,6 +15,7 @@
     ],
     "data": [
         "wizards/purchase_order_recommendation_view.xml",
+        "views/assets.xml",
         "views/purchase_order_view.xml",
     ],
 }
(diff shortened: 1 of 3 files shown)


sergio-teruel (approved):
Tested on private instance

HaraldPanten (approved):

pedrobaeza:
Why not doing this in web_widget_numeric_step?

ernestotejeda:
Pedro you are right. I just did it in web_widget_numeric_step.
I took advantage of the fact that I was going to touch the module of the widget to set a min-width to the widget field cell in the list views but ONLY when the field is edited, to have a better visualization of the cell.
https://github.com/OCA/web/pull/1760

PR #1045, opened 2020-12-17
[13.0][FIX] purchase_manual_currency: Misspell "refrese" should be "refresh"

PrapassornS:


kittiu:
Label only. Fast track merge.
/ocabot merge patch

PR #1047, opened 2020-12-17
[12.0][FIX] purchase_discount

miikanissi:
Added names for th and td fields following the style of base odoo template.
Makes it easier to xpath to the field if needed.


pedrobaeza (approved):

kittiu (approved):

kittiu:
This should be forward port to v13.

PR #1052, opened 2020-12-22
[12.0][IMP] procurement_purchase_no_grouping: Add product_category in procured_purchase_grouping field

victoralmau:
Add product_category in `procured_purchase_grouping` field

@Tecnativa TT27393

### DIFF
--- a/procurement_purchase_no_grouping/i18n/es.po
+++ b/procurement_purchase_no_grouping/i18n/es.po
@@ -8,35 +8,41 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Odoo Server 10.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2017-11-24 07:53+0000\n"
-"PO-Revision-Date: 2017-11-24 07:53+0000\n"
+"POT-Creation-Date: 2021-01-19 13:00+0000\n"
+"PO-Revision-Date: 2021-01-20 10:03+0100\n"
 "Last-Translator: OCA Transbot <transbot@odoo-community.org>, 2017\n"
 "Language-Team: Spanish (https://www.transifex.com/oca/teams/23907/es/)\n"
 "Language: es\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
-"Content-Transfer-Encoding: \n"
+"Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1);\n"
+"X-Generator: Poedit 2.3\n"
 
 #. module: procurement_purchase_no_grouping
 #: selection:product.category,procured_purchase_grouping:0
 msgid "No line grouping"
-msgstr ""
+msgstr "Sin agrupación de líneas"
 
 #. module: procurement_purchase_no_grouping
 #: selection:product.category,procured_purchase_grouping:0
 msgid "No order grouping"
-msgstr ""
+msgstr "Sin agrupación de pedidos"
 
 #. module: procurement_purchase_no_grouping
 #: model:ir.model.fields,field_description:procurement_purchase_no_grouping.field_product_category__procured_purchase_grouping
 msgid "Procured purchase grouping"
-msgstr ""
+msgstr "Agrupación de la compra abastecida"
 
 #. module: procurement_purchase_no_grouping
 #: model:ir.model,name:procurement_purchase_no_grouping.model_product_category
 msgid "Product Category"
-msgstr ""
+msgstr "Categoría de producto"
+
+#. module: procurement_purchase_no_grouping
+#: selection:product.category,procured_purchase_grouping:0
+msgid "Product category grouping"
+msgstr "Agrupación por categoría de producto"
 
 #. module: procurement_purchase_no_grouping
 #: model:ir.model,name:procurement_purchase_no_grouping.model_purchase_order_line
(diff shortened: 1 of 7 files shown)


procurement_purchase_no_grouping/models/stock_rule.py
```diff
@@ -15,16 +16,19 @@ class StockRule(models.Model):
...
-            if values.get("move_dest_ids"):
-                domain += (("id", "=", -values["move_dest_ids"][:1].id),)
-            # The minimum is imposed by PG int4 limit
-            domain += (("id", "=", random.randint(-2147483648, 0)),)
+        if not self.env.context.get('skip_domain'):
+            if self.env.context.get('grouping', 'standard') == 'order':
+                if values.get("move_dest_ids"):
+                    domain += (("id", "=", -values["move_dest_ids"][:1].id),)
+                # The minimum is imposed by PG int4 limit
+                domain += (("id", "=", random.randint(-2147483648, 0)),)
```
Tardo: What kind of magic is being done here? negative ids? O_o
```suggestion
        if not self.env.context.get('skip_domain') and self.env.context.get('grouping', 'standard') == 'order':
            if values.get("move_dest_ids"):
                domain += (("id", "=", -values["move_dest_ids"][:1].id),)
            # The minimum is imposed by PG int4 limit
            domain += (("id", "=", random.randint(-2147483648, 0)),)
```

procurement_purchase_no_grouping/models/product_category.py
```diff
@@ -11,7 +11,8 @@ class ProductCategory(models.Model):
     procured_purchase_grouping = fields.Selection(
         [('standard', 'Standard grouping'),
          ('line', 'No line grouping'),
-         ('order', 'No order grouping')],
+         ('order', 'No order grouping'),
+         ('product_category', 'Product category')],
```
pedrobaeza:
```suggestion
         ('product_category', 'Product category grouping')],
```

procurement_purchase_no_grouping/models/product_category.py
```diff
@@ -22,4 +23,5 @@ class ProductCategory(models.Model):
              "the same supplier, it will be reused, but lines won't be "
              "merged.\n"
              "* No order grouping: This option will prevent any kind of "
-             "grouping.")
+             "grouping.\n"
+             "* Product category: This option grouping lines by product category.")
```
pedrobaeza:
```suggestion
             "* Product category grouping: This option groups products in the same purchase order that belongs to the same product category.")
```

procurement_purchase_no_grouping/readme/CONFIGURE.rst
```diff
@@ -6,3 +6,4 @@ Go to each product category, and select one of these values in the field
 * *No line grouping*: With this value, if there are any open purchase order
   for the same supplier, it will be reused, but lines won't be merged.
 * *No order grouping*: This option will prevent any kind of grouping.
+* *Product category*: This option grouping lines by product category.
```
pedrobaeza:
```suggestion
* *Product category grouping*: This option groups products in the same purchase order that belongs to the same product category.
```

procurement_purchase_no_grouping/models/stock_rule.py
pedrobaeza: The trick is used for having no results at all from the search.

pedrobaeza (changes requested):
You are checking nowhere the category of the product against the category of the existing draft purchase orders, so this can't work and the test you have introduced is not enough. You have to test both positive and negative matching to see that everything works.

pedrobaeza (changes requested):
Please change Spanish translation + POT for completion, and the test is still not testing the new feature. It has to test:

- Having 2 products with same vendor and same category and running the procurement for both, there's only one PO for both.
- Having 2 products with same vendor and different category, there are 2 POs.

procurement_purchase_no_grouping/i18n/es.po
```diff
@@ -8,35 +8,41 @@ msgid ""
...
 #: selection:product.category,procured_purchase_grouping:0
 msgid "No order grouping"
-msgstr ""
+msgstr "Sin agrupación de pedidos"
 
 #. module: procurement_purchase_no_grouping
 #: model:ir.model.fields,field_description:procurement_purchase_no_grouping.field_product_category__procured_purchase_grouping
 msgid "Procured purchase grouping"
-msgstr ""
+msgstr "Agrupación por pedido de compra"
```
pedrobaeza:
```suggestion
msgstr "Agrupación de la compra abastecida"
```

procurement_purchase_no_grouping/i18n/es.po
```diff
@@ -52,18 +58,30 @@ msgid ""
...
 #: selection:product.category,procured_purchase_grouping:0
 msgid "Standard grouping"
-msgstr ""
+msgstr "Agrupación estándar"
 
 #. module: procurement_purchase_no_grouping
 #: model:ir.model,name:procurement_purchase_no_grouping.model_stock_rule
 msgid "Stock Rule"
-msgstr ""
+msgstr "Regla de Inventario"
```
pedrobaeza: On Spanish there's no upper in the middle of the sentence:
```suggestion
msgstr "Regla de inventario"
```

procurement_purchase_no_grouping/tests/test_procurement_purchase_no_grouping.py
```diff
@@ -89,3 +90,42 @@ def test_procurement_no_grouping_order_purchase(self):
...
+        product2 = self.product.copy()
+        product2.write({
+            'seller_ids': [
+                (0, 0, {
+                    'name': self.partner.id,
+                    'min_qty': 1.0,
+                })
+            ]
+        })
+        product2.categ_id = category2.id
```
pedrobaeza: Why not including this in the previous write?

pedrobaeza (changes requested):
Almost there!

pedrobaeza (approved):

pedrobaeza:
@Tardo please update your review

marcelsavegnago (approved):
LGTM

pedrobaeza:
/ocabot merge major

Please fw-port it to v13

PR #1055, opened 2020-12-22
[14.0] [ADD] New module to make a default expense account field as non-required field.

Chandresh-SerpentCS:
In previous versions of Odoo, the system did not require a default expense account. We would like to investigate the feasibility of a simple solution of just making this field not-required once again in v14.

CC @dreispt


kittiu:
Thank you for your contribution.

Some questions:

* What is the case for not having the expense account in account.journal?
* This is not related to purchase-workflow, right?, may be the right place should be in account-financial-tools?

dreispt:
@kittiu Thanks for reviewing:


> What is the case for not having the expense account in account.journal?

This will force the user to manually select an account, rather than than a possibly wrong account is used by default.

> This is not related to purchase-workflow, right?, may be the right place should be in account-financial-tools?

It is true that the setting is done in Accounting app, but the user feature is in Purchase app, so I think this is the right place.

purchase_account_journal_modification/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+    "author": "Open Source Integrators, Odoo Community Association (OCA)",
+    "summary": "Remove default expense account for vendor bills journal",
+    "website": "https://github.com/OCA/purchase-workflow",
+    "license": "AGPL-3",
+    "depends": ["account"],
+    "category": "Accounting/Accounting",
+    "data": [
+        "views/account_journal_views.xml",
+    ],
+    "installable": True,
```
dreispt, line 16: Add maintainers and development_status keys? (see https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py)

dreispt (approved):
LGTM, module is simple and straightforward

dreispt (changes requested):
I just noticed the module name is vague. It should summarize the feature provided.
I suggest changing to `purchase_account_no_default`.

dreispt:
@Chandresh-SerpentCS Actually, thinking better about this, @kittiu is right:
This module add a change to `account.journal`, to be used in Accounting vendor bills, and does not depend on the Purchase app.
So it should be an `account_` module, rather than a `purchase_` one, and go to https://github.com/OCA/account-financial-tools

PR #1061, opened 2021-01-08
[IMP] purchase_request: Setting default payment term from supplier

alessandro-fiorino:
When creating a RFQ from a PR the payment term should be copied from the default supplier payment term.

### DIFF
--- a/purchase_request/tests/test_purchase_request_to_rfq.py
+++ b/purchase_request/tests/test_purchase_request_to_rfq.py
@@ -80,8 +80,9 @@ class TestPurchaseRequestToRfq(common.TransactionCase):
         purchase_request.button_to_approve()
         purchase_request.button_approved()
 
+        supplier = self.env.ref('base.res_partner_12')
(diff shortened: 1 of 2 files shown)


HviorForgeFlow:
Functional review LGTM :+1:

Are you planning to add a test that covers this feature? 

Just ensuring payment term is correctly populated to PO.

kittiu:
+1 for the test coverage.

alessandro-fiorino:
Sure, studying how to do it.

HviorForgeFlow:
I would inspire taking a test from https://github.com/OCA/purchase-workflow/blob/12.0/purchase_request/tests/test_purchase_request_to_rfq.py

Create supplier with an specific payment terms, then follow the workflow until you create the PO and check that the payment terms of that PO is the same that you set in the supplier previously.

HviorForgeFlow (approved):

marcelsavegnago (approved):

PR #1063, opened 2021-01-11
[12.0][ADD] procurement_purchase_sale_no_grouping: New module

carlosdauden:
This module adds compatibility between sale_purchase and
procurement_purchase_no_grouping modules when 'No order grouping' option
is selected an confirm sales generate a service purchase.

@Tecnativa TT27622


pedrobaeza:
Not needed anymore in this version. See upper versions.

PR #1070, opened 2021-01-15
[13.0][IMP] purchase_stock_secondary_unit: Allow update secondary unit after a purchase order is confirmed

sergio-teruel:
cc @Tecnativa TT27362
ping @carlosdauden @chienandalu @alan196 

Depends on:

- [x] https://github.com/OCA/stock-logistics-warehouse/pull/1053


pedrobaeza:
This is not adding a module, so it should be `[FIX]`, `[REF]` or `[IMP]`, not `[ADD]`.

carlosdauden (approved):

pedrobaeza (approved):

PR #1072, opened 2021-01-18
[14.0][ADD] purchase_request_type

PrapassornS:


HviorForgeFlow (commented):
Align copyright headers

kittiu:
I think this module can be useful, if it can pass the Type to PO.

**Note:** Currently, when we use both `purchase_request` and `purchase_order_type`. The PO is created automatically, but with default purchase_type, which is not ideal. (the sequence already assigned).

So, I think this module help in this problem, and can act as a glue module between `purchase_request` and `purchase_order_type`.

So I am thinking that, instead of being independent module like this, it better be glue module.

* `purchase_order_type_from_request` -> which auto_install when [purchase_request, purchase_order_type]
* The PR will also use the same purchase_type (but adding PR sequence), so it can be passed to PO easily.

Let's talk about the concept before continue the review.

kittiu:
@HviorForgeFlow WDYT about this PR?

HviorForgeFlow:
I think that as `purchase_request` is a development to ensure a previous step to create a PO, and `purchase_order_type` is already adding the feature on Purchase Order that this PR is adding to Purchase Request, it would be nice that this module links both.

I would keep this module called `purchase_request_type`.

By the other hand Purchase Request allows to group several Purchase Request Lines to a PO, how it will deal when grouping different lines from different Purchase Request types.

kittiu:
@PrapassornS just summarize what I think you should start with.

* Keep the name as `purchase_request_type`
* Reuse the same type model as purchase_order_type, in `purchase_request`. But adding more option, ie.., PR's sequence, PR picking type.
* Passing the select type, to PO.
* At the time, may be we just give warning message, if user mix PR lines from different types.
* Other usability features (filter, group by, etc. as your think appropriate).

Will this comply with your requirement? WDYT?

PrapassornS:
> WDYT?

Yes, I think so.

PrapassornS:
Please review @kittiu

kittiu:
@PrapassornS There is no change according to this summary yet.
https://github.com/OCA/purchase-workflow/pull/1072#issuecomment-762783611

purchase_request_type/views/purchase_request_type_view.xml
```diff
@@ -0,0 +1,115 @@
...
+        </field>
+    </record>
+    <record id="view_purchase_request_type_form" model="ir.ui.view">
+        <field name="model">purchase.request.type</field>
+        <field name="arch" type="xml">
+            <form string="Type">
+                <sheet>
+                    <div class="oe_button_box" name="button_box">
+                        <button
+                            name="toggle_active"
```
newtratip, line 25: toggle_active no need in v.14, archived button is in under action button
Selection_291

**example fix**
https://github.com/OCA/purchase-workflow/pull/1147/files#diff-6155a254e04cee4f8c777946add4303b338735a57f7bd4ee0cedd1d1bfab9ae1L23

PrapassornS: Noted

kittiu:
After trying to use this in real life situations, I think it is ok to server only request type as it is now, and not to relate with purchase type yet.

purchase_request_type/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,5 @@
+This module adds the following functions:
+
+    - Adds a configurable *type* on the purchase requests.
+    - Choose reduce step not to create rfq.
```
kittiu, line 4: @newtratip 
```suggestion
```

kittiu (approved):

purchase_request_type/readme/DESCRIPTION.rst
newtratip: Fixed.

Issue #1053, opened 2020-12-22
[BUG][13.0] purchase_order_secondary_unit: Error when selecting sec. uom without selected product


gaikaz:
Affected versions: 13.0 (probably others too)
Possibly same issue could happen with Sales adaptation too.

Steps to reproduce:
1. Create a new Purchase Order
2. Select "Add a product"
3. Before selecting product, try selecting the secondary unit

<details>
<summary>Error from Runbot</summary>
```
Odoo Server Error
Traceback (most recent call last):
  File "/.repo_requirements/odoo/odoo/http.py", line 624, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/.repo_requirements/odoo/odoo/http.py", line 310, in _handle_exception
    raise pycompat.reraise(type(exception), exception, sys.exc_info()[2])
  File "/.repo_requirements/odoo/odoo/tools/pycompat.py", line 14, in reraise
    raise value
  File "/.repo_requirements/odoo/odoo/http.py", line 669, in dispatch
    result = self._call_function(**self.params)
  File "/.repo_requirements/odoo/odoo/http.py", line 350, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/.repo_requirements/odoo/odoo/service/model.py", line 94, in wrapper
    return f(dbname, *args, **kwargs)
  File "/.repo_requirements/odoo/odoo/http.py", line 339, in checked_call
    result = self.endpoint(*a, **kw)
  File "/.repo_requirements/odoo/odoo/http.py", line 915, in __call__
    return self.method(*args, **kw)
  File "/.repo_requirements/odoo/odoo/http.py", line 515, in response_wrap
    response = f(*args, **kw)
  File "/.repo_requirements/odoo/addons/web/controllers/main.py", line 1327, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/.repo_requirements/odoo/addons/web/controllers/main.py", line 1319, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/.repo_requirements/odoo/odoo/api.py", line 383, in call_kw
    result = _call_kw_model(method, model, args, kwargs)
  File "/.repo_requirements/odoo/odoo/api.py", line 356, in _call_kw_model
    result = method(recs, *args, **kwargs)
  File "/home/odoo/dependencies/product-attribute/product_secondary_unit/models/product_second_unit.py", line 45, in name_search
[...]

PR #1066, opened 2021-01-12
[13.0][IMP] purchase_order_secondary_unit: Less code. Use product_secondary _unit mixin model from product-attribute repository.

sergio-teruel:
cc @Tecnativa TT27363
Also in this commit set secondary_uom as default is the product has informed a default purchase secondary uom for purchases.
ping @carlosdauden @chienandalu 

Depends on:

- [x] https://github.com/OCA/product-attribute/pull/773

PR #1074, opened 2021-01-19
[13.0][FIX] purchase_order_secondary_unit, error selecting 2nd_uom when no product_id

kittiu:
Fix -> https://github.com/OCA/purchase-workflow/issues/1053

### DIFF
--- a/purchase_order_secondary_unit/views/purchase_order_views.xml
+++ b/purchase_order_secondary_unit/views/purchase_order_views.xml
@@ -35,7 +35,7 @@
                     name="secondary_uom_id"
                     domain="[('product_tmpl_id.product_variant_ids', 'in', [product_id])]"
                     options="{'no_create': True}"
-                    attrs="{'readonly': [('state', 'in', ('done', 'cancel'))]}"
+                    attrs="{'readonly': ['|',('state', 'in', ('done', 'cancel')), ('product_id', '=', False)]}"
                 />
             </xpath>
         </field>


HviorForgeFlow:
What about don't allow user to select anything until product_id is selected, using an or domain in readonly attribute?
```
attrs="{'readonly': ['|',('state', 'in', ('done', 'cancel')), ('product_id', '=', False)]}"
```

Form view needs to be fixed also

kittiu:
> What about don't allow user to select anything until product_id is selected, using an or domain in readonly attribute?
> 
> ```
> attrs="{'readonly': ['|',('state', 'in', ('done', 'cancel')), ('product_id', '=', False)]}"
> ```
> 
> Form view needs to be fixed also

What ia the benefit? It change behavior, not just fix bug. :p

HviorForgeFlow:
Maybe the current behavior is not the proper solution. Does it make sense to query for product_id equal 0? :P :P

HviorForgeFlow:
I mean by design, you should not change any field that requires that the product was set until you set it.

kittiu:
> I mean by design, you should not change any field that requires that the product was set until you set it.

Fixed!

HviorForgeFlow:
I noticed Odoo use another strategy on sale model. They add a domain on the UoM to use same UoM category that the UoM in the product id: https://github.com/odoo/odoo/blob/13.0/addons/sale/models/sale.py#L1273-L1274

But as far I can see product_secondary_unit allows adding a UoM from another category as secondary.

So I'm not sure about the fix.

ping @pedrobaeza @rousseldenis

pedrobaeza:
First you should see #1066 if it fixes the problem

kittiu:
> First you should see #1066 if it fixes the problem

I tested those, they are not related. Same error there.

gaikaz (approved):
:shipit: Perfect

HviorForgeFlow:
Add same domain here: 
https://github.com/OCA/purchase-workflow/pull/1074/files#diff-0e2f71df6f79dc375b0e0e9479c9f6c69ca1b85d5908a9fa2e6a2b574c866ecfR23

kittiu:
> Add same domain here:
> https://github.com/OCA/purchase-workflow/pull/1074/files#diff-0e2f71df6f79dc375b0e0e9479c9f6c69ca1b85d5908a9fa2e6a2b574c866ecfR23

Sorry Hvior, not sure what is needed here. :P

HviorForgeFlow (approved):
Not problem, purchase order line form can't be accessed on edit mode

PR #1076, opened 2021-01-20
[IMP] purchase_last_price_info: Add last_purchase_price field dependa…

Daniel-CA:
…t to the "Product Price" decimal precision

### DIFF
--- a/purchase_last_price_info/models/product.py
+++ b/purchase_last_price_info/models/product.py
@@ -3,13 +3,14 @@
 #   (http://www.eficent.com)
 
 from odoo import api, fields, models
+import odoo.addons.decimal_precision as dp
 
 
 class ProductProduct(models.Model):
     _inherit = 'product.product'
 
     last_purchase_price = fields.Float(
-        string='Last Purchase Price')
+        string='Last Purchase Price', digits=dp.get_precision('Product Price'))
     last_purchase_date = fields.Date(
         string='Last Purchase Date')
     last_supplier_id = fields.Many2one(
@@ -63,7 +64,7 @@ class ProductTemplate(models.Model):
     _inherit = 'product.template'
 
     last_purchase_price = fields.Float(
-        string='Last Purchase Price')
+        string='Last Purchase Price', digits=dp.get_precision('Product Price'))
     last_purchase_date = fields.Date(
         string='Last Purchase Date')
     last_supplier_id = fields.Many2one(


oihane (approved):
LGTM

purchase_last_price_info/__manifest__.py
```diff
@@ -2,7 +2,7 @@
 
 {
     "name": "Purchase Product Last Price Info",
-    "version": "12.0.1.0.0",
+    "version": "12.0.1.1.0",
```
bodedra: Remove this changes. It will taken care by merge command.

bodedra (approved):
LG :+1:

purchase_last_price_info/__manifest__.py
Daniel-CA: > Remove this changes. It will taken care by merge command.

Done!

bodedra (approved):

HviorForgeFlow (approved):

HviorForgeFlow:
Have you tried if this change have side effects when updating the module from previous version?

Daniel-CA:
Hi @HviorForgeFlow, yes, there is no side effects, but you should update the field column to view the changes if you have a different decimal precision than the default.

HviorForgeFlow:
Manually? What happens if you have the module installed in master version and you apply this update. 

Do you lost column data, or nothing happens until you recompute the price?

 If it is the second option maybe we need to apply some migration script through this version.

HviorForgeFlow (changes requested):
Waiting an answer

bodedra (changes requested):
Agree with @HviorForgeFlow comment.

HviorForgeFlow:
ping @Daniel-CA

Daniel-CA:
Hey there, I have this issue pending, there is no data lost but I was going to make a script to update the column data format to match like it should be.

Daniel-CA:
I have been testing previous version and looks like that there is no need of change the column data format, it works like it is, looks like that in my previous test I didn't reload the view. There is no new data recalculation or anything similar, just updates the view to match with the current "product price" decimal precision.

HviorForgeFlow (approved):
Ok, fair enough :smile:

bodedra (approved):

marcelsavegnago (approved):

PR #1077, opened 2021-01-21
[13.0][IMP] procurement_purchase_no_grouping: Add product_category in procured_purchase_grouping field

victoralmau:
Add product_category in `procured_purchase_grouping` field

Related to 12.0: https://github.com/OCA/purchase-workflow/pull/1052

@Tecnativa TT27393

### DIFF
--- a/procurement_purchase_no_grouping/i18n/procurement_purchase_no_grouping.pot
+++ b/procurement_purchase_no_grouping/i18n/procurement_purchase_no_grouping.pot
@@ -6,6 +6,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Odoo Server 13.0\n"
 "Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2021-02-02 12:31+0000\n"
+"PO-Revision-Date: 2021-02-02 12:31+0000\n"
 "Last-Translator: \n"
 "Language-Team: \n"
 "MIME-Version: 1.0\n"
@@ -65,6 +67,12 @@ msgid "Product Category"
 msgstr ""
 
 #. module: procurement_purchase_no_grouping
+#: model:ir.model.fields.selection,name:procurement_purchase_no_grouping.selection__product_category__procured_purchase_grouping__product_category
+#: model:ir.model.fields.selection,name:procurement_purchase_no_grouping.selection__res_company__procured_purchase_grouping__product_category
+msgid "Product category grouping"
+msgstr ""
+
+#. module: procurement_purchase_no_grouping
 #: model_terms:ir.ui.view,arch_db:procurement_purchase_no_grouping.res_config_settings_view_form_procurement_purchase_grouping
 msgid "Purchase"
 msgstr ""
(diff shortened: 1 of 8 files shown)


procurement_purchase_no_grouping/i18n/procurement_purchase_no_grouping.pot
```diff
@@ -65,8 +67,13 @@ msgid "Product Category"
 msgstr ""
 
 #. module: procurement_purchase_no_grouping
+<<<<<<< HEAD
```
pedrobaeza: Incorrect cherry-picking with conflicts

procurement_purchase_no_grouping/readme/CONFIGURE.rst
```diff
@@ -6,8 +6,10 @@ Go to each product category, and select one of these values in the field
 * *No line grouping*: With this value, if there are any open purchase order
   for the same supplier, it will be reused, but lines won't be merged.
 * *No order grouping*: This option will prevent any kind of grouping.
+* *Product category grouping*: This option groups products in the same purchase order that belongs to the same product category.
 * *<empty>*: If you select nothing, default value set up in System
   settings will be applied.
 
 System default behaviour can be set up in System settings / Purchase / Procurement
 Purchase Grouping
+=======
```
pedrobaeza: Why adding this?

pedrobaeza (approved):

procurement_purchase_no_grouping/models/stock_rule.py
```diff
@@ -23,13 +24,18 @@ def _run_buy(self, procurements):
...
         """
         domain = super()._make_po_get_domain(company_id, values, partner)
-        if values.get("grouping") == "order":
+        if values.get("grouping") == "product_category":
+            if values.get("product_id"):
+                product = self.env["product.product"].browse(values["product_id"])
+                domain += (
+                    ("order_line.product_id.categ_id", "=", product.categ_id.id),
+                )
+        elif values.get("grouping") == "order":
```
Tardo: In the worst case, only take the value once
```suggestion
        grouping = values.get("grouping")
        if grouping == "product_category":
            if values.get("product_id"):
                product = self.env["product.product"].browse(values["product_id"])
                domain += (
                    ("order_line.product_id.categ_id", "=", product.categ_id.id),
                )
        elif grouping == "order":
```

Tardo (approved):
LGTM :+1: code review. Only a non-blocking comment

procurement_purchase_no_grouping/models/stock_rule.py
pedrobaeza: This barely optimizes, as both is an access to a variable or a value in a dictionary. There are microseconds of difference. It's no the same as caching other things in variables that requires a more complicated access (ORM fields for example), or for clearer code.

victoralmau:
Aplly PRs:
- [x] https://github.com/OCA/purchase-workflow/pull/1052 
- [x] https://github.com/OCA/purchase-workflow/pull/1080 
- [x] Adapt 13.0

Please @pedrobaeza and @Tardo, review it again.

procurement_purchase_no_grouping/readme/CONFIGURE.rst
pedrobaeza: This still continues here

pedrobaeza (changes requested):
You have split commits, and it can make sense if you also separate pre-commit stuff. As is, the 13.0 adaptation commit doesn't make too much sense, as it mixes things that are not pure 13.0 changes, but formatting one. This also learns you to try to use in 12.0 already black syntax (specially quotes) for avoiding a lot of changes.

Being in this situation, I would squash the 3 commits into one.

pedrobaeza (approved):

PR #1078, opened 2021-01-21
[14.0][IMP] purchase_work_acceptance: show accepted qty, show button create wa

SurachetS:
- show accepted qty on purchase order line
- show button create wa when available
    - qty_accepted != product_qty
    - purchase_order.state in ['purchase', 'done']

### DIFF
--- a/purchase_work_acceptance/models/purchase.py
+++ b/purchase_work_acceptance/models/purchase.py
@@ -1,7 +1,7 @@
 # Copyright 2019 Ecosoft Co., Ltd. (http://ecosoft.co.th)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import _, fields, models
+from odoo import _, api, fields, models
 
 
 class PurchaseOrder(models.Model):
@@ -19,6 +19,7 @@ class PurchaseOrder(models.Model):
         string="WA Lines",
         readonly=True,
     )
+    wa_accepted = fields.Boolean(string="WA Accepted", compute="_compute_wa_accepted")
 
     def _compute_wa_ids(self):
         for order in self:
@@ -84,6 +85,11 @@ class PurchaseOrder(models.Model):
             }
         return super().action_create_invoice()
 
+    def _compute_wa_accepted(self):
+        for order in self:
+            lines = order.order_line.filtered(lambda l: l.qty_to_accept > 0)
+            order.wa_accepted = not any(lines)
+
 
 class PurchaseOrderLine(models.Model):
     _inherit = "purchase.order.line"
@@ -93,6 +99,20 @@ class PurchaseOrderLine(models.Model):
         inverse_name="purchase_line_id",
         string="WA Lines",
         readonly=True,
+    )
+    qty_accepted = fields.Float(
+        compute="_compute_qty_accepted",
+        string="Accepted Qty.",
+        store=True,
+        readonly=True,
+        digits="Product Unit of Measure",
+    )
+    qty_to_accept = fields.Float(
+        compute="_compute_qty_accepted",
+        string="To Accept Qty.",
+        store=True,
+        readonly=True,
+        digits="Product Unit of Measure",
     )
 
     def _get_product_qty(self):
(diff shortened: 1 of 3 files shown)


purchase_work_acceptance/models/purchase.py
```diff
@@ -84,6 +87,13 @@ def action_create_invoice(self):
             }
         return super().action_create_invoice()
 
+    def _compute_invisible_create_wa(self):
+        for order in self:
+            invisible_create_wa = False
+            if any([not qty for qty in order.order_line.mapped("qty_to_accept")]):
```
kittiu: What if qty is negative? It think the button will still show here.

purchase_work_acceptance/models/purchase.py
```diff
@@ -109,3 +133,27 @@ def _prepare_account_move_line(self, move=False):
...
+        "wa_line_ids.product_qty",
+        "product_qty",
+        "product_uom_qty",
+        "order_id.state",
+    )
+    def _compute_qty_accepted(self):
+        for line in self:
+            # compute qty_accepted
+            qty = 0.0
+            for wa_line in line.wa_line_ids:
```
kittiu, line 148:
```suggestion
            for wa_line in line.wa_line_ids.filtered(lambda l: l.state == "accept"):
```

purchase_work_acceptance/models/purchase.py
```diff
@@ -109,3 +133,27 @@ def _prepare_account_move_line(self, move=False):
...
+                if wa_line.wa_id.state in ["accept"]:
```
kittiu:
```suggestion
            for wa_line in line.wa_line_ids.filtered(lambda l: l.state == "accept"):
```

purchase_work_acceptance/models/purchase.py
```diff
@@ -19,6 +19,9 @@ class PurchaseOrder(models.Model):
         string="WA Lines",
         readonly=True,
     )
+    invisible_create_wa = fields.Boolean(
```
kittiu:
```suggestion
    wa_accepted = fields.Boolean(
```
And add help = All quantity in purchase order are fully accepted.

kittiu: This field can be more useful with this,
image

kittiu:
@SurachetS I understand this use case.
But what if, we want to accept more Qty then qty in purchase? For example, we buy 100 pcs, but then, we got 102 (giveaway).
This is not the stopper, just try to think for more case. WDYT?

What about, making this an option (in Purchase Setting)?

purchase_work_acceptance/models/purchase.py
kittiu: I think following is a bit easier to read (try not to use double negate)
Note: if we use field `wa_accepted` as suggested above.
```
for order in self:
    lines = order.order_line.filtered(lambda l: l.qty_to_accept > 0)  # lines not done wa
    order.wa_accepted = not any(lines)
```
Then, you can use this field to invisible the button too.

purchase_work_acceptance/models/purchase.py
```diff
@@ -109,3 +133,27 @@ def _prepare_account_move_line(self, move=False):
...
+                    qty += wa_line.product_uom._compute_quantity(
+                        wa_line.product_qty, line.product_uom
+                    )
+            line.qty_accepted = qty
+
+            # compute qty_to_accept
+            if line.order_id.state in ["purchase", "done"]:
+                line.qty_to_accept = line.product_uom_qty - line.qty_accepted
+            else:
+                line.qty_to_accept = 0
```
kittiu: Do we need to worry about state? Even if PO state is cancelled, I think qty_to_accept can still show a number. I think it will be more simple. And with `column_invisible`, the column will be hidden on draft/cancelled anyway.
```suggestion
            line.qty_to_accept = line.product_uom_qty - line.qty_accepted
```
From the code in view, the button will be hide because of state check anyway, right?
```
attrs="{'invisible': ['|', ('state', 'not in', ('purchase', 'done')), ('wa_accepted', '=', True)]}"
```

purchase_work_acceptance/views/purchase_views.xml
```diff
@@ -6,14 +6,15 @@
         <field name="inherit_id" ref="purchase.purchase_order_form" />
         <field name="arch" type="xml">
             <xpath expr="//button[@name='button_approve']" position="after">
+                <field name="wa_accepted" invisible="1" />
```
kittiu: Are you having this field twice in the same form?

SurachetS: ah, sorry let me fix

purchase_work_acceptance/models/purchase.py
SurachetS: fixed

purchase_work_acceptance/models/purchase.py
SurachetS: fixed

purchase_work_acceptance/models/purchase.py
SurachetS: fixed

purchase_work_acceptance/models/purchase.py
SurachetS: fixed

purchase_work_acceptance/views/purchase_views.xml
SurachetS: fixed

kittiu (approved):
LGTM

ps-tubtim (approved):
Code and Functional Test are looked good to me.

PR #1079, opened 2021-01-21
[14.0][ADD] purchase_rfq_number

PrapassornS:
- This module is similar to sale_quotation_number, but for purchase/rfq
- I want to separate the document number (rfq and po), But do not want to create a new purchase order.


purchase_rfq_number/models/purchase_order.py
```diff
@@ -0,0 +1,62 @@
...
+                continue
+            else:
+                order.write(
+                    {
+                        "name": self.env["ir.sequence"].next_by_code("purchase.order"),
+                    }
+                )
+
+            # save rfq pdf as attachment
+            order.action_get_rfq_attachment()
```
kittiu: I think we shouldn't make attachment. Because when you say form "report_purchase_quotation", it is too specific (in real project, as user want to use other form. **Or you need to make it default but configurable in the settings menu.** 

**Note:** in report action, there is standard feature to make a pdf attachment on your first print. User will use that instead?

purchase_rfq_number/models/purchase_order.py
```diff
@@ -0,0 +1,62 @@
...
+
+        return super().button_confirm()
+
+    def action_get_rfq_attachment(self):
```
kittiu, line 50: As mentioned, this should be configurable if we will have it.

purchase_rfq_number/models/purchase_order.py
```diff
@@ -0,0 +1,62 @@
...
+        if "company_id" in vals:
+            company = self.env["res.company"].browse(vals.get("company_id"))
+        else:
+            company = self.env.company
+        if not company.keep_name_po and vals.get("name", "New") == "New":
+            vals["name"] = self.env["ir.sequence"].next_by_code("purchase.rfq") or "New"
+            vals["rfq_number"] = vals["name"]
+        return super().create(vals)
+
+    def button_confirm(self):
```
kittiu, line 34: Can you shorten the code (make it work as multi).
**Note:** I saw this code in sale_quotation_number, but I think we can make it better.

purchase_rfq_number/models/purchase_order.py
PrapassornS: Yes

PrapassornS:
@kittiu  please  review

purchase_rfq_number/models/purchase_order.py
kittiu: I tested and see that, the file name is not correct.
image

purchase_rfq_number/models/res_company.py
```diff
@@ -0,0 +1,29 @@
...
+    _inherit = "res.company"
+
+    keep_name_po = fields.Boolean(
+        string="Use Same Enumeration",
+        help="If this is unchecked, purchase rfq use a different sequence from "
+        "Purchase orders",
+        default=True,
+    )
+    auto_attachment_rfq = fields.Boolean(
+        string="Auto Attachment RFQ",
```
kittiu, line 17: Print RFQ as attachment to confirmed order.
```suggestion
        string="Print RFQ as attachment to confirmed order",
```

purchase_rfq_number/models/res_company.py
```diff
@@ -0,0 +1,29 @@
...
+        help="Auto attachment requests for quotation after confirm",
```
kittiu, line 18:
```suggestion
        help="Auto attach RFQ after confirm",
```

kittiu:
@PrapassornS Almost ready, just few fix.

kittiu (approved):

newtratip:
@PrapassornS, please change to same pattern of license's url of all python files.
`# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html)`

purchase_rfq_number/__manifest__.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2021 ProThai Technology Co.,Ltd. (http://prothaitechnology.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+
```
newtratip, line 3: Remove one empty line.

purchase_rfq_number/models/purchase_order.py
```diff
@@ -0,0 +1,62 @@
...
+import base64
+
+from odoo import api, fields, models
+
+
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+
+    rfq_number = fields.Char(
+        "RFQ Reference",
```
newtratip:
```suggestion
        string="RFQ Reference",
```

newtratip (approved):
Little comment.

Functional / code review :+1:

HviorForgeFlow (commented):
When printing the RFQ once Purchase Order is Confirmed use PO Sequence. This is non-blocking comment, is this ok? Can RFQ number printed?
image
image

kittiu:
> When printing the RFQ once Purchase Order is Confirmed use PO Sequence. This is non-blocking comment, is this ok? Can RFQ number printed?
> image
> image

Good idea!

kittiu:
@HviorForgeFlow Done!
https://github.com/OCA/purchase-workflow/compare/c6f339c30b8716c1e7ffd75d13e6aa26a2ca7306..f6f5795195f034f37d06f7f1aa666bb2a00c0586

HviorForgeFlow (commented):
I was testing your changes, when I came up with the cancel + set to draft procedure.
Starting with RFQ002:
image
Confirm it, everything seems to be right :+1: 
image
But once I cancel it and use set to draft, I expect RFQ come back to RFQ002:
image

Unfortunately this is not happening, and when I confirm it again:
image

RFQ Reference is PO00013 and PO reference changed to PO00014

Do you think it can be improved?

:pleading_face: :pray:

kittiu:
@HviorForgeFlow thank you for thorough test, sure I will fix it.

Note: But when confirm again, the number will be PO00014, which mean, PO00013 will be skipped. I guess this is OK for purchase order document? (or we will need to keep PO00013 is hidden field, to use it again)

HviorForgeFlow (approved):

HviorForgeFlow:
Not sure, as I noticed then PO00013 reference is lost because there is no record referencing them, which in my opinion it can create confusion.

HviorForgeFlow (approved):

newtratip:
**Test**
Setting purchase menu (Display on UI)
* Use Same Enumeration > OK
* Auto Attachment RFQ > Not ok, This field should invisible when `Use Same Enumeration` (Fixed https://github.com/ecosoft-odoo/purchase-workflow/pull/1/commits/913dfce1298fd46378532af7c00ef1a498404a1c)

Case 1 Use Same Enumeration (Standard PO)
* RFQ Reference should invisible (Fixed https://github.com/ecosoft-odoo/purchase-workflow/pull/1/commits/de271a4e29a8bb5779381a193eaa1bd211234a05)
Selection_351

* Set to Draft not should be RFQ Number = 'New' (Fixed https://github.com/ecosoft-odoo/purchase-workflow/pull/1/commits/438173b4063936ba9bcc7779129ab873809661ef)
Selection_352

Case 2 Not Use Same Enumeration and Not Auto Attachment RFQ
`OK For me`

Case 3 Not Use Same Enumeration and Auto Attachment RFQ
* Attachment file name should be RFQxxx (Fixed https://github.com/ecosoft-odoo/purchase-workflow/pull/1/commits/cebb3319466871471b30892cf0f3aae93313c861)
Peek 2021-04-09 02-13

Fix form (if RFQ Reference = 'New' will show number else will show RFQ Number) 
> Case Use Same Enumeration RFQ Reference will be 'New'
(Fixed https://github.com/ecosoft-odoo/purchase-workflow/pull/1/commits/873cdc945527cb1e7a7682f6a5cf1fdeefddcdcc)

@kittiu @HviorForgeFlow @PrapassornS Do you agree if I fixed this ?

kittiu:
@newtratip thanks for the improvement.

PR #1080, opened 2021-01-21
[12.0][FIX] procurement_purchase_no_grouping: In product_category grouping need to separate in some purchase orders

victoralmau:
Some test indicate that product_category grouping don't work fine (https://github.com/OCA/purchase-workflow/blob/12.0/procurement_purchase_no_grouping/tests/test_procurement_purchase_no_grouping.py#L110)

In product_category grouping need to separate in some purchase orders.

In 13.0: https://github.com/OCA/purchase-workflow/pull/1081

Please @pedrobaeza and @carlosdauden review it.

@Tecnativa TT27393

### DIFF
--- a/procurement_purchase_no_grouping/models/stock_rule.py
+++ b/procurement_purchase_no_grouping/models/stock_rule.py
@@ -5,7 +5,6 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
 
 import random
-
 from odoo import api, models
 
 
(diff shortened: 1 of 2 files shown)


procurement_purchase_no_grouping/models/stock_rule.py
```diff
@@ -24,11 +23,18 @@ def _run_buy(self, product_id, product_qty, product_uom, location_id,
...
+                purchase = self.env["purchase.order"].search(
+                    [
+                        ("order_line.product_id.categ_id", "!=", product.categ_id.id),
+                        ("partner_id", "=", partner.id),
+                        ("date_planned", ">=", values['date_planned']),
+                        ("state", "=", 'draft'),
+                    ], limit=1
                 )
+                if purchase:
+                    domain += (("id", "!=", purchase.id),)
```
carlosdauden: Would it work fine this way?
```suggestion
                domain += (
                        ("order_line.product_id.categ_id", "=", product.categ_id.id),
                        ("partner_id", "=", partner.id),
                        ("date_planned", ">=", values['date_planned']),
                        ("state", "=", 'draft'),
                )
```

victoralmau: Thanks but with your suggested changes don't work because only need to add domain to NOT add line in other purchase order, need to separate according to category.

pedrobaeza (approved):

procurement_purchase_no_grouping/models/stock_rule.py
pedrobaeza: Fixed with a simplified domain, but using other value (not `product_id`, but `supplier`).

marcelsavegnago (approved):

PR #1082, opened 2021-01-22
[12.0][FIX] purchase_order_product_recommendation: Add vendor for new line

Lopsanz:
If line.partner_id is not defined, unit price is catched from product instead of vendor pricelist.

### DIFF
--- a/purchase_order_product_recommendation/wizards/purchase_order_recommendation.py
+++ b/purchase_order_product_recommendation/wizards/purchase_order_recommendation.py
@@ -376,5 +376,8 @@ class PurchaseOrderRecommendationLine(models.TransientModel):
         return {
             'order_id': self.wizard_id.order_id.id,
             'product_id': self.product_id.id,
+            # set this related field manually, as there's a glitch
+            # in ORM that doesn't set its initial value
+            'partner_id': self.partner_id.id,
             'sequence': sequence,
         }


chienandalu (commented):
That field is a related one. I couldn't reproduce the fail. Do you have steps to reproduce?

Lopsanz:
Of course, @chienandalu 

1. Take a product with: Cost price: X and several vendor pricelists. Vendor 1: Y , vendor 2. Z ...
2. Make a purchase order to vendor 1 or 2.
3. Launch "Recommended products". The wizard looks fine... prices are right, but....
4. Validate and look at unit price that arrives to order line. 

Debugging this code I've seen that lines that come from the wizard haven't parnter_id yet when po_line._onchange_quantity() is called in action_accept procedure.

Perhaps this is not the perfect fix, but it's an easy fix lol!

chienandalu (approved):

chienandalu:
Ok :+1:

purchase_order_product_recommendation/wizards/purchase_order_recommendation.py
```diff
@@ -376,5 +376,6 @@ def _prepare_new_po_line(self, sequence):
         return {
             'order_id': self.wizard_id.order_id.id,
             'product_id': self.product_id.id,
+            'partner_id': self.partner_id.id,
```
pedrobaeza, line 381: Always add a comment for conflicting things:
```suggestion
            # set this related field manually, as there's a glitch
            # in ORM that doesn't set its initial value
            'partner_id': self.partner_id.id,
```

pedrobaeza (changes requested):

chienandalu (approved):

pedrobaeza:
But better in only one commit :wink:

marcelsavegnago (approved):

kittiu:
@Lopsanz would you merge commits as suggested by @pedrobaeza then we can merge this, I think.

Lopsanz:
Only a commit, @pedrobaeza

pedrobaeza (approved):

pedrobaeza:
Great, thanks!

/ocabot merge patch

PR #1097, opened 2021-02-08
[14.0][FIX] purchase_work_acceptance: stock picking domain wa_id

PrapassornS:
- Cannot search field wa_id in stock picking page because wa_id domain by active_id of purchase_order
- Fix by change domain filter by field purchase_id in table stock picking

### DIFF
--- a/purchase_work_acceptance/models/stock_picking.py
+++ b/purchase_work_acceptance/models/stock_picking.py
@@ -13,10 +13,6 @@ class Picking(models.Model):
         comodel_name="work.acceptance",
         string="WA Reference",
         copy=False,
-        domain=lambda self: [
-            ("state", "=", "accept"),
-            ("purchase_id", "=", self._context.get("active_id")),
-        ],
     )
 
     def _compute_require_wa(self):
--- a/purchase_work_acceptance/views/purchase_views.xml
+++ b/purchase_work_acceptance/views/purchase_views.xml
@@ -55,6 +55,7 @@
             <field name="fiscal_position_id" position="after">
                 <field
                     name="wa_accepted"
+                    groups="purchase_work_acceptance.group_enable_wa_on_po"
                     attrs="{'invisible': [('state', 'not in', ('purchase', 'done'))]}"
                 />
             </field>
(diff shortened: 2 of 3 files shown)


kittiu:
This case shouldn't be fixed in the view. Initial reasons that it was using `active_id` it is because `purchase_id` may not available until stock.picking is saved. Here is the field purchase_id form stock module.
```
class StockPicking(models.Model):
    _inherit = 'stock.picking'

    purchase_id = fields.Many2one('purchase.order', related='move_lines.purchase_line_id.order_id',
        string="Purchase Orders", readonly=True)
```

Please fix in .py file, and test all cases.

```
    wa_id = fields.Many2one(
        comodel_name="work.acceptance",
        string="WA Reference",
        copy=False,
        domain="[('state', '=', 'accept'), ('purchase_id', '=', purchase_id)]"
    )
```

PrapassornS:
Noted ka.

kittiu:
@PrapassornS is this ready for review?

PrapassornS:
> This case shouldn't be fixed in the view. Initial reasons that it was using `active_id` it is because `purchase_id` may not available until stock.picking is saved. Here is the field purchase_id form stock module.
> 
> ```
> class StockPicking(models.Model):
>     _inherit = 'stock.picking'
> 
>     purchase_id = fields.Many2one('purchase.order', related='move_lines.purchase_line_id.order_id',
>         string="Purchase Orders", readonly=True)
> ```
> 
> Please fix in .py file, and test all cases.
> 
> ```
>     wa_id = fields.Many2one(
>         comodel_name="work.acceptance",
>         string="WA Reference",
>         copy=False,
>         domain="[('state', '=', 'accept'), ('purchase_id', '=', purchase_id)]"
>     )
> ```

I  tried but field purchase_id is a compute field that cannot domain if not store field.

kittiu:
> I tried but field purchase_id is a compute field that cannot domain if not store field.

You can make it store=True.

kittiu (approved):

ps-tubtim (approved):

PR #1105, opened 2021-02-10
[13.0][FIX] purchase_order_uninvoiced_amount: Allow amount_uninvoiced negative

victoralmau:
Steps to reproduce:

- Create a purchase order with quantity 10
- Receive 10 units
- Create an invoice and define 20 units
`amount_uninvoiced` is 0 but it should be negative (-100 € for example)

In 12.0 happen same.

@Tecnativa TT28185


sergio-teruel:
@victoralmau I think that do this by this way can you get negative units when user re-invoices the purchase order

sergio-teruel:
@victoralmau The comment is no related to this PR... sorry..

sergio-teruel (approved):

ernestotejeda (approved):

PR #1113, opened 2021-02-15
[13.0][IMP] purchase_request: emails conditionals

glitchov:
The goal is to make the message_posts conditionals in case you don't want emails to be sent.

### DIFF
--- a/purchase_request/models/purchase_order.py
+++ b/purchase_request/models/purchase_order.py
@@ -55,7 +55,8 @@ class PurchaseOrder(models.Model):
                 message = po._purchase_request_confirm_message_content(
                     request, requests_dict[request_id]
                 )
-                request.message_post(body=message, subtype="mail.mt_comment")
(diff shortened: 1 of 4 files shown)


ThomasBinsfeld (approved):

BenjaHe (approved):

HviorForgeFlow:
Code changes do not seems to alter any functionality, unless you override the *_content method separately. Couldn't be better to add some kind of company settings that allow you to decide which messages are wanted to be sent?

ThomasBinsfeld:
> Code changes do not seems to alter any functionality, unless you override the *_content method separately. Couldn't be better to add some kind of company settings that allow you to decide which messages are wanted to be sent?

It would be a good new feature. This PR just aims to make the code more easy to override.

HviorForgeFlow (approved):
> It would be a good new feature. This PR just aims to make the code more easy to override.

Fair enough, let's merge it

PR #1120, opened 2021-02-23
[14.0][IMP] purchase_request, auto sequence only when saved

kittiu:
This PR change behavior when create new Purchase Request. Instead of immediately run the number, it now only run number when user click "Save".

**New Record**

image

**Saved Record (but still editable)**

image

**Note:**

* Keeping inline with current behavior which use `is_editable` to determine if the `name` is editable during draft. But this means, user can change name to anything.
* Personally, I prefer to make the name field readonly at all time. WDTY?

### DIFF
--- a/purchase_request/views/purchase_request_view.xml
+++ b/purchase_request/views/purchase_request_view.xml
@@ -102,10 +102,11 @@
                     </div>
                     <h1>
                         <field name="is_editable" attrs="{'invisible': True}" />
+                        <field name="is_name_editable" invisible="1" />
                         <field
                             name="name"
                             class="oe_inline"
-                            attrs="{'readonly': [('is_editable','=', False)]}"
+                            attrs="{'readonly': [('is_name_editable','=', False)]}"
                         />
                     </h1>
                     <group>
--- a/purchase_request/models/purchase_request.py
+++ b/purchase_request/models/purchase_request.py
@@ -56,8 +56,11 @@ class PurchaseRequest(models.Model):
     name = fields.Char(
         string="Request Reference",
         required=True,
-        default=_get_default_name,
+        default=lambda self: _("New"),
         track_visibility="onchange",
+    )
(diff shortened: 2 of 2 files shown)


HviorForgeFlow (commented):
I also think read-only is the way to go.
Thanks for the contribution

kittiu:
> I also think read-only is the way to go.
> Thanks for the contribution

OK, changed to readonly with last amend.

image

HviorForgeFlow:
Could be possible to define if users can edit or not the name for a technical group? Then fix the test and move forward.

kittiu:
> Could be possible to define if users can edit or not the name for a technical group? Then fix the test and move forward.

Fixed

kittiu:
cc @newtratip

Saran440 (approved):
Functional Test :+1:

purchase_request/views/purchase_request_view.xml
```diff
@@ -102,10 +102,11 @@
                     </div>
                     <h1>
                         <field name="is_editable" attrs="{'invisible': True}" />
+                        <field name="is_name_editable" invisible="1" />
                         <field
                             name="name"
                             class="oe_inline"
-                            attrs="{'readonly': [('is_editable','=', False)]}"
+                            attrs="{'readonly': [('is_name_editable','=', False)]}"
```
newtratip, line 109:
```suggestion
                            
```

Remove this attrs.

newtratip (commented):
:+1:

newtratip (approved):

newtratip:
I think, the number should be read only like other documents (PO, WH, ...)

Internal User inherited with Technical group, when I create a new user with PR User group. I can edit it (**I don't need this**)
Selection_078
Selection_079

kittiu:
> I think, the number should be read only like other documents (PO, WH, ...)

I think it is better inline with the existing one, which is editable. At least for now.

PR #1123, opened 2021-02-24
[MIG] purchase_order_approval_block: Migration to 13.0

albariera:
Standard migration. No real changes.


purchase_order_approval_block/models/purchase_order.py
```diff
@@ -0,0 +1,53 @@
...
+        comodel_name="purchase.approval.block.reason", string="Approval Block Reason",
+    )
+    approval_blocked = fields.Boolean(
+        "Approval Blocked", compute="_compute_approval_blocked",
+    )
+
+    def _compute_approval_blocked(self):
+        for rec in self:
+            if rec.approval_block_id:
+                rec.approval_blocked = True
```
HviorForgeFlow: Compute fields on version 13.0 must always return a value, add an equal False if no approval_block_id is found.

HviorForgeFlow (changes requested):

purchase_order_approval_block/models/purchase_order.py
albariera: Thanks, comment attended

purchase_order_approval_block/models/purchase_order.py
```diff
@@ -0,0 +1,53 @@
...
+        comodel_name="purchase.approval.block.reason", string="Approval Block Reason",
+    )
+    approval_blocked = fields.Boolean(
+        "Approval Blocked", compute="_compute_approval_blocked",
+    )
+
+    @api.depends("approval_block_id")
+    def _compute_approval_blocked(self):
+        for rec in self:
+            rec.approval_blocked = rec.approval_block_id
```
HviorForgeFlow, line 20: Approval Blocked boolean field expects True or False
```suggestion
        for rec in self:
            if rec.approval_block_id:
                rec.approval_blocked = True
            else:
                rec.approval_blocked = False
```

HviorForgeFlow (changes requested):

purchase_order_approval_block/models/purchase_order.py
etobella: When assigning the value, it is transforming the new value to cache. So It is using the following function:
https://github.com/odoo/odoo/blob/13.0/odoo/fields.py#L1180

It is not needed.

etobella (approved):
LGTM :+1:

purchase_order_approval_block/models/purchase_order.py
HviorForgeFlow: Ok!

HviorForgeFlow (approved):

PR #1130, opened 2021-03-08
[13.0][FIX] purchase_discount: fix acl error for low level users

Cedric-Pigeon:
## Way to reproduce

1. Create a purchase order with one line with a discount
2. Confirm the purchase order
3. Configure a user with "Inventory User" and no purchase rights
4. Try to confirm the incoming shipment with this user

## Current behavior

You get an ACL error on purchase.order.line for a write action.

## Expected behavior

You should be able to confirm the incoming shipment.
This PR intends to make it possible.

### DIFF
--- a/purchase_discount/models/stock_move.py
+++ b/purchase_discount/models/stock_move.py
@@ -22,8 +22,8 @@ class StockMove(models.Model):
             if price != po_line.price_unit:
                 # Only change value if it's different
                 price_unit = po_line.price_unit
-                po_line.price_unit = price
+                po_line.sudo().price_unit = price
         res = super()._get_price_unit()
         if price_unit:
-            po_line.price_unit = price_unit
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Isn't the sudo changing the cache "section" and thus provoking incorrect amounts?

Cedric-Pigeon:
@pedrobaeza  hum good question. not sure. Any better idea?

pedrobaeza:
Let's see first is tests complain. If not, then it shouldn't be a problem. If yes, I will think about an alternative.

Cedric-Pigeon:
I just checked in a debug session and the cache does not seem altered so tests should be green.

pedrobaeza (approved):

HviorForgeFlow (approved):

PR #1137, opened 2021-03-12
[13.0][mig][purchase_manual_delivery]

JordiBForgeFlow:
Standard migration.

cc @AaronHForgeFlow @LoisRForgeFlow @AdriaGForgeFlow


CasVissers-360ERP:
@JordiBForgeFlow 
My main concern with this module is that pending_to_receive quantities are not visible in other parts of Odoo (i.e. virtual qty).
 
Would it be an option to add this to _compute_quantities_dict are do you think is it too intrusive? Which I think it is actually. Just wanted to raise to concern.

JordiBForgeFlow:
@CasVissers-360ERP I agree, but at the same time virtual quantity is really a terrible idea in the context where you have a long delivery schedule. What good does it make to say that the virtual quantity is 10000 if you are pending to receive this quantity over 1 year?

But that's how odoo works, and perhaps you are right that we should include it in _compute_quantities_dict. Will do that.

purchase_manual_delivery/views/purchase_order_views.xml
```diff
@@ -0,0 +1,126 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
```
CasVissers-360ERP, line 2: XML in this file is not easy to read, can this be improved?

purchase_manual_delivery/wizard/create_manual_stock_picking.xml
```diff
@@ -0,0 +1,60 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
```
CasVissers-360ERP: XML in this file is not easy to read, can this be improved?

purchase_manual_delivery/views/purchase_order_views.xml
```diff
@@ -0,0 +1,126 @@
...
+                <attribute name="invisible">1</attribute>
+            </button>
+            <button id="draft_confirm" position="after">
+                <button
+                    name="button_confirm_manual"
+                    type="object"
+                    states="draft"
+                    string="Confirm Order"
+                    id="draft_confirm_manual"
+                />
```
CasVissers-360ERP, line 28: Don't understand the need to create a separate method and button for this. Why not just update the context in button_confirm directly.

purchase_manual_delivery/wizard/create_manual_stock_picking.py
```diff
@@ -0,0 +1,250 @@
...
+                    "date_planned": line.date_planned,
+                    "price_unit": line.price_unit,
+                    "product_qty": line.product_qty,
+                    "existing_qty": line.existing_qty,
+                    "remaining_qty": line.product_qty - line.existing_qty,
+                    "qty": line.product_qty - line.existing_qty,
+                    "product_uom": line.product_uom.id,
+                    "currency_id": line.currency_id.id,
+                    "partner_id": line.partner_id.id,
+                    # 'taxes_id': line.taxes_id.ids,
```
CasVissers-360ERP, line 83: Can be removed?

purchase_manual_delivery/models/purchase_order.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+class PurchaseOrderLine(models.Model):
+    _inherit = "purchase.order.line"
+
+    existing_qty = fields.Float(
+        compute="_compute_existing_qty",
+        string="Existing Qty",
+        compute_sudo=True,
+        digits=dp.get_precision("Product Unit of Measure"),
```
CasVissers-360ERP: digits="Product Unit of Measure",

purchase_manual_delivery/models/purchase_order.py
```diff
@@ -0,0 +1,91 @@
...
+            return
+        return super()._create_picking()
+
+
+class PurchaseOrderLine(models.Model):
+    _inherit = "purchase.order.line"
+
+    existing_qty = fields.Float(
+        compute="_compute_existing_qty",
+        string="Existing Qty",
```
CasVissers-360ERP: string="Receiving" to keep in line with schedule lines naming

purchase_manual_delivery/models/purchase_order.py
```diff
@@ -0,0 +1,91 @@
...
+        compute_sudo=True,
+        digits=dp.get_precision("Product Unit of Measure"),
+        help="Quantity already planned or shipped " "(stock movements already created)",
+    )
+    pending_to_receive = fields.Boolean(
+        compute="_compute_existing_qty",
+        compute_sudo=True,
+        store=True,
+        string="Pending Qty to Receive",
```
CasVissers-360ERP: string="Pending to Receive" to keep in line with schedule lines naming

purchase_manual_delivery/models/purchase_order.py
CasVissers-360ERP: Still open, after refactoring.

purchase_manual_delivery/wizard/create_manual_stock_picking.py
CasVissers-360ERP: Feedback still valid in my opinion.

purchase_manual_delivery/wizard/create_manual_stock_picking.xml
CasVissers-360ERP: Feedback still valid in my opinion.

purchase_manual_delivery/views/purchase_order_views.xml
CasVissers-360ERP: Non-important feedback. Still valid maybe though.

purchase_manual_delivery/models/purchase_order.py
CasVissers-360ERP: Still open, after refactoring.

HviorForgeFlow:
Hello @CasVissers-360ERP can you take a look on new incoming changes? Thanks!

lfreeke (approved):
:+1: Functional test

HviorForgeFlow (approved):

PR #1138, opened 2021-03-12
[10.0][FIX] purchase_request_to_rfq: domain on existing purchase order lines

AaronHForgeFlow:
Two corner cases not well solved:
* no product in the purchase request line: then use the name, but remove the domain on the product
* if there is a procurement location then we check if there is a procurement, the condition is incorrect.


AaronHForgeFlow:
cc @ForgeFlow

HviorForgeFlow (approved):

PR #1139, opened 2021-03-12
[13.0][add][purchase_order_line_schedule]

JordiBForgeFlow:
This module allows users to introduce schedule lines associated to a purchase order line, providing the ability to split the quantities that are expected to be received, by date.

Over the years we found that the majority of companies expect to be able to enter, at order line level, a schedule of the planned delivery dates for the various quantities that will be received, regardless if this is is for storable, consumable or services.

In other modules I will implement:
* purchase_order_line_schedule_stock -> Manage de dependency with purchase_stock. Create stock moves for each schedule line instead of each purchase order line.
* purchase_order_line_schedule_manual_delivery -> Depends on purchase_manual_delivery -> This is the most important module, as it will allow to manually create incoming shipments based on the schedule lines that the user chooses to receive.

@ForgeFlow 

Todo:
* [x] Add test scripts
* [x] Implement init hook to pre-create schedule lines for existing PO's.


CasVissers-360ERP:
@JordiBForgeFlow Nice can't wait for the other modules, especially purchase_order_line_schedule_manual_delivery.

Some initial questions/feedback for this module:

1. The schedule lines don't update the product_qty on the PO line?
2. date_planned should be required on schedule lines?  Traceback when setting empty (ol.date_planned = max(ol.schedule_line_ids.mapped("date_planned")))
3. date_planned could default to last date_planned from schedule lines or PO date_planned
3. Would be nice to add product_id to schedule lines for reporting
4. Activating the calendar view might be nice for schedule lines

Just some quick feedback because I am very enthousiastic about this module. Next week I will spend some more time reviewing.

JordiBForgeFlow:
@CasVissers-360ERP thanks for your feedback. I have added some more features

purchase_order_line_schedule/models/purchase_order.py
```diff
@@ -0,0 +1,37 @@
...
+                for ol in order.order_line
+            ):
+                order.has_qty_to_schedule = True
+            else:
+                order.has_qty_to_schedule = False
+
+    def button_confirm(self):
+        for order in self:
+            if order.has_qty_to_schedule:
+                raise UserError(
```
OpenCode, line 31: Is it possible to collect all orders with `has_qty_to_schedule` and show numbers in error message?

CasVissers-360ERP:
@JordiBForgeFlow 
Are the modules test-ready?

purchase_order_line_schedule/models/purchase_order_line.py
```diff
@@ -0,0 +1,101 @@
...
+                    if qty != rec.product_qty:
+                        sl.product_qty += rec.product_qty - qty
+                    if sl.date_planned != rec.date_planned:
+                        sl.date_planned = rec.date_planned
+                    break
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        res = super(PurchaseOrderLine, self).create(vals_list)
+        self._update_schedule_lines()
```
CasVissers-360ERP: Do we need a post init script to generate lines for open RFQ's?

purchase_order_line_schedule/views/purchase_order_views.xml
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="purchase_order_form" model="ir.ui.view">
+        <field name="name">purchase.order.form.inherit</field>
+        <field name="model">purchase.order</field>
+        <field name="inherit_id" ref="purchase.purchase_order_form" />
+        <field name="arch" type="xml">
+            <field name="date_approve" position="after">
+                <field name="has_qty_to_schedule" invisible="True" />
```
CasVissers-360ERP: You use 1 everywhere :)

purchase_order_line_schedule/views/purchase_order_views.xml
```diff
@@ -0,0 +1,42 @@
...
+                    string="Delivery Schedule"
+                    name="action_open_schedule_lines"
+                    type="object"
+                    icon="fa-list"
+                    optional="hide"
+                />
+                <field
+                    name="qty_to_schedule"
+                    attrs="{'column_invisible': [('parent.has_qty_to_schedule', '=', False)]}"
+                />
```
CasVissers-360ERP, line 26: Would be nice to always see the schedule not only when there is something to schedule.

purchase_order_line_schedule/views/purchase_order_views.xml
```diff
@@ -0,0 +1,42 @@
...
+                expr="//notebook//field[@name='order_line']/tree/field[@name='date_planned']"
+                position="after"
+            >
+                <field name="schedule_line_ids" invisible="1" />
+                <button
+                    string="Delivery Schedule"
+                    name="action_open_schedule_lines"
+                    type="object"
+                    icon="fa-list"
+                    optional="hide"
```
CasVissers-360ERP: Why hiding this important feature by default?

JordiBForgeFlow: @CasVissers-360ERP because the field 'date_planned' is also optional. Either we show or hide both to keep consistency.

CasVissers-360ERP: Show both makes the module more intuitive in my opinion.

purchase_order_line_schedule/models/purchase_order_line_schedule.py
```diff
@@ -0,0 +1,84 @@
...
+        inverse="_inverse_qty_received",
+        compute_sudo=True,
+        digits="Product Unit of Measure",
+    )
+    qty_to_receive = fields.Float(
+        "To Receive",
+        compute="_compute_qty_to_receive",
+        compute_sudo=True,
+        digits="Product Unit of Measure",
+    )
```
CasVissers-360ERP, line 38: Should negative quantities be allowed? Or is in addition, i.e. purchase_order_line_schedule_rental

purchase_order_line_schedule/models/purchase_order_line.py
```diff
@@ -0,0 +1,101 @@
...
+                qty = sum(rec.schedule_line_ids.mapped("product_qty"))
+                # Update the last scheduled date of the schedule lines
+                # And update the quantity of schedule lines so as to match
+                # that of the purchase order line.
+                for sl in sls.sorted(lambda x: x.date_planned, reverse=True):
+                    if qty != rec.product_qty:
+                        sl.product_qty += rec.product_qty - qty
+                    if sl.date_planned != rec.date_planned:
+                        sl.date_planned = rec.date_planned
+                    break
```
CasVissers-360ERP, line 86: This auto spread feature might be confusing to end-users. Maybe it's better to only update scheduled lines if there is just one scheduled line. If not the user should do it manually?

purchase_order_line_schedule/models/purchase_order_line.py
JordiBForgeFlow: Yes, this is pending.

purchase_order_line_schedule/models/purchase_order_line_schedule.py
JordiBForgeFlow: This is necessary to ensure compatibility with existing modules / flows. If a user changes the quantity to order by default it allocates the extra quantity to the most recent schedule line. The user can always to go to the schedule lines and adjust.

purchase_order_line_schedule/models/purchase_order_line.py
JordiBForgeFlow: This is necessary to ensure compatibility with existing modules / flows. If a user changes the quantity to order by default it allocates the extra quantity to the most recent schedule line. The user can always to go to the schedule lines and adjust.

CasVissers-360ERP: Maybe we need a constraint in that case on product_qty on SL? The qty should be qty of related PO line. Right now a loop is created if you schedule more or less, the quantities are not saved.

purchase_order_line_schedule/models/purchase_order.py
```diff
@@ -0,0 +1,37 @@
...
+            "Product Unit of Measure"
+        )
+        for order in self:
+            if any(
+                not float_is_zero(ol.qty_to_schedule, precision_digits=precision)
+                for ol in order.order_line
+            ):
+                order.has_qty_to_schedule = True
+            else:
+                order.has_qty_to_schedule = False
```
CasVissers-360ERP, line 26: Will always be false, since _update_schedule_lines will keep correcting it.

purchase_order_line_schedule/views/purchase_order_views.xml
```diff
@@ -0,0 +1,42 @@
...
+            <xpath
+                expr="//notebook//field[@name='order_line']/tree/field[@name='date_planned']"
+                position="after"
+            >
+                <field name="schedule_line_ids" invisible="1" />
+                <button
+                    string="Delivery Schedule"
+                    name="action_open_schedule_lines"
+                    type="object"
+                    icon="fa-list"
```
CasVissers-360ERP: I prefer fa-table

purchase_order_line_schedule/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,9 @@
+This module allows users to introduce a delivery schedule associated to
+a purchase order line. A schedule is used when a quantity on an order line
+is requested to be delivered in multiple dates.
+
+The Delivery schedule allows you to specify the number of dates in
+which the total quantity of the order line will be delivered from the vendor.
+
+By default, the system copies the total quantity and scheduled date
+of the original purchase line into the first delivery schedule line.
```
CasVissers-360ERP, line 9: Minor changes in text:

This module allows users to associate a delivery schedule to
a purchase order line. The schedule is used when quantities on an order line
are requested to be delivered on multiple dates.

The delivery schedule allows you to specify the dates on
which the quantities of the order line will be delivered from the vendor.

By default, the system copies the total quantity and scheduled date
of the original purchase line into the first delivery schedule line.

purchase_order_line_schedule/models/purchase_order_line.py
JordiBForgeFlow: @CasVissers-360ERP added the init script.

purchase_order_line_schedule/views/purchase_order_views.xml
JordiBForgeFlow: I am adding only the button. The other field is anyway useless when you use this module.

purchase_order_line_schedule/views/purchase_order_views.xml
JordiBForgeFlow: Fixed!

JordiBForgeFlow:
@CasVissers-360ERP I have added a wizard to schedule order lines that makes it much more usable.
image

purchase_order_line_schedule/models/purchase_order.py
JordiBForgeFlow: Works after adding the wizard.

purchase_order_line_schedule/views/purchase_order_views.xml
JordiBForgeFlow: Fixed.

purchase_order_line_schedule/readme/DESCRIPTION.rst
JordiBForgeFlow: Fixed.

purchase_order_line_schedule/models/purchase_order_line.py
JordiBForgeFlow: Fixed after adding the wizard.

CasVissers-360ERP:
Nice, will check Thursday morning.

purchase_order_line_schedule/init_hooks.py
```diff
@@ -0,0 +1,29 @@
...
+        INSERT INTO purchase_order_line_schedule
+            (create_uid, create_date, write_uid, write_date,
+             order_id, order_line_id, date_planned, product_qty,
+             qty_received_manual, product_id, company_id)
+        SELECT create_uid, create_date, write_uid, write_date,
+               order_id, id, date_planned, product_qty,
+               qty_received_manual, product_id, company_id
+        FROM purchase_order_line
+    """
+    )
```
CasVissers-360ERP, line 24: `ERROR: null value in column "date_planned" violates not-null constraint `

Seems like date_planned can be null on POL, I suggest:


```
INSERT INTO purchase_order_line_schedule
            (create_uid, create_date, write_uid, write_date,
             order_id, order_line_id, date_planned, product_qty,
             qty_received_manual, product_id, company_id)
        SELECT pol.create_uid, pol.create_date, pol.write_uid, pol.write_date,
               pol.order_id, pol.id, coalesce(pol.date_planned, po.date_planned, po.date_order) , pol.product_qty,
               pol.qty_received_manual, pol.product_id, pol.company_id
        FROM purchase_order_line AS pol, purchase_order AS po
        WHERE pol.order_id = po.id

```

purchase_order_line_schedule/models/purchase_order_line_schedule.py
```diff
@@ -0,0 +1,107 @@
...
+        precision = self.env["decimal.precision"].precision_get(
+            "Product Unit of Measure"
+        )
+        for rec in self:
+            total_scheduled_qty = sum(
+                rec.order_line_id.schedule_line_ids.mapped("product_qty")
+            )
+            line_qty = rec.order_line_id.product_qty
+            if (
+                float_compare(total_scheduled_qty, line_qty, precision_digits=precision)
```
CasVissers-360ERP, line 101: @JordiBForgeFlow 
This still allows for negative quantities? I think this should be blocked as well?

purchase_order_line_schedule/models/purchase_order.py
CasVissers-360ERP: But why do we need it, if qty in schedule lines is constrained to be the same as order qty?

purchase_order_line_schedule/wizards/schedule_order_line_views.xml
```diff
@@ -0,0 +1,46 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="schedule_order_line_form" model="ir.ui.view">
+        <field name="name">schedule.order.line.form</field>
+        <field name="model">schedule.order.line</field>
+        <field name="arch" type="xml">
+            <form string="Schedule Lines">
+                <group>
+                    <field name="order_line_id" readonly="1" />
```
CasVissers-360ERP, line 9: Can we also show ordered quantity here? Makes it easier to spread the quantities.

CasVissers-360ERP:
@JordiBForgeFlow 
Minor and some bigger feedback points on all PR's related to this module.

JordiBForgeFlow:
@CasVissers-360ERP I have done a significant refactoring in all the schedule* modules, and added tests.

purchase_order_line_schedule/init_hooks.py
CasVissers-360ERP: @JordiBForgeFlow 
Why didn't you process this feedback?

JordiBForgeFlow: @CasVissers-360ERP I missed it. Will review.

purchase_order_line_schedule/wizards/schedule_order_line_views.xml
CasVissers-360ERP: Feedback still valid in my opinion.

purchase_order_line_schedule/models/purchase_order_line_schedule.py
CasVissers-360ERP: Feedback still valid in my opinion.

purchase_order_line_schedule/init_hooks.py
CasVissers-360ERP: Feedback still valid in my opinion. Still got the error.

CasVissers-360ERP:
Btw, I did an addition for some reporting fields, maybe you like it too. Good for liquidity forecast:


```
    @api.depends('qty_to_receive', 'order_line_id', 'order_line_id.price_unit')
    def _compute_value_to_receive(self):
        for line in self:
            company_currency = self.company_id.currency_id
            currency = line.order_id.currency_id
            line.currency_id = currency
            price = line.order_line_id.price_unit
            if currency == company_currency:
                line.value_to_receive = price * line.qty_to_receive
                line.value_to_receive_fc = price * line.qty_to_receive
            else:
                line.value_to_receive = price * line.qty_to_receive
                price_eu = currency._convert(price, company_currency, line.order_id.company_id, line.order_id.date_order or fields.Date.today())
                line.value_to_receive_fc = price_eu * line.qty_to_receive
```

purchase_order_line_schedule/views/purchase_order_line_schedule_views.xml
```diff
@@ -0,0 +1,57 @@
...
+        <field name="name">purchase.order.line.schedule.tree</field>
+        <field name="model">purchase.order.line.schedule</field>
+        <field name="arch" type="xml">
+            <tree editable="top" create="true" edit="true" delete="true">
+                <field name="order_id" invisible="1" />
+                <field name="purchase_state" invisible="1" />
+                <field name="qty_received_method" invisible="1" />
+                <field name="order_line_id" readonly="1" />
+                <field name="date_planned" />
+                <field name="product_qty" />
```
CasVissers-360ERP, line 13: add sum="Total"

purchase_order_line_schedule/views/purchase_order_line_schedule_views.xml
```diff
@@ -0,0 +1,57 @@
...
+                <field
+                    name="qty_received"
+                    attrs="{'invisible': [('purchase_state', 'not in', ('purchase', 'done'))], 'readonly': [('qty_received_method', '!=', 'manual')]}"
+                />
+                <field name="qty_received_manual" invisible="1" />
+                <field name="qty_to_receive" />
```
CasVissers-360ERP, line 19: add sum="Total"

CasVissers-360ERP:
@JordiBForgeFlow Any plans to continue this chain of PR's?

purchase_order_line_schedule/models/purchase_order_line_schedule.py
```diff
@@ -0,0 +1,109 @@
...
+    @api.onchange("qty_received")
+    def _inverse_qty_received(self):
+        for line in self:
+            if line.order_line_id.qty_received_method == "manual":
+                line.qty_received_manual = line.qty_received
+            else:
+                line.qty_received_manual = 0.0
+
+    def _get_order_lines_to_update(self):
+        return self.mapped("order_line_id")
```
CasVissers-360ERP, line 73: This method doesn't make sense and doesn't work in _update_order_line_date_planned the date of POL is not updated.

purchase_order_line_schedule/models/purchase_order_line_schedule.py
```diff
@@ -0,0 +1,109 @@
...
+
+    order_id = fields.Many2one(
+        comodel_name="purchase.order", related="order_line_id.order_id", store=True
+    )
+    purchase_state = fields.Selection(related="order_line_id.order_id.state",)
+    qty_received_method = fields.Selection(related="order_line_id.qty_received_method",)
+    order_line_id = fields.Many2one(
+        comodel_name="purchase.order.line", required=True, ondelete="cascade"
+    )
+    date_planned = fields.Datetime(string="Scheduled Date", index=True, required=True)
```
CasVissers-360ERP, line 22: The operation cannot be completed:
- Create/update: a mandatory field is not set.
- Delete: another model requires the record being deleted. If possible, archive it instead.

Model: Purchase Order Line Schedule (purchase.order.line.schedule), Field: Scheduled Date (date_planned)

Should be required.

clementmbr:
Hi @JordiBForgeFlow @CasVissers-360ERP, I'm starting to look at these PRs and I'm a bit surprised it has been inactive since few months... Is the project on hold on your sides? Is there finally a better alternative to PO reception scheduling than these modules ?

Thank you very much for the update ! :pray:

JordiBForgeFlow:
@clementmbr Although I love this approach I got some push back from the project I intended to put it in, and finally the PR has been a bit abandoned from my side. But hey, feel free to review, use it and provide your feedback. I am proud of this baby and I thinl that it is a good step forward in the proper scheduling of receptions.

clementmbr:
@JordiBForgeFlow Thanks for your feedback !
Just one more thing before I go deeper into your modules:
I'm wondering why it is better to _subdivide the PO lines_ like you do with the `purchase.order.line.schedule` instead of _creating many PO lines_ for the same product but each one with a distinct `date_planned` ?

I'm seeing that using many PO lines would need at least some work to duplicate lines easily and to improve the relation between the Order's date and these `date_planned`, but can you explain quickly why you preferred the `purchase.order.line.schedule` strategy ?

Thanks !

JordiBForgeFlow:
@clementmbr The PO lines are created many times from reordering rules or manually, and the vendor expects to receive a single line with the full quantity. If you start to split the PO lines I believe that the PO can become quite a mess. 

SAP, for example, uses the concept of schedule lines associated to a PO line https://help.sap.com/viewer/65b1545de4cf409db91e81c6da53db5a/6.18.15/en-US/08e5c353b677b44ce10000000a174cb4.html

Same with MS Navision: https://docs.microsoft.com/en-us/dynamics365/supply-chain/procurement/tasks/create-purchase-order-delivery-schedule

PR #1140, opened 2021-03-14
[13.0][add][purchase_order_line_schedule_stock]

JordiBForgeFlow:
This is a glue module between purchase_order_line_schedule (https://github.com/OCA/purchase-workflow/pull/1139) and purchase_stock.

In itself the module is not very useful becasue odoo will always merge the stock moves that have different scheduled for the same product date into one, but it is a basis to the module purchase_order_line_schedule_manual_delivery.

@ForgeFlow @CasVissers-360ERP


purchase_order_line_schedule_stock/README.rst
```diff
@@ -0,0 +1,88 @@
+===============================
+Purchase Line Procurement Group
```
CasVissers-360ERP, line 2: Incorrect readme?

purchase_order_line_schedule_stock/models/purchase_order_line.py
```diff
@@ -0,0 +1,40 @@
+# Copyright 2021 ForgeFlow, S.L.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+from odoo import models
+
+
+class PurchaseOrderLine(models.Model):
+
+    _inherit = "purchase.order.line"
+
+    def _prepare_stock_moves(self, picking):
```
CasVissers-360ERP, line 10: I still only get one stock move per POL, I think it might be good to use purchase_delivery_split_date for this.

purchase_order_line_schedule_stock/README.rst
CasVissers-360ERP: Still open, after refactoring.

purchase_order_line_schedule_stock/models/purchase_order_line.py
CasVissers-360ERP: Non prio feedback.

PR #1146, opened 2021-03-18
[13.0]  purchase_delivery_split_date: fix move assign after split

santostelmo:
### Context
When a confirmed PO line has its scheduled date changed, the related stock move is moved to another picking.
Before the move is assigned to another picking, it is unreserved. However, after the change of picking_id and setting the new date, it is not reserved back again which is causing the reception to not be ready.

### DIFF
--- a/purchase_delivery_split_date/models/purchase.py
+++ b/purchase_delivery_split_date/models/purchase.py
@@ -106,6 +106,7 @@ class PurchaseOrder(models.Model):
                             move._do_unreserve()
                             move.picking_id = pickings_by_date[date_key]
                             move.date_expected = date_key
+                            move._action_assign()
             for picking in pickings_by_date.values():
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -47,3 +47,7 @@ script:
 
 after_success:
   - travis_after_tests_success
+
+branches:
+  except:
+    - /^merge-branch-.*$/
```
carlosserratoro: Is this required for this PR?

santostelmo: No is not. I remote it

TDu: It seems your PR includes an unwanted commit, it is the cause of this addition

simahawk (approved):
seems legit

HviorForgeFlow (approved):

PR #1151, opened 2021-04-06
[14.0][ADD] purchase_order_line_product_state

rousseldenis:
Depends on:

- [ ] https://github.com/OCA/product-attribute/pull/725

image


rousseldenis:
Do you think a group could be a nice addition in order to not overload the view ?

kittiu:
> Do you think a group could be a nice addition in order to not overload the view ?

Not sure what is meant by group. But I think this should be optional="hide" and to use the badge as it is a kind of state.
image

WDYT?

PR #1156, opened 2021-04-14
[12.0][ADD] Add purchase_requisition_line_description module

ajaniszewska-dev:
The aim of this module is to show description in Purchase Agreement form and 'Call for Tender' report.

### DIFF
--- a/purchase_requisition_line_description/models/purchase_requisition_line.py
+++ b/purchase_requisition_line_description/models/purchase_requisition_line.py
@@ -1,0 +1,28 @@
+# Copyright 2021 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+from odoo.fields import first
+
+
+class PurchaseRequisitionLine(models.Model):
+    _inherit = "purchase.requisition.line"
+
+    name = fields.Text(string="Description")
+
+    @api.onchange('product_id')
+    def _onchange_product_id(self):
+        res = super()._onchange_product_id()
+        if self.product_id:
+
+            partner = (
+                self.requisition_id.purchase_ids.partner_id
+                or first(self.requisition_id.mapped("purchase_ids")).partner_id
+            )
+            product_lang = self.product_id.with_context(
+                lang=partner.lang, partner_id=partner.id,
+            )
+            self.name = product_lang.display_name
+            if product_lang.description_purchase:
+                self.name += '\n' + product_lang.description_purchase
+        return res
--- a/purchase_requisition_line_description/tests/test_purchase_requisition_line_description.py
+++ b/purchase_requisition_line_description/tests/test_purchase_requisition_line_description.py
@@ -1,0 +1,28 @@
+# Copyright 2021 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo.tests import common, Form
+
+
+class TestPurchaseRequisitionLineDescription(common.SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super(TestPurchaseRequisitionLineDescription, cls).setUpClass()
+        cls.partner = cls.env['res.partner'].create({
+            'name': 'Test partner',
+        })
+        cls.product = cls.env['product.product'].create({
+            'name': 'Product',
+            'standard_price': 10,
+            'description_purchase': 'description for purchase',
+        })
+
+    def test_onchange_product_id(self):
+        with Form(self.env['purchase.requisition']) as purchase_form:
+            purchase_form.vendor_id = self.partner
+            with purchase_form.line_ids.new() as line_form:
+                line_form.product_id = self.product
+            self.assertEqual(
+                line_form.name,
+                self.product.name+'\n'+self.product.description_purchase
+            )
--- a/purchase_requisition_line_description/views/purchase_requisition.xml
+++ b/purchase_requisition_line_description/views/purchase_requisition.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0"?>
+<odoo>
+    <record model="ir.ui.view" id="purchase_requisition_chatter_view">
+      <field name="name">purchase_requisition_form</field>
+      <field name="model">purchase.requisition</field>
+      <field name="inherit_id" ref="purchase_requisition.view_purchase_requisition_form"/>
+      <field name="arch" type="xml">
+        <xpath expr="//field[@name='product_id']" position="after">
+          <field name="name"/>
+        </xpath>       
+      </field>
+    </record>
+</odoo>
--- a/purchase_requisition_line_description/__init__.py
+++ b/purchase_requisition_line_description/__init__.py
@@ -1,0 +1,1 @@
+from . import models
--- a/purchase_requisition_line_description/__manifest__.py
+++ b/purchase_requisition_line_description/__manifest__.py
@@ -1,0 +1,22 @@
+# Copyright 2021 Camptocamp
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Purchase Requisition Line Description',
+    'summary': 'Extends the functionality of Purchase Agreements to '
+               'show line description.',
+    'version': '12.0.1.0.0',
+    'category': 'Purchases',
+    'website': 'https://github.com/OCA/purchase-workflow',
+    'author': 'Camptocamp, Odoo Community Association (OCA)',
+    'license': 'AGPL-3',
+    'application': False,
+    'installable': True,
+    'depends': [
+        'purchase_requisition',
+    ],
+    'data': [
+        'views/purchase_requisition.xml',
+        'report/purchase_requisition_report.xml',
+    ],
+}
--- a/purchase_requisition_line_description/models/__init__.py
+++ b/purchase_requisition_line_description/models/__init__.py
@@ -1,0 +1,1 @@
+from . import purchase_requisition_line
--- a/purchase_requisition_line_description/readme/CONTRIBUTORS.rst
+++ b/purchase_requisition_line_description/readme/CONTRIBUTORS.rst
@@ -1,0 +1,1 @@
+* Anna Janiszewska <anna.janiszewska@camptocamp.com>
--- a/purchase_requisition_line_description/readme/DESCRIPTION.rst
+++ b/purchase_requisition_line_description/readme/DESCRIPTION.rst
@@ -1,0 +1,1 @@
+This module allows to use product purchase description on the purchase requisition lines.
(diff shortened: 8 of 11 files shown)


purchase_requisition_line_description/models/product.py
```diff
@@ -0,0 +1,33 @@
...
+                    product_ids = self._search(
+                        [('seller_ids.product_name', '=', name)] + args,
+                        limit=limit,
+                        access_rights_uid=name_get_uid,
+                    )
+        else:
+            product_ids = self._search(
+                args, limit=limit, access_rights_uid=name_get_uid
+            )
+        return self.browse(product_ids).name_get()
```
i-vyshnevska:
```suggestion
            )
            return self.browse(self+product_ids).name_get()
        return super()._name_search(name, args, operator, limit, name_get_uid)
```
I am actually not sure if that is correct to return name_gate

purchase_requisition_line_description/models/purchase_requisition_line.py
```diff
@@ -0,0 +1,31 @@
...
+
+class PurchaseRequisitionLine(models.Model):
+    _inherit = "purchase.requisition.line"
+
+    name = fields.Text()
+
+    @api.onchange('product_id')
+    def _onchange_product_id(self):
+        desc = ""
+        res = super(PurchaseRequisitionLine, self)._onchange_product_id()
```
i-vyshnevska:
```suggestion
        res = super()._onchange_product_id()
```

purchase_requisition_line_description/models/purchase_requisition_line.py
```diff
@@ -0,0 +1,31 @@
...
+        if self.product_id:
+            product_lang = self.product_id.with_context(
+                lang=self.requisition_id.purchase_ids.partner_id.lang,
+                partner_id=self.requisition_id.purchase_ids.partner_id.id,
+            )
+            vendor_data = product_lang.seller_ids.filtered(
+                lambda a: a.product_id == self.product_id
+                and a.product_name
+                or a.product_code
```
i-vyshnevska:
```suggestion
                and (a.product_name
                or a.product_code)
```
one of the is mandatory?

purchase_requisition_line_description/models/purchase_requisition_line.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class PurchaseRequisitionLine(models.Model):
+    _inherit = "purchase.requisition.line"
+
+    name = fields.Text()
+
+    @api.onchange('product_id')
+    def _onchange_product_id(self):
+        desc = ""
```
i-vyshnevska:
```suggestion
```

purchase_requisition_line_description/models/purchase_requisition_line.py
```diff
@@ -0,0 +1,31 @@
...
+        res = super(PurchaseRequisitionLine, self)._onchange_product_id()
+        if self.product_id:
```
i-vyshnevska, line 16:
```suggestion
        desc = self.name
        if self.product_id:
```

purchase_requisition_line_description/models/purchase_requisition_line.py
```diff
@@ -0,0 +1,31 @@
...
+from odoo import api, fields, models
+
+
+class PurchaseRequisitionLine(models.Model):
+    _inherit = "purchase.requisition.line"
+
+    name = fields.Text()
+
+    @api.onchange('product_id')
+    def _onchange_product_id(self):
```
i-vyshnevska, line 14: can you add a comment how name will look like after

purchase_requisition_line_description/models/product.py
ajaniszewska-dev: afair name_get is called at the end of default name_search - not really sure why we need this change?

purchase_requisition_line_description/models/purchase_requisition_line.py
ajaniszewska-dev: @i-vyshnevska  that won't work. I already checked it yesterday with @grindtildeath , somehow self.name is type bool. If later we try to concat str with bool we run into issues with TypeError Invalid operand. Coulnd't find out why it is bool type instead of 'str', so we made sure that desc is string by assigning empty string to it and update self.name with desc in latest step.


```
File "/odoo/external-src/purchase-workflow/purchase_requisition_line_description/models/purchase_requisition_line.py", line 28, in _onchange_product_id
    desc += '[' + vendor_data.product_code + '] '
TypeError: unsupported operand type(s) for +=: 'bool' and 'str'
```

purchase_requisition_line_description/models/product.py
ajaniszewska-dev:
```
  File "/odoo/external-src/purchase-workflow/purchase_requisition_line_description/models/product.py", line 29, in _name_search
    return self.browse(self.product_ids).name_get()
AttributeError: 'product.product' object has no attribute 'product_ids'
```

purchase_requisition_line_description/models/purchase_requisition_line.py
i-vyshnevska: desc = (self.name or "")

ajaniszewska-dev: works fine, thank you

ajaniszewska-dev: @i-vyshnevska  with this once you have line with product and you update it you get description not properly updated (contains still data from product before change):
image
so currently it works fine only if your name is empty. i'll revert it back.

purchase_requisition_line_description/views/purchase_requisition.xml
```diff
@@ -0,0 +1,16 @@
...
+<odoo>
+    <record model="ir.ui.view" id="purchase_requisition_chatter_view">
+      <field name="name">purchase_requisition_form</field>
+      <field name="model">purchase.requisition</field>
+      <field name="inherit_id" ref="purchase_requisition.view_purchase_requisition_form"/>
+      <field name="arch" type="xml">
+        <xpath expr="//notebook/page/field[@name='line_ids']/tree/field[@name='product_id']" position="attributes">
+          <attribute name="context">{'vendor_data_search':True}</attribute>
+        </xpath>
+        <xpath expr="//notebook/page/field[@name='line_ids']/tree/field[@name='product_id']" position="after">
```
i-vyshnevska:
```suggestion
        <xpath expr="//field[@name='product_id']" position="attributes">
          <attribute name="context">{'vendor_data_search':True}</attribute>
        </xpath>
        <xpath expr="//field[@name='product_id']" position="after">
```

purchase_requisition_line_description/models/product.py
i-vyshnevska:
```suggestion
        if name and self.env.context.get('vendor_data_search'):
            product_ids = self._search(
                    [('name', operator, name)] + args,
                    limit=limit,
                    access_rights_uid=name_get_uid,
                )
            positive_operators = ['=', 'ilike', '=ilike', 'like', '=like']
            if operator in positive_operators:
                product_ids += self._search(
                    [('seller_ids.product_code', operator, name)] + args,
                    limit=limit,
                    access_rights_uid=name_get_uid,
                )
                if not product_ids:
                    # this condition will never run
                    product_ids += self._search(
                        [('seller_ids.product_name', operator, name)] + args,
                        limit=limit,
                        access_rights_uid=name_get_uid,
                    )
           return self.browse(product_ids).name_get()
        return super()._name_search(name, args, operator, limit, name_get_uid)
```

ajaniszewska-dev: @i-vyshnevska  this code doesn't work. Why do we need to return name_get here?

ajaniszewska-dev: so far this works:
```

        product_ids = []
        if name and self.env.context.get('vendor_data_search'):
            product_ids = self._search(
                    [('name', operator, name)] + args,
                    limit=limit,
                    access_rights_uid=name_get_uid,
                )
            positive_operators = ['=', 'ilike', '=ilike', 'like', '=like']
            if operator in positive_operators:
                product_ids += self._search(
                    ['|',('seller_ids.product_code', operator, name),('seller_ids.product_name', operator, name)] + args,
                    limit=limit,
                    access_rights_uid=name_get_uid,
                )
        return self.browse(product_ids).name_get()
```
pls clarify what you plan to achieve with name_get and name_search instead of browsing on record, how it is done in default method.

i-vyshnevska (approved):

i-vyshnevska:
pls check lint to fix build

purchase_requisition_line_description/models/purchase_requisition_line.py
```diff
@@ -0,0 +1,32 @@
...
+        if self.product_id:
+            product_lang = self.product_id.with_context(
+                lang=self.requisition_id.purchase_ids.partner_id.lang,
+                partner_id=self.requisition_id.purchase_ids.partner_id.id,
+            )
+            vendor_data = product_lang.seller_ids.filtered(
+                lambda a: a.product_id == self.product_id
+                and (a.product_name
+                or a.product_code)
+            )[0]
```
grindtildeath: This will probably raise an error if the recordset is empty. You can instead use `fields.first` to safely select first occurrence in a recordset.
Moreover, as this is an OCA module, IMO we should respect the default behaviour of Odoo from the purchase module and only use the display name: https://github.com/odoo/odoo/blob/12.0/addons/purchase/models/purchase.py#L595-L597

purchase_requisition_line_description/models/product.py
```diff
@@ -0,0 +1,28 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, models
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    @api.model
+    def _name_search(
```
grindtildeath: IMO this looks like a specific behavior we don't necessarily want in an OCA module.

grindtildeath (commented):
Can you please add readme sections and a test?

purchase_requisition_line_description/tests/test_purchase_requisition_line_description.py
```diff
@@ -0,0 +1,37 @@
...
+                'product_qty': 15.0,
+                'product_uom': cls.env.ref('uom.product_uom_unit').id,
+                'date_planned': fields.Date.today(),
+            })]
+        })
+
+    def test_onchange_product_id(self):
+        self.assertEqual(self.product.name, self.purchase.line_ids[0].name)
+        # Test onchange product
+        self.purchase.line_ids[0]._onchange_product_id()
```
grindtildeath: IMO would be better to use `Form` (https://github.com/odoo/odoo/blob/12.0/odoo/tests/common.py#L1075)  to ensure the onchange is triggered automatically. :wink:

grindtildeath (approved):
LGTM, some nitpicking on the tests, and travis still to fix

leemannd:
Hello @kittiu, @HviorForgeFlow, If you have the time, could you have a look and maybe help us to have it merged?

HviorForgeFlow (approved):
Code review LGTM :+1:

PR #1158, opened 2021-04-16
[8.0][FIX] purchase_order_reorder_lines: Add purchase_partial_invoicing depends to prevent error in tests

victoralmau:
Add `purchase_partial_invoicing` depends to prevent error in tests.

Please @joao-p-marques can you review it?

@Tecnativa TT28516

### DIFF
--- a/purchase_order_reorder_lines/tests/test_invoice_on_lines.py
+++ b/purchase_order_reorder_lines/tests/test_invoice_on_lines.py
@@ -27,6 +27,8 @@ from openerp.tools.safe_eval import safe_eval
 
 class testPurchasePartialInvoicing(common.TransactionCase):
 
+    post_install = True
+
     def setUp(self):
         super(testPurchasePartialInvoicing, self).setUp()
         self.context = self.env['res.users'].context_get()


joao-p-marques:
@victoralmau can you explain a little bit better why this is needed please? To leave it as reference and to see if maybe it needs to be forward-ported.

pedrobaeza:
Is there a real dependency or just for tests? If the latter, this fix is incorrect, and maybe the problem is to execute tests in post-install.

purchase_order_reorder_lines/tests/test_invoice_on_lines.py
```diff
@@ -27,6 +27,8 @@
 
 class testPurchasePartialInvoicing(common.TransactionCase):
 
+    post_install = True
```
pedrobaeza, line 30: Well, I was saying the contrary... Can you start putting the stack trace of the error, for knowing what is happening?

victoralmau: Ok, sorry.

Stack trace when running test is:
`ValueError: Invalid field 'fully_invoiced' in leaf "<osv.ExtendedLeaf: ('fully_invoiced', '=', False) on purchase_order_line (ctx: )>"`

pedrobaeza: I'm afraid that's not enough. Where is that risen? As we are doing this as inherited and temporary while we migrate to 13.0 version, I'm thinking of just disabling the execution of this test in our integration CI.

PR #1159, opened 2021-04-16
[12.0][FIX]purchase_open_qty do not take into account the direction(outgoing/incoming) of the movePurchase open qty fix

RodolfoLima:
When a return is made in picking the calculation does not take into account if the item is incoming or outgoing
Example:
|Steps|Description|
|---|---|
|1.| Purchase Order, and receive items: qty:10, qty to receive:0, received qty:10|
|2.| Return items(3 in this case), check “To Refund (update SO/PO)”|
|3.|Receive the items returned, check “To Refund (update SO/PO)”|

Expected behavior after step 2:
|Steps|Expected Behaviour|
|---|---|
|2.| purchase order: qty:10, qty to receive:3, received qty:7|
|3.|purchase order: qty:10,	qty to receive:0, received qty:10|

Current behavior after step 2:
|Steps|Current Behaviour|
|---|---|
|2.| purchase order: qty:10, qty to receive:-3 received qty:7|
|3.|purchase order: qty:10,	qty to receive:-6, received qty:10|

In the current implementation quantity to receive does not take into account the direction of the move and always subtract the line.move ‘done’ no matter if is a return or not, as can be seen in the following code.

https://github.com/OCA/purchase-workflow/blob/ac39e8561abe4bc76f3a3ab85b001343add59ee4/purchase_open_qty/models/purchase_order.py#L33-L34
This patch fixes this issue.


HviorForgeFlow:
Travis is :red_circle: , can you squash commits?  Thanks!

@RodolfoLima

jpoa:
Can this be marked as "no stale"?

PR #1161, opened 2021-04-27
[12.0][FIX] purchase_discount error on creating product.supplierinfo

Zar-SW:
record where 'product_tmp_id' not always in the vals

### DIFF
--- a/purchase_discount/models/product_supplierinfo.py
+++ b/purchase_discount/models/product_supplierinfo.py
@@ -32,7 +32,7 @@ class ProductSupplierInfo(models.Model):
         """ Insert discount (or others) from context from purchase.order's
         _add_supplier_to_product method """
         for vals in vals_list:
-            product_tmpl_id = vals['product_tmpl_id']
+            product_tmpl_id = vals.get('product_tmpl_id', False)
             po_line_map = self.env.context.get('po_line_map', {})
             if product_tmpl_id in po_line_map:
                 po_line = po_line_map[product_tmpl_id]


purchase_discount/models/product_supplierinfo.py
```diff
@@ -32,7 +32,7 @@ def create(self, vals_list):
         """ Insert discount (or others) from context from purchase.order's
         _add_supplier_to_product method """
         for vals in vals_list:
-            product_tmpl_id = vals['product_tmpl_id']
+            product_tmpl_id = vals.get('product_tmpl_id', False)
```
legalsylvain, line 35: I don't understand in wich case the value is not set. 

I think that if not defined, Odoo will not work correctly, as the link between the product.template and the product.supplierinfo is lost. don't you think ? 

https://github.com/odoo/odoo/blob/12.0/addons/product/models/product_template.py#L114

legalsylvain (commented):
one remark inline.

purchase_discount/models/product_supplierinfo.py
Zar-SW: @legalsylvain 
If there is a custom module that creates a product.supplierinfo then link it to the product the vals on create function can be without product_tmpl_id

HviorForgeFlow (approved):

HviorForgeFlow:
IMO Change does not hurt and give more flexibility for "custom" modules

pedrobaeza (changes requested):
I think the same as Sylvain. Which are the steps to reproduce a problem? If third-party modules modify the dictionary values, then such module should care about this.

Zar-SW:
@pedrobaeza The Change does not hurt and give more flexibility for third-party modules

pedrobaeza:
This is not giving flexibility, it's having an invalid data structure.

ypapouin:
Does this happen on a migrated database ?

PR #1162, opened 2021-04-27
[12.0][FIX] date approved missing when order is approved. An order is 'approved' but Odoo default 'date approve' is blank, so it confuse the user

sergiocorato:


legalsylvain:
why not reuse and populate ``date_approve`` native field. I see it a little confusing to have 2 fields. ``date_approved`` and ``date_approve`` in the same table. 

Don't you think ?

sergiocorato:
I thought the same, but this will be overwritten [here](https://github.com/odoo/odoo/blob/5cf7acb17566a28f59351400b04496acad50a95c/addons/purchase/models/purchase.py#L319) and the information will be lost.
From the POV of this module, the default field `date_approve` become a `Confirmation Date`, when the RFQ become a PO.
So another possible solution is changing the string of the field? And add a more specified name in table?

legalsylvain:
I understand. An alternative could be, in the overload of ``button_approve`` to 
- get the values of ``date_approve``, and store it in a dict {purchase_id: date_approve}
- call super
- rewrive the erased values of ``date_approve``.

that way one single field.

What do you think ?

sergiocorato:
Ok, good approach.

legalsylvain (approved):
thanks. LGTM. Technical review only. I personally don't use this feature.

sergiocorato:
@OCA/procurement-purchase-maintainers reviews?

HviorForgeFlow (approved):

PR #1163, opened 2021-04-27
[13.0][FIX] purchase_order_product_recommendation: Fix tests

CarlosRoca13:
cc @Tecnativa TT29377

After commit odoo/odoo@4625f3d
the warehouse is taken into account when computing the quantities.

As we are limiting the test by warehouse, the final value should be updated to the new
correctly computed.

Please @sergio-teruel @carlosdauden review this.

### DIFF
--- a/purchase_order_product_recommendation/tests/test_recommendation.py
+++ b/purchase_order_product_recommendation/tests/test_recommendation.py
@@ -237,9 +237,9 @@ class RecommendationCaseTests(RecommendationCase):
         line_prod_3 = wizard.line_ids.filtered(lambda x: x.product_id == self.prod_3)
         self.assertEqual(line_prod_3.times_delivered, 1)
         self.assertEqual(line_prod_3.units_delivered, 13)
-        self.assertEqual(line_prod_3.units_included, 9)
-        self.assertEqual(line_prod_3.units_available, 4)
-        self.assertEqual(line_prod_3.units_virtual_available, 4)
+        self.assertEqual(line_prod_3.units_included, 8)
+        self.assertEqual(line_prod_3.units_available, 5)
+        self.assertEqual(line_prod_3.units_virtual_available, 5)
 
     def test_action_accept(self):
         """Open wizard when there are PO Lines and click on Accept"""


purchase_order_product_recommendation/i18n/es.po
```diff
@@ -138,6 +176,11 @@ msgstr "Producto"
 msgid "Product Categories"
 msgstr "Categorías de productos"
 
+#. module: purchase_order_product_recommendation
+#: model:ir.actions.report,name:purchase_order_product_recommendation.recommendation_xlsx
+msgid "Product Recommendation XLSX"
+msgstr "Productos recomendados XLSX"
```
joao-p-marques:
```suggestion
msgstr "XLSX de productos recomendados"
```

joao-p-marques (approved):
Code review :+1: 

Just a small suggestion for the translations.

purchase_order_product_recommendation/__manifest__.py
```diff
@@ -11,9 +11,10 @@
     "license": "AGPL-3",
     "application": False,
     "installable": True,
-    "depends": ["purchase_stock"],
+    "depends": ["purchase_stock", "report_xlsx"],
```
chienandalu: Isn't this adding a very off the main target dependency? Why not an extra module?

CarlosRoca13: Ok so, I will preserve this PR for the fix and do a new one for the XLS export module

joao-p-marques (approved):

CarlosRoca13:
@chienandalu review it again please

purchase_order_product_recommendation/tests/test_recommendation.py
```diff
@@ -237,7 +237,7 @@ def test_recommendations_by_warehouse(self):
         line_prod_3 = wizard.line_ids.filtered(lambda x: x.product_id == self.prod_3)
         self.assertEqual(line_prod_3.times_delivered, 1)
         self.assertEqual(line_prod_3.units_delivered, 13)
-        self.assertEqual(line_prod_3.units_included, 9)
+        self.assertEqual(line_prod_3.units_included, 8)
```
chienandalu, line 240: Tracing this change, I see it was intruduced in the module migration. I wonder why it passed then :thinking:

joao-p-marques: It is due to a commit introduced in odoo 2 days ago: https://github.com/odoo/odoo/commit/4625f3d

chienandalu: Ok. But I wonder why this was forced in the first place: https://github.com/OCA/purchase-workflow/commit/e79a33913644c0e1af79ac2104f0781352a62133#diff-34f3284cd0131dc2295261002115060ea32adfa877cfba1bf2255f76447a036dR239-R241

joao-p-marques: Yes, maybe this bug was introduced in v13 and that change had to be done to adapt to it...

CarlosRoca13: It was solved two days ago on this commit https://github.com/odoo/odoo/commit/4625f3d

So that's why wasn't failing since now

chienandalu: I mean: the test was tricked in the migration to fit the result and pass then. Look that the triggering condition is the change of a date. Could it be one of those cases that depend on the hour of the day you run the test?

chienandalu (approved):
:+1:

sergio-teruel (approved):

PR #1164, opened 2021-04-27
[13.0][FIX] purchase_order_product_recommendation_brand: There is not products when a brand is selected

sergio-teruel:
cc @Tecnativa TT29448

ping @chienandalu @CarlosRoca13

### DIFF
--- a/purchase_order_product_recommendation_brand/wizards/purchase_order_recommendation.py
+++ b/purchase_order_product_recommendation_brand/wizards/purchase_order_recommendation.py
@@ -14,8 +14,11 @@ class PurchaseOrderRecommendation(models.TransientModel):
         # Filter products by brand if set.
         # It will apply to show_all_partner_products as well
         if self.product_brand_ids:
+            # We are in onchange context, so to avoid to compare records with
+            # new_id instances we use self.product_brand_ids.ids,
+            # self.product_brand_ids = newid but self.product_brand_ids.ids is a real id
             products = products.filtered(
-                lambda x: x.product_brand_id in self.product_brand_ids
+                lambda x: x.product_brand_id.id in self.product_brand_ids.ids
             )
         return products
 


purchase_order_product_recommendation_brand/wizards/purchase_order_recommendation.py
```diff
@@ -14,8 +14,11 @@ def _get_products(self):
         # Filter products by brand if set.
         # It will apply to show_all_partner_products as well
         if self.product_brand_ids:
+            # We are in onchange context, so to avoid to compare records with
+            # new_id instances we use self.product_brand_ids.ids,
+            # self.product_brand_ids = newid but self.product_brand_ids.ids is a real id
             products = products.filtered(
-                lambda x: x.product_brand_id in self.product_brand_ids
+                lambda x: x.product_brand_id.id in self.product_brand_ids.ids
```
chienandalu, line 21: Maybe better something like this?:

```suggestion
                lambda x: x.product_brand_id._origin in self.product_brand_ids
```

sergio-teruel: @CarlosRoca13 please, can you explain to @chienandalu why is better use .ids instead origin.
Anyway, the newid instance is self.product_brand_ids not x.product_brand_id

CarlosRoca13: The field `ids` stores the ids involved on the NewId object. So it is a direct way to get the value. We have to use the field `_origin` when we have a cache record, this normally has `ids = []`

chienandalu: ok, ok. Tests are failing anyway

CarlosRoca13:
I fix the test error on https://github.com/OCA/purchase-workflow/pull/1163

sergio-teruel:
@chienandalu @CarlosRoca13 Rebase done after https://github.com/OCA/purchase-workflow/pull/1163 has been merged!

CarlosRoca13 (approved):

chienandalu (approved):

PR #1166, opened 2021-04-29
[14.0][ADD] purchase_advance_payment

MateuGForgeFlow:
The module allows to add advance payments on purchase orders.
cc @LoisRForgeFlow

### DIFF
--- a/purchase_advance_payment/readme/DESCRIPTION.rst
+++ b/purchase_advance_payment/readme/DESCRIPTION.rst
@@ -1,0 +1,3 @@
+The module allows to add advance payments on purchase orders and Request
+for Quotation. The advance payments are allowed even before confirmation
+and before starting the billing process.
--- a/purchase_advance_payment/__init__.py
+++ b/purchase_advance_payment/__init__.py
@@ -1,0 +1,2 @@
+from . import models
+from . import wizard
--- a/purchase_advance_payment/__manifest__.py
+++ b/purchase_advance_payment/__manifest__.py
@@ -1,0 +1,19 @@
+# Copyright (C) 2021 ForgeFlow S.L.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html)
+
+{
+    "name": "Purchase Advance Payment",
+    "version": "14.0.1.0.0",
+    "author": "Forgeflow, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/purchase-workflow",
+    "category": "Purchase",
+    "license": "AGPL-3",
+    "summary": "Allow to add advance payments on purchase orders",
+    "depends": ["purchase"],
+    "data": [
+        "wizard/purchase_advance_payment_wizard_view.xml",
+        "views/purchase_view.xml",
+        "security/ir.model.access.csv",
+    ],
+    "installable": True,
+}
--- a/purchase_advance_payment/models/__init__.py
+++ b/purchase_advance_payment/models/__init__.py
@@ -1,0 +1,2 @@
+from . import payment
+from . import purchase_order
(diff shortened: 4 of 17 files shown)


kittiu:
Is this same or different use case with `purchase_deposit` https://github.com/OCA/purchase-workflow/tree/13.0/purchase_deposit

MateuGForgeFlow:
Hi @kittiu. It is very similar. However, the purchase_advance_payment only registers payments and then you link them with the bill. It also allows to register payments in other currencies.

kittiu:
> Hi @kittiu. It is very similar. However, the purchase_advance_payment only registers payments and then you link them with the bill. It also allows to register payments in other currencies.

Is it possible to enhance that one to also cover your need instead of adding a new module. By no mean object this one, just in case it can.

LoisRForgeFlow:
Hi @kittiu thanks for reviewing here :)

I know the two modules apparently overlap a bit, but we think they are 2 different approaches for very similar needs, it is a almost a matter of taste of the customer (and/or the implementator). So having it as a 2 separate modules makes more sense to me in this case, it is not necessarily bad that the OCA provides different alternatives, don't you think?

kittiu:
> Hi @kittiu thanks for reviewing here :)
> 
> I know the two modules apparently overlap a bit, but we think they are 2 different approaches for very similar needs, it is a almost a matter of taste of the customer (and/or the implementator). So having it as a 2 separate modules makes more sense to me in this case, it is not necessarily bad that the OCA provides different alternatives, don't you think?

@LoisRForgeFlow. Per say, I have no objection, just to make sure it is properly reviewed. Please continue, and I will be happy to review it once ready :).

purchase_advance_payment/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+The module allows to add advance payments on purchase orders.
```
kittiu: More detail description of this module will be helpful. I..e, I see the main idea (to use this instead of other module) of this module is to allow adding payment for this particular PO much in advance, even before the vendor bill. Is it right?

kittiu (commented):
Functional Review:

I can see the idea of this module is to allow making payment **much** in advance, before any Bill is created (full amount even in draft will hide button "Pay purchase advanced"), right?

image

But if go normal path to create Bill and then Register Payment.

Even Bill is fully Paid (1), user can still see (2) Residual Amount in full, which is misleading as it should be 0.0 and (3) Advance Tab also open, though nothing there.

I think,

1. Residual amount should take paid amount from billing into account?
2. If advance payment is not used at all in this PO, should we hide Residual amount?
3. Payment advanced tab should be hidden if no advance?

---

For code review, I can't find things to complain :smile:  :+1:

purchase_advance_payment/readme/DESCRIPTION.rst
nmoturi: I see the cleaner implementation in purchase_deposit module, which is inline to customer down payments functionality. I understand whilst the alternatives solutions could be encouraging but would be good if they align to overall workflow in other functions.

LoisRForgeFlow:
> Functional Review:
> 
> I can see the idea of this module is to allow making payment **much** in advance, before any Bill is created (full amount even in draft will hide button "Pay purchase advanced"), right?
> 
> image
> 
> But if go normal path to create Bill and then Register Payment.
> 
> Even Bill is fully Paid (1), user can still see (2) Residual Amount in full, which is misleading as it should be 0.0 and (3) Advance Tab also open, though nothing there.
> 
> I think,
> 
> 1. Residual amount should take paid amount from billing into account?
> 2. If advance payment is not used at all in this PO, should we hide Residual amount?
> 3. Payment advanced tab should be hidden if no advance?
> 
> For code review, I can't find things to complain smile +1

Hey @kittiu 

I agree with you, we should consider the regular payments to compute the residual amount, I have just corrected that. Can you have a look and update your review?

Many thanks!

LoisRForgeFlow (approved):
From my side this is good to merge, future improvements can be added for sure, but it is a functional module already.

kittiu (approved):

PR #898, opened 2020-04-22
[13.0][MIG] purchase_order_type

ps-tubtim:
Add company_id

PR #1170, opened 2021-05-01
[14.0][FIX] purchase_order_type, default type not set

kittiu:
The original _default_order_type() that set default order type no longer function because of the enhancement about multi-company https://github.com/OCA/purchase-workflow/blob/14.0/purchase_order_type/models/purchase_order.py#L65

(And since the default required field is not set, this also make test in other module red)

**Note:** this fix should back port to v13 too.

@ps-tubtim can you review please. Looks like company was from your commit - https://github.com/OCA/purchase-workflow/pull/898/commits/c4904605b934baa6ebb7be5affa3723cb985ba3e


HviorForgeFlow (approved):

ps-tubtim (approved):
Code LGTM

PR #1172, opened 2021-05-05
[13.0][MIG] purchase_propagate_qty

mmequignon:
Migration from v12 to v13


purchase_propagate_qty/tests/test_purchase_delivery.py
```diff
@@ -0,0 +1,88 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo.exceptions import UserError
+from odoo.tests.common import SavepointCase
+
+
+class TestQtyUpdate(SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
```
TDu, line 16: > cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))

Always good to add it in tests

purchase_propagate_qty/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,2 @@
+* Alexandre Fayolle <alexandre.fayolle@camptocamp.com>
+* Matthieu Mequignon <matthieu.mequignon@camptocamp.com>
```
simahawk, line 2: I think we should see original authors too here

purchase_propagate_qty/readme/HISTORY.rst
```diff
@@ -0,0 +1,4 @@
+12.0.1.0.0 (2020-04-30)
```
simahawk: if you don't update this manually or add a towcrier compatible this file is useless

purchase_propagate_qty/tests/test_purchase_delivery.py
simahawk: AMEN :wink:

purchase_propagate_qty/tests/test_purchase_delivery.py
```diff
@@ -0,0 +1,88 @@
...
+        """decrease qty on confirmed po -> decreased reception"""
+        line1 = self.po.order_line[0]
+        move1 = self.env["stock.move"].search([("purchase_line_id", "=", line1.id)])
+        line1.write({"product_qty": 30})
+        self.assertEqual(move1.product_uom_qty, 30)
+
+    def test_purchase_line_unlink(self):
+        """decrease qty on confirmed po -> decreased reception"""
+        line1 = self.po.order_line[0]
+        with self.assertRaises(UserError):
```
simahawk: use `assertRaiseRegex` and check you get a specific msg

purchase_propagate_qty/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2014-2016 Numérigraphe SARL
+# Copyright 2017 Eficent Business and IT Consulting Services, S.L.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Purchase Propagate Quantity",
+    "version": "13.0.1.0.0",
+    "summary": "Quantity decrease on purchase line are propagated "
+    "to the corresponding stock.move",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
```
simahawk: what about the authors in the copyright? :thinking:

simahawk (changes requested):

purchase_propagate_qty/readme/CONTRIBUTORS.rst
mmequignon: Well, I just appended my name there…
https://github.com/OCA/purchase-workflow/blob/12.0/purchase_propagate_qty/README.rst

simahawk:
build is :red_circle:

simahawk (approved):

simahawk:
Build is red because Odoo core is broken by this commit https://github.com/odoo/odoo/commit/c347f679d6a635428b12ffa04b5e16b4f2d3d78f
Fix https://github.com/odoo/odoo/pull/70800

purchase_propagate_qty/models/purchase.py
```diff
@@ -0,0 +1,30 @@
...
+            if line.state != "purchase":
+                continue
+            for move in line.move_dest_ids | line.move_ids:
+                if move.state in ("cancel", "done"):
+                    continue
+                if move.product_id != line.product_id:
+                    continue
+                move.product_uom_qty = line.product_uom_qty
+                if float_is_zero(move.product_uom_qty, line.product_uom.rounding):
+                    move._action_cancel()
```
mmequignon, line 30: This doesn't work when increasing the POL qty.
In purchase stock, a new move is created in such case, with the diff between old POL qty and new POL qty.
This method here sets both moves qties to the new POL qty.
e.g.
POL line with qty set to 100.
User sets the qty to 110.
Odoo creates a new move with qty = 10.
This method sets both qties to 110.

purchase_propagate_qty/models/purchase.py
```diff
@@ -20,11 +20,46 @@ def _propagage_qty_to_moves(self):
...
                 continue
-            for move in line.move_dest_ids | line.move_ids:
-                if move.state in ("cancel", "done"):
-                    continue
-                if move.product_id != line.product_id:
-                    continue
-                move.product_uom_qty = line.product_uom_qty
-                if float_is_zero(move.product_uom_qty, line.product_uom.rounding):
-                    move._action_cancel()
+            moves = line.move_dest_ids | line.move_ids
```
sebalix: I don't like this fix... it seems overcomplicated. If I understand well, it has been done to handle a situation where we increase/decrease qty of a purchase order line having several moves.
The question is: in which situation we have several available moves related to a purchase order line? Isn't the issue is to have several moves here?

mmequignon: still, if it can happen, then we have to fix this.
I'll investigate further.

mmequignon: I'll create a new pr for that, let's discuss about this on it.

sebalix (approved):

purchase_propagate_qty/models/purchase.py
sebalix: Link to the new PR: https://github.com/OCA/purchase-workflow/pull/1233

PR #1175, opened 2021-05-11
[13.0][FIX] purchase_delivery_split_date

TDu:
On an order already confirmed, if a scheduled date is changed on an
existing line, the move is correctly assign to the picking.
But if a line is added with it is not the case, this changes fixes that.

### DIFF
--- a/purchase_delivery_split_date/tests/test_purchase_delivery.py
+++ b/purchase_delivery_split_date/tests/test_purchase_delivery.py
@@ -17,6 +17,9 @@ class TestDeliverySingle(TransactionCase):
         p2 = self.product2 = self.product_model.create(
             {"name": "Test Product 2", "type": "product", "default_code": "PROD2"}
         )
+        self.p3 = self.product2 = self.product_model.create(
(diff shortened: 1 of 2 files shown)


purchase_delivery_split_date/tests/test_purchase_delivery.py
```diff
@@ -147,3 +150,37 @@ def test_purchase_line_date_change_split_picking(self):
...
+
+        """
+        self.po.button_confirm()
+        self.assertEqual(self.po.state, "purchase")
+        new_date = "2016-01-30"
+        moves_before = self.env["stock.move"].search(
+            [("purchase_line_id", "in", self.po.order_line.ids)]
+        )
+        self.assertEqual(len(moves_before.mapped("picking_id")), 1)
+        self.po.order_line = [
```
simahawk, line 169: cosmetic: an explicit call to create line or write order would be more clear IMO.

simahawk (approved):

HviorForgeFlow (approved):

PR #1178, opened 2021-05-19
[14.0][MIG] purchase_deposit

kittiu:
Standard migration, ready for review.


kittiu:
Test passed (ci/runbot has problem currently?)

```
+=======================================
|  Tests summary:
|---------------------------------------
| test_server.py              [1;32mSuccess[0;m
+=======================================
 * Stopping PostgreSQL 9.6 database server       [80G 
```

kittiu:
cc @meksrv please review.

meksrv:
Functional Tested: Look good to me.

ygol:
LGTM functionally.

kittiu:
> LGTM functionally.

Would you approve? Thanks!

ygol:
Question: 
- It will be very usefull to have the "uninvoiced amount" stored in a field to allow showing it in the tree view.
- It would be nice to add in the "vendor bills" smart button, the invoiced amount or alternatively, as in the https://github.com/OCA/sale-workflow/tree/13.0/sale_order_invoice_amount

image


I tested the module https://github.com/OCA/purchase-workflow/tree/14.0/purchase_order_uninvoiced_amount but it doesn't work because it re-calculate the amount from the "order.order_line".

should this feature be part of this module? 
what do you think?

kittiu:
> order_line

I would think it should be on separated module. i.e., in purchase_order_uninvoiced_amount  as you pointed.
This module should be kept simple, WDYT?

But...why you think it should be here? may be I miss something.

ygol:
> > order_line
> 
> I would think it should be on separated module. i.e., in purchase_order_uninvoiced_amount as you pointed.
> This module should be kept simple, WDYT?
> 
> But...why you think it should be here? may be I miss something.

My bad. You are correct that this must be a separate module. There are indeed cases where you will have multiple bills and no  deposits.

nmoturi:
@kittiu - From a usability perspective, do we need to have a seperate button for "Register Deposit", why cannot be integrated into "create bill" option similar to customer invoice option. This would look much cleaner and align to similar look and feel of sales.

image

kittiu:
> @kittiu - From a usability perspective, do we need to have a seperate button for "Register Deposit", why cannot be integrated into "create bill" option similar to customer invoice option. This would look much cleaner and align to similar look and feel of sales.

Mostly for the modularity purposes. Changing the way the existing button behave can conflict other modules that might inherited from it.

Although it is possible (that the original was written in a way that allows it). We have to be more careful.

nmoturi:
> > @kittiu - From a usability perspective, do we need to have a seperate button for "Register Deposit", why cannot be integrated into "create bill" option similar to customer invoice option. This would look much cleaner and align to similar look and feel of sales.
> 
> Mostly for the modularity purposes. Changing the way the existing button behave can conflict other modules that might inherited from it.
> 
> Although it is possible (that the original was written in a way that allows it). We have to be more careful.

Thanks @kittiu, that makes sense but this is the downside when migrating from previous versions not taking the advantage of UI improvements from newer version. I believe as much as functional importance was given need to be considerate on the UI as well. All good.

kittiu:
> Thanks @kittiu, that makes sense but this is the downside when migrating from previous versions not taking the advantage of UI improvements from newer version. I believe as much as functional importance was given need to be considerate on the UI as well. All good.

Thanks for sharing ideas, OK, I think we can try that. But let's do it as another PR with minor version. I have many other custom modules in use this module, so I can fix version if needed.

Can you help review/approve this so we could move one? @ygol @nmoturi 

Thanks!

nmoturi:
Thanks @Kitti. I reviewed the functional flow and I haven't found any issue and it does exactly what needed to be.

kittiu:
> Thanks @kitti. I reviewed the functional flow and I haven't found any issue and it does exactly what needed to be.

Can you click to approve too :pray:

ygol (approved):

kittiu:
@nmoturi @newtratip can you help review/approve this. This is a while and it could be moved on.
Thanks!

nmoturi (approved):

kittiu:
ping @OCA/procurement-purchase-maintainers

PR #1180, opened 2021-05-19
[14.0][IMP] purchase_invoice_plan, misc improvements

kittiu:
Improvement:

* When create invoice by invoice plan, to better handle the last invoice (handle decimals)
* Make a hook method (to be extended with purchase_invoice_plan_deposit)
* Use function account_create_invoice (in v14) instead of module own account_invoice_create (removed)
* Add amount in invoice plan, as helper to calculate percent
* Add sum invoice plan total, and percent

### DIFF
--- a/purchase_invoice_plan/models/purchase.py
+++ b/purchase_invoice_plan/models/purchase.py
@@ -96,59 +96,20 @@ class PurchaseOrder(models.Model):
         next_date = fields.Date.to_string(next_date)
         return next_date
 
-    def action_invoice_create(self):
-        journal = (
-            self.env["account.move"]
-            .with_context(
-                default_move_type="in_invoice",
-                default_currency_id=self.currency_id.id,
-                default_company_id=self.env.user.company_id.id,
-            )
-            ._get_default_journal()
-        )
-        pre_inv = self.env["account.move"].new(
-            {
-                "move_type": "in_invoice",
-                "purchase_id": self.id,
-                "journal_id": journal.id,
-                "currency_id": self.currency_id.id,
-                "company_id": self.company_id.id,
-                "invoice_origin": self.name,
-                "ref": self.partner_ref or "",
-                "narration": self.notes,
-                "fiscal_position_id": self.fiscal_position_id.id,
-                "invoice_payment_term_id": self.payment_term_id.id,
-            }
-        )
-        pre_inv._onchange_purchase_auto_complete()
-        inv_data = pre_inv._convert_to_write(pre_inv._cache)
-        invoice = self.env["account.move"].create(inv_data)
-        if not invoice.invoice_line_ids:
-            raise UserError(
-                _(
-                    "There is no invoiceable line. If a product has a"
-                    "Delivered quantities invoicing policy, please make sure"
-                    "that a quantity has been delivered."
-                )
-            )
-        invoice._onchange_partner_id()
-        invoice.message_post_with_view(
-            "mail.message_origin_link",
-            values={"self": invoice, "origin": self},
-            subtype_id=self.env.ref("mail.mt_note").id,
-        )
+    def action_view_invoice(self, invoices=False):
         invoice_plan_id = self._context.get("invoice_plan_id")
         if invoice_plan_id:
             plan = self.env["purchase.invoice.plan"].browse(invoice_plan_id)
-            plan._compute_new_invoice_quantity(invoice)
-            invoice.write(
-                {
-                    "date": plan.plan_date,
-                    "invoice_date": plan.plan_date,
-                }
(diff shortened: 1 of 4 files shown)


kittiu:
cc @AaronHForgeFlow @newtratip

AaronHForgeFlow:
Would it be possible to move the hook to a separate module and then purchase_invoice_plan and purchase_invoice_plan_deposit would inherit that module? It is better organized that way.

kittiu:
> Would it be possible to move the hook to a separate module and then purchase_invoice_plan and purchase_invoice_plan_deposit would inherit that module? It is better organized that way.

But if you look at the hook, it is just a normal subroutine. Wouldn't notice if I don't comment it. :) WDYT?

AaronHForgeFlow:
> > Would it be possible to move the hook to a separate module and then purchase_invoice_plan and purchase_invoice_plan_deposit would inherit that module? It is better organized that way.
> 
> But if you look at the hook, it is just a normal subroutine. Wouldn't notice if I don't comment it. :) WDYT?

Yes, please dismiss my comment, I was thinking on a hook on a hook of a standard method, but it is not the case

kittiu:
@AaronHForgeFlow there is another enhancement I like to do.

* Now, there is no need to have our own method `action_invoice_create`, as v14 already use its own `action_create_invoice`.

If you don't mind to merge this one first, I will include that as another commit here. OK?

kittiu:
Ready for review now.

kittiu:
@newtratip can you review please.

purchase_invoice_plan/models/purchase.py
```diff
@@ -210,6 +188,16 @@ class PurchaseInvoicePlan(models.Model):
...
 
+    @api.depends("percent")
+    def _compute_amount(self):
+        for rec in self:
+            rec.amount = rec.percent * rec.purchase_id.amount_untaxed / 100
+
+    @api.onchange("amount")
+    def _inverse_amount(self):
+        for rec in self:
+            rec.percent = rec.amount / rec.purchase_id.amount_untaxed * 100
```
newtratip: if amount_untaxed = 0, This will error. Please add condition 

```
if rec.purchase_id.amount_untaxed != 0:
     rec.percent = rec.amount / rec.purchase_id.amount_untaxed * 100
```

kittiu: Fixed

kittiu:
cc @newtratip @AaronHForgeFlow should be ready.

kittiu:
Another UI improvement.

image

kittiu:
@newtratip @AaronHForgeFlow too many improvements here already. Would appreciate your reviews.

AaronHForgeFlow:
There is a small rounding issue. I have a PO with one product and subtotal 150$, however, bacause rounding issues, the invoice plan looks like this (small difference):
image

AaronHForgeFlow:
> There is a small rounding issue. I have a PO with one product and subtotal 150$, however, bacause rounding issues, the invoice plan looks like this (small difference):
> image

This is still happening, the all the other improvements are ok to me

kittiu:
@AaronHForgeFlow good catch, thanks!

Now, we make sure amount also consider being last installment into account. In fact, it will auto adjust itself using the remaining amount/percent (no matter what user type in).

kittiu:
@AaronHForgeFlow is this good to go?

purchase_invoice_plan/models/purchase.py
```diff
@@ -29,6 +29,20 @@ class PurchaseOrder(models.Model):
         compute="_compute_ip_invoice_plan",
         help="At least one invoice plan line pending to create invoice",
     )
+    ip_total_percent = fields.Float(
+        compute="_compute_ip_total",
+        string="Percent",
+    )
+    ip_total_amount = fields.Monetary(
+        compute="_compute_ip_total",
+        string="Total",
```
AaronHForgeFlow: I think this is giving a warning in the logs and that is why runbot is red:

2021-07-17 08:00:38,675 167 INFO openerp_test odoo.modules.loading: Loading module purchase_order_line_sequence (92/93)
2021-07-17 08:00:39,132 167 INFO openerp_test odoo.modules.registry: module purchase_order_line_sequence: creating or updating database tables
2021-07-17 08:00:39,147 167 WARNING openerp_test odoo.addons.base.models.ir_model: Two fields (ip_total_amount, amount_total) of purchase.order() have the same label: Total.

AaronHForgeFlow (commented):
@kittiu , yes I think it is fine now, just one comment

AaronHForgeFlow (approved):

PR #1184, opened 2021-05-20
[14.0][FIX] purchase_request, action.sudo().read()

kittiu:
For v14, calling action.read() will need sudo(), otherwise, access error will occur.


kittiu:
**Note:** I saw this problem in v14 in so many place. Anyone can verify that, the way this PR doing is correct?
cc @pedrobaeza @newtratip @AaronHForgeFlow

pedrobaeza (approved):

AaronHForgeFlow (approved):
Did not know that. Thanks for the fix.

PR #1186, opened 2021-05-24
[14.0][ADD] purchase_invoice_plan_expand

kittiu:
Depends:
- [ ] purchase_invoice_plan #1180

TODO:
- [x] Test
- [x] Readme
- [x] Test with purchase_request, make sure links stays on expand mode.

## Invoice Plan without Group Expand

image

## Invoice plan with expand group

image

image

When create invoice using invoice plan, system will look into the group and will create invoice based on that group only (with correct percent)


kittiu:
Not intuitive enough to use.

PR #1189, opened 2021-05-27
[FIX] purchase_discount: access error

chienandalu:
On certain situations this compute could be triggered by non purchase
users (e.g: Inventory users) so we'll get an access error on it.

cc @Tecnativa TT29910

please review @carlosdauden @pedrobaeza

### DIFF
--- a/purchase_discount/models/stock_move.py
+++ b/purchase_discount/models/stock_move.py
@@ -16,7 +16,7 @@ class StockMove(models.Model):
         is not merged.
         """
         price_unit = False
-        po_line = self.purchase_line_id
+        po_line = self.purchase_line_id.sudo()
         if po_line and self.product_id == po_line.product_id:
             price = po_line._get_discounted_price_unit()
             if price != po_line.price_unit:


purchase_discount/models/stock_move.py
```diff
@@ -16,7 +16,7 @@ def _get_price_unit(self):
         is not merged.
         """
         price_unit = False
-        po_line = self.purchase_line_id
+        po_line = self.purchase_line_id.sudo()
```
victoralmau, line 19:
```suggestion
        po_line = self.purchase_line_id
        if (
            self.user_has_groups("purchase.group_purchase_user")
            or self.user_has_groups("account.group_account_invoice")
        ):
            po_line = self.purchase_line_id.sudo()
```

victoralmau (changes requested):

purchase_discount/models/stock_move.py
pedrobaeza: I think there's no need of over-complicating this. Anyway, my main concern is bout the cache/computed fields glitches doing this sudo change.

chienandalu: Mmm, maybe we could put the `sudo` just on the write parts, as every user has read access anyway. What do you think?

pedrobaeza: As other times, first we need steps to reproduce the problem. They can be depicted in a test, at least for these attempts of solving them.

chienandalu: The minimum steps to reproduce it:

- Create a PO with a line with an storable product and any discount
- Confirm it to create the picking

Now with a user which only has Inventory User permissions and no accounting permissions:

- Try to validate the picking

Access error raises:

```
Sorry, you are not allowed to modify this document. Only users with the following access level are currently allowed to do that:
- Accounting & Finance/Billing
- Purchases/Manager
- Purchases/User

(Document model: purchase.order.line) - (Operation: write, User: 8)
```

`_get_price_unit` is always triggered on the moves `_action_done` so when we try to store the price to do the trick of the discount compute, we can't do in behalf of a user with no `write` permissions on `purchase.order.line`

chienandalu:
Can you update your review on this?

purchase_discount/models/stock_move.py
pedrobaeza: OK, got it, and yes, we don't have better option for now.

PR #1201, opened 2021-06-10
[12.0][ADD] purchase_stock_picking_note

alexeirivera87:
This module allows to add a note in the purchase order which will be set in the generated Receipt


alexeirivera87:
Hello @victoralmau 

Could you please check this MR ? Or someone that could approve/reject it? 

Regards

PR #1204, opened 2021-06-15
[14.0][REF] purchase_work_acceptance, code refactor

kittiu:
Improve code, 

1. when Create Bill, to use `purchase.action_create_invoice()` as suggested in v14.

    In v13, Create Bill was using onchange("purchase_id") on new invoice, which could be wrongly calculate lines when some module also do the onchange on the invoice also.

2. Choosing WA in invoice / picking, to list only unused WA to reduce user error.
3. Click Accept in WA now open wizard for user to enter date (instead of use now)
4. Add purchase order in WA's list view, and filters.

cc @ps-tubtim


kittiu:
This PR contain some refactoring of WA from real use and has been here quite a while. I think this should be merged soon.
For those of us that are using WA, any further comment please?

@albariera @etobella @ps-tubtim

albariera (approved):
Functional review LGTM :+1:

ps-tubtim (approved):
Functional test ok.

Saran440 (approved):
LGTM :+1:

Issue #1209, opened 2021-06-21
[13.0][BUG?] procurement_purchase_no_grouping: missing option?


Rad0van:
Not sure if I get this right. In this module I am missing another option (or at least I cannot get it to work that way). Let's assume this scenario:
1. I create SO1 with 3 Products P1(Supplier 1), P2(Supplier 1), P3(Supplier 2).
2. When I confirm I expect 2 PO to be created: PO1(Supplier 1)=[P1, P2] and PO2(Supplier 2)=[P3]
3. I create SO2 with the same content as SO1.
4. When I confirm it I'd expect to have PO3(Supplier 1)=[P1, P2] and PO4(Supplier 2)=[P3]

So what I want to achieve is basically for each SO to have minimal amount of its own(!) POs - that is grouping lines from the same Supplier into same PO. The closest to it is 'order' option but that will cause each product to be in a separate PO.

Can you please confirm I am not missing anything? If so I will develop appropriate code. Any suggestions for naming option that would do that?

Also a side question (@pedrobaeza maybe?): if I assign analytic code to SO that gets propagated to SOL and that gets added to domain in function _make_po_get_domain in procurement_mto_analytic module which prevents any kind of grouping in any case. Is that intentional?

pedrobaeza:
That option is not covered right now. Analytic is taken intentionally for not grouping lines.

Rad0van:
Thanks for quick answer. Will try to implement the required code.

Rad0van:
Implemented (hopefully) in #1211

PR #1211, opened 2021-06-22
[13.0][IMP] procurement_purchase_no_grouping - add minimal type of grouping

Rad0van:
Following my question in #1209 I have created initial version that implements this kind of grouping:
1. I create SO1 with 3 Products P1(Supplier 1), P2(Supplier 1), P3(Supplier 2).
2. When I confirm I expect 2 PO to be created: PO1(Supplier 1)=[P1, P2] and PO2(Supplier 2)=[P3]
3. I create SO2 with the same content as SO1.
4. When I confirm it I'd expect to have PO3(Supplier 1)=[P1, P2] and PO4(Supplier 2)=[P3]

The idea is to have PO follow individual SO in the minimal way. If all the products are from single vendor, there only will be single PO for individual SO. If there are products from more suppliers in SO, as many POs as there are suppliers will be created.

I have also cleaned a bit of help in Purchase Settings where there was superficial value described that only makes sense for setting per Product Category.

@pedrobaeza would you please provide some comments/review? Especially - is there better way to get `origin` into `_make_po_get_domain` other than as I did inserting it in `_run_buy`? Also I couldn't come up with better name for that option than **minimal** Maybe **origin**? Thank you a lot.

### DIFF
--- a/procurement_purchase_no_grouping/models/product_category.py
+++ b/procurement_purchase_no_grouping/models/product_category.py
@@ -14,6 +14,7 @@ class ProductCategory(models.Model):
             ("line", "No line grouping"),
             ("order", "No order grouping"),
             ("product_category", "Product category grouping"),
+            ("minimal", "Minimal grouping"),
         ],
         string="Procured purchase grouping",
         default="standard",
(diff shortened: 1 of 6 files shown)


Rad0van:
@pedrobaeza would you care to have a look? Thanx.

Rad0van:
Pinging to find some reviewers. @pedrobaeza @MiquelRForgeFlow @Tardo @sergio-teruel would you care to have a look? Thanx.

MiquelRForgeFlow:
It failed in ci:
Selection_091

Rad0van:
> It failed in ci: Selection_091
How did I miss that? Will check.

Rad0van:
@MiquelRForgeFlow intersetingly it:

1. fails in a different module (purchase_line_procurement_group)
2. fails only in odoo/odoo but not in oca/ocb run

Will have to investigate this further. Wonder what's going on here.

PR #1212, opened 2021-06-22
[14.0][IMP] purchase_work_acceptance_invoice_plan, miscellaneous improvements.

kittiu:
1.  Because WA is now based on invoice plan, the option "Create Bill by Plan" will be removed.
  User will be left with only option "Create Bill" (and to choose WA), the chosen WA is now used to control the quantity in invoice instead of the invoice plan.

2. Better control when create WA based on installment.

Depends On,

- [x] https://github.com/OCA/purchase-workflow/pull/1243


Saran440:
@kittiu https://github.com/OCA/purchase-workflow/pull/1212/commits/ba866e2b31e236dfc7bf971a7480c770b761dbf6

This PR hide button `Create Invoice Plan`.
when we create invoices with plan, invoice plan table should be related from invoice

Selection_004

Is it make sense or any ideas?

Saran440 (approved):

ps-tubtim (approved):
LGTM.

PR #1046, opened 2020-12-17
[MIG] 14.0 purchase_quick

PierrickBrun:
depends on https://github.com/OCA/product-attribute/pull/755

PR #1213, opened 2021-06-22
[14.0][MIG+REF] purchase_quick: migration to 14.0

kevinkhao:
supersedes https://github.com/OCA/purchase-workflow/pull/1046 due to messy history


sebastienbeau:
@kevinkhao test are broken can you fix them?

kittiu:
Functional tested, LGTM

kittiu (approved):

kevinkhao:
thanks @kittiu if you found this interesting maybe you will find this one interesting too https://github.com/OCA/multi-company/pull/296

sbidoul:
The test exclusion does not sound right here. What you are doing here is 2 test groups (x2 for Odoo and OCB):

1. all modules except purchase_quick
2. all modules except purchase_order_type

And purchase_quick and purchase_order_type are never tested.

That is probably not what you want.

If you want to have 3 tests groups (x2 for Odoo and OCB):

1. purchase_quick only
2. purchase_order_type
3. all modules *except* purchase_quick and purchase_order_type

You should run `copier update` and answer to the question `rebel_module_groups`: `["purchase_quick", "purchase_order_type"]`.
This will generate the correct configuration for Travis, and also for GitHub actions in the future.

kevinkhao:
> The test exclusion does not sound right here. What you are doing here is 2 test groups (x2 for Odoo and OCB):
> 
> 1. all modules except purchase_quick
> 2. all modules except purchase_order_type
> 
> And purchase_quick and purchase_order_type are never tested.
> 
> That is probably not what you want.
> 
> If you want to have 3 tests groups (x2 for Odoo and OCB):
> 
> 1. purchase_quick only
> 2. purchase_order_type
> 3. all modules _except_ purchase_quick and purchase_order_type
> 
> You should run `copier update` and answer to the question `rebel_module_groups`: `["purchase_quick", "purchase_order_type"]`.
> This will generate the correct configuration for Travis, and also for GitHub actions in the future.

Hi Stéphane, I will implement the changes as you listed, but this leads to another question, how about module inter-compatibility ? If I understand correctly in this case purchase_quick and purchase_order_type are tested in isolation, so we don't test those against all other modules ?

sbidoul:
Actually I don't know why you want to exclude this two modules (I've not looked closely).

In any case we never really test that modules in a repo work together because what we do is 1/ installing dependencies 2/ running with --test-enable for all modules in the repo which test one at a time, with previously installed modules.
So for modules that don't depend on each other in a repo, it depends on the installation order which is not defined.

kevinkhao:
> Actually I don't know why you want to exclude this two modules (I've not looked closely).

purchase_quick uses tests.Form, purchase_order_type adds a new field that is not required Python-side, but required in the view definition. It has no default value, so the test breaks.

> In any case we never really test that modules in a repo work together because what we do is 1/ installing dependencies 2/ running with --test-enable for all modules in the repo which test one at a time, with previously installed modules.
> So for modules that don't depend on each other in a repo, it depends on the installation order which is not defined.

OK, I understand now.

.travis.yml
```diff
@@ -18,12 +18,27 @@ stages:
...
+    - stage: test
+      env:
+        - TESTS=1 ODOO_REPO="odoo/odoo" MAKEPOT=1 INCLUDE="purchase_quick"
+    - stage: test
+      env:
+        - TESTS=1 ODOO_REPO="OCA/OCB" INCLUDE="purchase_quick"
+    # Test all other addons together
+    - stage: test
+      env:
+        - TESTS=1 ODOO_REPO="odoo/odoo" EXCLUDE="purchase_order_type,purchase_quick"
```
sbidoul: Hm, it looks like there is a bug in the template: there should be `MAKEPOT=1` here. Fixed in https://github.com/OCA/oca-addons-repo-template/pull/65

sbidoul: I made a template release. You can run `copier -f update` to get it.

sebastienbeau:
@bealdav @PierrickBrun @bguillot  can you review ?

PierrickBrun (approved):
code + functional review.

Not sure about the `.travis.yml` changes. But I think this is how to do it

Also, please squash commits before merging

PR #1121, opened 2021-02-23
 [13.0] [MIG] purchase_exception

albariera:
Supersedes #1007 
@HviorForgeFlow @Saran440 @kittiu @etobella

PR #1215, opened 2021-06-25
[14.0][MIG] purchase_exception

kittiu:
Standard migration

cc @newtratip @albariera


kittiu:
Exception is about `next_state` field.
```
2021-06-25 03:46:35,713 1 ERROR devel odoo.addons.purchase_exception.tests.test_purchase_exception: ERROR: TestPurchaseException.test_purchase_order_exception
  File "/opt/odoo/auto/addons/purchase_exception/tests/test_purchase_exception.py", line 57, in test_purchase_order_exception
    self.exception_noemail.next_state = False
```
I only can find the field `next_state` exists back in version 11.0.
https://github.com/OCA/server-tools/blob/11.0/base_exception/models/base_exception.py#L49

And I found the removal since the refactor here,
https://github.com/OCA/server-tools/commit/254391987cea07e93c668399ee1c6d51d8491fb2#diff-62560178a1a102d754def34699dc218643e202aa2eec43470f753f38c945705bL56

** I wonder why there is no test error on [13.0][MIG] which arrive much later?
https://github.com/OCA/purchase-workflow/pull/1121

Do you know @albariera @florian-dacosta 

Note: just remove the problem next_state line and test will pass.

kittiu:
I think the Purchase Exception Rule shouldn't be under specific Exception menu, because if we have other purchase module (like purchase_request_exception), we don't want to inherit purchase_exception. I think, we should move out.

WDYT?

image

albariera:
Reviewing the possible source of the problem, we have seen that the problem comes from base_exception. It seems that a refactory removing the field next_state was done at 12.0 and 14.0, but not at 13.0. 
In view of this situation, what do you want to do? @kittiu @florian-dacosta  

Regarding the menu change, I agree with you. :+1:

kittiu:
Fixed menu, and also enhance the message, same as what been done in sale_exception.

image

Ready for review

sebastienbeau (approved):
LGTM (code review)

florian-dacosta (approved):
> Reviewing the possible source of the problem, we have seen that the problem comes from base_exception. It seems that a refactory removing the field next_state was done at 12.0 and 14.0, but not at 13.0.
> In view of this situation, what do you want to do? @kittiu @florian-dacosta
> 
> Regarding the menu change, I agree with you. +1

Hi, Removing the next state from here is the good way to go I think.
Indeed, it has been removed form base_exception and so it should be removed from here. If it was still there, it was a mistake.
Like stated in https://github.com/OCA/server-tools/commit/254391987cea07e93c668399ee1c6d51d8491fb2#diff-62560178a1a102d754def34699dc218643e202aa2eec43470f753f38c945705bL56
It was used only in `purchase_order_approval_block` module and should be part of separated module like "base_exception_next_state" or something, if still needed.
But it seems `purchase_order_approval_block` also dropped the feature.

Anyway, I saw you've cleaned this module from `next_state` already, so LGTM for me

kittiu:
Thanks @sebastienbeau @florian-dacosta

PR #1222, opened 2021-06-30
[FIX] purchase_discount: supplierinfo with no template

chienandalu:
Actually, product_tmpl_id isn't a required field, so in theory a dummy
record could be created prior to assign it to a product template.

cc @Tecnativa TT23732

please review @victoralmau @carlosdauden

### DIFF
--- a/purchase_discount/models/product_supplierinfo.py
+++ b/purchase_discount/models/product_supplierinfo.py
@@ -32,7 +32,7 @@ class ProductSupplierInfo(models.Model):
         """ Insert discount (or others) from context from purchase.order's
         _add_supplier_to_product method """
         for vals in vals_list:
-            product_tmpl_id = vals['product_tmpl_id']
+            product_tmpl_id = vals.get('product_tmpl_id')
             po_line_map = self.env.context.get('po_line_map', {})
             if product_tmpl_id in po_line_map:
                 po_line = po_line_map[product_tmpl_id]


pedrobaeza:
The same as #1161, with the same doubts

chienandalu:
Well, it clashes with this type of declaration https://github.com/OCA/stock-logistics-warehouse/blob/12.0/stock_orderpoint_manual_procurement/tests/test_stock_orderpoint_manual_procurement.py#L49-L51

Which is something absurd, indeed, but it's permited by the core, thus my choice to fix it in this module

pedrobaeza:
For me that denotes an incorrect usage of the model, so the fix should be there in `stock_orderpoint_manual_procurement` instead.

chienandalu:
I won't say the contrary, but it's a problem within Odoo core code itself. Some few examples (there are more):

https://github.com/odoo/odoo/blob/14.0/addons/purchase_stock/tests/test_create_picking.py#L114-L117
https://github.com/odoo/odoo/blob/14.0/addons/purchase_stock/tests/test_onchange_product.py#L40-L45
https://github.com/odoo/odoo/blob/14.0/addons/stock_dropshipping/tests/test_dropship.py#L18-L21

Even the view doesn't requiere such field (I don't know why functionally this can be useful):
image

And the main problem here is that a test or worst a bit of functional code could pass under the radar and bite your CI in the worst unexpected moment.

This change doesn't let you do anything that the core already permits.

carlosdauden (approved):

pedrobaeza (approved):
OK, thanks for extra explanations. I hope Odoo normalizes soon this.

PR #1233, opened 2021-07-21
[13.0][FIX] purchase propagate qty handle split moves

mmequignon:
based on https://github.com/OCA/purchase-workflow/pull/1172, to be merged first.

In some cases (to be defined), odoo creates multiple moves for a single purchase order line.
Since the purchase_propagate_qty propagates the new purchased qty to all moves, we get two moves with a same quantity, which is the same as the one on the purchase line.

e.g 
 - user adds 10 to a purchase line with qty 100 -> new qty 110
 - for some reason, odoo creates a new move with qty 10 -> 2 moves with qties 100 and 10
 - purchase_propagate_qty does what it's meant to, propagates 110 onto the moves -> two moves with 110

This PR fixes this behaviour

### DIFF
--- a/purchase_propagate_qty/models/purchase.py
+++ b/purchase_propagate_qty/models/purchase.py
@@ -3,8 +3,8 @@
 # Copyright 2021 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from odoo import models
-from odoo.tools import float_is_zero
+from odoo import _, exceptions, models
+from odoo.tools import float_compare, float_is_zero
 
 
 class PurchaseOrderLine(models.Model):
@@ -20,11 +20,37 @@ class PurchaseOrderLine(models.Model):
         for line in self:
             if line.state != "purchase":
                 continue
-            for move in line.move_dest_ids | line.move_ids:
+            rounding = line.product_uom.rounding
+            for move in line.move_dest_ids:
                 if move.state in ("cancel", "done"):
                     continue
                 if move.product_id != line.product_id:
                     continue
                 move.product_uom_qty = line.product_uom_qty
-                if float_is_zero(move.product_uom_qty, line.product_uom.rounding):
+                if float_is_zero(move.product_uom_qty, precision_rounding=rounding):
                     move._action_cancel()
+            moves = line.move_ids
+            previous_qty = sum(moves.mapped("product_uom_qty"))
+            new_qty = line.product_uom_qty
+            # Do nothing is qty has been increased, since odoo handles this case
+            if new_qty >= previous_qty:
+                continue
+            # If qty has been decreased, cancel full moves if possible
+            moves = moves.filtered_domain(
+                [
+                    ("state", "not in", ("cancel", "done")),
+                    ("product_id", "=", line.product_id.id),
+                ]
+            ).sorted(lambda m: (m.product_uom_qty, m.quantity_done))
+            qty_to_remove = previous_qty - new_qty
+            removable_qty = moves._get_removable_qty()
+            if (
+                float_compare(qty_to_remove, removable_qty, precision_rounding=rounding)
+                == 1
+            ):
+                exception_text = _(
+                    "You cannot remove more that what remains to be done.\n"
+                    "Max removable quantity {}."
+                ).format(removable_qty)
+                raise exceptions.UserError(exception_text)
+            moves._deduce_qty(qty_to_remove, rounding)
--- a/purchase_propagate_qty/models/purchase.py
+++ b/purchase_propagate_qty/models/purchase.py
@@ -29,7 +29,7 @@ class PurchaseOrderLine(models.Model):
                 move.product_uom_qty = line.product_uom_qty
                 if float_is_zero(move.product_uom_qty, precision_rounding=rounding):
                     move._action_cancel()
-            moves = line.move_ids
+            moves = line.move_ids.filtered(lambda r: r.state not in ("cancel", "done"))
             previous_qty = sum(moves.mapped("product_uom_qty"))
             new_qty = line.product_uom_qty
             # Do nothing is qty has been increased, since odoo handles this case
@@ -49,7 +49,7 @@ class PurchaseOrderLine(models.Model):
                 == 1
             ):
                 exception_text = _(
-                    "You cannot remove more that what remains to be done.\n"
+                    "You cannot remove more that what remains to be done. "
                     "Max removable quantity {}."
                 ).format(removable_qty)
                 raise exceptions.UserError(exception_text)
--- a/purchase_propagate_qty/models/stock_move.py
+++ b/purchase_propagate_qty/models/stock_move.py
@@ -1,0 +1,38 @@
+# Copyright 2021 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from odoo import models
+from odoo.tools import float_compare, float_is_zero
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    def _deduce_qty(self, qty_to_remove, rounding):
+        """Deduce the provided qty with respect to done qties."""
+        self = self.with_context(do_not_unreserve=True)
+        for move in self:
+            if float_is_zero(qty_to_remove, precision_rounding=rounding):
+                break
+            # we cannot deduce more than the "not done" qty
+            removable_qty = move._get_removable_qty()
+            # If removable_qty <= qty_to_remove, deduce removable_qty
+            if (
+                float_compare(removable_qty, qty_to_remove, precision_rounding=rounding)
+                < 1
+            ):
+                move.product_uom_qty -= removable_qty
+                qty_to_remove -= removable_qty
+            # Else, deduce qty_to_remove from it
+            else:
+                move.product_uom_qty -= qty_to_remove
+                qty_to_remove = 0
+            # if new move product_uom_qty is 0, cancel it
+            if float_is_zero(move.product_uom_qty, precision_rounding=rounding):
+                move._action_cancel()
+
+    def _get_removable_qty(self):
+        assert (
+            len(set(self.mapped("product_uom"))) == 1
+        ), "moves doesn't share the same UoM"
+        return sum([move.product_uom_qty - move.quantity_done for move in self])
--- a/purchase_propagate_qty/models/stock_move.py
+++ b/purchase_propagate_qty/models/stock_move.py
@@ -32,7 +32,14 @@ class StockMove(models.Model):
                 move._action_cancel()
 
     def _get_removable_qty(self):
+        moves = self.filtered(
+            lambda move: move.location_dest_id.usage != "supplier"
+            and (
+                not move.origin_returned_move_id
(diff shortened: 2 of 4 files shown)


sebalix:
See my comment on the migration PR: https://github.com/OCA/purchase-workflow/pull/1172#discussion_r673922419

mmequignon:
The reason why moves are splitted is probably because of an issue in odoo. Rounding is wrong when computing the unit price of the new move and therefore can't be merge with the one that already exists.
An issue has been opened here https://github.com/odoo/odoo/issues/74106

mmequignon:
more informations about duplicated moves :

- user creates the PO with a single line qty 1000, price 10 and confirms it.
- odoo creates a new supplierinfo on the product with price 10.
- another user updates the supplierinfo and sets the price to 9
- user adds qty 100 to the POL, which triggers a recompute on unit price, which is updated to 9.
- odoo creates a new move with unit price = 9 and qty = 100, and isn't able to merge the two moves since their price is different.

purchase_propagate_qty/models/purchase.py
```diff
@@ -20,11 +19,30 @@ def _propagage_qty_to_moves(self):
...
                 continue
-            for move in line.move_dest_ids | line.move_ids:
-                if move.state in ("cancel", "done"):
-                    continue
-                if move.product_id != line.product_id:
-                    continue
-                move.product_uom_qty = line.product_uom_qty
-                if float_is_zero(move.product_uom_qty, line.product_uom.rounding):
-                    move._action_cancel()
+            moves = line.move_dest_ids | line.move_ids
```
sebalix: You can not sum qties of `move_dest_ids` and `move_ids`, or you will double the expected qty for sure if there is a downstream move.
I would process them separately, the original implementation was fine I think for downstream moves:
```suggestion
            for move in line.move_dest_ids:
                if move.state in ("cancel", "done"):
                    continue
                if move.product_id != line.product_id:
                    continue
                move.product_uom_qty = line.product_uom_qty
                if float_is_zero(move.product_uom_qty, line.product_uom.rounding):
                    move._action_cancel()
            moves = line.move_ids
```

purchase_propagate_qty/models/purchase.py
```diff
@@ -20,11 +19,30 @@ def _propagage_qty_to_moves(self):
...
+                self.env["stock.move"]
+                .search(
+                    [
+                        ("id", "in", moves.ids),
+                        ("state", "not in", ("cancel", "done")),
+                        ("product_id", "=", line.product_id.id),
+                    ]
+                )
+                .sorted(lambda m: (m.product_uom_qty, m.quantity_done))
+            )
```
sebalix:
```suggestion
            moves = moves.filtered_domain(
                [("state", "not in", ("cancel", "done")), ("product_id", "=", line.product_id.id)]
            ).sorted(lambda m: (m.product_uom_qty, m.quantity_done))
```

purchase_propagate_qty/tests/test_purchase_delivery.py
```diff
@@ -94,3 +94,57 @@ def test_purchase_line_qty_decrease_to_zero(self):
...
+
+    def test_purchase_line_splitted_move(self):
+        # Add qty, create a new move with the diff
+        line1 = self.po.order_line[0]
+        move1 = self.env["stock.move"].search([("purchase_line_id", "=", line1.id)])
+        self.assertEqual(len(move1), 1)
+        line1.write({"product_qty": 64})
+        self.assertEqual(move1.product_uom_qty, 64)
+        # split the move
+        move1._split(22)
```
sebalix: Why not split the move the way Odoo does it, by changing the qty + the price unit of the PO line? To stick to the standard workflow.

purchase_propagate_qty/tests/test_purchase_delivery.py
```diff
@@ -94,3 +94,57 @@ def test_purchase_line_qty_decrease_to_zero(self):
...
+
+    def test_purchase_line_splitted_move_w_reserved_qty(self):
+        # Add qty, create a new move with the diff
+        line1 = self.po.order_line[0]
+        move1 = self.env["stock.move"].search([("purchase_line_id", "=", line1.id)])
+        self.assertEqual(len(move1), 1)
+        line1.with_context(unittests=True).write({"product_qty": 64})
+        self.assertEqual(move1.product_uom_qty, 64)
+        # split the move
+        move1._split(22)
```
sebalix: Same remark

mmequignon:
@sebalix I adressed your comments, could you please review it again? Thanks!

purchase_propagate_qty/models/purchase.py
```diff
@@ -20,11 +20,33 @@ def _propagage_qty_to_moves(self):
...
                     continue
                 if move.product_id != line.product_id:
                     continue
                 move.product_uom_qty = line.product_uom_qty
                 if float_is_zero(move.product_uom_qty, line.product_uom.rounding):
                     move._action_cancel()
+            moves = line.move_ids
+            previous_qty = sum(moves.mapped("product_uom_qty"))
+            new_qty = line.product_uom_qty
+            # Do nothing is qty has been increased, since odoo handles this case
```
simahawk, line 35:
```suggestion
            # Do nothing if qty has been increased, since odoo handles this case
```
?

purchase_propagate_qty/models/stock_move.py
```diff
@@ -0,0 +1,32 @@
...
+            # Else, deduce qty_to_remove from it
+            else:
+                move.product_uom_qty -= qty_to_remove
+                qty_to_remove = 0
+            # if new move product_uom_qty is 0, cancel it
+            if float_is_zero(move.product_uom_qty, rounding):
+                move._action_cancel()
+
+    def _get_removable_qty(self):
+        return sum([move.product_uom_qty - move.quantity_done for move in self])
```
simahawk: this should always run a single record no?

purchase_propagate_qty/tests/test_purchase_delivery.py
```diff
@@ -94,3 +94,56 @@ def test_purchase_line_qty_decrease_to_zero(self):
         line1.write({"product_qty": 0})
         self.assertEqual(move1.product_uom_qty, 0)
         self.assertEqual(move1.state, "cancel")
+
+    def test_purchase_line_splitted_move(self):
```
simahawk:
```suggestion
    def test_purchase_line_split_move(self):
```

purchase_propagate_qty/tests/test_purchase_delivery.py
```diff
@@ -94,3 +94,56 @@ def test_purchase_line_qty_decrease_to_zero(self):
...
+        self.assertEqual(moves.mapped("product_uom_qty"), [42.0, 11.0])
+        # Again, decrease 11
+        line1.write({"product_qty": 42})
+        self.assertEqual(moves.mapped("product_uom_qty"), [42.0, 0.0])
+        active_move = moves.filtered(lambda m: m.state != "cancel")
+        inactive_move = moves.filtered(lambda m: m.state == "cancel")
+        self.assertEqual(active_move.product_uom_qty, 42)
+        self.assertEqual(inactive_move.product_uom_qty, 0)
+
+    def test_purchase_line_splitted_move_w_reserved_qty(self):
```
simahawk:
```suggestion
    def test_purchase_line_split_move_w_reserved_qty(self):
```

purchase_propagate_qty/tests/test_purchase_delivery.py
```diff
@@ -94,3 +94,56 @@ def test_purchase_line_qty_decrease_to_zero(self):
...
+        self.assertEqual(move2._get_removable_qty(), 10)
+        line1.write({"product_qty": 60})
+        self.assertEqual(moves.mapped("product_uom_qty"), [42.0, 18.0])
+        # We cannot remove more than 10 on move2, deduce the max from it,
+        # then deduce from move1
+        line1.write({"product_qty": 50})
+        self.assertEqual(moves.mapped("product_uom_qty"), [38.0, 12.0])
+        # We should not be able to set a qty < to 12, since 12 products are reserved
+        with self.assertRaises(UserError):
+            line1.write({"product_qty": 11})
```
simahawk, line 148: check the message w/ `assertRaiseRegex`

simahawk (changes requested):
LG overall

purchase_propagate_qty/models/stock_move.py
sebalix: I would add an `assert` to check that the same UoM is shared between all moves

purchase_propagate_qty/models/purchase.py
```diff
@@ -20,11 +20,33 @@ def _propagage_qty_to_moves(self):
...
+            # If qty has been decreased, cancel full moves if possible
+            moves = moves.filtered_domain(
+                [
+                    ("state", "not in", ("cancel", "done")),
+                    ("product_id", "=", line.product_id.id),
+                ]
+            ).sorted(lambda m: (m.product_uom_qty, m.quantity_done))
+            qty_to_remove = previous_qty - new_qty
+            removable_qty = moves._get_removable_qty()
+            if qty_to_remove > removable_qty:
```
sebalix: Use `float_compare` for such comparison

purchase_propagate_qty/models/stock_move.py
```diff
@@ -0,0 +1,32 @@
...
+    def _deduce_qty(self, qty_to_remove, rounding):
+        """Deduce the provided qty with respect to done qties."""
+        self = self.with_context(do_not_unreserve=True)
+        for move in self:
+            if qty_to_remove == 0:
+                break
+            # we cannot deduce more than the "not done" qty
+            removable_qty = move._get_removable_qty()
+            # If removable_qty <= qty_to_remove, deduce removable_qty
+            if removable_qty <= qty_to_remove:
```
sebalix: `float_compare` is welcome too

purchase_propagate_qty/models/stock_move.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    def _deduce_qty(self, qty_to_remove, rounding):
+        """Deduce the provided qty with respect to done qties."""
+        self = self.with_context(do_not_unreserve=True)
+        for move in self:
+            if qty_to_remove == 0:
```
sebalix: Use `float_is_zero`

purchase_propagate_qty/models/stock_move.py
sebalix: No it is used on a recordset of `moves` in `_propagage_qty_to_moves` method, as a PO line could have several moves linked (if the unit price changed over time for instance when you update the PO line qty, Odoo create another move).

mmequignon: No, when called from purchase_order_line._propagage_qty_to_moves, it is called on a recordset.

purchase_propagate_qty/tests/test_purchase_delivery.py
TDu: It has been done

simahawk (approved):

sebalix (approved):

PR #1234, opened 2021-07-27
[13.0][ADD] purchase_stock_picking_show_currency_rate: Show currency rate in purchase stock pickings

victoralmau:
Show currency rate in purchase stock pickings.

Locked by:
- [x] https://github.com/OCA/purchase-workflow/pull/1238

@Tecnativa TT30786


purchase_stock_picking_show_currency_rate/README.rst
```diff
@@ -0,0 +1,108 @@
...
+.. |badge4| image:: https://img.shields.io/badge/weblate-Translate%20me-F47D42.png
+    :target: https://translation.odoo-community.org/projects/purchase-workflow-13-0/purchase-workflow-13-0-purchase_stock_picking_show_currency_rate
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/142/13.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+This module shows the currency rate applied in purchase stock picking, so you can visually verify what is going to be applied for valuation layer, or which one was applied once converted to company currency.
```
joao-p-marques:
```suggestion
This module shows the currency rate applied in a purchase stock picking, so that you can visually verify what is going to be applied to the valuation layer, or which one was applied once converted to company currency.
```

purchase_stock_picking_show_currency_rate/README.rst
```diff
@@ -0,0 +1,108 @@
...
+
+#. Go to Invoicing > Configuration > Settings > Currencies > Multi-Currencies
+#. Go to any request for quotation
+#. Change the purchase currency.
+#. Add any line with storable product.
+#. Confirm Order.
+#. Go to Receipt.
+#. The proper currency rate, based on today date and the selected currency, will be shown.
+
+It's really necessary to exists some rates about not company currency distinct to 1.0, for example from EUR:
```
joao-p-marques:
```suggestion
Rates for non-company currencies must be available and be distinct to 1.0. For example, for EUR:
```

purchase_stock_picking_show_currency_rate/README.rst
```diff
@@ -0,0 +1,108 @@
...
+
+
+To use this module, you need to:
+
+#. Go to Purchase > Orders > Request for quotation
+#. Create Purchase and set Currency distinct to company currency (EUR for example)
+#. Add any line with storable product.
+#. Confirm Order.
+#. Go to Receipt.
+#. Rate account show according to currency defined.
```
joao-p-marques:
```suggestion
#. The account rate should be shown according to the currency defined previously.
```

purchase_stock_picking_show_currency_rate/models/stock_picking.py
```diff
@@ -0,0 +1,38 @@
...
+    )
+    currency_rate_amount = fields.Float(
+        string="Rate amount", compute="_compute_currency_rate_amount", digits=0,
+    )
+
+    @api.depends("purchase_currency_id", "company_currency_id")
+    def _compute_show_currency_rate_amount(self):
+        for item in self:
+            item.show_currency_rate_amount = False
+            if item.purchase_currency_id != item.company_id.currency_id:
```
joao-p-marques: If you are creating the related field, you can use it instead:
```suggestion
            if item.purchase_currency_id != item.company_currency_id:
```

OTOH, are the related fields really necessary? Or do they provide any performance improvement? If not, maybe you could avoid them and consult the fields directly...

purchase_stock_picking_show_currency_rate/models/stock_picking.py
```diff
@@ -0,0 +1,38 @@
...
+                item.show_currency_rate_amount = True
```
joao-p-marques: You can also avoid this extra assignment:
```suggestion
            if item.purchase_currency_id != item.company_id.currency_id:
                item.show_currency_rate_amount = True
            else:
                item.show_currency_rate_amount = False
```

purchase_stock_picking_show_currency_rate/models/stock_picking.py
```diff
@@ -0,0 +1,38 @@
...
+
+    @api.depends("purchase_id", "show_currency_rate_amount", "company_id")
```
joao-p-marques: Shouldn't these dependencies be like this?
```suggestion
    @api.depends("purchase_currency_id", "show_currency_rate_amount", "company_id")
```

purchase_stock_picking_show_currency_rate/models/stock_picking.py
```diff
@@ -0,0 +1,38 @@
...
+    def _compute_currency_rate_amount(self):
+        for item in self:
+            item.currency_rate_amount = 1
+            if item.show_currency_rate_amount:
+                rates = item.purchase_currency_id._get_rates(
+                    item.company_id, fields.Date.context_today(self)
+                )
+                item.currency_rate_amount = rates.get(item.purchase_currency_id.id)
```
joao-p-marques: Again, you can avoid the extra assignment by doing:
```suggestion
            if item.show_currency_rate_amount:
                rates = item.purchase_currency_id._get_rates(
                    item.company_id, fields.Date.context_today(self)
                )
                item.currency_rate_amount = rates.get(item.purchase_currency_id.id)
            else:
                item..currency_rate_amount = 1
```

purchase_stock_picking_show_currency_rate/tests/test_purchase_stock_picking.py
```diff
@@ -0,0 +1,63 @@
...
+        cls.product = cls.env["product.product"].create(
+            {
+                "name": "Product Test",
+                "type": "product",
+                "list_price": 10,
+                "taxes_id": [(6, 0, [cls.account_tax.id])],
+            }
+        )
+        # Create custom rates to USD + EUR
+        today = datetime.today().strftime(DF)
```
joao-p-marques: You can use the default Odoo tools here: https://github.com/odoo/odoo/blob/13.0/odoo/fields.py#L1694
```suggestion
        today = fields.Date.today()
```

joao-p-marques (changes requested):
Code review

victoralmau:
Thanks @joao-p-marques , changes done.

purchase_order_type/models/purchase_order.py
```diff
@@ -11,17 +11,14 @@ class PurchaseOrder(models.Model):
     _inherit = "purchase.order"
 
     def _default_order_type(self):
-        return self.env["purchase.order.type"].search(
-            [("company_id", "in", [False, self.env.company.id])], limit=1,
-        )
+        return self.env["purchase.order.type"].search([], limit=1)
```
pedrobaeza: You have to explain why this, and it seems incorrect at first sight.

purchase_order_type/models/res_partner.py
```diff
@@ -7,6 +7,11 @@
...
     _inherit = "res.partner"
 
+    def _default_purchase_type(self):
+        return self.env["purchase.order.type"].search([], limit=1)
+
     purchase_type = fields.Many2one(
-        comodel_name="purchase.order.type", string="Purchase Order Type"
+        comodel_name="purchase.order.type",
+        string="Purchase Order Type",
+        default=lambda self: self._default_purchase_type(),
```
pedrobaeza: This shouldn't have a default.

purchase_stock_picking_show_currency_rate/README.rst
```diff
@@ -0,0 +1,108 @@
...
+.. contents::
+   :local:
+
+Configuration
+=============
+
+Enable the option for multiple currencies in your instance:
+
+#. Go to Invoicing > Configuration > Settings > Currencies > Multi-Currencies
+#. Go to any request for quotation
```
pedrobaeza: From this line to the latest point up to next paragraph, they are not configuration steps, but usage ones.

purchase_stock_picking_show_currency_rate/models/stock_picking.py
```diff
@@ -0,0 +1,43 @@
...
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    purchase_currency_id = fields.Many2one(
+        string="Purchase currency", related="purchase_id.currency_id", readonly=True,
+    )
+    company_currency_id = fields.Many2one(
+        string="Company Currency", related="company_id.currency_id", readonly=True,
+    )
```
pedrobaeza: Why do you need this field?

purchase_stock_picking_show_currency_rate/models/stock_picking.py
```diff
@@ -0,0 +1,43 @@
...
+    @api.depends("purchase_currency_id", "company_currency_id")
+    def _compute_show_currency_rate_amount(self):
+        for item in self:
+            if (
+                item.purchase_currency_id
+                and item.purchase_currency_id != item.company_currency_id
+            ):
+                item.show_currency_rate_amount = True
+            else:
+                item.show_currency_rate_amount = False
```
pedrobaeza:
```suggestion
            item.show_currency_rate_amount = (
                item.purchase_id.currency_id
                and item.purchase_id.currency_id != item.company_id.currency_id
            )
```

purchase_stock_picking_show_currency_rate/models/stock_picking.py
```diff
@@ -0,0 +1,43 @@
...
+    @api.depends("purchase_currency_id", "show_currency_rate_amount", "company_id")
+    def _compute_currency_rate_amount(self):
+        for item in self:
+            if item.show_currency_rate_amount:
+                rates = item.purchase_currency_id._get_rates(
+                    item.company_id, fields.Date.context_today(self)
+                )
+                item.currency_rate_amount = rates.get(item.purchase_currency_id.id)
+            else:
+                item.currency_rate_amount = 1
```
pedrobaeza:
```suggestion
        self.currency_rate_amount = 1
        for item in self.filtered("show_currency_rate_amount"):
            rates = item.purchase_id.currency_id._get_rates(
                item.company_id, fields.Date.context_today(self)
            )
            item.currency_rate_amount = rates.get(item.purchase_currency_id.id)
```

purchase_stock_picking_show_currency_rate/views/stock_picking_view.xml
```diff
@@ -0,0 +1,28 @@
...
+<odoo>
+    <record id="stock_picking_form" model="ir.ui.view">
+        <field name="name">stock.picking.form</field>
+        <field name="model">stock.picking</field>
+        <field name="inherit_id" ref="stock.view_picking_form" />
+        <field name="arch" type="xml">
+            <field name="origin" position="after">
+                <field name="purchase_id" invisible="1" />
+                <field
+                    name="purchase_currency_id"
```
pedrobaeza: Put both in the same line with class `oe_inline` for better display. Check other examples in core.

purchase_stock_picking_show_currency_rate/views/stock_picking_view.xml
```diff
@@ -0,0 +1,28 @@
...
+                    groups="base.group_multi_currency"
+                    attrs="{'invisible':[('purchase_id', '=', False)]}"
```
pedrobaeza:
```suggestion
                    attrs="{'invisible':[('show_currency_rate_amount', '=', False)]}"
```

pedrobaeza (changes requested):

purchase_order_type/models/purchase_order.py
victoralmau: Purchase order type model have the same rule (https://github.com/OCA/purchase-workflow/blob/13.0/purchase_order_type/security/security.xml) and IMO is not necessary to apply these domain.

pedrobaeza: You don't know if the operation is done in sudo mode, so please keep it.

purchase_order_type/models/res_partner.py
victoralmau: I have made this change because using `Form()` in the test to create a purchase order, the test always fails because there is no `order_type` defined.

Any suggestions on how to remove this change so that the test does not fail?

pedrobaeza: I don't get the problem. Please always put the error.

victoralmau: https://app.travis-ci.com/github/OCA/purchase-workflow/jobs/527335286#L1826

victoralmau:
Now, Travis error of the tests is related to `purchase_order_type`.
What is the best option to prevent error in tests @pedrobaeza ?

purchase_stock_picking_show_currency_rate/models/stock_picking.py
```diff
@@ -0,0 +1,36 @@
...
+        for item in self:
+            item.show_currency_rate_amount = (
+                item.purchase_id.currency_id
+                and item.purchase_id.currency_id != item.company_id.currency_id
+            )
+
+    @api.depends("purchase_currency_id", "show_currency_rate_amount", "company_id")
+    def _compute_currency_rate_amount(self):
+        for item in self:
+            item.currency_rate_amount = 1
```
pedrobaeza: Can be assigned in one shot
```suggestion
        self.currency_rate_amount = 1
```

purchase_stock_picking_show_currency_rate/README.rst
```diff
@@ -0,0 +1,102 @@
...
+.. |badge4| image:: https://img.shields.io/badge/weblate-Translate%20me-F47D42.png
+    :target: https://translation.odoo-community.org/projects/purchase-workflow-13-0/purchase-workflow-13-0-purchase_stock_picking_show_currency_rate
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/142/13.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+This module shows the currency rate applied in a purchase stock picking, so that you can visually verify what is going to be applied to the valuation layer, or which one was applied once converted to company currency.
```
pedrobaeza, line 28: I don't see code for the rate once the picking is done and the valuation layer is created.

purchase_order_type/models/res_partner.py
pedrobaeza: Sorry, I'm seeing now the problem and indeed although the solution is a bit dirty, we need to add it for compatibility. Please re-introduce it. I don't have any other alternative right now.

purchase_order_type/models/purchase_order.py
pedrobaeza: This comment still applies

victoralmau: Sorry, it is not necessary to apply a domain to filter by company because this domain is already defined as a rule of the model.

pedrobaeza: Such domain doesn't apply here. It's just for UI.

purchase_order_type/models/purchase_order.py
```diff
@@ -21,7 +21,6 @@ def _default_order_type(self):
         states=Purchase.READONLY_STATES,
         string="Type",
         ondelete="restrict",
-        domain="[('company_id', 'in', [False, company_id])]",
```
joao-p-marques: Why do you remove this? Please explain, for example in the commit message body... Also, please provide a clearer commit message (title)

pedrobaeza: This shouldn't be removed. No matter record rules, you can have more than one company displayed on your user, and thus, this will show incorrect types.

victoralmau:
Travis is green now.

purchase_order_type/models/purchase_order.py
pedrobaeza: I still see this missing

victoralmau: Sorry, it is already restored.

purchase_order_type/models/res_partner.py
pedrobaeza: Looking again the error, the problem was on the purchase order, not on the partner, so I see again that this is not needed. In fact, I mixed both when I checked. There should be something preventing the default in purchase order (probably the computed writable, switching to `False` when changing partner, but should be fixed there.

joao-p-marques (approved):
Apart from @pedrobaeza's comments, looks good :+1: 

Code and functional review

purchase_order_type/models/res_partner.py
victoralmau: Ok, thanks, changes done (removed `_onchange_company()` to prevent set to `False`).

pedrobaeza (approved):

pedrobaeza:
Please rebase to see the problem.

pedrobaeza:
As suspected, main branch is now red. Can you take a look to see if it's an easy fix? If not, please ping the author of the module.

victoralmau:
Although I thought that the commit I added would solve the problem (locally it solves it) it seems that it is not enough.
Ping @grindtildeath please, can you check the error in the `purchase_line_procurement_group` addon tests?

victoralmau:
I have made some changes to `purchase_line_procurement_group` tests and Travis is now green.
It is ready to merge, right?

PR #1235, opened 2021-07-30
[13.0][IMP] purchase_last_price_info: Add last_purchase_currency and last_purchase_price_currency in products

victoralmau:
Add `last_purchase_currency` and `last_purchase_price_currency` in products.

**Refactor code**: Now used `last_purchase_line_id` field and is auto-set wit all info.
When adding new purchase order lines `last_purchase_line_ids` field will be used to re-compute `last_purchase_line_id` field and change all information.

Locked by:
- [x] https://github.com/OCA/purchase-workflow/pull/1238

@Tecnativa TT30742


purchase_last_price_info/models/product.py
```diff
@@ -71,12 +96,56 @@ class ProductTemplate(models.Model):
     last_supplier_id = fields.Many2one(
         comodel_name="res.partner", string="Last Supplier"
     )
+    last_purchase_company_id = fields.Many2one(
```
pedrobaeza: This field is not needed.

pedrobaeza (changes requested):
The general approach of this module is totally incorrect (not your fault), but it's not taken into account a multi-company environment. Imagine this case:

- You have a product valid for 2 companies.
- You purchase in company A from vendor X something.
- You go to company B, and show such statistics in the product:
  * It doesn't serve for you (unless you are viewing both companies at the same time).
  * It will probably give an access error due to the supplier belonging to the other company.

This must be totally refactored for having such fields computed on the fly, and depending on the allowed companies, so remove all the "update" stuff and switch all them to computed. Do it also on a smart-way for avoiding too much duplicated code between template and variant.

purchase_last_price_info/hooks.py
```diff
@@ -1,15 +0,0 @@
-# Copyright 2019 ForgeFlow S.L.
-#   (http://www.forgeflow.com)
-# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
-
-from odoo import SUPERUSER_ID, api
-
-
-def set_last_price_info(cr, registry):
```
joao-p-marques, line 8: Can you clarify why this is not needed anymore?

purchase_last_price_info/models/product_product.py
```diff
@@ -0,0 +1,84 @@
...
+        inverse_name="product_id",
+        domain=[("state", "in", ["purchase", "done"])],
+        string="Last Purchase Order Lines",
+    )
+    last_purchase_line_id = fields.Many2one(
+        comodel_name="purchase.order.line",
+        compute="_compute_last_purchase_line_id",
+        string="Last Purchase Line",
+        store=True,
+    )
```
joao-p-marques: Also, can you explain a bit better why these 2 fields are necessary and how they are set up? I don't see them in the view, so maybe a quick explanation in the readme would help. OTOH, why does `last_purchase_line_id` need to be stored? If you are just getting the last one, maybe a simple compute is enough, right? Or does it have a big influence in performance?

purchase_last_price_info/hooks.py
victoralmau: It's not necessary because now `last_purchase_line_id` field is auto-set wit all info.

joao-p-marques: Ok, then maybe put that information in the commit message body and in the PR description.

purchase_last_price_info/models/product_product.py
victoralmau: Now `last_purchase_line_ids` field is necessary to auto-set `last_purchase_line_id` field (when `last_purchase_line_ids` have changed according to new purchase lines), and these is the explanation to set as compute and `store=True`.

joao-p-marques: Ok, I didn't think about the `inverse_name="product_id"` and I didn't know where `last_purchase_line_ids` came from :sweat_smile: 
However, I still don't understand the need for the `store=True` in `last_purchase_line_id`. Can you explain a little better?

joao-p-marques (approved):
Code review :+1: 

Travis is red (the `swig` problem again). Try re-launching the build, as it should be fixed now.

victoralmau:
Thanks @joao-p-marques changes done.

victoralmau:
Rebase done, ready to merge now.

purchase_last_price_info/models/product_template.py
```diff
@@ -0,0 +1,54 @@
...
+        digits=0,
+    )
+
+    @api.depends("last_purchase_line_ids")
+    def _compute_last_purchase_line_id(self):
+        for item in self:
+            if item.last_purchase_line_ids:
+                item.last_purchase_line_id = item.last_purchase_line_ids[:1]
+            else:
+                item.last_purchase_line_id = item.last_purchase_line_id
```
pedrobaeza: No need to do all of this, and better to use this for not losing cache:
```
from odoo.fields import first
```

```suggestion
            item.last_purchase_line_id = first(item.last_purchase_line_ids)
```

purchase_last_price_info/models/product_product.py
```diff
@@ -0,0 +1,84 @@
...
+        comodel_name="purchase.order.line",
+        inverse_name="product_id",
+        domain=[("state", "in", ["purchase", "done"])],
+        string="Last Purchase Order Lines",
+    )
+    last_purchase_line_id = fields.Many2one(
+        comodel_name="purchase.order.line",
+        compute="_compute_last_purchase_line_id",
+        string="Last Purchase Line",
+        store=True,
```
pedrobaeza: Why is this stored? This is going to trigger a huge computation when installing the module.

purchase_last_price_info/models/product_product.py
```diff
@@ -0,0 +1,84 @@
...
+        digits=0,
+    )
+
+    @api.depends("last_purchase_line_ids")
+    def _compute_last_purchase_line_id(self):
+        for item in self:
+            if item.last_purchase_line_ids:
+                item.last_purchase_line_id = item.last_purchase_line_ids[:1]
+            else:
+                item.last_purchase_line_id = item.last_purchase_line_id
```
pedrobaeza:
```suggestion
            item.last_purchase_line_id = first(item.last_purchase_line_ids)
```

victoralmau:
Thanks @pedrobaeza, changes done.

pedrobaeza (approved):

PR #1237, opened 2021-08-02
[12.0][FIX] purchase_request: compute preferred supplier in multicompany

astirpe:
This PR fixes an access rights error that may occur in a multicompany environment.
For example, when user tries to update a vendor price, method `def _compute_supplier_id()` tries to recompute the 'Preferred supplier' on all purchase request lines. Some lines may belong to different companies and that causes the error.

### DIFF
--- a/purchase_request/models/purchase_request_line.py
+++ b/purchase_request/models/purchase_request_line.py
@@ -239,6 +239,7 @@ class PurchaseRequestLine(models.Model):
     product_id = fields.Many2one(
         'product.product', 'Product',
         domain=[('purchase_ok', '=', True)],
+        index=True,
         track_visibility='onchange')
 
     @api.onchange('product_id')
--- a/purchase_request/models/purchase_request_line.py
+++ b/purchase_request/models/purchase_request_line.py
@@ -74,6 +74,7 @@ class PurchaseRequestLine(models.Model):
     supplier_id = fields.Many2one('res.partner',
                                   string='Preferred supplier',
                                   compute="_compute_supplier_id",
+                                  compute_sudo=True,
                                   store=True)
     cancelled = fields.Boolean(
         string="Cancelled", readonly=True, default=False, copy=False)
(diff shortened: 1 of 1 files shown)


MPishpecki (approved):

kittiu:
Code reviewed, LGTM.

kittiu (approved):

kittiu:
This important fix should be foward port to 13, 14 also.

MPishpecki (approved):

StefanRijnhart (approved):
Wondering why the design of this feature is such that the supplier_id is recomputed on processed request lines, as it must be slowing down the updates on supplierinfo. This is aggravated by the fact that there is no index on purchase_request_line.product_id. Would you consider adding that one in this PR?

But in any case, thanks!

astirpe:
@StefanRijnhart we are browsing here: `rec.product_id.seller_ids`. So I don't think an index on `purchase_request_line.product_id` would be helpful in this particular case.

astirpe:
@StefanRijnhart or do you mean because of `@api.depends('product_id', 'product_id.seller_ids')` ?

astirpe:
I think that a proper fix would be to set supplier_id as a non-stored field instead.

StefanRijnhart:
Thanks for adding the index on product_id. I asked for this index, not because it is used in the compute method itself (which as you mention is not the case), but because it is used in the recompute trigger on product_id  (--edit--: ah you mention this in a follow up comment as well).

astirpe:
@StefanRijnhart would setting the supplier_id as a non-stored field work for you? That would avoid the overall slowing down by the compute method each time the `seller_ids` on products is updated.

StefanRijnhart:
@astirpe Thanks for considering that. However, I'm not sure whether it was stored for a reason (grouping, searching, performance of the request views). So having the index is an improvement already, and the further refactoring is out of scope as far as I'm concerned.

PR #1238, opened 2021-08-05
Init copier template

joao-p-marques:
Also, install the `swig` package, now needed when installing `reporting-engine`

@pedrobaeza @Yajo @victoralmau can you review please?


victoralmau (approved):

pedrobaeza:
Instead of putting `swig` everywhere, we should change OCA/reporting-engine this way:

- Remove `cryptography` and `endesive` from requirements.txt.
- Install them inside . travis.yml, indicating why it's done there.

joao-p-marques:
@pedrobaeza Ok, the problem in doing that is that the `requirements.txt` file is generated from the `external_dependencies` key in the modules' manifest, so we would have to remove it there as well, and loose that constraint...

yajo:
IMHO this reveals a weakness in OCA dependency system. Maybe using a tool such as https://pypi.org/project/bindep/ for declaring system dependencies would help propagating them automatically across linked repos.

yajo (approved):
In any case, this is OK to be merged.

joao-p-marques:
> IMHO this reveals a weakness in OCA dependency system. Maybe using a tool such as https://pypi.org/project/bindep/ for declaring system dependencies would help propagating them automatically across linked repos.

Yes, it would be a good idea, although it would involve a bigger refactor in the OCA's infrastructure, so maybe for now we should stick with this forced manual update...

pedrobaeza:
We need to remove such external dependency then, or don't generate such requirements.txt. We can't afford to change every OCA repo due to this.

joao-p-marques:
Just removed the `swig` dependency, as requested, but still this can be merged to add the copier template @pedrobaeza

PR #1243, opened 2021-08-13
[14.0][IMP] purchase_invoice_plan, misc improvements to be more accurate.

kittiu:
For invoice plan to be more flexible and accurate, following improvement are added,

1. Ensure that, only unused invoice plan are editable.
2. Validate that, total plan percent = 100
3. Add decimal prec "Purchase Invoice Plan Percent" to gain more accuracy of computed amount. (now, default perent prec = 6 digits, this is much need for big amount with lots of installments).


Saran440 (approved):
Functional Test :+1:

ps-tubtim (approved):
LGTM.

PR #1246, opened 2021-08-17
[13.0] Refactor purchase_split_by_date (support new PO line or increased qty on a confirmed PO)

jbaudoux:
Support new PO line or increased qty on a confirmed PO. Instead of changing _create_stock_moves that is called only when the PO is confirmed, modify _prepare_stock_move to properly ensure the move is directly inserted in the right picking.

Improve PO line date change to properly insert move in the right picking and ensure to have one picking per date.

Improve date comparison taking care of the timezone.

Be compliant with ~~propagate_date and (after discussing with @jgrandguillaume, we better do not care of that feature as it is currently in previous versions)~~ display_type from odoo standard. 

cc @sebalix @jgrandguillaume @gurneyalex @LoisRForgeFlow @JordiBForgeFlow

### DIFF
--- a/purchase_delivery_split_date/models/purchase.py
+++ b/purchase_delivery_split_date/models/purchase.py
@@ -1,8 +1,7 @@
 # Copyright 2014-2016 Numérigraphe SARL
 # Copyright 2017 ForgeFlow, S.L.
+# Copyright 2021 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
-
-from itertools import groupby
 
 from odoo import api, fields, models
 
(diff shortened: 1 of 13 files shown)


purchase_delivery_split_date/models/purchase.py
```diff
@@ -1,138 +1,100 @@
...
 # Copyright 2017 ForgeFlow, S.L.
+# Copyright 2021 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
-from itertools import groupby
+from odoo import api, fields, models, SUPERUSER_ID
 
-from odoo import api, fields, models
+
+def date(picking, datetime):
```
TDu: This function should have a more explicit name

purchase_delivery_split_date/models/purchase.py
```diff
@@ -1,138 +1,100 @@
...
-    def _first_picking_copy_vals(self, key, lines):
-        """The data to be copied to new pickings is updated with data from the
-        grouping key.  This method is designed for extensibility, so that
-        other modules can store more data based on new keys."""
-        vals = {"move_lines": []}
-        for key_element in key:
-            if "date_planned" in key_element.keys():
-                vals["scheduled_date"] = key_element["date_planned"]
-        return vals
+    def _is_valid_picking(self, picking):
```
TDu, line 12: Needs a `self.ensure_one()`

purchase_delivery_split_date/models/purchase.py
```diff
@@ -1,138 +1,100 @@
...
-from itertools import groupby
+from odoo import api, fields, models, SUPERUSER_ID
 
-from odoo import api, fields, models
+
+def date(picking, datetime):
+    # TODO: extract the tz field on the warehouse from the module
+    # sale_cutoff_time_delivery in OCA/sale-workflow to make a generic module
+    # on which this module can depend on. At the moment, we take the tz of the
+    # SUPERUSER. This is safer than the tz of the user (purchaser)
```
TDu: How is it safer ? Checking on the project those changes are intended for, I see the superuser has no tz set !

TDu (commented):
Also the unit test about timezone need to be updated, has the timezone used is not the one of the current user but the superuser.
https://github.com/OCA/purchase-workflow/blob/57a99827ac3affe6b414d1d04114cad3ae58d38c/purchase_delivery_split_date/tests/test_purchase_delivery.py#L206

TDu:
@jbaudoux If I understand well this refactor also works for the case mentioned on this pr https://github.com/OCA/purchase-workflow/pull/1233 ?
If so could you add the new test cases from it ?

purchase_delivery_split_date/models/purchase.py
jbaudoux: I find it safer because it is not anymore based on a user that can freely change his timezone in his preferences.

TDu:
I did some tests in the UI and when I change the Scheduled Date on a line, the Scheduled date on the related picking is not changed. I think this was a functionality of the previous implementation.

purchase_delivery_split_date/models/purchase.py
jbaudoux: I'm adding a fallback on context tz or user tz if not defined on admin

purchase_delivery_split_date/models/purchase.py
jbaudoux: Renamed and moved to stock.picking model

purchase_delivery_split_date/models/purchase.py
jbaudoux: I don't think it is useful to add in internal functions

jbaudoux:
> I did some tests in the UI and when I change the Scheduled Date on a line, the Scheduled date on the related picking is not changed. I think this was a functionality of the previous implementation.

@TDu Thanks for the testing. I'll redo that test then.

purchase_delivery_split_date/models/purchase.py
TDu: There is already kind of one in `context_today` https://github.com/odoo/odoo/blob/aff7694c56185212c17ceaf358ad99b3d14dee8b/odoo/fields.py#L1716

jbaudoux:
The test is failing due to a regression on orderpoints since this commit in odoo https://github.com/odoo/odoo/commit/006ab22aae0e6d8e9bd88d827282d6755f808cf0

jbaudoux:
> The test is failing due to a regression on orderpoints since this commit in odoo [odoo/odoo@006ab22](https://github.com/odoo/odoo/commit/006ab22aae0e6d8e9bd88d827282d6755f808cf0)

Issue opened here https://github.com/odoo/odoo/issues/75481

TDu:
@jbaudoux Could you integrate the last commit of this branch https://github.com/camptocamp/purchase-workflow/commits/13-split_date.
https://github.com/camptocamp/purchase-workflow/commit/0684fb866927a730702545aeb2ee71147394f5fe
Thank you

jbaudoux:
> @jbaudoux Could you integrate the last commit of this branch https://github.com/camptocamp/purchase-workflow/commits/13-split_date. [camptocamp@0684fb8](https://github.com/camptocamp/purchase-workflow/commit/0684fb866927a730702545aeb2ee71147394f5fe) Thank you

Done. Rebased & squashed

gurneyalex:
@jbaudoux can you check the CI ? I see errors in the tests in Travis which should be easy to fix ( for instance `NameError: name 'SUPERUSER_ID' is not defined`)

jbaudoux:
> @jbaudoux can you check the CI ? I see errors in the tests in Travis which should be easy to fix ( for instance `NameError: name 'SUPERUSER_ID' is not defined`)

obvious fix done. Thanks for the ping ;)

purchase_delivery_split_date/models/purchase.py
```diff
@@ -1,138 +1,109 @@
...
+        date_planned = picking.dt2date(self.date_planned)
+        expected_dates = {
+            picking.dt2date(expected_date)
+            for expected_date in picking.move_lines.mapped("date_expected")
+        }
+        return (
+            picking.state not in ("done", "cancel")
+            and picking.dt2date(picking.scheduled_date) == date_planned
+            and len(expected_dates) == 1
+            and picking.location_dest_id.usage in ("internal", "transit", "customer")
```
gurneyalex, line 22: maybe add a comment that "customer" is there to handle drop shipping purchases?

purchase_delivery_split_date/models/purchase.py
```diff
@@ -1,138 +1,109 @@
...
         moves = self.env["stock.move"]
-        # Group the order lines by group key
-        order_lines = sorted(
-            self.filtered(lambda l: not l.display_type), key=lambda l: l.date_planned
-        )
-        date_groups = groupby(
-            order_lines, lambda l: self._get_group_keys(l.order_id, l, picking=picking)
-        )
+        for line in self:
+            moves |= super(PurchaseOrderLine, line)._create_stock_moves(picking)
```
gurneyalex: I expect += to be a little faster here: these are new records so there should not be any duplicates.

gurneyalex (approved):

purchase_delivery_split_date/models/purchase.py
jbaudoux: changed

purchase_delivery_split_date/models/purchase.py
```diff
@@ -1,138 +1,109 @@
...
+                if not picking:
+                    copy_vals = self._prepare_picking_copy_vals()
+                    picking = move.picking_id.copy(copy_vals)
+                    pickings |= picking
+            move._do_unreserve()
+            picking_to_cancel |= move.picking_id
+            move.picking_id = picking
+            move._merge_moves()
+            move._action_assign()
+        picking_to_cancel.filtered(lambda p: not p.move_lines).write(
```
kos94ok-3D, line 95: I think better call `action_cancel()`

jbaudoux: No, it cannot. It would do nothing as there are no move lines

ndevas: But if you need to do something with a transfer when canceling it, you will add it in `action_cancel` or in `write`?
Not always cancel action related to move_lines

jbaudoux: @ndevas The transfer is not canceled. The moves are reassigned to other transfer. Nothing is canceled! The empty picking is closed in a cancel state

PR #1249, opened 2021-08-23
[14.0][ADD] purchase_global_discount

omar7r:
Depends on https://github.com/OCA/account-invoicing/pull/990


omar7r:
Migrated account_multi_vat from 13.0 in https://github.com/OCA/account-fiscal-rule/pull/266

PR #1253, opened 2021-08-30
[13.0][ADD] purchase_order_qty_change_no_recompute: Prevent recompute if only quantity has changed in purchase order line

victoralmau:
Prevent recompute if only quantity has changed in purchase order line.

Locked by:

@Tecnativa TT31537


victoralmau:
The current travis error is from the `purchase_line_procurement_group` addon and is not related to this PR.

chienandalu (commented):
What if the onchange is legit? Shouldn't we flag this exceptional behavior only when the price is changed on hand?

purchase_order_qty_change_no_recompute/__manifest__.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2021 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Purchase Order Qty change no recompute",
+    "summary": "Prevent recompute if only quantity has changed in purchase order line",
+    "version": "13.0.1.0.1",
```
joao-p-marques:
```suggestion
    "version": "13.0.1.0.0",
```

purchase_order_qty_change_no_recompute/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module prevent recompute if only quantity has changed in purchase order line.
```
joao-p-marques:
```suggestion
This module prevents recompute if only quantity has changed in purchase order line.
```

joao-p-marques (commented):
As @chienandalu said, maybe ignoring the onchange completely is a bit dangerous, as it may introduce some other changes that make sense, like the `date_planned` (https://github.com/odoo/odoo/blob/13.0/addons/purchase/models/purchase.py#L715).

So what if, instead of forcing the onchange to be skipped, you overwrote it and applied only what makes sense here?

Apart from that, just some minor comments to the README and manifest.

purchase_order_qty_change_no_recompute/readme/DESCRIPTION.rst
pedrobaeza:
```suggestion
This module prevents to recompute if only quantity has changed in purchase order line.
```

pedrobaeza:
The approach to fade out completely the onchange is on purpose, as we don't want any data to be changed when changing quantity, so it's OK to leave it as is.

chienandalu (approved):
Ok, maybe a safeguard to skip this could be added with a context. Otherwise :+1:

pedrobaeza:
Travis is red. Please check.

victoralmau:
Sorry, it is related to the `purchase_line_procurement_group` addon https://github.com/OCA/purchase-workflow/pull/1234#issuecomment-909980018

purchase_order_qty_change_no_recompute/models/purchase_order.py
```diff
@@ -0,0 +1,23 @@
...
+class PurchaseOrderLine(models.Model):
+    _inherit = "purchase.order.line"
+
+    def _onchange_eval(self, field_name, onchange, result):
+        """Remove the trigger for the undesired onchange method with this field.
+        We have to act at this place, as `_onchange_methods` is defined as a
+        property, and thus it can't be inherited due to the conflict of
+        inheritance between Python and Odoo ORM, so we can consider this as a HACK.
+        """
+        if field_name == "product_qty":
```
pedrobaeza: Please add a safeguard for tests evaluating `config["test_enable"]` and only acting if a special context key is passed.

joao-p-marques (approved):

pedrobaeza (approved):

purchase_order_qty_change_no_recompute/models/purchase_order.py
victoralmau: Do you think we should apply this change also to the `sale_order_qty_change_no_recompute` addon?

pedrobaeza: Yeah, that would prevent the problem you have with the other module in tests.

joao-p-marques:
@victoralmau Travis is still red. Can you please check it?

Looks like a rebase could solve it...

victoralmau:
The travis bug is related to the `purchase_line_procurement_group` addon and is not resolved yet.

In another PR https://github.com/OCA/purchase-workflow/pull/1234 I have added a commit that IMO solves the problem, so if we merge that PR first, then you could rebase and merge this one as well.

pedrobaeza:
Please rebase

victoralmau:
Rebase done, ready to merge now.

PR #1254, opened 2021-08-31
[14.0][FIX] purchase_work_acceptance_late_fines : Check Late Fines is not negative value.

newtratip:
Selection_211


kittiu:
@newtratip I think it is much easier to do this as _sql_constraints

I have added it here in another PR -> https://github.com/OCA/purchase-workflow/pull/1307/files#diff-d58c37f206b4f0f4748c4d9c64184bab8a4c76dcb597f5f6e5e3e1ea7605812dR42-R54

PR #1260, opened 2021-09-09
[13.0][ADD] purchase_landed_cost_volumetric_weight: Add volumetric_weight option in calculation_method from expense types.

victoralmau:
Add volumetric_weight option in calculation_method from expense types.

@Tecnativa TT30782


pedrobaeza:
Let's not go to this way, but `stock_landed_costs` way. We can talk about this later.

PR #1261, opened 2021-09-09
[14.0] purchase_order_line_price_history needs create permission

oa-eska:
set price doesn't work without this permission


purchase_order_line_price_history/security/ir.model.access.csv
```diff
@@ -1,3 +1,3 @@
 id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
-access_purchase_order_line_price_history_line,access_purchase_order_line_price_history_line,model_purchase_order_line_price_history_line,base.group_user,1,0,0,0
-access_purchase_order_line_price_history,access_purchase_order_line_price_history,model_purchase_order_line_price_history,base.group_user,1,0,0,0
+access_purchase_order_line_price_history_line,access_purchase_order_line_price_history_line,model_purchase_order_line_price_history_line,base.group_user,1,1,1,1
+access_purchase_order_line_price_history,access_purchase_order_line_price_history,model_purchase_order_line_price_history,base.group_user,1,1,1,1
```
kittiu, line 3: This line don't need.
```suggestion
access_purchase_order_line_price_history,access_purchase_order_line_price_history,model_purchase_order_line_price_history,base.group_user,1,0,0,0
```

kittiu (approved):
Thanks for the finding.

pedrobaeza (changes requested):
Aborting. This is not totally correct. Permissions has been put on transient model on purpose, for restricting those who has accessed to potential confidential information. In this case, the purchase information, so you should change permissions to be full for purchase users, but none for the rest of the regular users.

PR #1266, opened 2021-09-15
[12.0][NEW] Module purchase_add_products_from_bom.

mbcosta:
New module to allow add products from a BoM in the Purchase Order.

Go to Purchase Order click the button 'Add Products from BoM' 

image

Inform the Product, BoM and quantity and click 'Get Products from BoM', will appear the list of products where you can change the Price or Quantity if necessary

image

In case of Purchase UoM are different from Product UoM the Product Quantity are automatic change

image

In the case of Finished Product has another Product with BoM inside the type Kit/phanton will be automaticlly exploded the type Manufacture this product/normal will appear with button 'Add Products from BoM' to make possible also exploded this product,

image

 in the case the product have more than one BoM the user need choose one

image

After explode the BoM the line become bold and appears an button to use if the user want to make reverse process

image

The user need to mark the products to be add, if everything is right click in 'Add Products'

image

image

If include more than one BoM and the products has some common items those will be add in exist purchase line instead to duplicate

image

image

cc @renatonlima @rvalyi


purchase_add_products_from_bom/wizard/add_products_from_bom.py
```diff
@@ -0,0 +1,387 @@
...
+    product_qty_after_buy = fields.Float(
+        string='Quantity After Buy',
+        compute='_compute_product_qty_after_buy',
+    )
+    total_product_value = fields.Monetary(
+        string='Total Product Value',
+        currency_field='currency_id',
+        compute='_compute_total_product_value',
+    )
+    selected_product = fields.Boolean(string='Selected Product')
```
quentinDupont: I think it should be True by default. It was confusing for me : if i selected a product (and its bom), it is because i want it to be import in the purchase order :) And it's not perfectly clear that we have to select the lines for that !

purchase_add_products_from_bom/wizard/add_products_from_bom.py
```diff
@@ -0,0 +1,387 @@
...
+
+        factor = product_uom_id._compute_quantity(
+            product_qty,
+            bom_id.product_uom_id) / bom_id.product_qty
+
+        boms, lines = bom_id.explode(
+            finished_product, factor, picking_type=bom_id.picking_type_id)
+
+        for line in lines:
+            # LINE: (mrp.bom.line(19,), {
```
quentinDupont: you can remove comments i think :)

purchase_add_products_from_bom/wizard/add_products_from_bom.py
```diff
@@ -0,0 +1,387 @@
...
+                else:
+                    values['main_bom_product_id'] = self.product_id.id
+
+            else:
+                values['main_bom_product_id'] = self.product_id.id
+
+            self.create(values)
+
+        return {
+            # 'context': self.env.context,
```
quentinDupont: comment to remove

purchase_add_products_from_bom/wizard/add_products_from_bom.py
```diff
@@ -0,0 +1,387 @@
...
+    def return_finished_products_from_bom(self):
+        products_to_delete = self.product_finished_id.raw_product_line_ids.search([
+            ('main_bom_product_id', '=', self.product_id.id),
+            ('product_finished_id', '=', self.product_finished_id.id)
+        ])
+
+        products_to_delete.unlink()
+        self.exploded_product = False
+        return {
+            # 'context': self.env.context,
```
quentinDupont: same

purchase_add_products_from_bom/wizard/add_products_from_bom.py
```diff
@@ -0,0 +1,387 @@
...
+        comodel_name='mrp.bom',
+        string='Bill of Material',
+    )
+    # Used to create Domain
+    bom_ids = fields.One2many(
+        comodel_name='mrp.bom',
+        related='product_id.product_tmpl_id.bom_ids',
+        string='Bill of Materials'
+    )
+    product_qty = fields.Float(
```
quentinDupont, line 37: default=1 seems fine for me !

quentinDupont (changes requested):
Thanks for the work ! i will add french translation later :)

quentinDupont:
I add FR translation here : https://github.com/akretion/purchase-workflow/pull/14

purchase_add_products_from_bom/wizard/add_products_from_bom.py
mbcosta: It was made in this way beacuse some products can has enough quantity in stock and don't need to buy more, that the reason to show in wizard the "Quantity Avaliable in Stock" and "Quantity after Buy" so the users should be able to decide based in those quantitys which products they want to import. But I don't see any problem in mark all lines with True and the users uncheck the lines

mbcosta:
Thank you @quentinDupont for review and the french translation, I was near to close this PR because we are not sure if the mainteners of the repository has interesing in put this module here, I think it's important confirm this question ( Do you know if this module can be include here? ), my other option are try to put the module in https://github.com/akretion/odoo-usability

purchase_add_products_from_bom/wizard/add_products_from_bom.py
quentinDupont: Yes i understand. I think that it's quicker for a lot of use cases, to have all lines to True and uncheck them :)

quentinDupont:
> Thank you @quentinDupont for review and the french translation, I was near to close this PR because we are not sure if the mainteners of the repository has interesing in put this module here, I think it's important confirm this question ( Do you know if this module can be include here? ), my other option are try to put the module in https://github.com/akretion/odoo-usability

Well, this module is about mrp and purchase workflow. But everything action is in purchase workflow so I would say that this is the right place :+1: and i could be interesting in this module if it was in OCA !

mbcosta:
@quentinDupont I made the changes, field "Select Product" default True, "Product Qty" default 1, removed the comments and another one change in wizard Tree View moved the field "Select True" to first column

image

The error in Travis don't seems related with PR 

image

Maybe it's should be solve by the pendent action to replace CI Travis for CI Github Actions ( https://odoo-community.org/groups/contributors-15/contributors-253387?mode=date&date_begin=2022-09-01&date_end=2022-10-01 ), it's done in this repo only in version 13 and foward

mbcosta:
Hi @rvalyi @renatonlima @quentinDupont 

Do you see any problem or anything to be change or fixed in the module for be approved?

renatonlima (approved):

quentinDupont (approved):

PR #1268, opened 2021-09-17
[13.0][ADD] purchase_order_supplierinfo_update: new module

ernestotejeda:
Cc @Tecnativa TT31625

This module extends the functionality of purchase to allow the price of the supplier-info to be updated automatically for each product of the a purchase when it is confirmed. This will allow that when a purchase order is created that the suggested price for those products is the last one purchased.


purchase_order_supplierinfo_update/models/purchase_order.py
```diff
@@ -0,0 +1,50 @@
...
+from odoo import api, models
+
+
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+
+    def write(self, vals):
+        res = super().write(vals)
+        if vals.get("state", "") in ["purchase", "done"]:
+            self.mapped("order_line").update_supplierinfo_prince()
```
JordiBForgeFlow: prince --> price

JordiBForgeFlow:
To review @ForgeFlow

purchase_order_supplierinfo_update/models/purchase_order.py
```diff
@@ -0,0 +1,50 @@
...
+                # The same domain used in purchase.order.line to suggest the price.
+                # see _suggest_quantity method in models/purchase.py of odoo core
+                # purchase module.
+                supplierinfo = self.product_id.seller_ids.filtered(
+                    lambda r: (
+                        r.name == self.order_id.partner_id
+                        and (not r.product_id or r.product_id == self.product_id)
+                    )
+                ).sorted(key=lambda r: r.min_qty)
+                if supplierinfo and line.price_unit != supplierinfo[0].price:
```
legalsylvain: Please create a function for that. ``def _update_supplierinfo(self, line)``.

that will allow inheritance. otherwise, it's not possible to make glue modules with purchase_discount and purchase_triple_discount modules.

legalsylvain (changes requested):
One suggestion to make module modular.

A general question. i see this module very similar to account_invoice_supplier_info_update, but the design is completely different. In the other module, we have a wizard that display differences, and we can update (or remove some lines we don't want to update).
In your implementation, that's not possible. I see it as a limitation, and also something quite dangerous. (update automatically data without saying the update to the user).

I understand that Odoo make that for creation, but update is something more sensible.

what your point of view ? 

did you thought to intherit from account_invoice_supplierinfo_update to reuse wizard and algorighm ? 

Ref module : https://github.com/OCA/account-invoicing/tree/12.0/account_invoice_supplierinfo_update#readme

kind regards.

purchase_order_supplierinfo_update/models/purchase_order.py
```diff
@@ -0,0 +1,50 @@
+# Copyright 2021 Tecnativa - Ernesto Tejeda
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, models
+
+
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+
+    def write(self, vals):
```
carlosdauden, line 10: Why don't extend button_approve instead of write:
https://github.com/odoo/odoo/blob/d9ff2e8ec7200aa15830e92e5083251c4e81dd1c/addons/purchase/models/purchase.py#L349

What happen when the user clicks in Lock button in an approved PO?
will it called without changes?

AlvaroTForgeFlow:
Hello. We have been testing the code and, while it works fine for single-product orders, it doesn't seem to work for multi-product orders.

An example to ilustrate: say that I create a purchase order request on company A for products P1 and P2.
    P1 has the current supplier-info price tag set to 10
    P2 has the current supplier-info price tag set to 5
    In this one single order I purchase:
        -P1 for 15
        -P2 for 25
    Then I confirm order, and save.
    Later I set another order, for company A, again I purchase P1 and P2 (I fill the blanks by that order).
    -P1: loads 25 as the supplier-info price tag (the price of P2 from the previous order)
    -P2: loads 5 as the supplier-info price tag (the price of P2 from the beginning)

Could it be something related wtih indexes? Could it be this line of code? In purchase_order.py, at the update_supplierinfo_prince function.
```
                if supplierinfo and line.price_unit != supplierinfo[0].price:
                    supplierinfo[0].price = line.price_unit
```

Also, a suggestion: when you select a product, the suggested price tag is updated, could it be possible to make it happen also when changing the Vendor?

purchase_order_supplierinfo_update/models/purchase_order.py
ernestotejeda: done

ernestotejeda:
> One suggestion to make module modular.
> 
> A general question. i see this module very similar to account_invoice_supplier_info_update, but the design is completely different. In the other module, we have a wizard that display differences, and we can update (or remove some lines we don't want to update).
> In your implementation, that's not possible. I see it as a limitation, and also something quite dangerous. (update automatically data without saying the update to the user).
> 
> I understand that Odoo make that for creation, but update is something more sensible.
> 
> what your point of view ?
> 
> did you thought to intherit from account_invoice_supplierinfo_update to reuse wizard and algorighm ?
> 
> Ref module : https://github.com/OCA/account-invoicing/tree/12.0/account_invoice_supplierinfo_update#readme
> 
> kind regards.

Sylvain thank you very much for your suggestion. I think you can do a PR to incorporate the functionality of the wizard, but that must be optional per company since our client wants it to be done automatically.

purchase_order_supplierinfo_update/models/purchase_order.py
ernestotejeda: @legalsylvain I have extract the supplierinfo update in a separate method to make it extensible as you suggested, could you check it, please?

purchase_order_supplierinfo_update/models/purchase_order.py
ernestotejeda: @carlosdauden check the changes I made regarding this, please.

ernestotejeda:
> Hello. We have been testing the code and, while it works fine for single-product orders, it doesn't seem to work for multi-product orders.
> ..

@AlvaroTForgeFlow could you test that use case again to see if it still does'n work fine, thanks

ernestotejeda:
@carlosdauden @legalsylvain @AlvaroTForgeFlow @JordiBForgeFlow  changes done.

I have used the same _select_seller method of product.product , what do you think @carlosdauden

carlosdauden (approved):

pedrobaeza:
OK for you @legalsylvain ?

legalsylvain (commented):
Hi. Thanks for the changes ! I still think that refactor and mutualization of code should be done with ``account_invoice_supplierinfo_update`` but it's ok for that release. I'll not work on V13, and migrate directly to V16, so I'll do that job for that release.

@ernestotejeda : could you just add a file ``ROADMAP.rst`` file in the readme folder, containing a text like : 

* refactor that module to share algorithm with similar module ``account_invoice_supplierinfo_update``.

Except that little addition, :+1: to merge that PR. 
Thanks for sharing your work and for the changes !

ernestotejeda:
@legalsylvain thank you, changes done

PR #1272, opened 2021-09-22
[FIX] purchase_order_line_sequence: do not reset sequence

robinkeunen:
Resetting the line sequence forbids the user to set his own line sequence. It makes the module incompatible with module purchase_order_line_sequence.

I guess `reset_sequence` has a reason to be there but I do not understand it. This PR is a way to ask that question.


robinkeunen:
@cjallais  @damdam-s can you take a look at this ?

PR #1273, opened 2021-09-24
[14.0][purchase_return][new] module to manage purchase return orders

JordiBForgeFlow:
This module allows to create Purchase Return Orders. Typically vendors expect
to receive Purchase Return Orders as the starting event for claims associated
with products or services delivered.

The process work pretty much the same as with Purchase Orders, but the flow
is inverted. Upon indication of quantities delivered (representing the
services that have been accepted to be refunded for, or products returned) you
can create the subsequent refund bills.

TODO:
* [x] Add tests

@ForgeFlow


legalsylvain (commented):
Hi. 

thanks for sharing this module. I don't understand why you marked " Use inheritance from purchase.order" as done. I don't see the inheritance mechanism here.
Did I missed something ?

JordiBForgeFlow:
@legalsylvain I removed this check. We tried to use the inheritance but was a disaster. It's best not to inherit from a non-abstract model

legalsylvain:
hi. thanks for your answer. 
Could you elaborate, why you not used ``purchase.order`` model ? Naive question, but this module seems a fork of the purchase feature (+3500 lines !).
I think it will be incompatible with all the other modules present in purchase-workflow. (``purchase_discount``, ``purchase_triple_discount`` (for instance) but I guess others modules also).

I'm quite dubious about this kind of implementation, and I think it will require a lot of maintenance.
what your point of view ?

purchase_return/__manifest__.py
```diff
@@ -10,7 +10,7 @@
     "license": "LGPL-3",
     "application": False,
     "installable": True,
-    "depends": ["purchase"],
+    "depends": ["purchase", "mail"],
```
legalsylvain, line 13: Purchase install mail, so it should not be necessary. Dont you think ?

purchase_return/models/product_template.py
```diff
@@ -0,0 +1,26 @@
...
+        help="Keep this field empty to use the default value from the product category.",
+    )
+
+    def _get_product_accounts(self):
+        accounts = super(ProductTemplate, self)._get_product_accounts()
+        accounts["vendor_returns"] = (
+            self.property_account_vendor_return_id
+            or self.categ_id.property_account_vendor_return_categ_id
+        )
+        return accounts
```
legalsylvain, line 26: Well, in most case, return account is the same as the expense account.
In my opinion, this function should return default expense account, if no account return is fund.

pedrobaeza:
Does it make sense to reuse `rma` module to perform this? This way, you will have covered picking + invoice part, and only have to inject the parts for the invoice type and similar things.

PR #1275, opened 2021-09-28
[13.0][IMP] purchase_stock_picking_show_currency_rate + purchase_last_price_info: Improve tests according to different currencies

victoralmau:
Improve tests according to different currencies.

@Tecnativa TT30631


pedrobaeza:
Same comments as the other PR.

pedrobaeza:
Anyways, do we know which module is changing the currency of the main company in the integrated environment? Maybe we should avoid that in first place.

joao-p-marques (approved):

pedrobaeza (approved):

PR #1278, opened 2021-10-04
[FIX] purchase_work_acceptance: fix view_move_form view

albariera:


purchase_work_acceptance/views/account_move_views.xml
```diff
@@ -3,7 +3,7 @@
     <record id="view_move_form" model="ir.ui.view">
         <field name="name">account.move.form</field>
         <field name="model">account.move</field>
-        <field name="inherit_id" ref="account.view_move_form" />
+        <field name="inherit_id" ref="purchase_stock.stock_move_purchase" />
```
etobella: maybe it would be better to use `purchase.view_move_form_inherit_purchase`

etobella (changes requested):

purchase_work_acceptance/views/account_move_views.xml
albariera: You are right! Thanks!

kittiu:
@albariera what is the benefit of this vs the old one?
Thanks!

albariera:
@kittiu There was a bug because it was trying to get the field "purchase_vendor_bill_id" from the "account.view_move_form" view, but it was not present in that view. This field is defined at "view_move_form_inherit_purchase" from the module "purchase". Thanks!

kittiu (approved):

etobella (approved):

etobella:
> @kittiu There was a bug because it was trying to get the field "purchase_vendor_bill_id" from the "account.view_move_form" view, but it was not present in that view. This field is defined at "view_move_form_inherit_purchase" from the module "purchase". Thanks!

Also, an error was raised if a user without access to purchase accessed an account.move.

kittiu:
Should this port to v14 as well?

albariera:
Done @kittiu! #1282

kittiu:
> Done @kittiu! #1282

🙏

PR #1279, opened 2021-10-14
[IMP][FIX] purchase_stock_price_unit_sync: handle kit cases

chienandalu:
If we receive a kit, we shouldn't write the kit purchase price on the
components price units.

TT32040

ping @sergio-teruel

### DIFF
--- a/purchase_stock_price_unit_sync/models/purchase_order.py
+++ b/purchase_stock_price_unit_sync/models/purchase_order.py
@@ -16,6 +16,15 @@ class PurchaseOrderLine(models.Model):
 
     def stock_price_unit_sync(self):
         for line in self.filtered(lambda l: l.state in ['purchase', 'done']):
+            # When the affected product is a kit we do nothing, which is the
+            # default behavior on the standard: the move is exploded into moves
+            # for the components and those get the default price_unit for the
+            # time being. We avoid a hard dependency as well.
+            if (
+                hasattr(line.product_id, "bom_ids")
+                and line.product_id._is_phantom_bom()
+            ):
+                continue
             line.move_ids.write({
                 'price_unit': line.with_context(skip_stock_price_unit_sync=True
                                                 )._get_stock_move_price_unit(),


pedrobaeza:
Why `_get_stock_move_price_unit` is not returning the product standard price?

chienandalu:
It gets the kit cost, not the components ones

pedrobaeza:
But that's because the standard? Shouldn't this be changed in Odoo itself?

chienandalu:
Well, as far as we could find out, the exploded component moves default to the current standard price in the moment of the validation as the price_unit of the purchase line isn't propagated (and that it's correct, or otherwise we'll be having wrong valuations, which is the point of the fix).

I'm going to try and get the immediate previous historic price so the cost is more reliable.

We can't do much better I'm afraid.

pedrobaeza (approved):

purchase_stock_price_unit_sync/models/purchase_order.py
```diff
@@ -16,6 +16,18 @@ def write(self, vals):
...
         for line in self.filtered(lambda l: l.state in ['purchase', 'done']):
+            # When we receive a kit, we can only default to its registered cost.
+            # We are avoiding a hard dependency here though.
+            if (
+                hasattr(line.product_id, "bom_ids")
+                and line.product_id._is_phantom_bom()
+            ):
+                for move in line.move_ids:
+                    previous_instant = move._previous_instant_date()
+                    move.price_unit = move.product_id.get_history_price(
```
sergio-teruel: Why do we update the price with the previous price??

sergio-teruel (approved):

sergio-teruel (changes requested):
Sorry

purchase_stock_price_unit_sync/models/purchase_order.py
chienandalu: Ok, as we debated internally it's better to do nothing as that's what core does anyway and we don't want resync the whole move chain

sergio-teruel (approved):

PR #1285, opened 2021-10-21
[13.0][IMP] purchase_order_supplierinfo_update: Fix random test failure

carlosdauden:
When 2 purchase orders are created by the test at the same time can have same date order and this condition:
https://github.com/OCA/purchase-workflow/blob/d30f00b3d646c9d969561d25e3bd9612aed4dc2a/purchase_order_supplierinfo_update/models/purchase_order.py#L33-L40
causes the price to be updated here:
https://github.com/OCA/purchase-workflow/blob/d30f00b3d646c9d969561d25e3bd9612aed4dc2a/purchase_order_supplierinfo_update/tests/test_purchase_order_supplierinfo_update.py#L74
and this assert fails:
https://github.com/OCA/purchase-workflow/blob/d30f00b3d646c9d969561d25e3bd9612aed4dc2a/purchase_order_supplierinfo_update/tests/test_purchase_order_supplierinfo_update.py#L75

@Tecnativa

### DIFF
--- a/purchase_order_supplierinfo_update/tests/test_purchase_order_supplierinfo_update.py
+++ b/purchase_order_supplierinfo_update/tests/test_purchase_order_supplierinfo_update.py
@@ -1,6 +1,9 @@
 # Copyright 2021 Tecnativa - Ernesto Tejeda
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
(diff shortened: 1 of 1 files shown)


legalsylvain (approved):
Makes sense

pedrobaeza (approved):

PR #1286, opened 2021-10-22
[14.0] Add purchase_partner_incoterm

sonhd91:


purchase_incoterm/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2020 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    'name': 'Purchase Incoterm',
```
sebalix: I would rename the addon `purchase_partner_incoterm` to be inline with the existing [`sale_partner_incoterm`](https://github.com/OCA/sale-workflow/tree/14.0/sale_partner_incoterm).
+ Update the descriptions/user labels as well to be coherent:
```suggestion
    "name": "Default purchase incoterm per partner",
    "summary": "Set the supplier preferred incoterm on each purchase order",
```

purchase_incoterm/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+    'summary': 'Add a an incoterm field for supplier and use it on purchase order',
+    'version': '14.0.1.0.0',
+    'category': 'Purchase',
+    'author': 'Camptocamp, Odoo Community Association (OCA)',
+    'license': 'AGPL-3',
+    'depends': [
+        'account',
+        'base',
```
sebalix:
```suggestion
```

purchase_incoterm/models/purchase_order.py
```diff
@@ -0,0 +1,14 @@
...
+
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+
+    @api.onchange('partner_id')
+    def _onchange_partner(self):
+        if self.partner_id:
+            self.incoterm_id = self.partner_id.purchase_incoterm_id
+        else:
+            self.incoterm_id = False
```
sebalix: There is already a `onchange_partner_id` on `purchase.order`, better to re-use it and simplify the implementation:
```suggestion
    def onchange_partner_id(self):
        res = super().onchange_partner_id()
        self.incoterm_id = self.partner_id.purchase_incoterm_id
        return res
```

purchase_incoterm/models/res_partner.py
```diff
@@ -0,0 +1,13 @@
...
+from odoo import fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    purchase_incoterm_id = fields.Many2one(
+        comodel_name="account.incoterms",
+        ondelete="restrict",
+        string="Purchase Incoterm",
```
sebalix: Use the same terms than `sale_partner_incoterm`:
```suggestion
        string="Default Purchase Incoterm",
        help="The default incoterm for new purchase orders for this supplier.",
```

sebalix (changes requested):

purchase_incoterm/models/res_partner.py
sonhd91: Thanks for reivew, branch updated

purchase_incoterm/models/purchase_order.py
sonhd91: Thanks for reivew, branch updated

purchase_incoterm/__manifest__.py
sonhd91: Thanks for reivew, branch updated

TDu: I wonder it the name in the manifest should not be the name of the module without the underscore ?

TDu:
Renaming the module as proposed by @sebalix into `purchase_partner_incoterm` seems like a good idea

sebalix:
@sonhd91 could you rename the module `purchase_partner_incoterm`?

sonhd91:
> @sonhd91 could you rename the module `purchase_partner_incoterm`?

module name changed, thanks

TDu (approved):

sebalix (approved):

PR #1288, opened 2021-10-25
[13.0][FIX] purchase_reception_notify: Show multiple lines for the same PO line on Receipt confirmation message.

HviorForgeFlow:
When stock move A and B belongs to same purchase order line, delivery quantity is shown only once with the last line information because dict purchase order line key is overwritten. 

Use stock move id as key seems better to handle quantities correctly in messages.

CC @ForgeFlow

### DIFF
--- a/purchase_reception_notify/models/stock_picking.py
+++ b/purchase_reception_notify/models/stock_picking.py
@@ -8,20 +8,20 @@ class StockPicking(models.Model):
     _inherit = "stock.picking"
 
     @api.model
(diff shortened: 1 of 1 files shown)


JordiMForgeFlow (commented):
Code review LGTM 👍🏼

JordiMForgeFlow (approved):

AaronHForgeFlow (approved):
:+1:  code review LGTM

PR #1296, opened 2021-10-29
[12.0][ADD] sale_purchase_service_grouping

ivantodorovich:
This module allows to configure the grouping of generated Purchase Orders for `service_to_purchase` products.

This configuration is done in the Product Category.

Possible values are:

* *Sales Order*: Group by Sales Order.
* *Product Category*: Group by Product Category.
* *Product Template*: Group by Product Template.
* *Product*: Group by Product.

An empty value defaults to the standard behaviour, which groups by Supplier.

---

From a functional perspective, this module is similar to`procurement_purchase_no_grouping`.

Technically, they're quite different, as service product don't generate procurements.
In fact this module doesn't even depend on `stock`.

### DIFF
--- a/sale_purchase_service_grouping/__init__.py
+++ b/sale_purchase_service_grouping/__init__.py
@@ -1,0 +1,1 @@
+from . import models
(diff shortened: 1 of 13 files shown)


ajaniszewska-dev (approved):
LG

PR #1300, opened 2021-11-16
[13.0][ADD] stock_landed_costs_valuation_real_time_bypass_check: Remove the need to define inventory valuation in real cost for the use of landed cost.

victoralmau:
Remove the need to define inventory valuation in real cost for the use of landed cost.

@Tecnativa TT32954


stock_landed_cost_valuation_real_time_bypass_check/models/stock_landed_cost.py
```diff
@@ -0,0 +1,46 @@
...
+    def compute_landed_cost(self):
+        _self = self.with_context(bypass_valuation_real_time=True)
+        return super(LandedCost, _self).compute_landed_cost()
+
+    def get_valuation_lines(self):
+        if not self.env.context.get("bypass_valuation_real_time"):
+            return super().get_valuation_lines()
+        lines = []
+
+        for move in self.mapped("picking_ids").mapped("move_lines"):
```
pedrobaeza: Why do you need to do all this stuff?

pedrobaeza (commented):
Yesterday I discover that account entries are created always:

https://github.com/odoo/odoo/blob/6f7e9a93048155e403e6cc56d24595c281297124/addons/stock_landed_costs/models/stock_landed_cost.py#L181

so you should shadow them as well.

stock_landed_cost_valuation_real_time_bypass_check/models/stock_landed_cost.py
victoralmau: According to https://github.com/odoo/odoo/blob/6f7e9a93048155e403e6cc56d24595c281297124/addons/stock_landed_costs/models/stock_landed_cost.py#L215-L216
we need to avoid that lines are not generated for products with category `valuation='real_time'`

Can you think of another possibility to do this better?

victoralmau: I have tried to avoid completely redoing this function and changing product categories to `real_time` previously, but that causes a big change that is NOT what we want (https://github.com/odoo/odoo/blob/13.0/addons/stock_account/models/product.py#L728-L770). I think it is better to redo `get_valuation_lines()` function than to avoid by context those changes in the category of a product.
What do you think?

pedrobaeza:
@victoralmau I'm now thinking that the host repo should be stock-logistics-workflow

pedrobaeza:
Closing as it has been put in the core in odoo/odoo#80200.

PR #1306, opened 2021-11-27
[13.0][FIX] procurement_purchase_no_grouping: Take into account SO updates

pedrobaeza:
Steps to reproduce:

- Create a product with MTO route and "No line grouping" product category.
- Create an SO with such product and quantity = 1.
- Confirm it.
- A purchase order is created with such product and qty = 1.
- Edit the sales order, and change quantity = 2.

Current behavior:

A new purchase line is created with quantity = 2, having ordered 3 units instead of 2.

Expected behavior:

The same existing purchase line is updated with quantity = 2.

------

For solving this problem, we are going to group by procurement group instead of simply discarding any possible line grouping.

NOTE: This makes that if you manually assign the same procurement group to several different sales orders, the grouping will be done no matter the grouping criteria, but this is the only way to do it without having to put a lot of glue modules, and on standard operation mode, procurement groups are not reused between sales orders.

@Tecnativa TT33031

### DIFF
--- a/procurement_purchase_no_grouping/README.rst
+++ b/procurement_purchase_no_grouping/README.rst
@@ -52,6 +52,14 @@ Go to each product category, and select one of these values in the field
 System default behaviour can be set up in System settings / Purchase / Procurement
 Purchase Grouping
 
+Known issues / Roadmap
+======================
+
+- If you reuse the same procurement group between several sales orders, and
+  using "No line grouping", they will be grouped anyways, as the criteria for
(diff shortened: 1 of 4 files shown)


victoralmau (approved):

carlosdauden (approved):

pedrobaeza:
Cherry-picked to v14 at 8df3b459da7c5f8d51c21831a186a4f43b927489

PR #1313, opened 2021-12-06
[13.0][FIX][IMP] purchase_security: prevent access error & rules for order lines

dalonsod:
This PR comes with a bugfix and an enhancement:
* _Prevent access error_. When a Purchase User (own orders) create or edit an order, if changes Representative user to another and save it, an access error is raised. With this change, those users cannot change the default Representative, then orders are initially assigned to them, and no errors were fired.
* _Rules for purchase order lines_. With this improvement, order lines are also restricted for Own Reponsible Users, and this addon becomes more compatible with e.g. `purchase_order_line_menu`.

### DIFF
--- a/purchase_security/views/purchase_order_views.xml
+++ b/purchase_security/views/purchase_order_views.xml
@@ -1,0 +1,26 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="purchase_order_form_own_orders" model="ir.ui.view">
+        <field name="name">purchase.order.form (own orders)</field>
+        <field name="model">purchase.order</field>
+        <field name="inherit_id" ref="purchase.purchase_order_form" />
+        <field name="groups_id" eval="[(4, ref('group_purchase_own_orders'))]" />
+        <field name="arch" type="xml">
+            <field name="user_id" position="attributes">
+                <attribute name="readonly">1</attribute>
+                <attribute name="force_save">1</attribute>
+            </field>
+        </field>
+    </record>
+    <record id="purchase_order_form_all_orders" model="ir.ui.view">
+        <field name="name">purchase.order.form (all orders)</field>
+        <field name="model">purchase.order</field>
+        <field name="inherit_id" ref="purchase_order_form_own_orders" />
+        <field name="groups_id" eval="[(4, ref('purchase.group_purchase_manager'))]" />
+        <field name="arch" type="xml">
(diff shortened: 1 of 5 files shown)


dalonsod:
@flook-chaiwat @joao-p-marques could you review? Thanks!

joao-p-marques (approved):
Code review. LGTM :+1:

purchase_security/views/purchase_order_views.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="purchase_order_form_own_orders" model="ir.ui.view">
+        <field name="name">purchase.order.form (own orders)</field>
+        <field name="model">purchase.order</field>
+        <field name="inherit_id" ref="purchase.purchase_order_form" />
+        <field name="groups_id" eval="[(4, ref('group_purchase_own_orders'))]" />
+        <field name="arch" type="xml">
+            <field name="user_id" position="attributes">
+                <attribute name="readonly">1</attribute>
```
pedrobaeza, line 10:
```suggestion
                <attribute name="readonly">1</attribute>
                <attribute name="force_save">1</attribute>
```

purchase_security/views/purchase_order_views.xml
```diff
@@ -0,0 +1,26 @@
...
+            </field>
+        </field>
+    </record>
+    <record id="purchase_order_form_all_orders" model="ir.ui.view">
+        <field name="name">purchase.order.form (all orders)</field>
+        <field name="model">purchase.order</field>
+        <field name="inherit_id" ref="purchase.purchase_order_form" />
+        <field name="groups_id" eval="[(4, ref('purchase.group_purchase_manager'))]" />
+        <field name="priority" eval="100" />
```
pedrobaeza:
```suggestion
        <field name="inherit_id" ref="purchase_order_form_own_orders" />
        <field name="groups_id" eval="[(4, ref('purchase.group_purchase_manager'))]" />
```

pedrobaeza (changes requested):

purchase_security/views/purchase_order_views.xml
dalonsod: Tested on runbot, without `force_save` still works, but I think it's a good point adding it anyway, so I've also included it.

pedrobaeza (approved):

PR #1316, opened 2021-12-13
[14.0][ADD] purchase_invoice_create_security_group

rousseldenis:


pedrobaeza:
Shouldn't this be called `purchase_invoicing_security` and go to `account-invoicing` the same as other invoicing related things?

rousseldenis:
> 

purchase_invoicing_security is too generic, maybe something more accurate like 'purchase_invoice_create_security_group'

IMHO, as the entry point is in purchase and moreover purchase module pulls account, this module should be here.

pedrobaeza:
Usually *-workflow repositories are very crowded, and that's why I prefer to put invoicing flows on that repo, but as you prefer.

rousseldenis:
> Usually *-workflow repositories are very crowded, and that's why I prefer to put invoicing flows on that repo, but as you prefer.

Yes, but that's quite the same for invoicing one.

IMHO, we maybe need to extract common modules that have the same scope in other repos as soon as possible

rousseldenis:
> IMHO, we maybe need to extract common modules that have the same scope in other repos as soon as possible

e.g: purchase_request*, purchase_work*

rousseldenis:
@OCA/procurement-purchase-maintainers

HviorForgeFlow (approved):

rousseldenis:
@HviorForgeFlow I will try to add a setting in order to activate the feature as it breaks tests in other modules that try to confirm.

HviorForgeFlow:
Sounds great!

rousseldenis:
@HviorForgeFlow Ready and  :green_apple:

rousseldenis:
@HviorForgeFlow

HviorForgeFlow (approved):
Code review :+1:

Issue #964, opened 2020-08-18
[11.0, 12.0] purchase_landed_cost only working for cost type average


jjscarafia:
Is there any reason why this module is only working for cost_method = average?
Check [this line](https://github.com/OCA/purchase-workflow/blob/11.0/purchase_landed_cost/models/purchase_cost_distribution.py#L271)

@pedrobaeza any idea about this?
Thanks!

pedrobaeza:
The module is simply not prepared for other costing methods. For the others, you have stock_landed_cost

jjscarafia:
Thanks

PR #1318, opened 2021-12-20
[14.0][MIG] purchase landed cost

jonerikceberio:
Migration of the 13.0 purchase_landed_cost.
The 12.0 functionality that updates the product price has not been reintroduced, but moved to another module purchase_landed_cost_price_avg that depends from this one, as to address #964 . If PR goes on I will make another PR with it.


pedrobaeza:
Since https://github.com/odoo/odoo/pull/80200, we are deprecating this module in favor of the standard one.

jonerikceberio:
Didn't know, closing the PR then

PR #1319, opened 2021-12-20
[15.0][MIG] purchase_tier_validation

JasminSForgeFlow:
Standard Migration
@ForgeFlow


etobella:
Tests are failing. Can you rebase and try again?

JasminSForgeFlow:
> Tests are failing. Can you rebase and try again?

Done, Thanks

purchase_tier_validation/tests/common.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2018 ForgeFlow S.L.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+
+def setup_test_model(env, model_clses):
```
simahawk: this approach is deprecated. Can you please use https://github.com/OCA/odoo-test-helper ?

purchase_tier_validation/tests/test_tier_validation.py
```diff
@@ -0,0 +1,65 @@
...
+
+from .common import setup_test_model, teardown_test_model
+from .tier_validation_tester import TierValidationTester
+
+
+@tagged("post_install", "-at_install")
+class TestPurchaseTierValidation(common.TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super(TestPurchaseTierValidation, cls).setUpClass()
```
simahawk, line 14: can you set `tracking_disable` on the env?

purchase_tier_validation/tests/test_tier_validation.py
```diff
@@ -0,0 +1,65 @@
...
+                "definition_domain": "[('test_field', '>', 1.0)]",
+            }
+        )
+
+        cls.test_record = cls.test_model.create({"test_field": 2.5})
+
+    @classmethod
+    def tearDownClass(cls):
+        teardown_test_model(cls.env, [TierValidationTester])
+        return super(TestPurchaseTierValidation, cls).tearDownClass()
```
simahawk, line 63: no need to return anything here

simahawk (commented):
@JasminSForgeFlow only technical remarks from me.
If you want to tackle them in a second PR, I won't block a PR that is here since so long for these small reasons (if you promise you'll take care :stuck_out_tongue_winking_eye: )

Camille0907 (approved):
LGTM, thanks

purchase_tier_validation/tests/common.py
JasminSForgeFlow: Updated, Thanks

purchase_tier_validation/tests/test_tier_validation.py
JasminSForgeFlow: Updated, Thanks

purchase_tier_validation/tests/test_tier_validation.py
JasminSForgeFlow: Pre-commit is failing if not add return

purchase_tier_validation/tests/test_tier_validation.py
simahawk: it won't work like that. You must override it in setUpClass like
```
        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
```

purchase_tier_validation/tests/test_tier_validation.py
simahawk: then pre-commit is wrong and you can work around it w/ `# pylint: disable=$err_code`

simahawk (changes requested):

JasminSForgeFlow:
@simahawk Changes have been updated.

Thanks

purchase_tier_validation/tests/test_tier_validation.py
simahawk: this has not changed. I won't block the PR but please rewrite this w/ disable comment in a new PR to not encourage this approach.

simahawk (approved):

PR #1336, opened 2022-01-17
[14.0][FIX] purchase_rfq_number : Keep po name from action cancel po > reset to draft > confirm order

newtratip:
This PR is make for keep po name when click cancel > reset to draft > confirm order

@kittiu @Nantikan23 Could you review this ?


Saran440 (approved):
Functional Test :+1: 

This function is helpful with import PO data from other system to odoo only

ps-tubtim (approved):
Functional test ok.

PR #1340, opened 2022-01-18
[IMP][14.0] Purchase_deposit:added button create_deposit

JoanMForgeFlow:
Improved the module of purchase deposit, by adding the possibility to just create the deposit and not the bill, or the two things.
@AaronHForgeFlow 
@JoanSForgeFlow

### DIFF
--- a/purchase_deposit/README.rst
+++ b/purchase_deposit/README.rst
@@ -23,7 +23,7 @@ Purchase Deposit
     :target: https://runbot.odoo-community.org/runbot/142/14.0
     :alt: Try me on Runbot
 
-|badge1| |badge2| |badge3| |badge4| |badge5| 
+|badge1| |badge2| |badge3| |badge4| |badge5|
 
 This module allow purchase order to register deposit similar to that in sales order
 
--- a/purchase_deposit/models/purchase.py
+++ b/purchase_deposit/models/purchase.py
@@ -23,7 +23,7 @@ class PurchaseOrderLine(models.Model):
 
     is_deposit = fields.Boolean(
         string="Is a deposit payment",
-        help="Deposit payments are made when creating invoices from a purhcase"
+        help="Deposit payments are made when creating bills from a purhcase"
         " order. They are not copied when duplicating a purchase order.",
     )
 
--- a/purchase_deposit/readme/CONFIGURATION.rst
+++ b/purchase_deposit/readme/CONFIGURATION.rst
@@ -1,5 +1,5 @@
 After install this module, you need to select the product "Purchase Deposit" in
-configuration window to be used as default product when create deposit invoice.
+configuration window to be used as default product when create deposit bill.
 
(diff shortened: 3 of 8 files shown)


mariadforgeflow (approved):
LGTM!

HviorForgeFlow:
WDYT @kittiu

AaronHForgeFlow (commented):
Would it be good to hide the Create deposit button in the PO after the vendor bill is created?

Edit: well let's keep it just in case there is some late deposit

JoanSForgeFlow (approved):
Tested! LGTM

AaronHForgeFlow (commented):
I am not sure if I understand, but it seems the button "Create deposit" is not fine:

1.- Create PO
2.- Create Deposit (without bills) --> a PO line is added
3.- Create vendor bill --> the deposit is not deducted on the bill.

JoanMForgeFlow:
@AaronHForgeFlow 
you have two options, the first one is to create the deposit (a PO line is added), but latter manually you have to create your deposit bill (invocing -> vendors -> bill), to be deducted from the total amount of the vendor bill.
The second option, is to create the deposit and the deposit bill, and in the vendor bill it will be deducted.

So, @AaronHForgeFlow , you are missing the deposit bill.

AaronHForgeFlow:
Ok, I would like some assistance on the creation of the deposit bill after the deposit is created. But I guess that is not trivial as the bill will be created for accountants.

AaronHForgeFlow (approved):
Functional + code

JordiBForgeFlow (approved):

vvrossem:
> @AaronHForgeFlow you have two options, the first one is to create the deposit (a PO line is added), but latter manually you have to create your deposit bill (invocing -> vendors -> bill), to be deducted from the total amount of the vendor bill. The second option, is to create the deposit and the deposit bill, and in the vendor bill it will be deducted.
> 
> So, @AaronHForgeFlow , you are missing the deposit bill.

Hi @JoanMForgeFlow !
A late functional review: 
I've tested the option n°1 you mentioned (first use "Create deposit", then manually create the vendor bill):

https://user-images.githubusercontent.com/28013700/160088883-5423d87c-d7f8-4204-92dd-dbd016163619.mov

Is it a wanted behaviour to manually encode, on the vendor bill, a negative quantity to have the deposit deducted from the total amount? Or am I missing something?

JoanMForgeFlow:
@vvrossem thanks for the review!
The idea of first creating the deposit and later manually making the bill is creating 2 bills in total, not just one.
So you make the deposit, then you go to invoicing -->bills and fill the autocomplete field,  for example in your case 3 new lines has been created, so you have to delete the 2 lines with the product (because you just want to bill the deposit), and only keep the deposit line. So when you register the payment and make the final bill, you are going to see how the deposit amount is deducted from the total amount.

vvrossem:
@JoanMForgeFlow thank you for the explanation, got it !

PR #1342, opened 2022-01-24
[MIG] purchase_request: Migration to 15.0

PabloPaezSheridan:


HviorForgeFlow (changes requested):
Right now there are 2 places where a similar message is received, when a picking is done and when a move line is mark as done

image

I would drop the stock picking one and keep the one coming from the move line.

HviorForgeFlow (commented):
I think Purchase Request form needs some refactoring taking advantage of Odoo view improvements. They way you add products to the Purchase Request could be speed up if in line edit is enable, but we could still access the form for more information by a button.

image

PabloPaezSheridan:
> I think Purchase Request form needs some refactoring taking advantage of Odoo view improvements. They way you add products to the Purchase Request could be speed up if in line edit is enable, but we could still access the form for more information by a button.
> 
> image
@HviorForgeFlow  I have just change the code for this functionality, take a look

Now the form view of the purchase request line is over here
https://github.com/OCA/purchase-workflow/blob/9770a1241c601486863c5a558b3d35bb10f634f7/purchase_request/views/purchase_request_line_view.xml#L548

The button is here
https://github.com/OCA/purchase-workflow/blob/9770a1241c601486863c5a558b3d35bb10f634f7/purchase_request/views/purchase_request_view.xml#L197

and the method which call the view is here
https://github.com/OCA/purchase-workflow/blob/9770a1241c601486863c5a558b3d35bb10f634f7/purchase_request/models/purchase_request_line.py#L395

PabloPaezSheridan:
> Right now there are 2 places where a similar message is received, when a picking is done and when a move line is mark as done
> 
> image
> 
> I would drop the stock picking one and keep the one coming from the move line.

@HviorForgeFlow i can not make this happen, when i validate the income of a purchase order picking only the message of the stock.piking is showed 
image

florenciafrigieri2:
> > Right now there are 2 places where a similar message is received, when a picking is done and when a move line is mark as done
> > image
> > I would drop the stock picking one and keep the one coming from the move line.
> 
> @HviorForgeFlow i can not make this happen, when i validate the income of a purchase order picking only the message of the stock.piking is showed image

@HviorForgeFlow Hi! If I understood correctly, what you mean before is that those two message should be the same one. If changing that part of the code is going to be complicated, I'd prefer for now to prioritize the migration for now. The messages are pretty similar and appear in different objects (picking and stock request)...

florenciafrigieri2:
Another comment... can we please change the domain & context of the supplier when creating the RFQ to align this module to Odoo standard for PO? Currently the domain is asking for the supplier to be a company:

image

While Odoo standard when creating a new PO is accepting any company as vendor: 
image

purchase_request/models/purchase_request.py
```diff
@@ -0,0 +1,299 @@
...
+        for rec in self:
+            rec.estimated_cost = sum(rec.line_ids.mapped("estimated_cost"))
+
+    @api.depends("line_ids")
+    def _compute_purchase_count(self):
+        for rec in self:
+            rec.purchase_count = len(rec.mapped("line_ids.purchase_lines.order_id"))
+
+    def action_view_purchase_order(self):
+        action = self.env.ref("purchase.purchase_rfq").sudo().read()[0]
```
i-vyshnevska:
```suggestion
        action = self.env["ir.actions.actions"]._for_xml_id("purchase.purchase_rfq")
```
use this instead sudo+read all over places

HviorForgeFlow:
> > Right now there are 2 places where a similar message is received, when a picking is done and when a move line is mark as done
> > image
> > I would drop the stock picking one and keep the one coming from the move line.
> 
> @HviorForgeFlow i can not make this happen, when i validate the income of a purchase order picking only the message of the stock.piking is showed image



> Now the form view of the purchase request line is over here https://github.com/OCA/purchase-workflow/pull/1342/files#diff-40b0317e677c2ab3bca46661a02996c87f0f9fe946ab73fb19cf511232c70c36R548
> 
> The button is here https://github.com/OCA/purchase-workflow/pull/1342/files#diff-40b0317e677c2ab3bca46661a02996c87f0f9fe946ab73fb19cf511232c70c36R548

I think company field in PR Lines is not so useful, it could be added as optional if it is possible/or remove and use Preferred supplier instead. 
Preferred supplier only can be found on PR Lines model tree view. It should be added also to PR Line form.

https://github.com/OCA/purchase-workflow/pull/1342/files#diff-40b0317e677c2ab3bca46661a02996c87f0f9fe946ab73fb19cf511232c70c36R548 I see all Migration commit please update direct link to lines

Show Details text can be hidden? (v15 standard feature?)

Nitpicking: When editing/creating PR Lines Purchase Status shows an empty string bubble (not an issue)

Thanks for improving Purchase Request creation!

HviorForgeFlow:
> > > Right now there are 2 places where a similar message is received, when a picking is done and when a move line is mark as done
> > > image
> > > I would drop the stock picking one and keep the one coming from the move line.
> > 
> > 
> > @HviorForgeFlow i can not make this happen, when i validate the income of a purchase order picking only the message of the stock.piking is showed image
> 
> @HviorForgeFlow Hi! If I understood correctly, what you mean before is that those two message should be the same one. If changing that part of the code is going to be complicated, I'd prefer for now to prioritize the migration for now. The messages are pretty similar and appear in different objects (picking and stock request)...

Okey then if it is going to be kept then just fix this:
image

On my side when I validate a picking to done, it creates a message when stock.move.line of some product is validated. Then when stock picking start their action to set to done, it repeats the message. I considered a duplicity, not sure why it is not happening on your side, same behaviour with runboat?

You can see [purchase_request/models/stock_picking.py](https://github.com/OCA/purchase-workflow/pull/1342/files#diff-7079c979984d6272a65b4b8025a33fde47a38fac951590d7eaa40a0191744653R1-R69) _action_done method don't do any operation it is just take information about move_lines in picking and add the message, in my use case again.

Thanks!

HviorForgeFlow (commented):
image

Should be changed to show pickings or keep like this?

purchase_request/models/purchase_request.py
PabloPaezSheridan: @i-vyshnevska thanks irina, all done

PabloPaezSheridan:
> image
> 
> Should be changed to show pickings or keep like this?

I believe is a good idea, i have re-define this method in order to show pickings
https://github.com/OCA/purchase-workflow/blob/430604c5247a43de791b159a60b0c7688776a2f6/purchase_request/models/purchase_request.py#L184

PabloPaezSheridan:
> > > > Right now there are 2 places where a similar message is received, when a picking is done and when a move line is mark as done
> > > > image
> > > > I would drop the stock picking one and keep the one coming from the move line.
> > > 
> > > 
> > > @HviorForgeFlow i can not make this happen, when i validate the income of a purchase order picking only the message of the stock.piking is showed image
> > 
> > 
> > @HviorForgeFlow Hi! If I understood correctly, what you mean before is that those two message should be the same one. If changing that part of the code is going to be complicated, I'd prefer for now to prioritize the migration for now. The messages are pretty similar and appear in different objects (picking and stock request)...
> 
> Okey then if it is going to be kept then just fix this: image
> 
> On my side when I validate a picking to done, it creates a message when stock.move.line of some product is validated. Then when stock picking start their action to set to done, it repeats the message. I considered a duplicity, not sure why it is not happening on your side, same behaviour with runboat?
> 
> You can see [purchase_request/models/stock_picking.py](https://github.com/OCA/purchase-workflow/pull/1342/files#diff-7079c979984d6272a65b4b8025a33fde47a38fac951590d7eaa40a0191744653R1-R69) _action_done method don't do any operation it is just take information about move_lines in picking and add the message, in my use case again.
> 
> Thanks!

I have fix this method to show the correct information
https://github.com/OCA/purchase-workflow/blob/430604c5247a43de791b159a60b0c7688776a2f6/purchase_request/models/stock_picking.py#L11

PabloPaezSheridan:
> image
> 
> Should be changed to show pickings or keep like this?

PabloPaezSheridan:
@HviorForgeFlow i have deleted stock_picking.py in order to not show the duplicate message anymore when validate picking.  Can you give me a hand with the translations? i dont know why but pre commit is showing error in the translations files in a line that i have not touched at all

is over here
https://github.com/OCA/purchase-workflow/blob/883ef1cc91366575311bf28ed0821c966d9027f3/purchase_request/wizard/purchase_request_line_make_purchase_order.py#L55

and the error is :
Translation string couldn't be parsed correctly using string%variables TypeError('not all arguments converted during string formatting'

PabloPaezSheridan:
Well @HviorForgeFlow this PR is now all green (except for the runbot which i believe is not working in v15). If you can give me the +1 i will appreciate as you are one of the dvps of this module. If you can merge it better for us. Thanks a lot for everything

HviorForgeFlow (approved):
LGTM :+1:

nicomacr (approved):
LGTM

HviorForgeFlow:
@sbidoul It seems that runboat are not fetching last push force update code, even if I try to reset the instance manually.
Any idea?

sbidoul:
@HviorForgeFlow each runboat build is bound to a specific commit. So when you click reset, it reinstalls and recreates a fresh database for the exact same commit.

purchase_request/models/purchase_request_line.py
```diff
@@ -0,0 +1,410 @@
...
+                lambda si: not si.company_id or si.company_id == rec.company_id
+            )
+            rec.supplier_id = sellers[0].name if sellers else False
+
+    @api.onchange("product_id")
+    def onchange_product_id(self):
+        if self.product_id:
+            name = self.product_id.name
+            if self.product_id.code:
+                name = "[{}] {}".format(name, self.product_id.code)
```
HviorForgeFlow:
```suggestion
                name = "[{}] {}".format(self.product_id.code, name)
```

purchase_request/views/purchase_request_view.xml
```diff
@@ -0,0 +1,406 @@
...
+                                        groups="analytic.group_analytic_accounting"
+                                    />
+                                    <field name="date_required" />
+                                    <field name="estimated_cost" widget="monetary" />
+                                    <field name="currency_id" invisible="1" />
+                                    <field
+                                        name="company_id"
+                                        groups="base.group_multi_company"
+                                        widget="selection"
+                                    />
```
HviorForgeFlow, line 184: Replace for preferred_supplier
company_id field is a related readonly from the order, it does not give additional information.

purchase_request/views/purchase_request_view.xml
```diff
@@ -0,0 +1,406 @@
...
+                                        widget="badge"
+                                        decoration-success="purchase_state == ('done')"
+                                        decoration-muted="purchase_state == ('draft')"
+                                        decoration-info="purchase_state in ('sent', 'purchase')"
+                                        decoration-warning="purchase_state == ('to_approve')"
+                                        decoration-danger="purchase_state == 'cancelled'"
+                                    />
+                                    <button
+                                        name="action_show_details"
+                                        string="More details"
```
HviorForgeFlow: drop string "More details", let users discover the feature :smile:

purchase_request/views/purchase_request_line_view.xml
```diff
@@ -0,0 +1,651 @@
...
+                            <field
+                                name="date_required"
+                                attrs="{'readonly': [('is_editable','=', False)]}"
+                            />
+                            <field
+                                name="estimated_cost"
+                                widget="monetary"
+                                attrs="{'readonly': [('is_editable','=', False)]}"
+                            />
+                            <field name="currency_id" attrs="{'invisible': True}" />
```
HviorForgeFlow, line 148: Add preferred_supplier to purchase_request_line form view

HviorForgeFlow (changes requested):
Code review

purchase_request/views/purchase_request_view.xml
nicomacr: This fields doesn't exist in the code, you mean create it?

HviorForgeFlow: supplier_id technical name Preferred Supplier as String name

https://github.com/OCA/purchase-workflow/pull/1342/files#diff-3ce1fbd00b838bb7ff1423c460773e0a8c5bb02b2e31a945fdbd132cb8282f11R89-R95

HviorForgeFlow (approved):

lk-eska:
There were problems?

PR #1348, opened 2022-02-02
[15.0][ADD] purchase_receipt_expectation

SilvioC2C:
New module `purchase_receipt_expectation`.

Base module for managing custom workflows for incoming picking creation when a purchase order gets approved.


purchase_receipt_expectation/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2021 Camptocamp SA
```
ajaniszewska-dev:
```suggestion
# Copyright 2022 Camptocamp SA
```

ajaniszewska-dev (approved):

purchase_receipt_expectation/models/purchase_order.py
```diff
@@ -0,0 +1,44 @@
...
+    _inherit = "purchase.order"
+
+    receipt_expectation = fields.Selection(
+        [("automatic", "Automatic")],
+        help="Defines how reception pickings are managed when the order is"
+        " approved.\nDefault value is 'automatic', which means the"
+        " picking will be created following the standard Odoo workflow.\n"
+        "New values added to this selection also require the definition"
+        " of methods that manage such values.\nThese methods must be"
+        " named `_create_picking_for_<value>_receipt_expectation`.",
```
ivantodorovich:
```suggestion
```

This part is too technical.
Help messages are intended for users, not developpers

purchase_receipt_expectation/models/purchase_order.py
```diff
@@ -0,0 +1,44 @@
...
+        default="automatic",
+        required=True,
+    )
+
+    def _create_picking(self):
+        if self.env.context.get("standard_create_picking"):
```
ivantodorovich: Not blocking.
IMO  `standard_create_picking` is too vague, and could potentially be used as a context key in other modules, causing conflicts with this one if both are installed.

I'd go for something like this instead. They it's used is reserved, set False rather than True.

```suggestion
        if not self.env.context.get("handle_receipt_expectation", True):
```

PR #1108, opened 2021-02-11
[14.0][FIX] purchase_discount: Allow move to be False if not assigned

HviorForgeFlow:
Solves https://github.com/OCA/purchase-workflow/issues/1107
Can you confirm it @Alliasys ?

PR #1362, opened 2022-02-15
[FIX] purchase_discount: Fix undefined argument on purchase_discount: original function use move=False and inherited function just use move

rolandojduartem:
The original code caused the following error https://gist.github.com/rolandojduartem/64f4c0775d1eb11ff9c6bcfda02b94be at the moment to create a bill from a purchase order. This code fix that error. The original function is https://github.com/odoo/odoo/blob/14.0/addons/purchase/models/purchase.py#L1143
purchase-wokflow_fix

### DIFF
--- a/purchase_discount/models/purchase_order.py
+++ b/purchase_discount/models/purchase_order.py
@@ -104,8 +104,8 @@ class PurchaseOrderLine(models.Model):
             return
         self.discount = seller.discount
 
-    def _prepare_account_move_line(self, move):
-        vals = super(PurchaseOrderLine, self)._prepare_account_move_line(move)
+    def _prepare_account_move_line(self, move=False):
+        vals = super(PurchaseOrderLine, self)._prepare_account_move_line(move=move)
         vals["discount"] = self.discount
         return vals
 


purchase_discount/models/purchase_order.py
```diff
@@ -104,7 +104,7 @@ def _apply_value_from_seller(self, seller):
             return
         self.discount = seller.discount
 
-    def _prepare_account_move_line(self, move):
+    def _prepare_account_move_line(self, move=False):
         vals = super(PurchaseOrderLine, self)._prepare_account_move_line(move)
```
pedrobaeza:
```suggestion
        vals = super()._prepare_account_move_line(move=move)
```

pedrobaeza (commented):
Thanks for the patch. Pleas sign the CLA.

rolandojduartem:
Hi, @luisg123v, could you review this pull request, please?

purchase_discount/models/purchase_order.py
pedrobaeza: Can you attend this comment?

pedrobaeza (approved):

pedrobaeza:
Oh, there are conflicts.

luisg123v (changes requested):
It seems the fix you're trying to submit was already performed one year ago on #1108.

rolandojduartem:
This pull request  will be closed because the issue was fixed one year ago.

PR #1381, opened 2022-03-15
[IMP] purchase_request: update existing line if any

ajaniszewska-dev:
Problem:
with current module state there are many request lines created for same product and same day.
Aim of this PR is to prevent creating new line if Purchase Request with line for same product and date was found - in that case we update only product_qty on existing line.

### DIFF
--- a/purchase_request/models/stock_rule.py
+++ b/purchase_request/models/stock_rule.py
@@ -112,6 +112,7 @@ class StockRule(models.Model):
                 .search([dom for dom in domain])
                 .filtered(
                     lambda x: procurement.product_id in x.line_ids.mapped("product_id")
+                    and x.purchase_count == 0
                 )
             )
             pr = pr[0] if pr else False
@@ -138,6 +139,11 @@ class StockRule(models.Model):
             [
                 ("product_id", "=", request_line_data["product_id"]),
                 ("date_required", "=", request_line_data["date_required"].date()),
+                (
+                    "purchase_state",
+                    "=",
+                    False,
+                ),  # avoid updating if there is RFQ or PO linked
             ],
         )
         if same_product_date_request_line:
--- a/purchase_request/models/stock_rule.py
+++ b/purchase_request/models/stock_rule.py
@@ -52,9 +52,10 @@ class StockRule(models.Model):
         :return: False
         """
         domain = (
-            ("state", "=", "draft"),
             ("picking_type_id", "=", self.picking_type_id.id),
             ("company_id", "=", values["company_id"].id),
+            ("product_id", "=", values["product_id"].id),
+            ("date_start", "=", values["date_planned"]),
         )
         gpo = self.group_propagation_option
         group_id = (
@@ -101,6 +102,7 @@ class StockRule(models.Model):
         purchase_request_line_model = self.env["purchase.request.line"]
         cache = {}
         pr = self.env["purchase.request"]
+        procurement.values["product_id"] = procurement.product_id
         domain = rule._make_pr_get_domain(procurement.values)
         if domain in cache:
             pr = cache[domain]
@@ -124,4 +126,20 @@ class StockRule(models.Model):
                 pr.write({"origin": procurement.origin})
         # Create Line
         request_line_data = rule._prepare_purchase_request_line(pr, procurement)
-        purchase_request_line_model.create(request_line_data)
+        # check if request has lines for same product and data
+        # if yes, update qty instead of creating new line
+        same_product_date_request_line = pr.line_ids.filtered_domain(
+            [
+                ("product_id", "=", request_line_data["product_id"]),
+                ("date_required", "=", request_line_data["date_required"]),
+            ],
+        )
+        if same_product_date_request_line:
+            new_product_qty = (
+                same_product_date_request_line.product_qty
+                + request_line_data["product_qty"]
+            )
+            same_product_date_request_line.write({"product_qty": new_product_qty})
+        else:
+            # Create Line
+            purchase_request_line_model.create(request_line_data)
--- a/purchase_request/models/stock_rule.py
+++ b/purchase_request/models/stock_rule.py
@@ -1,6 +1,8 @@
 # Copyright 2018-2020 ForgeFlow, S.L.
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl-3.0)
 
+from datetime import datetime
+
 from odoo import api, fields, models
 
 
@@ -54,8 +56,7 @@ class StockRule(models.Model):
         domain = (
             ("picking_type_id", "=", self.picking_type_id.id),
             ("company_id", "=", values["company_id"].id),
-            ("product_id", "=", values["product_id"].id),
-            ("date_start", "=", values["date_planned"]),
+            ("date_start", "=", datetime.today().date()),
         )
         gpo = self.group_propagation_option
         group_id = (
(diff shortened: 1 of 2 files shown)


purchase_request/models/stock_rule.py
```diff
@@ -124,4 +126,22 @@ def create_purchase_request(self, procurement_group):
                 pr.write({"origin": procurement.origin})
         # Create Line
         request_line_data = rule._prepare_purchase_request_line(pr, procurement)
+        # check if request has lines for same product and data
+        # if yes, update qty instead of creating new line
+        same_product_date_request_line = pr.line_ids.search(
+            [
+                ("product_id.id", "=", request_line_data["product_id"]),
```
simahawk:
```suggestion
                ("product_id", "=", request_line_data["product_id"]),
```

purchase_request/models/stock_rule.py
```diff
@@ -124,4 +126,22 @@ def create_purchase_request(self, procurement_group):
...
+        )
+        if same_product_date_request_line:
+            new_product_qty = (
+                same_product_date_request_line.product_qty
+                + request_line_data["product_qty"]
+            )
+            same_product_date_request_line.write({"product_qty": new_product_qty})
+        else:
+            # Create Line
+            purchase_request_line_model.create(request_line_data)
```
simahawk, line 157: the same line will be created again right after, no?

purchase_request/models/stock_rule.py
```diff
@@ -52,9 +52,10 @@ def _make_pr_get_domain(self, values):
         :return: False
         """
         domain = (
-            ("state", "=", "draft"),
+            ("state", "=", "approved"),
```
simahawk: hmm are we sure about this change? Shouldn't you select both states? Or exclude approved?

simahawk (changes requested):
@ajaniszewska-dev on your next push can you please rewrite the commit to:
1. be short and move explanation to the description
2. remove the odoo version

purchase_request/models/stock_rule.py
gurneyalex: I tend to agree here

purchase_request/models/stock_rule.py
```diff
@@ -52,9 +52,10 @@ def _make_pr_get_domain(self, values):
...
             ("picking_type_id", "=", self.picking_type_id.id),
             ("company_id", "=", values["company_id"].id),
+            ("product_id", "=", values["product_id"].id),
```
gurneyalex: this is not a stored field, you will not be able to search it. You need to search the lines.product_id

purchase_request/models/stock_rule.py
ajaniszewska-dev: ah yes, i forgot to remove line 147, will do it with commit update.

purchase_request/models/stock_rule.py
```diff
@@ -124,4 +126,22 @@ def create_purchase_request(self, procurement_group):
                 pr.write({"origin": procurement.origin})
         # Create Line
         request_line_data = rule._prepare_purchase_request_line(pr, procurement)
+        # check if request has lines for same product and data
+        # if yes, update qty instead of creating new line
+        same_product_date_request_line = pr.line_ids.search(
```
gurneyalex: WARNING: Model::search is a model method not a recordset method. if you want to filter a recordset with a restriction, use Model::filtered_domain
```suggestion
        same_product_date_request_line = pr.line_ids.filtered_domain(
```

purchase_request/models/stock_rule.py
gurneyalex: yes needs fixing too.

gurneyalex (changes requested):
please add an automated test.

purchase_request/models/stock_rule.py
ajaniszewska-dev: @gurneyalex  correct me if i am wrong but at this stage of request creation lines are not available yet i think?

purchase_request/tests/test_purchase_request_procurement.py
```diff
@@ -110,3 +113,39 @@ def test_procure_purchase_request(self):
...
+            "Test Purchase Request Single Line",
+            "Test Purchase Request Single Line",
+            self.product_1,
+            5,
+        )
+        self.assertTrue(has_route)
+        self.env["procurement.group"].run_scheduler()
+        # make sure no new PR is created
+        self.assertEqual(self.env["purchase.request"].search_count([]),1)
+        self.assertEqual(len(pr.line_ids), 1)
```
ajaniszewska-dev, line 149: @simahawk @gurneyalex  any idea why it creates 2nd line here even if it should not?  Getting `AssertionError: 2 != 1`. Not sure if whole TC scenario is correct but looking at the model code, there should be procurement handling in the background - didn't find better starting point.

//edit:
problem with date in filtered_domain, doesn't pass in the test. Maybe due to FakeDate used by freeze_time?

simahawk: wild guess: the procurement group date is not frozen?

ajaniszewska-dev: @simahawk in filtered domain i was comparing datetime.datetime with datetime.date, just fixed by extracting date(), let's see.

```
-> if same_product_date_request_line:
(Pdb) pr.line_ids
purchase.request.line(72,)
(Pdb) pr.line_ids.date_required
datetime.date(2022, 3, 16)
(Pdb) request_line_data["date_required"]
datetime.datetime(2022, 3, 16, 8, 15, 14)
```

purchase_request/models/stock_rule.py
```diff
@@ -52,9 +54,9 @@ def _make_pr_get_domain(self, values):
         :return: False
         """
         domain = (
-            ("state", "=", "draft"),
             ("picking_type_id", "=", self.picking_type_id.id),
             ("company_id", "=", values["company_id"].id),
+            ("date_start", "=", datetime.today().date()),
```
mmequignon, line 59: not blocking, but it's generally better to use `fields.Date.today()` instead of `datetime.datetime.today()`

purchase_request/models/stock_rule.py
```diff
@@ -105,7 +107,13 @@ def create_purchase_request(self, procurement_group):
         if domain in cache:
             pr = cache[domain]
         elif domain:
-            pr = self.env["purchase.request"].search([dom for dom in domain])
+            pr = (
+                self.env["purchase.request"]
+                .search([dom for dom in domain])
```
mmequignon, line 112: Why the list comprehension?

purchase_request/models/stock_rule.py
```diff
@@ -105,7 +107,13 @@ def create_purchase_request(self, procurement_group):
...
+                .filtered(
+                    lambda x: procurement.product_id in x.line_ids.mapped("product_id")
+                )
```
mmequignon: could we add this condition in the domain instead?
`("line_ids.product_id", "in", procurement_product_id.ids)` or something like that

simahawk: If think it's better to filter on the result unless you have a join on line_ids which is probably not the case

purchase_request/models/stock_rule.py
simahawk: I'm still not sure about this domain.... let's talk about it on the next iteration.

mmequignon:
to close, prior to https://github.com/OCA/purchase-workflow/pull/1451

PR #1392, opened 2022-03-24
[14.0][FIX] vendor_transport_lead_time: fix error

sebalix:
When coming from an onchange, the 'order_id.date_order' could be False, even if there is a value in it.

Ref. 3468


sebalix:
Tried to add the "order_id.date_order" in the compute above, but this doesn't solve the issue.

TDu (approved):

flotho (approved):
code review
LGTM

PR #1396, opened 2022-03-28
purchase_propagate_qty: support 2 steps propagation

sebastienbeau:
@sebalix @TDu @simahawk @gurneyalex

### DIFF
--- a/purchase_propagate_qty/models/stock_move.py
+++ b/purchase_propagate_qty/models/stock_move.py
@@ -10,6 +10,8 @@ class StockMove(models.Model):
 
     def _deduce_qty(self, qty_to_remove, rounding):
         """Deduce the provided qty with respect to done qties."""
+        if self.move_dest_ids:
+            self.move_dest_ids._deduce_qty(qty_to_remove, rounding)
         self = self.with_context(do_not_unreserve=True)
         for move in self:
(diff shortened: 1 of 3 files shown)


purchase_propagate_qty/models/stock_move.py
```diff
@@ -10,6 +10,8 @@ class StockMove(models.Model):
 
     def _deduce_qty(self, qty_to_remove, rounding):
         """Deduce the provided qty with respect to done qties."""
+        if self.move_dest_ids:
```
simahawk, line 13: wouldn't be better to explicitly check for the two_steps settings?

simahawk (changes requested):
@sebastienbeau tnx for you contrib.

Please, rewrite the commit msg to include the module name, like: 

`purchase_propagate_qty: support 2 steps propagation`

Thanks :pray: 

FTR there are also Odoo guidelines regarding commits https://www.odoo.com/documentation/15.0/developer/misc/other/guidelines.html#commit-message-structure

Personally I think that the "tag" like "[ADD]" or "[IMP]" is a waste of chars IF you write the commit msg properly w/ the right verb / word (such as "add something") but that's probably a matter of taste :wink:

PR #1399, opened 2022-03-29
[14.0][ADD] purchase_request_group_by_date

ajaniszewska-dev:


purchase_request_group_by_date/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+    "summary": "Use this module to group request for same product by date",
+    "website": "https://github.com/OCA/purchase-workflow",
+    "category": "Purchase Management",
+    "depends": ["purchase_request", "stock", "product", "purchase_stock"],
+    "data": [
+        "views/res_config_settings.xml",
+    ],
+    "license": "LGPL-3",
+    "installable": True,
+    "application": True,
```
mmequignon:
```suggestion
```

purchase_request_group_by_date/models/res_company.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2022 Camptocamp
```
mmequignon: I don't think this is required. If we don't want to group PR, then we uninstall the module.

purchase_request_group_by_date/tests/test_purchase_request_group_by.py
```diff
@@ -0,0 +1,183 @@
...
+            "Test Purchase Request Single Line",
+            "Test Purchase Request Single Line",
+            self.product_1,
+            5,
+        )
+        self.assertTrue(has_route)
+        self.env["procurement.group"].run_scheduler()
+        # make sure no new PR is created
+        self.assertEqual(self.env["purchase.request"].search_count([]), 1)
+        self.assertEqual(len(pr.line_ids), 1)
```
ajaniszewska-dev, line 112:
```
        # make sure no new PR is created
        self.assertEqual(self.env["purchase.request"].search_count([]), 1)
        import pdb;pdb.set_trace();
>       self.assertEqual(len(pr.line_ids), 1)
E       AssertionError: 2 != 1

odoo/external-src/purchase-workflow/purchase_request_group_by_date/tests/test_purchase_request_group_by.py:113: AssertionError

```

ajaniszewska-dev: afais in model:
```
(Pdb) self.company_group_purchase_request
False
```

so parameter isn't properly set in test case, that's why mechanism doesnt work.

ajaniszewska-dev: @simahawk  if you would have some time, can you take a look here? Issue is that i am moving what we did recently for gzt to another oca module where it can be enabled/diabled by parameter. But test fails due to assertion which is caused by fact that travis doesnt see flag set on the company (i tested manually in UI and it seems to works correctly). 
I am not sure if it is enough to do in test:     `self.company_id.group_purchase_request = True ` like it is in l.28 or i should simulate it with some res.config records?

purchase_request_group_by_date/models/stock_rule.py
```diff
@@ -0,0 +1,98 @@
...
+                ("date_start", "=", datetime.today().date()),
+            )
+            gpo = self.group_propagation_option
+            group_id = (
+                (gpo == "fixed" and self.group_id.id)
+                or (gpo == "propagate" and values["group_id"].id)
+                or False
+            )
+            if group_id:
+                domain += (("group_id", "=", group_id),)
```
mmequignon: better append in the domain here, so if the `_make_pr_get_domain` changes, we don't have to modify this module as well.
```suggestion
            domain += ("date_start", "=", datetime.today().date())
            gpo = self.group_propagation_option
            group_id = (
                (gpo == "fixed" and self.group_id.id)
                or (gpo == "propagate" and values["group_id"].id)
                or False
            )
            if group_id:
                domain += (("group_id", "=", group_id),)
```

purchase_request_group_by_date/models/stock_rule.py
```diff
@@ -0,0 +1,98 @@
...
+        if self.company_group_purchase_request:
+            procurement = procurement_group[0]
+            rule = procurement_group[1]
+            purchase_request_model = self.env["purchase.request"]
+            purchase_request_line_model = self.env["purchase.request.line"]
+            cache = {}
+            pr = self.env["purchase.request"]
+            domain = rule._make_pr_get_domain(procurement.values)
+            if domain in cache:
+                pr = cache[domain]
```
mmequignon: cache is empty here. I don't understand this.

purchase_request_group_by_date/models/stock_rule.py
```diff
@@ -0,0 +1,98 @@
...
+                (gpo == "fixed" and self.group_id.id)
+                or (gpo == "propagate" and values["group_id"].id)
+                or False
+            )
+            if group_id:
+                domain += (("group_id", "=", group_id),)
+        return domain
+
+    # override to change existing pr handling
+    def create_purchase_request(self, procurement_group):
```
mmequignon, line 24: It's quite hard to find out what has changed in this method.
Why not splitting the method in `purchase_request` and override just what you need?
Would be easier to understand, I think.

ajaniszewska-dev: i'd prefer to not go too deep with modification, esp. in original module.
in case of _make_pr_get_domain i think due to method construction is it clear what we want to change. 
things get bit worse in create_purchase_request method but crucial part of method is imo well commented.

purchase_request_group_by_date/tests/test_purchase_request_group_by.py
```diff
@@ -0,0 +1,228 @@
...
+                "prod_1",
+                "prod_1",
+                self.product_1,
+                5,
+            )
+            self.assertTrue(has_route)
+            self.env["procurement.group"].run_scheduler()
+            pr = self.env["purchase.request"].search([("origin", "=", "prod_1")])
+            self.product_1.purchase_request = True
+            self.assertTrue(pr.to_approve_allowed)
```
ajaniszewska-dev: to check:
```
line 173, in test_existing_purchase_request

    self.assertTrue(pr.to_approve_allowed)

AssertionError: False is not true
```

ajaniszewska-dev: to check:
```
line 173, in test_existing_purchase_request

    self.assertTrue(pr.to_approve_allowed)

AssertionError: False is not true
```

purchase_request_group_by_date/tests/test_purchase_request_group_by.py
```diff
@@ -0,0 +1,226 @@
...
+            self.env["procurement.group"].run_scheduler()
+            pr = self.env["purchase.request"].search(
+                [("origin", "=", "Test Purchase Request Single Line, prod_1")]
+            )
+            self.product_1.purchase_request = True
+            # self.assertTrue(pr.to_approve_allowed)
+            self.assertEqual(pr.origin, "Test Purchase Request Single Line, prod_1")
+            prl = self.env["purchase.request.line"].search([("request_id", "=", pr.id)])
+            self.assertEqual(len(prl), 2)
+            self.assertEqual(sum(pr.line_ids.mapped("product_qty")), 9)
```
ajaniszewska-dev, line 179: @mmequignon  to make travis happy i had to modify this TC to calculate with sum of quantities instead of treating them separately. Do you have any idea why it didn't pass initially? this is mainly copy & paste of code from your PR inside new addon, no idea what causes different result :confused:

ajaniszewska-dev:
@mmequignon ping :)
pls check changes in this PR when you have some time (not urgent).

yankinmax (approved):

purchase_request_group_by_date/__manifest__.py
```diff
@@ -0,0 +1,15 @@
...
+
+{
+    "name": "Purchase Request Group by Date",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "version": "14.0.1.0.0",
+    "summary": "Use this module to group request for same product by date",
+    "website": "https://github.com/OCA/purchase-workflow",
+    "category": "Purchase Management",
+    "depends": ["purchase_request", "stock", "product", "purchase_stock"],
+    "data": [],
```
TDu, line 12: Can be removed

purchase_request_group_by_date/models/stock_rule.py
```diff
@@ -0,0 +1,82 @@
...
+    # override method from purchase_request to change existing pr handling
+    def create_purchase_request(self, procurement_group):
+        procurement = procurement_group[0]
+        rule = procurement_group[1]
+        purchase_request_model = self.env["purchase.request"]
+        purchase_request_line_model = self.env["purchase.request.line"]
+        cache = {}
+        pr = self.env["purchase.request"]
+        domain = rule._make_pr_get_domain(procurement.values)
+        if domain in cache:
```
TDu, line 32: Cache is empty at this stage, should be removed ?

purchase_request_group_by_date/models/stock_rule.py
```diff
@@ -0,0 +1,82 @@
...
+            pr = cache[domain]
+        elif domain:
+            pr = (
+                self.env["purchase.request"]
+                .search([dom for dom in domain])
+                .filtered(
+                    lambda x: procurement.product_id in x.line_ids.mapped("product_id")
```
TDu, line 39: Is it not possible to search on `purchase.request.line` and avoid doing a search with a filter ?

purchase_request_group_by_date/models/stock_rule.py
```diff
@@ -0,0 +1,82 @@
...
+                    and x.purchase_count == 0
+                )
+            )
+            pr = pr[0] if pr else False
+            cache[domain] = pr
```
TDu, line 44: This cache seems to never be used.

purchase_request_group_by_date/models/stock_rule.py
```diff
@@ -0,0 +1,82 @@
...
+            pr = purchase_request_model.create(request_data)
+            cache[domain] = pr
+        elif not pr.origin or procurement.origin not in pr.origin.split(", "):
+            if pr.origin:
+                if procurement.origin:
+                    pr.write({"origin": pr.origin + ", " + procurement.origin})
+                else:
+                    pr.write({"origin": pr.origin})
+            else:
+                pr.write({"origin": procurement.origin})
```
TDu, line 58: This is a copy paste from the dependency, I wonder if there is not a better way to handle all this, and including a call to the overriden function.

purchase_request_group_by_date/models/stock_rule.py
```diff
@@ -0,0 +1,82 @@
...
+                ("product_id", "=", request_line_data["product_id"]),
+                ("date_required", "=", request_line_data["date_required"].date()),
+                (
+                    "purchase_state",
+                    "=",
+                    False,
+                ),  # avoid updating if there is RFQ or PO linked
+            ],
+        )
+        if same_product_date_request_line:
```
TDu, line 74: Could we have multiple lines, which would multiple line here ?
Which would multiply the order with the write below ?

PR #1407, opened 2022-04-05
[14.0] purchase_request: Fix wrong method name

mmequignon:
In 14.0, the `action_done` method has been renamed `_action_done`


simahawk (approved):
which means there's no test running this piece of code... I guess? :cry:

mmequignon:
> which means there's no test running this piece of code... I guess? cry

I think so…

PR #1409, opened 2022-04-06
14.0 Imp last purchase price with discount

michotm:
when the modute purchase_discount is installed the las purchase price doesn't include the discount that why I propose to compute the last puchase price an other way from the line total and the qty


purchase_last_price_info/models/product_product.py
```diff
@@ -53,7 +53,10 @@ def _compute_last_purchase_line_id(self):
     @api.depends("last_purchase_line_id")
     def _compute_last_purchase_line_id_info(self):
         for item in self:
-            item.last_purchase_price = item.last_purchase_line_id.price_unit
+            item.last_purchase_price = (
+                item.last_purchase_line_id.price_subtotal
+                / item.last_purchase_line_id.product_qty
```
sebastienbeau: Becarefull we should take care of the case where item.last_purchase_line_id.product_qty is 0
The best is to change the domain of last_purchase_line_ids

```

    last_purchase_line_ids = fields.One2many(
        comodel_name="purchase.order.line",
        inverse_name="product_id",
        domain=[("state", "in", ["purchase", "done"]), ("product_qty", ">", 0)],
        string="Last Purchase Order Lines",
)

```

flotho (approved):
LGTM , agree with @sebastienbeau

flotho (approved):
LGTM

flotho:
Hi @sebastienbeau , code has been refactored, any chance to review ?

PR #1410, opened 2022-04-06
[14.0][ADD] purchase_order_price_recalculation: New addon.

victoralmau:
Module that adds a button to recalculate prices of lines in purchase orders.

@Tecnativa TT33369


purchase_order_price_recalculation/models/purchase_order.py
```diff
@@ -0,0 +1,26 @@
...
+            vals = line._convert_to_write(line.read()[0])
+            if "product_tmpl_id" in line._fields:
+                vals["product_tmpl_id"] = line.product_tmpl_id
+            # we make this to isolate changed values:
+            line2 = self.env["purchase.order.line"].new(vals)
+            line2.env.add_to_compute(
+                line2.product_id._fields["price"], line2.product_id
+            )
+            line2.onchange_product_id()
+            line.write({field: line2[field] for field in fields})
```
chienandalu: I'd say a better way to recalculate the lines would be to just trigger `_onchange_quantity` line by line. This way, you'll get the whole pack: discounts, new reception days and anything that could come from the supplierinfo.

https://github.com/odoo/odoo/blob/421e9d9100e9593999c9490342070a81b1786078/addons/purchase/models/purchase.py#L1071

chienandalu (approved):
Tested :+1: Thanks!

purchase_order_price_recalculation/README.rst
```diff
@@ -0,0 +1,101 @@
...
+.. |badge4| image:: https://img.shields.io/badge/weblate-Translate%20me-F47D42.png
+    :target: https://translation.odoo-community.org/projects/purchase-workflow-14-0/purchase-workflow-14-0-purchase_order_price_recalculation
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/142/14.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+This module add one button on purchase orders (below sale order lines) that:
```
pedrobaeza:
```suggestion
This module adds one button on purchase orders (below the lines) that:
```

purchase_order_price_recalculation/README.rst
```diff
@@ -0,0 +1,101 @@
...
+
+* Recalculates the prices of the order lines that contain a product in them.
+
+It is launched manually as a button to get the user decide if he/she wants to
```
pedrobaeza:
```suggestion
It is launched manually as a button to get the user to decide if he/she wants to
```

purchase_order_price_recalculation/models/purchase_order.py
```diff
@@ -0,0 +1,27 @@
...
+    @api.onchange("partner_id")
+    def _onchange_partner_id(self):
+        if (
+            self.partner_id
+            and self._origin.partner_id
+            and self._origin.partner_id != self.partner_id
+        ):
+            self.show_update_line_info = True
+        else:
+            self.show_update_line_info = False
```
pedrobaeza: This is anyways not covering all cases. Maybe it's better to simply mark as True on the onchange.

```suggestion
    def _onchange_partner_id(self):
        self.show_update_line_info = (
            self.partner_id
            and self._origin.partner_id
            and self._origin.partner_id != self.partner_id
        )
```

purchase_order_price_recalculation/models/purchase_order.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 2022 Tecnativa - Víctor Martínez
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+from odoo import api, fields, models
+
+
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+
+    show_update_line_info = fields.Boolean()
```
pedrobaeza:
```suggestion
    show_update_line_info = fields.Boolean(store=False)
```

pedrobaeza (changes requested):

purchase_order_price_recalculation/models/purchase_order.py
```diff
@@ -0,0 +1,24 @@
...
+    _inherit = "purchase.order"
+
+    show_update_line_info = fields.Boolean(store=False)
+
+    @api.onchange("partner_id")
+    def _onchange_partner_id(self):
+        self.show_update_line_info = (
+            self.partner_id
+            and self._origin.partner_id
+            and self._origin.partner_id != self.partner_id
```
pedrobaeza: What about my suggestion of always marking this as True when changed?

legalsylvain (changes requested):
Hi. Thanks for sharing this module.

Functional review : 
- I don't see the interest to display the button only when the supplier changed. The user can want to update prices in differents moments.
- Also if the user duplicate the PO, and then save it, the button disappear (so, if the user want to update the price, he has to cancel the current PO, duplicate it again)
-> Why not do the same as for other OCA modules ? (for tax recomputation, for pricelist recomputation, etc...) Just having a button, and if the user click on it, it recompute the purchase prices. There is a warning on the button, so the user can't do it by chance.

Other point (smaller) : the other OCA modules regarding recomputation put the button in the ``<header>`` and it seems odoo doesn't use anymore inline buttons. I so think classic button in header is a better choice.

regards.

purchase_order_price_recalculation/models/purchase_order.py
victoralmau: In my opinion, it should not be set to True in the onchange because that would cause the button to be displayed when a record is created (unnecessary).

victoralmau:
I have placed the button in the header.
Odoo core in sales does a similar thing by displaying the button only when needed. https://github.com/odoo/odoo/blob/14.0/addons/sale/views/sale_views.xml#L334-L344

pedrobaeza:
@legalsylvain do you agree then?

legalsylvain:
>  Odoo core in sales does a similar thing by displaying the button only when needed. https://github.com/odoo/odoo/blob/14.0/addons/sale/views/sale_views.xml#L334-L344

OK. But do you think it's a great feature ? I mean : 
- it make the code more complex, because it requires an extra field
- it makes the button disappear in situations where we would like it to remain displayed. (specially if user duplicates PO and saves the PO).

What is the interest ? Or said differently, what is the problem with the simplification I propose ?

victoralmau:
In my opinion I think it's a good idea to follow the same example of odoo and show the button only when it is necessary, showing the buttons always I think is not very functional.

Anyway, it's just my opinion, what do you think @pedrobaeza ?

pedrobaeza:
Although I tend to homogenize the behavior with core functionality, if there's a functional gap as @legalsylvain suggested, we should be feature-complete. Is such case happening?

victoralmau:
I have reviewed the use case suggested by @legalsylvain and it would be correct. What do you think?

example

pedrobaeza:
I think he's saying that if you update prices outside the PO (for example, changing product.supplierinfo), you are not able to press on recompute for getting the new prices, as no apparent change has been made on the PO. I tend to agree with that, but the killer one is that it simplifies a lot the code. The only cons is that the button is appearing all the time and can confuse people thinking they need to recompute. That's the main reason I think why Odoo has made all this stuff.

purchase_order_price_recalculation/views/purchase_order_view.xml
```diff
@@ -0,0 +1,19 @@
...
+        <field name="model">purchase.order</field>
+        <field name="inherit_id" ref="purchase.purchase_order_form" />
+        <field name="arch" type="xml">
+            <button name="button_unlock" position="after">
+                <button
+                    name="update_lines_info"
+                    type="object"
+                    string=" Update lines info"
+                    icon="fa-refresh"
+                    confirm="This will update all line information based on the currently set vendor."
```
pedrobaeza:
```suggestion
                    confirm="This will update all the line prices based on the vendor currently set. Do you want to continue?"
```

pedrobaeza (approved):

PR #1418, opened 2022-04-12
[13.0][FIX] purchase_propagate_qty: Take correct quantity if propagate_uom is active

HviorForgeFlow:
CC @ForgeFlow


rafaelbn (approved):
LGTM 👍🏼 

@gurneyalex would you like to review as contributor of the module?

PR #1420, opened 2022-04-13
[15.0] [ADD] purchase_order_no_zero_price

Nikul-OSI:
@dreispt can you please take a look?


purchase_order_no_zero_price/models/purchase_order_line.py
```diff
@@ -0,0 +1,25 @@
...
+    _inherit = "purchase.order.line"
+
+    @api.constrains("price_unit", "display_type", "state")
+    def _check_price_unit_zero(self):
+        # Prevent zero prices on confirmed POs
+        # to help ensure proper Stock Valuation
+        # when receiving the products
+        lines = self.filtered(
+            lambda x: not x.price_unit
+            and x.discount != 100.0  # Allow zero proce for 100% discounts
```
legalsylvain: you could write a getattr(x, "discount", False). That way : 
- you don't have to set a dependency to ``purchase_discount``.
- you could make the module working with ``purchase_triple_discount`` writing similar code for ``discount2`` and ``discount3``.
-

legalsylvain (changes requested):
you have two modules proposed in the PR. 
could you remove the other one ? 
also a proposal inline.

LGTM otherwise, thanks !

purchase_order_no_zero_price/models/purchase_order_line.py
dreispt: This has actually been deployed along with `purchase_triple_discount`, and that's not a bad idea.

dreispt (approved):

purchase_order_no_zero_price/models/purchase_order_line.py
dreispt: @legalsylvain Done, please review.

purchase_order_no_zero_price/models/purchase_order_line.py
```diff
@@ -0,0 +1,30 @@
...
+        # to help ensure proper Stock Valuation
+        # when receiving the products
+        # Also supports `purchase_triple_discount` module
+        no_price_lines = self.filtered(
+            lambda x: not x.display_type
+            and x.state not in ["draft", "done", "cancel"]
+            and not x.price_unit
+            # Allow zero price for 100% discounts:
+            and (
+                x.discount + getattr(x, "discount2", 0.0) + getattr(x, "discount3", 0.0)
```
legalsylvain: if you replace : 

x.discount + getattr(x, "discount2", 0.0) + getattr(x, "discount3", 0.0)

by 
getattr(x, "discount", 0.0) + getattr(x, "discount2", 0.0) + getattr(x, "discount3", 0.0)

you can change the dependency module from ``purchase_discount`` to ``purchase`` and so make the module usable in more case.

purchase_order_no_zero_price/models/purchase_order_line.py
```diff
@@ -0,0 +1,30 @@
...
+            )
+            != 100.0
```
legalsylvain: that algorighm is wrong. 
you should test if any discount is different from 100%.

discount 50 + discount 20 + discount 30 != discount 100%

discount 50 + discount 20 + discount 30 = discount 88%

purchase_order_no_zero_price/models/purchase_order_line.py
```diff
@@ -19,7 +19,9 @@ def _check_price_unit_zero(self):
             and not x.price_unit
             # Allow zero price for 100% discounts:
             and (
-                x.discount + getattr(x, "discount2", 0.0) + getattr(x, "discount3", 0.0)
+                getattr(x, "discount", 0.0)
+                + getattr(x, "discount2", 0.0)
+                + getattr(x, "discount3", 0.0)
             )
             != 100.0
```
dreispt:
```
            and getattr(x, "discount", 0.0) != 100.0
            and getattr(x, "discount2", 0.0) != 100.0
            and getattr(x, "discount3", 0.0) != 100.0
```

Nikul-OSI:
@dreispt Here test case failed because in `purchase_request` module have the test case and changed the field value. When it's changed that time affected our module method.

purchase_order_no_zero_price/readme/USAGE.rst
```diff
@@ -0,0 +1,6 @@
+To use:
+
+- Create an RFQ, and add a line with zero price.
+- Confirm the RFQ, an error message will prevent that ans ask you to enter the missing
```
sswapnesh:
```suggestion
- Confirm the RFQ, an error message will prevent that and ask you to enter the missing
```

dreispt:
@Nikul-OSI This look good, and tests were fixed.
@legalsylvain @sswapnesh Please review.

legalsylvain (approved):
thanks.

sswapnesh (approved):

dreispt:
I think this needs a global setting, so that it can be enabled only when starting the module's tests.
Based on other feedback, a setting per Product might be useful. 
But at this point I'm not sure I'll be deploying this module, so I'm pausing effort on it.

sswapnesh:
@dreispt 
>  Based on other feedback, a setting per Product might be useful.

Indeed, That will be good option.

 > But at this point I'm not sure I'll be deploying this module, so I'm pausing effort on it.

I think it is just adding about a field on product and modifying domain on _check_price_unit_zero
(Not sure about test case efforts)

angelmoya (approved):
Code review

HviorForgeFlow:
@Nikul-OSI seems that other modules test got conflict with this module. Can you please take a look?

Nikul-OSI:
@HviorForgeFlow tests fixed, Can you please recheck?

AaronHForgeFlow (approved):
LGTM

PR #1423, opened 2022-04-13
[15.0] [IMP] purchase_default_terms_conditions: changed field type Text to Html on note

Nikul-OSI:
@dreispt can you please take a look?

### DIFF
--- a/purchase_default_terms_conditions/models/res_company.py
+++ b/purchase_default_terms_conditions/models/res_company.py
@@ -8,6 +8,6 @@ class ResCompany(models.Model):
 
     _inherit = "res.company"
 
-    purchase_note = fields.Text(
+    purchase_note = fields.Html(
         string="Purchase Default Terms and Conditions", translate=True
     )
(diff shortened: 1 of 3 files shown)


dreispt (approved):

sswapnesh (approved):

legalsylvain (changes requested):
don't you have to set ``widget="html"`` in related views ?

dreispt:
@legalsylvain Adding the widget is redundant
@Nikul-OSI Please fix pre-commit, thanks

Nikul-OSI:
@dreispt Fixed pre-commit.

purchase_default_terms_conditions/views/res_partner_views.xml
```diff
@@ -7,7 +7,7 @@
         <field name="priority">36</field>
         <field name="arch" type="xml">
             <group name="purchase" position="inside">
-                  <field name="purchase_note" />
+                  <field name="purchase_note" widget="html" />
```
dreispt: @Nikul-OSI the widget is redundant as the field is already of type Html

PR #1428, opened 2022-04-19
[13.0][ADD] purchase_v12_control_menu: New module

chienandalu:
This module recovers the Purchase control menu as it was in the former version for those who find it handy

cc @Tecnativa TT35990


purchase_control_menu/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,3 @@
+* `Tecnativa <https://www.tecnativa.com>`__:
+
+  * David Vidal
```
legalsylvain: as all the effective code comes from odoo, I think we should : 
- set it as contributors
- keep the license to LGPL-3.

Don't you think ?

purchase_control_menu/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module recovers the Purchase control menu as it was in the former version.
```
legalsylvain: could you make add a more descriptive text. For newcomers in V13 (and then V14...) it doesn't mean much.

legalsylvain (changes requested):
a legal remark and a suggestion. Otherwise, LGTM.

Thanks !

pedrobaeza:
The module name should be something more like `purchase_picking_control_menu` or similar, as the other is too generic.

chienandalu:
Well, it recovers the whole menu, with the supplier invoices option, which was on base, thus the generic name...

pedrobaeza:
purchase_v12_control_menu ?

purchase_control_menu/readme/CONTRIBUTORS.rst
chienandalu: Ok, I changed it in the manifest and readme regenerated

purchase_control_menu/readme/DESCRIPTION.rst
chienandalu: Right, I developed it a little more

chienandalu:
Changes done

pedrobaeza (approved):
@legalsylvain is it OK for you?

legalsylvain (approved):
Well regarding the technical name, maybe it will make no sense when we will migrate it in V20 ! But i dont have a better idea, so let's move on !

chienandalu:
The merge failed. Maybe the branch CIs need to be updated?

PR #1430, opened 2022-04-19
[14.0][MIG] purchase_order_analytic_search

JuanyDForgeflow:
@ForgeFlow 

Organizations often require to quickly find the purchase orders associated to
an analytic account, searching by it's code, name or project/account manager.

This module introduces the possibility to search purchase orders by analytic
account.

It also introduces a new menu entry in Purchasing to list purchase order lines.


AaronHForgeFlow (approved):
Functional review :+1: 

ping @HviorForgeFlow

purchase_order_analytic_search/README.rst
```diff
@@ -0,0 +1,90 @@
...
+    :target: http://www.gnu.org/licenses/lgpl-3.0-standalone.html
+    :alt: License: LGPL-3
+.. |badge3| image:: https://img.shields.io/badge/github-OCA%2Fpurchase--workflow-lightgray.png?logo=github
+    :target: https://github.com/OCA/purchase-workflow/tree/12.0/purchase_order_analytic_search
+    :alt: OCA/purchase-workflow
+.. |badge4| image:: https://img.shields.io/badge/weblate-Translate%20me-F47D42.png
+    :target: https://translation.odoo-community.org/projects/purchase-workflow-12-0/purchase-workflow-12-0-purchase_order_analytic_search
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/142/12.0
```
angelmoya: Check version 14.0

purchase_order_analytic_search/models/purchase_order.py
```diff
@@ -0,0 +1,53 @@
...
+        if not value:
+            return [("id", "=", False)]
+        if isinstance(value, (tuple, list)):
+            # we are searching on a list of ids
+            domain = [("order_id", "!=", False), ("account_analytic_id", "in", value)]
+        else:
+            if isinstance(value, int):
+                # we are searching on the id of the analytic_account
+                domain = [
+                    ("order_id", "!=", False),
```
angelmoya: Why you filter by order_id is not False? order_id is required, so you can delete this line

purchase_order_analytic_search/models/purchase_order.py
```diff
@@ -0,0 +1,53 @@
...
+        for purchase in self:
+            purchase.account_analytic_ids = purchase.mapped(
+                "order_line.account_analytic_id"
+            )
+
+    @api.model
+    def _search_analytic_accounts(self, operator, value):
+        po_line_obj = self.env["purchase.order.line"]
+        if not value:
+            return [("id", "=", False)]
```
angelmoya: Why you return this domain? Maybe you want to get purchase order without account analytic

purchase_order_analytic_search/views/purchase_order_view.xml
```diff
@@ -0,0 +1,77 @@
...
+#---------------------------------------------------------------------------------------------------------
+#   Extend menus
+#---------------------------------------------------------------------------------------------------------
+
+    <record id="purchase_line_form_action_3" model="ir.actions.act_window">
+        <field name="name">Purchase Order Lines</field>
+        <field name="type">ir.actions.act_window</field>
+        <field name="res_model">purchase.order.line</field>
+        <field name="view_mode">tree,form</field>
+        <field name="search_view_id" ref="purchase_order_line_search_analytic" />
```
angelmoya: Why do you create this new menu.

In runboat I checked that we have two menu items with same functions.

angelmoya (changes requested):
Small remarks, please check menu items.

purchase_order_analytic_search/views/purchase_order_view.xml
AaronHForgeFlow: I think this can be removed

purchase_order_analytic_search/models/purchase_order.py
AaronHForgeFlow: I think the standard behavior when there is not value is not to filter by anything. If filtering by no analytic is needed better to create a specific filter in the view.

AaronHForgeFlow:
Hi @angelmoya can you provide some feedback? Some of your suggestions were taken into account but not all of them.

purchase_order_analytic_search/models/purchase_order.py
angelmoya: If you filter id = False you will get any result.

But maybe you would like to filter purchase orders without analytic account... then value = False

AaronHForgeFlow: I see your point.  Would you do that @JuanyDForgeflow ?

JuanyDForgeflow: @AaronHForgeFlow @angelmoya I changed it please review

AaronHForgeFlow:
Hi @JuanyDForgeflow , now the tests are failing becuase there is a test that checks the search is returning no purchase orders. But it is finding purchase order. So the test must be updated to the new use case.

AaronHForgeFlow (changes requested):
Same as in v15, the filter is only available in RFQ, can you added it in purchase menu?

JuanyDForgeflow:
@AaronHForgeFlow It's added now here and on 15.0 as well

AaronHForgeFlow (approved):
Works good.

AaronHForgeFlow:
@angelmoya can you please update your review?

purchase_order_analytic_search/models/purchase_order.py
```diff
@@ -18,23 +18,23 @@ def _compute_analytic_accounts(self):
 
     @api.model
     def _search_analytic_accounts(self, operator, value):
-        po_line_obj = self.env["purchase.order.line"]
+        po_line_obj = self.env["purchase.order.line"].filtered(
+            lambda x: x.account_analytic_ids == value
```
AaronHForgeFlow: Use this instead:

`lambda x: x.account_analytic_id == value`

AaronHForgeFlow:
Keep this open please

HviorForgeFlow (approved):

marcelsavegnago (approved):
LGTM

PR #1431, opened 2022-04-19
[FIX] [15.0] `purchase_receipt_exceptation_manual_split`: Revert quantities with partial pickings

ivantodorovich:
The two cases are explained in the new unit tests:

1) Partial reception with a cancelled backorder picking.
2) Partial reception without a backorder picking.

Both cases should revert / merge cancelled quantities to the origin line.

Note: Also removed the use of unused context keys (probably left-over dead code)


SilvioC2C (approved):
LGTM

damdam-s (approved):

PR #1447, opened 2022-05-03
[14.0][ADD] sale_purchase_force_vendor: New addon.

victoralmau:
This module allows you to select a vendor at the sale order line level when a route is defined.

Please @pedrobaeza and  @chienandalu can you review it?

@Tecnativa TT35946


sale_purchase_force_vendor/readme/USAGE.rst
```diff
@@ -0,0 +1,6 @@
+#. Go to *Sale -> Orders -> Quotations* and create a new Quotation.
+#. Create a new line with the following options:
+    * `Route`: Buy + MTO.
```
chienandalu:
```suggestion
    * `Route`: Buy
```

chienandalu:
```suggestion
    * `Route`: Buy
```

sale_purchase_force_vendor/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,10 @@
+To configure this module, you need to:
+
+#. Install *sale_management* app.
+#. Go to *Inventory -> Configuration -> Settings* and check "Multi-Step Routes" option.
+#. Go to *Inventory -> Configuration -> Routes > Buy* and check "Sales Order Lines" option.
+#. Go to *Inventory -> Configuration -> Routes*, filter "Archived" an "Unarchived" MTO route.
+#. Go to *Sale -> Products -> Products*.
+#. Create a new product with the following options:
+    * [Puchase tab] `Vendors`: Set different vendors (Vendor A + Vendor B).
+    * [Iventory tab] `Routes`: Buy
```
chienandalu:
```suggestion
    * [Iventory tab] `Routes`: Buy and MTO
```

chienandalu (changes requested):
Tested. Just two readme tweaks:

sale_purchase_force_vendor/readme/USAGE.rst
pedrobaeza: It should be MTO for triggering the PO creation

pedrobaeza (changes requested):
This is missing the product.supplierinfo auto-creation

victoralmau:
> This is missing the product.supplierinfo auto-creation

Excuse me, can you explain a little what you mean? In order to select a vendor on sales order lines, `product.supplierinfo` records must exist.

pedrobaeza:
The idea is that you don't select a vendor, but a partner. Check FT custom.

victoralmau:
> The idea is that you don't select a vendor, but a partner. Check FT custom.

Ok, changes done.

sale_purchase_force_vendor/models/sale_order_line.py
```diff
@@ -0,0 +1,35 @@
...
+    vendor_id = fields.Many2one(
+        comodel_name="res.partner",
+        string="Vendor",
+    )
+
+    def _prepare_procurement_values(self, group_id=False):
+        res = super()._prepare_procurement_values(group_id=group_id)
+        if self.vendor_id:
+            addr = self.vendor_id.address_get(["purchase"])
+            supplier = self.env["res.partner"].browse(addr["purchase"])
```
pedrobaeza: This is not standard

victoralmau:
Fixed Travis errors.

chienandalu (approved):

sale_purchase_force_vendor/models/sale_order_line.py
```diff
@@ -0,0 +1,53 @@
...
+        comodel_name="res.partner",
+        string="Proveedor",
+        compute="_compute_allowed_vendor_ids",
+    )
+
+    @api.depends("product_id")
+    def _compute_allowed_vendor_ids(self):
+        supplierinfo_model = self.env["product.supplierinfo"]
+        self.allowed_vendor_ids = self.allowed_vendor_ids
+        for item in self.filtered(lambda x: x.product_id):
```
pedrobaeza, line 31: Make this configurable through ICP or similar for restricting or not the allowed vendors for covering the both cases we have.

victoralmau: Ok, I have added a checkbox in Purchase > Configuration > Products to restrict or not.

sale_purchase_force_vendor/models/stock_move.py
```diff
@@ -0,0 +1,22 @@
...
+from odoo import models
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    def _prepare_procurement_values(self):
+        res = super()._prepare_procurement_values()
+        if self.sale_line_id.vendor_id:
+            addr = self.sale_line_id.vendor_id.address_get(["invoice"])
```
pedrobaeza: Why are you selecting this address?

sale_purchase_force_vendor/models/stock_move.py
```diff
@@ -0,0 +1,22 @@
...
+            supplier = self.env["res.partner"].browse(addr["invoice"])
+            product = self.product_id
+            res["supplierinfo_id"] = product.with_company(
```
pedrobaeza: This seems duplicated with sale.order.line code

sale_purchase_force_vendor/views/sale_order_view.xml
```diff
@@ -0,0 +1,36 @@
...
+        <field name="arch" type="xml">
+            <xpath
+                expr="//field[@name='order_line']/tree/field[@name='price_unit']"
+                position="before"
+            >
+                <field name="allowed_vendor_ids" invisible="1" />
+                <field name="sale_purchase_force_vendor_restrict" invisible="1" />
+                <field
+                    name="vendor_id"
+                    attrs="{'invisible': ['|', ('route_id', '=', False), ('sale_purchase_force_vendor_restrict', '=', False)]}"
```
pedrobaeza: The question is not to totally hide the field, but restrict the list of possible elements to select or not. You can add 2 times the same field, one with domain and the other without, and put visibility depending on the parameter.

Another incorrect thing is to only check the visibility by route. Not all routes require vendor.

victoralmau: Ok done, the field is added 2 times and it is shown with the domain according to configuration.

sale_purchase_force_vendor/models/stock_move.py
victoralmau: You are right, I have removed the code and used the sale.order.line function.

sale_purchase_force_vendor/models/stock_move.py
victoralmau: Reviewing this, better to use the vendor_id field directly.

sale_purchase_force_vendor/models/sale_order_line.py
```diff
@@ -0,0 +1,54 @@
...
+                quantity=self.product_uom_qty,
+                uom_id=self.product_uom,
+            )
+            if not suppinfo:
+                suppinfo = self.env["product.supplierinfo"].create(
+                    {
+                        "product_tmpl_id": product.product_tmpl_id.id,
+                        "name": self.vendor_id.id,
+                        "min_qty": 0,
+                        "company_id": False,
```
pedrobaeza: Put the company of the SO, as there can't be incompatible combinations (like a product or partner restricted to one company and the supplierinfo not being restricted).

sale_purchase_force_vendor/models/sale_order_line.py
```diff
@@ -0,0 +1,54 @@
...
+            item.allowed_vendor_ids = supplierinfo_model.search(
+                [
+                    "|",
+                    ("product_tmpl_id", "=", item.product_id.product_tmpl_id.id),
+                    ("product_id", "=", item.product_id.id),
+                ]
+            ).mapped("name")
+
+    def _prepare_procurement_values(self, group_id=False):
+        res = super()._prepare_procurement_values(group_id=group_id)
```
pedrobaeza, line 44: Remember to add docstrings to indicate why you are doing this:
```suggestion
        """Inject in the procurement values the preferred vendor if any, and create
        supplierinfo record for it if it doesn't exist.
        """
        res = super()._prepare_procurement_values(group_id=group_id)
```

sale_purchase_force_vendor/models/stock_move.py
```diff
@@ -0,0 +1,16 @@
...
+    _inherit = "stock.move"
+
+    def _prepare_procurement_values(self):
+        res = super()._prepare_procurement_values()
+        if self.sale_line_id.vendor_id:
+            res_order_line = self.sale_line_id._prepare_procurement_values(
+                group_id=self.group_id
+            )
+            res.update({"supplierinfo_id": res_order_line["supplierinfo_id"]})
+        return res
```
pedrobaeza, line 19: The code is not correct, and please always add docstring for the WHYs. Is this case covered by tests?

```suggestion
        """Inject the preferred vendor in case of a MTO that first creates the OUT
        move.
        """
        res = super()._prepare_procurement_values()
        if self.sale_line_id.vendor_id:
            res["supplier_id"] = self.sale_line_id.vendor_id
        return res
```

sale_purchase_force_vendor/views/sale_order_view.xml
```diff
@@ -0,0 +1,41 @@
...
+                    options='{"no_create": True}'
+                />
+           </xpath>
+           <xpath
+                expr="//field[@name='order_line']/form/group/group/field[@name='price_unit']"
+                position="before"
+            >
+                <field name="allowed_vendor_ids" invisible="1" />
+                <field name="sale_purchase_force_vendor_restrict" invisible="1" />
+                <field
```
pedrobaeza, line 33: The double field is missing here.

pedrobaeza (changes requested):

sale_purchase_force_vendor/models/stock_move.py
victoralmau: Ok sorry, the docstring is already added.
The test uses this, but `supplierinfo_id` field is required for the test to work properly and create the order as per https://github.com/odoo/odoo/blob/14.0/addons/purchase_stock/models/stock_rule.py#L57

pedrobaeza: Please check, as this is not required in the previous code (but it was for v13).

pedrobaeza (approved):

francesco-ooops:
@victoralmau readme for this module is broken (it's an empty file), could you fix it?

pedrobaeza:
Fixed in e89330ed88b5b65f4f3ac739e79220407063c181

PR #1449, opened 2022-05-04
[15.0][MIG] purchase_analytic_global

yankinmax:


purchase_analytic_global/models/purchase.py
```diff
@@ -0,0 +1,32 @@
...
+            if len(account) == 1:
+                rec.account_analytic_id = account
+            else:
+                rec.account_analytic_id = False
+
+    def _inverse_analytic_account(self):
+        for rec in self:
+            if rec.account_analytic_id:
+                for line in rec.order_line:
+                    line.account_analytic_id = rec.account_analytic_id
```
ivantodorovich:
```suggestion
        for rec in self:
            rec.order_line.account_analytic_id = rec.account_analytic_id
```

I would do this instead

1) Fewer loc
2) Also allow to unset the line's analytic account. (unless there's a reason to prevent it?)

ivantodorovich (approved):

ajaniszewska-dev (approved):

purchase_analytic_global/models/purchase.py
HviorForgeFlow: @ivantodorovich @yankinmax any reason to don't apply this? Is OK for both of you?

yankinmax: @HviorForgeFlow I've just forgot about that, UPDATED, thx for kind reminder

PR #1453, opened 2022-05-05
[12.0][IMP] purchase_reception_status: Add header buttons to force/unforce Received

HviorForgeFlow:
CC @ForgeFlow

### DIFF
--- a/purchase_reception_status/models/purchase_order.py
+++ b/purchase_reception_status/models/purchase_order.py
@@ -22,6 +22,14 @@ class PurchaseOrder(models.Model):
         "even if some lines are not fully received. "
         "To be able to modify this field, you must first lock the order.")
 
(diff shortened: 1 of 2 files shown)


legalsylvain (commented):
Hi. thanks for this improvment. 

one remark, regarding consistency : 

- force_received field is hidden if state not in ["purchase,done"]
- the button are hidden if state not in ["done"]

I see it as inconsistent. Don't you think ?

HviorForgeFlow:
Thanks for reviewing @legalsylvain 

I followed `force_received` field definition logic, it is only editable (not readonly) on state 'done'
https://github.com/OCA/purchase-workflow/blob/12.0/purchase_reception_status/models/purchase_order.py#L20

PR #1455, opened 2022-05-05
[14.0][MIG] subcontracted_service

JuanyDForgeflow:
Migration to 14.0

This module allows a user to indicate that a service is subcontracted.
It provides the ability to create purchases from procurement processes.

This is a base module, upon specific modules for sale / manufacuturing, modules
will need to rely on. By itself it does not provide any function to the end user.

Possible uses of this module can be:

* Add subcontracted services to BOMs. When a manufacturing order is created a
  PO is triggered for the service to be subcontracted. See

* Add subcontracted services to sales order. When the SO is confirmed, it
  creates a PO for the service.


AaronHForgeFlow (approved):
Functional + code LGTM

AaronHForgeFlow (changes requested):
Sorry I meant to put the previous approval in a different PR, not this one.

I created a BOM with a subcontracted service and then created a manufacturing order and confirm it. However that did not create any requests for quotation. THe subcontracted service has a vendor specified on it.

AaronHForgeFlow:
The problem seems to be Odoo is banning services in the manufacturing order. Some refactor will be needed here :S

PR #1462, opened 2022-05-12
[14.0][ENH] purchase_invoice_plan, freeze plan amount if already invoiced

kittiu:
There are case when, even after some installment has been invoiced, user still add/edit purchase line amount which effect invoice plan. For example, given

Purchase Order = 400, with 2 Installments with the 1st installment already invoiced.
1. 50% = 200 (invoiced = 200)
2. 50% = 200

Then, after that, user change purchase so that amount become 800.

Before this enhancement (WRONG), the invoice plan table will be readjust as following (fixed percent as 50/50)
1. 50% = 400 (invoiced = 200)
2. 50% = 400

To be CORRECT, this PR ensure that the invoice plan table can become
1. 25% = 200 (invoiced = 200)
2. 75% = 600

Note: user can change installment 2 or even add installment 3

po_invoice_plan

### DIFF
--- a/purchase_invoice_plan/models/purchase.py
+++ b/purchase_invoice_plan/models/purchase.py
@@ -36,6 +36,7 @@ class PurchaseOrder(models.Model):
         string="Total Amount",
     )
 
+    @api.depends("invoice_plan_ids")
     def _compute_ip_total(self):
         for rec in self:
             installments = rec.invoice_plan_ids.filtered("installment")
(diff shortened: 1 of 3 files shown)


HviorForgeFlow (approved):
Code Review :+1:

Saran440 (approved):
Functional Test 👍

kittiu:
Rebased

Issue #1389, opened 2022-03-23
[13.0] purchase_last_price_info produce an error printing "Product Label (PDF)"


inprbau:
We have the same error on two types of installations: a multicompany installation and a single company installation when the purchase_last_price_info module is installed.

**_SINGLE COMPANY INSTALLATION_**
1. The purchase_last_price_info module is installed
2. We have a product that has not been purchased and the module fields are not filled in
![image](https://user-images.githubusercontent.com/50661813/159688228-a409a77c-cc7f-41cf-86c7-d37a311905af.png)
3. When we print the "Product Label (PDF)" of the article the following error appears
![image](https://user-images.githubusercontent.com/50661813/159688857-a6f75c57-f0b4-4163-84d6-e6f2765ee194.png)

Error:
Odoo Server Error

Traceback (most recent call last):
  File "/opt/odoo/odoo130/src/core/odoo/models.py", line 5124, in ensure_one
    _id, = self._ids
ValueError: not enough values to unpack (expected 1, got 0)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/odoo/odoo130/src/core/odoo/addons/base/models/qweb.py", line 333, in _compiled_fn
    return compiled(self, append, new, options, log)
  File "<template>", line 1, in template_product_report_simple_label_799
  File "/opt/odoo/odoo130/src/core/odoo/addons/base/models/ir_qweb.py", line 382, in _get_field
    content = converter.record_to_html(record, field_name, field_options)
  File "/opt/odoo/odoo130/src/core/odoo/addons/base/models/ir_qweb_fields.py", line 465, in record_to_html
    return super(MonetaryConverter, self).record_to_html(record, field_name, options)
  File "/opt/odoo/odoo130/src/core/odoo/addons/base/models/ir_qweb_fields.py", line 129, in record_to_html
    return False if value is False else record.env[self._name].value_to_html(value, options=options)
  File "/opt/odoo/odoo130/src/core/odoo/addons/base/models/ir_qweb_fields.py", line 435, in value_to_html
    formatted_amount = lang.format(fmt, display_currency.round(value),
  File "/opt/odoo/odoo130/src/core/odoo/addons/base/models/res_currency.py", line 133, in round
    self.ensure_one()
  File "/opt/odoo/odoo130/src/core/odoo/models.py", line 5127, in ensure_one
    raise ValueError("Expected singleton: %s" % self)
ValueError: Expected singleton: res.currency()

[...]

PR #1463, opened 2022-05-12
[FIX] purchase_last_price_info produces an error printing Product Label #1389

mmarchpunt:
[13.0] [FIX] Issue #1389


purchase_last_price_info/models/product_product.py
```diff
@@ -53,10 +53,15 @@ def _compute_last_purchase_line_id(self):
     @api.depends("last_purchase_line_id")
     def _compute_last_purchase_line_id_info(self):
         for item in self:
+            currency = (
+                item.last_purchase_line_id.currency_id
+                or item.company_id.currency_id
+                or item.currency_id
```
ivantodorovich, line 62:
```suggestion
                or item.currency_id
```

I think you should favour the product currency and stop there.
As the field is required it'll always be set

ivantodorovich (approved):

PR #1464, opened 2022-05-16
[14.0][FIX] po_secondary_unit: Drop 0 qty test assert on POL with secondary UoM

HviorForgeFlow:
Due to a change on https://github.com/OCA/purchase-workflow/commit/ef214d19706f15509f77abf047d1e1ea1487700b removed assert was failing from tests.

CC @pedrobaeza @sergio-teruel


pedrobaeza:
Indeed. Sorry for missing this one in 14.0:

/ocabot merge patch

PR #1465, opened 2022-05-17
[14.0][ADD] Purchase freight

bealdav:


purchase_freight_delay/models/purchase.py
```diff
@@ -0,0 +1,24 @@
...
+from datetime import timedelta
+
+
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+
+    freight_rule_id = fields.Many2one(comodel_name="freight.rule", copy=False)
+    freight_date = fields.Date(compute="_compute_freight", store=True, copy=False)
+
+    @api.depends("freight_rule_id", "freight_rule_id.duration", "state")
```
bealdav: missing date_planned

purchase_freight_delay/__manifest__.py
```diff
@@ -2,14 +2,17 @@
...
     "website": "https://github.com/OCA/purchase-workflow",
     "author": "Akretion",
     "license": "AGPL-3",
     "installable": True,
-    "maintainers": ["bealdav", "hparfr"],
+    "maintainers": [
+        "bealdav",
+        "hparfr",
+    ],
     "summary": "Bootstrap project quickly",
```
hparfr: this one too

purchase_freight_delay/models/purchase.py
```diff
@@ -9,42 +9,61 @@
...
-    already_computed_freight = fields.Boolean(readonly=True, help="Flag to ")
-    previous_date_planned = fields.Datetime(copy=False)
+    freight_rule_id = fields.Many2one(comodel_name="freight.rule", copy=False)
+    freight_duration = fields.Integer(
+        compute="_compute_freight_duration",
+        inverse="_inverse_freight_duration",
+        store=True,
+        copy=False,
+    )
+    initial_planned = fields.Datetime(copy=False)
```
hparfr: create a PO, add a POL, change date: it's reverted to initial_planned. Not what we expect

purchase_freight_delay/models/purchase.py
```diff
@@ -0,0 +1,69 @@
...
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+
+    freight_rule_id = fields.Many2one(comodel_name="freight.rule", copy=False)
+    freight_duration = fields.Integer(
+        compute="_compute_freight_duration",
+        inverse="_inverse_freight_duration",
+        store=True,
+        copy=False,
+    )
```
hparfr, line 27: It's missing a field like : 
```py
    date_dispatch = fields.Datetime('Date of dispatch', help="When the goods exit the boarding place" )
```

freight_duration:
* default from freight_rule
* can be set manually

date_dispatch
*  default from date_planned - freight_duration
*  can be set manually

date_planned:
* if date_dispatch or freigth_duration are changed then date_planned = freigth_duration + date_dispatch

bealdav: It's a possible behavior, but it makes UI and algo complex to tune on these 2 params, whereas only one could be sufficent for the same benefit. What think about the client ?

hparfr: Our process:
 
- (we) the client, agree with the supplier on a date of dispatch. It's part of the contract, and should be clearly displayed on the PO. The supplier is engaged to deliver the goods at this adress, on this date.
- we then choose the freight method. (and get a estimative freigth_duration).
- the freigth duration may be changed manually
- the date_planned is computed based on duration + date of shipment

There is also "normal" cases: the buyer set only the date_planned.

So for the buyer, it's quite clear, and he expects two fields:
- date planned = date when the goods should be in the destination warehouse
- date dispatch = date when the goods should be in the source warehouse

hparfr:
On runboat, create RFQ then :

```
   getattr(self, field.compute)()
  File "/mnt/data/odoo-addons-dir/purchase_freight_delay/models/purchase.py", line 125, in _compute_dispatch_date
    rec._update_freight_fields_and_co()
  File "/mnt/data/odoo-addons-dir/purchase_freight_delay/models/purchase.py", line 104, in _update_freight_fields_and_co
    days=vals["freight_duration"] or 0
Exception

The above exception was the direct cause of the following exception:

  File "/opt/odoo/odoo/http.py", line 639, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/opt/odoo/odoo/http.py", line 315, in _handle_exception
    raise exception.with_traceback(None) from new_cause
TypeError: unsupported operand type(s) for -: 'bool' and 'datetime.timedelta'
```

PR #1470, opened 2022-05-23
[13.0][FIX] purchase_blanket_order: Use purchasing UoM.

LoisRForgeFlow:
The default behavior in a PO is that the purchsing UoM is used
as the PO line UoM when set in the product. Blanket Orders should
do the same, otherwise it is a bit confusing.

@ForgeFlow


HviorForgeFlow (approved):
LGTM

AaronHForgeFlow (approved):
LGTM!

Issue #1421, opened 2022-04-13
Field product is not mandatory and the UoM is not filtered by product category of the product


hitrosol:
I try the Purchase Request V14 on runboat and found these cases:

Case 1. 
The Product I is not set as mandatory.
The result, after validate the PR, one can't create the RFQ and show this error message
![image](https://user-images.githubusercontent.com/5949936/163133959-24106942-5287-44c5-b3a9-dc07259ae328.png)

This will bring frustrated  of the user since they just know that have to fill the product after the PR approved.
IMO, the message error should shown whenever the user wants to save the form, not after validate it.

Case 2.
In multi UoM set, if the user input the Product, the UoM show all UoM regardless of the UoM category of the product.
It give risk where the user can choose UoM from different UoM category.
Still, the PR can be approved, but can't create the RFQ because of different UoM Category.
![image](https://user-images.githubusercontent.com/5949936/163135644-a5cd3020-44ac-4a38-bac5-219b0d4d9f43.png)

Regards,

HviorForgeFlow:
@hitrosol could you review this PR that is trying to solve your issues, thanks!

PR #1476, opened 2022-06-02
[14.0][FIX] purchase_request: UoM not required on PO creation

HviorForgeFlow:
Solves https://github.com/OCA/purchase-workflow/issues/1421

- [ ] Allow to create PO without UoM or selecting the PO ones by default.

- [ ] Restrict UoM selection to product UoM belonging category.

CC @ForgeFlow  @hitrosol

### DIFF
--- a/purchase_request/wizard/purchase_request_line_make_purchase_order.py
+++ b/purchase_request/wizard/purchase_request_line_make_purchase_order.py
@@ -352,7 +352,9 @@ class PurchaseRequestLineMakePurchaseOrderItem(models.TransientModel):
         string="UoM",
         related="product_id.uom_po_id",
         readonly=False,
+        domain="[('category_id', '=', product_uom_category_id)]",
     )
+    product_uom_category_id = fields.Many2one(related="product_id.uom_id.category_id")
     keep_description = fields.Boolean(
         string="Copy descriptions to new PO",
         help="Set true if you want to keep the "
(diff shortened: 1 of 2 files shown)


hitrosol:
hi @HviorForgeFlow, thanks for taking care the issues.
I try the runboat, found as follows:

1. I still can save and validate the PR eventhough I didn't fill the product_id, it's better to put the required in this field, so user understand that s/he has to fill this field. 
2. Other UoM categories still shown in the UoM field
gambar
It's better to limit the UoM shown only from the same category as product has.

Since information in the PR will be copied to RFQ, so better we follow the same requirement as in RFQ.

Regards,

purchase_request/wizard/purchase_request_line_make_purchase_order.py
```diff
@@ -348,8 +348,13 @@ class PurchaseRequestLineMakePurchaseOrderItem(models.TransientModel):
         string="Quantity to purchase", digits="Product Unit of Measure"
     )
     product_uom_id = fields.Many2one(
-        comodel_name="uom.uom", string="UoM", required=True
+        comodel_name="uom.uom",
+        string="UoM",
+        related="product_id.uom_po_id",
```
hparfr, line 353: A default value is may be better. Otherwise user can't decide which UoM to use.

hitrosol:
can we have the same behavior like in Purchase Order ? Where, the list UoM only shown in the same category as per product id uom category ?

purchase_request/wizard/purchase_request_line_make_purchase_order.py
HviorForgeFlow: Adding the readonly false allow the user to change it, using as default the Purchase UoM everytime the product_id is changed (default behavior). We also restrict by product_uom_category to keep consistency on selection. Also if UoM is not activated in the system the default selection will be the right one

hparfr: Just did a small test in runboat. The product_uom_id is always reset to product_id.uom_po_id

PR #1477, opened 2022-06-03
[IMP] purch_part_incoterm: add incoterm_address_id

bealdav:
Add fields incoterm_address _id on partner and purchase.order

Needs to depends on purchase_stock because field is not in defined in purchase view

Please @TDu @sonhd91 could you make review ?

Thanks

cc @rvalyi @hparfr

### DIFF
--- a/purchase_partner_incoterm/views/partner_view.xml
+++ b/purchase_partner_incoterm/views/partner_view.xml
@@ -8,6 +8,7 @@
     <field name="arch" type="xml">
       <group name="purchase" position="inside">
         <field name="purchase_incoterm_id" />
+        <field name="purchase_incoterm_address_id" />
       </group>
     </field>
   </record>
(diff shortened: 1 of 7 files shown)


purchase_partner_incoterm/views/partner_view.xml
```diff
@@ -8,6 +8,7 @@
     <field name="arch" type="xml">
       <group name="purchase" position="inside">
         <field name="purchase_incoterm_id" />
+        <field name="incoterm_address_id" />
```
bealdav, line 11: oops to set on purchase not on partner

hparfr:
Thx.

Can you please a default partner adress on the partner and set it during the onchange as the default incoterm ?

hparfr (approved):
Thanks

TDu (approved):

PR #1480, opened 2022-06-06
[13.0][IMP] purchase_discount: flag discount hack

chienandalu:
As Odoo isn't using its hook method yet, we still need to be using the price unit hack that makes two writes in order to be able to correctly compute the discount. This writes can confound other modules and even provoke an infinite loop in some situations. We should at least flag it to allow to ignore these writes.

We can see an example in this PR with `purchase_stock_price_unit_sync`

At the same time, we proposed again to Odoo to used its hook method which would avoid us this dirty hack:

- https://github.com/odoo/odoo/pull/92933

cc @Tecnativa TT36816

ping @pedrobaeza @sergio-teruel

### DIFF
--- a/purchase_discount/models/purchase_order.py
+++ b/purchase_discount/models/purchase_order.py
@@ -61,7 +61,7 @@ class PurchaseOrderLine(models.Model):
         value before calling super and restoring it later for assuring
         maximum inheritability.
 
-        HACK: This is needed while https://github.com/odoo/odoo/pull/29983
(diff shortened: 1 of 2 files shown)


cesar-tecnativa (approved):

pedrobaeza (approved):

sergio-teruel:
This commits have been ported to v14 and v15?

pedrobaeza:
It seems not. cc @chienandalu

PR #1483, opened 2022-06-09
[WIP][14.0] Add purchase_lead_hour

mmequignon:
This module allows to set more accurate vendor lead delays

I have to add tests

### DIFF
--- a/purchase_lead_hour/__manifest__.py
+++ b/purchase_lead_hour/__manifest__.py
@@ -1,0 +1,16 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+{
+    "name": "Purchase Lead Hour",
+    "summary": "Allows to set more accurate vendor lead delays",
+    "version": "14.0.1.0.0",
+    "category": "Purchases",
+    "website": "https://github.com/OCA/purchase-workflow",
+    "author": "Camptocamp SA, Odoo Community Association (OCA)",
+    "maintainers": ["mmequignon"],
+    "license": "AGPL-3",
+    "installable": True,
+    "auto_install": False,
+    "depends": ["product", "stock"],
+}
--- a/purchase_lead_hour/models/product_supplierinfo.py
+++ b/purchase_lead_hour/models/product_supplierinfo.py
@@ -1,0 +1,14 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from odoo import fields, models
+
+
+class ProductSupplierinfo(models.Model):
+    _inherit = "product.supplierinfo"
+
+    delay = fields.Float()
+
+    def get_next_availability_date(self):
+        # A hook to ease overrides
+        return fields.Datetime.now()
--- a/purchase_lead_hour/models/stock_orderpoint.py
+++ b/purchase_lead_hour/models/stock_orderpoint.py
@@ -1,0 +1,26 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from dateutil import relativedelta
+
+from odoo import fields, models
+
+
+class StockOrderpoint(models.Model):
+    _inherit = "stock.warehouse.orderpoint"
+
+    def _get_product_context(self):
+        # Since product.supplierinfo delay is now a float, we need to get the
+        # available qty as of now + delay
+        res = super()._get_product_context()
+        seller = fields.first(self.product_id.seller_ids)
+        if seller:
+            lead_days, __ = self.rule_ids._get_lead_days(self.product_id)
+            availability_date = seller.get_next_availability_date()
+            res.update(
+                {
+                    "to_date": availability_date
+                    + relativedelta.relativedelta(days=lead_days)
+                }
+            )
+        return res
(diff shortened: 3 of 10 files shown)


purchase_lead_hour/models/product_supplierinfo.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from odoo import fields, models
+
+
+class ProductSupplierinfo(models.Model):
+    _inherit = "product.supplierinfo"
+
+    delay = fields.Float()
```
sebalix, line 10: I would add a comment saying it's an override (from Integer to Float I guess?)

purchase_lead_hour/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+    "summary": "Allows to set more accurate vendor lead delays",
+    "version": "14.0.1.0.0",
+    "category": "Purchases",
+    "website": "https://github.com/OCA/purchase-workflow",
+    "author": "Camptocamp SA, Odoo Community Association (OCA)",
+    "maintainers": ["mmequignon"],
+    "license": "AGPL-3",
+    "installable": True,
+    "auto_install": False,
+    "depends": ["product", "stock"],
```
sebalix, line 15: As your module is named `purchase_*` we expect it to depend on `purchase`, or if the vendor delay can be used outside of a purchase flow, another module name could be found maybe.

purchase_lead_hour/models/stock_orderpoint.py
```diff
@@ -0,0 +1,26 @@
...
+        seller = fields.first(self.product_id.seller_ids)
+        if seller:
+            lead_days, __ = self.rule_ids._get_lead_days(self.product_id)
+            availability_date = seller.get_next_availability_date()
+            res.update(
+                {
+                    "to_date": availability_date
+                    + relativedelta.relativedelta(days=lead_days)
+                }
+            )
```
sebalix, line 25: I think you could break the expected behavior by forcing the date to today here (through the call to `get_next_availability_date()`) as the std code is:
```python
            'to_date': datetime.combine(self.lead_days_date, time.max)
```
This `lead_days_date` is computed by default from today, but it can be overridden by other submodules. We could of course create a glue module, but to avoid that should we instead propose a hook to Odoo to replace the call to `fields.Date.today()` here?
```python
    @api.depends('rule_ids', 'product_id.seller_ids', 'product_id.seller_ids.delay')
    def _compute_lead_days(self):
        for orderpoint in self.with_context(bypass_delay_description=True):
            if not orderpoint.product_id or not orderpoint.location_id:
                orderpoint.lead_days_date = False
                continue
            lead_days, dummy = orderpoint.rule_ids._get_lead_days(orderpoint.product_id)
            lead_days_date = fields.Date.today() + relativedelta.relativedelta(days=lead_days)
            orderpoint.lead_days_date = lead_days_date
```

I would also need this kind of hook, it could simplify another addon here: https://github.com/OCA/stock-logistics-warehouse/pull/1411 (which BTW is incompatible with your addon, it's something we should avoid).

mmequignon: I already created a hook on `product.supplierinfo`, which retrieves the seller's availability date.
Is that what you're talking about?

purchase_lead_hour/models/product_supplierinfo.py
```diff
@@ -0,0 +1,14 @@
...
+
+    def get_next_availability_date(self):
```
grindtildeath, line 12: The function name doesn't really match with its content :confused: 
Suggestion: `get_date_to_compute_delay` ?

purchase_lead_hour/models/stock_orderpoint.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class StockOrderpoint(models.Model):
+    _inherit = "stock.warehouse.orderpoint"
+
+    def _get_product_context(self):
+        # Since product.supplierinfo delay is now a float, we need to get the
+        # available qty as of now + delay
+        res = super()._get_product_context()
+        seller = fields.first(self.product_id.seller_ids)
```
grindtildeath, line 16: Use `_select_seller` on `product.product` instead.

purchase_lead_hour/models/stock_orderpoint.py
```diff
@@ -0,0 +1,26 @@
...
+from odoo import fields, models
+
+
+class StockOrderpoint(models.Model):
+    _inherit = "stock.warehouse.orderpoint"
+
+    def _get_product_context(self):
+        # Since product.supplierinfo delay is now a float, we need to get the
+        # available qty as of now + delay
+        res = super()._get_product_context()
```
grindtildeath, line 15: Probably need to add :
```
if not self.product_id or not self.location_id:
    return res
```
To mimic `_compute_lead_days` function

purchase_lead_hour/__manifest__.py
grindtildeath: If you're to use `product.supplierinfo` you have to depend on `purchase_stock` as `_get_lead_days` would filter on pull and pull_push rules if not installed.

grindtildeath (commented):
Can you please add some tests?

mmequignon:
replaced by https://github.com/OCA/purchase-workflow/pull/1484

PR #1484, opened 2022-06-10
[14.0] Add purchase_lead_hour

mmequignon:
supersedes https://github.com/OCA/purchase-workflow/pull/1483


purchase_lead_hour/models/product_supplierinfo.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class ProductSupplierinfo(models.Model):
+    _inherit = "product.supplierinfo"
+
+    # In odoo core, this field is an Integer, which in some cases,
+    # could not be accurate enough.
+    delay = fields.Float()
+
+    def get_next_availability_date(self):
```
ivantodorovich:
```suggestion
    def _get_next_availability_date(self):
```

This should be private.
In odoo, public methods are exposed through xmlrpc
rule of thumb is make everything private unless you really want this

purchase_lead_hour/models/stock_orderpoint.py
```diff
@@ -0,0 +1,28 @@
...
+class StockOrderpoint(models.Model):
+    _inherit = "stock.warehouse.orderpoint"
+
+    def _get_product_context(self):
+        # Since product.supplierinfo delay is now a float, we need to get the
+        # available qty as of now + delay
+        res = super()._get_product_context()
+        if not self.product_id or not self.location_id:
+            return res
+        seller = fields.first(self.product_id._select_seller(quantity=None))
```
ivantodorovich:
```suggestion
        seller = self.product_id._select_seller(quantity=None)
```

`_select_seller` never returns more than 1 record

mmequignon:
@ivantodorovich thanks Iván, I updated my pr

ivantodorovich (approved):

purchase_lead_hour/models/stock_orderpoint.py
```diff
@@ -0,0 +1,28 @@
...
+
+from odoo import models
+
+
+class StockOrderpoint(models.Model):
+    _inherit = "stock.warehouse.orderpoint"
+
+    def _get_product_context(self):
+        # Since product.supplierinfo delay is now a float, we need to get the
+        # available qty as of now + delay
```
sebalix: But here you only use your new hook `_get_next_availability_date` to compute the forecasted qty, and the forecasted date will remain unchanged if we override the hook. That's a blocking issue (the forecasted qty must comply with the lead date).

`_get_product_context` was using `lead_days_date` (computed with `fields.Date.today()`  in `_compute_lead_days`).
As you are overriding this lead date only in `_get_product_context`, maybe you will get the right qty, but the `lead_days_date` is still returning the std date.

We need a hook in Odoo core to replace the call to `fields.Date.today()` in `_compute_lead_days` so you can put the date/datetime you want, so you'll get coherent results regarding the lead date and the qty. Then your `_get_product_context` could be removed.

sebalix (changes requested):

purchase_lead_hour/models/product_supplierinfo.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class ProductSupplierinfo(models.Model):
+    _inherit = "product.supplierinfo"
+
+    # In odoo core, we can only set an amount of days,
+    # which could not be accurate enough.
+    delay_hour = fields.Integer("Delivery Lead Time Hour")
+
+    def _get_next_availability_date(self):
```
grindtildeath, line 14: > The function name doesn't really match with its content

https://github.com/OCA/purchase-workflow/pull/1483#discussion_r893636953 :stuck_out_tongue_winking_eye:

purchase_lead_hour/models/stock_orderpoint.py
```diff
@@ -0,0 +1,28 @@
...
+from dateutil import relativedelta
+
+from odoo import models
+
+
+class StockOrderpoint(models.Model):
+    _inherit = "stock.warehouse.orderpoint"
+
+    def _get_product_context(self):
+        # Since product.supplierinfo delay is now a float, we need to get the
```
grindtildeath: This comment looks outdated.

purchase_lead_hour/models/stock_rule.py
```diff
@@ -0,0 +1,23 @@
...
+        delay += delay_hour / 24
+        return delay, description
+
+    def _get_seller_lead_hours(self, product):
+        """Returns the main's seller delay_hour"""
+        buy_rule = self.filtered(lambda r: r.action == "buy")
+        seller = product.with_company(buy_rule.company_id)._select_seller(quantity=None)
+        if not buy_rule or not seller:
+            return 0.0
+        buy_rule.ensure_one()
```
grindtildeath: Any reason to use `ensure_one` here, as it would raise an error if you have more than one buy rule?

Wouldn't it be cleaner to check if the "buy" rule is in the orderpoint's rule_ids in your override of `_get_product_context` instead?

PR #1488, opened 2022-06-13
[14.0][MIG] purchase_stock_secondary_unit

mrgoetz:


mrgoetz:
Just a straight forward module migration. The test methods needed a small change, since action_done in stock_picking must be replaced by button_validate.

marylla (approved):
I made a functional review, like this:

1. Activate Unit of Measures here: Inventory > Configuration > Settings.
2. Create a new product, e.g. "Screws" with uom "Unit(s)".
3. Add a new secondary unit line with name, e.g. "Gram", with uom "g" and factor 10, so that 1 g = 10 Units.
4. Set this newly created secondary unit as the default secondary unit for purchases for this product.
5. Add a supplier for this product with price etc.
6. Create a new purchase order and select the newly created product.
7. Change either the "normal" quantity or the secondary quantity. It should update each other.
8. Confirm the purchase order and check the stock picking.
9. The stock picking has the same quantity, uom, secondary quantity and secondary uom.

So, it works as expected. :+1:

yweng8111:
@mrgoetz Can you please reorder the changes in the last 2 commits?
the following log is in the last commit. But the changes is in the previous commit
* stock_picking "action_done" call needed to be replaced by button_validate in tests

mrgoetz:
@yweng8111 wrote:

> @mrgoetz Can you please reorder the changes in the last 2 commits? the following log is in the last commit. But the changes is in the previous commit
> 
>     * stock_picking "action_done" call needed to be replaced by button_validate in tests

The last commits 87f757e,8345472,ad962f8 were done to accomplish that. Just the commit messages change, as clicking the compare button shows.

yweng8111 (approved):

pedrobaeza:
Please squash and name commits correctly (the migration commit doesn't contain the migration itself).

marylla:
@pedrobaeza Can we merge now? :)

rvalyi:
> @pedrobaeza Can we merge now? :)

It's merged already.

PR #1492, opened 2022-06-16
[14.0][FIX] purchase_invoice_plan: error when invoice=False

ps-tubtim:
- Fix error when `invoice = False`
- Add hook function for computing `amount_invoice`

### DIFF
--- a/purchase_invoice_plan/models/purchase.py
+++ b/purchase_invoice_plan/models/purchase.py
@@ -119,7 +119,7 @@ class PurchaseOrder(models.Model):
 
     def action_view_invoice(self, invoices=False):
         invoice_plan_id = self._context.get("invoice_plan_id")
-        if invoice_plan_id:
+        if invoice_plan_id and invoices:
             plan = self.env["purchase.invoice.plan"].browse(invoice_plan_id)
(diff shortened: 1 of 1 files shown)


purchase_invoice_plan/models/purchase.py
```diff
@@ -271,7 +271,11 @@ def _compute_invoiced(self):
                 lambda l: l.state in ("draft", "posted")
             )
             rec.invoiced = invoiced and True or False
-            rec.amount_invoiced = invoiced[:1].amount_untaxed
+            positive_lines = invoiced[:1].invoice_line_ids.filtered(
+                lambda l: l.price_subtotal > 0
```
kittiu: I think we should check with is_deposit flag (which can relate through po_line, with module purchase_invoice_plan_deposit).

Checking with negative amount, we can't be sure as user may just want to give discount, etc...

BUT, we may need to make a Hook functon to get the amount_invoiced.

kittiu (approved):
LGTM

Saran440 (approved):

HviorForgeFlow (approved):
Code review

PR #1498, opened 2022-06-21
[15.0][MIG] purchase_location_by_line

JuanyDForgeflow:
Purchase Order Lines.

This module is used to send the products to different locations, that may
not be children of the default location of the same PO picking type. Not be confused.


purchase_location_by_line/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+# © 2016 ForgeFlow S.L.
+#   (<http://www.forgeflow.com>)
+# © 2018 Hizbul Bahar <hizbul25@gmail.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Purchase Location by Line",
+    "summary": "Allows to define a specific destination location on each PO line",
+    "version": "15.0.1.0.1",
```
MiquelRForgeFlow:
```suggestion
    "version": "15.0.1.0.0",
```

MiquelRForgeFlow (approved):

HviorForgeFlow (commented):
Functional review.

It is possible to change Destination Location once Purchase order is confirm (equal to picking is also created). I guess blocking the capacity to edit the field when ['purchase', 'done', 'cancelled'] could be a solution.

HviorForgeFlow (approved):
Functional Review :+1: 
Thanks for the improvement

PR #1499, opened 2022-06-21
[15.0][MIG] purchase_open_qty

JuanyDForgeflow:
This module allows purchase users to identify what are the purchase orders
that currently have quantities pending to invoice or to receive.


purchase_open_qty/models/purchase_order.py
```diff
@@ -109,7 +109,7 @@ def _search_pending_qty_to_invoice(self, operator, value):
     )
     qty_to_receive = fields.Float(
         compute="_compute_qty_to_receive",
-        search="_search_qty_to_receive",
+        # search="_search_qty_to_receive",
```
chienandalu: Why?

chienandalu:
Can you add https://github.com/OCA/purchase-workflow/pull/1503 ?

purchase_open_qty/models/purchase_order.py
CasVissers-360ERP: _search_qty_to_receive is not defined.

chienandalu: The it should be simply removed

HviorForgeFlow:
@JuanyDForgeflow could you attend comments above? Thanks!

JuanyDForgeflow:
> Can you add #1503 ?

I added this in the migration commit

JuanyDForgeflow:
> @JuanyDForgeflow could you attend comments above? Thanks!

Done!

chienandalu (approved):
Thanks!:+1:

TheerayutEncoder:
Functional Test 👍

TheerayutEncoder (approved):

DhavalTforgeFlow:
https://github.com/OCA/purchase-workflow/blob/2ae6a826050701ed6c12d9035be14d1eec9fd87b/purchase_open_qty/models/purchase_order.py#L15

I think, We do not need qty_to_invoice field in this module. "qty_to_invoice" field is now standard field in purchase module. Can you @JuanyDForgeflow  please check and remove the declaration and computation of this field?

Standard field : https://github.com/odoo/odoo/blob/b3199e949ae70307e8181ba0d6237871d6d07af7/addons/purchase/models/purchase.py#L982

PR #1501, opened 2022-06-24
[IMP] purchase_open_qty: add fields to tree view

chienandalu:
This allows us to use anywhere this view is loaded. For example by the module ` purchase_order_line_menu`.

cc @Tecnativa TT37544

please review @pedrobaeza @victoralmau @MiquelRForgeFlow

### DIFF
--- a/purchase_open_qty/views/purchase_view.xml
+++ b/purchase_open_qty/views/purchase_view.xml
@@ -22,6 +22,19 @@
             </field>
         </field>
     </record>
+    <!-- Show open quantities fields wherever the tree view is used -->
+    <record model="ir.ui.view" id="view_purchase_order_line_tree">
+        <field name="model">purchase.order.line</field>
+        <field name="inherit_id" ref="purchase.purchase_order_line_tree" />
+        <field name="arch" type="xml">
+            <field name="product_qty" position="after">
+                <field name="qty_to_receive" optional="hide" />
+                <field name="qty_received" optional="hide" />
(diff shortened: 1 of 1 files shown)


pedrobaeza (changes requested):
There are computed non stored fields here, so the performance is not going to be OK

chienandalu:
Don't they compute only when they're shown?

chienandalu:
Well, checking it all these fields are stored indeed:


qty_invoiced/qty_received: https://github.com/OCA/OCB/blob/13.0/addons/purchase/models/purchase.py#L497-L503

qty_to_invoice/qty_to_receive: https://github.com/OCA/purchase-workflow/blob/13.0/purchase_open_qty/models/purchase_order.py#L44-L57

chienandalu:
> Don't they compute only when they're shown?

Responding myself, if hidden they're computed on the tree load anyway.

In any case, these fields are stored. So it should be no issue

pedrobaeza (approved):
True, the order of the models in the file has confused me, thinking that the ones not being stored are the line ones (but they are not stored at order level, not at line level). This is raising another question: installing this module will need a lot of time to compute such fields.

chienandalu:
There's an init hook for it: https://github.com/OCA/purchase-workflow/blob/eaf56c35067bdc8d6299389fa70f2da045b83566/purchase_open_qty/init_hook.py

victoralmau (approved):

pedrobaeza:
FW-port it to upper versions

PR #1512, opened 2022-07-01
[15.0][FIX] purchase_blanket_order: Set name by sequence only if is necessary.

victoralmau:
FWP from 14.0. https://github.com/OCA/purchase-workflow/pull/1510

Set name by sequence only if is necessary.

Steps to reproduce it before:

- Create a blanket order and confirm > PBO/0001
- Convert to draft
- Confirm > PBO/0002

@Tecnativa TT37792


pedrobaeza (approved):

pedrobaeza:
Fast-tracking the last in the chain:

/ocabot merge patch

PR #1514, opened 2022-07-04
[14.0][ADD] purchase_global_discount

omar7r:


monterl:
Hello, 

Is going to be resolve not successful check to mergue to 14.0? or is it discontinue?

Salutes.

omar7r:
Hello @monterl 

This pr is waiting for approvals for merging, if you test it, and works you can do the approval. The test's fail wasn't related to this addon.

ajaniszewska-dev:
@omar7r  if it is migration - i'd expect to see past commits & module existing on older branches - probably (looking at your commits) this is a new module then?

purchase_global_discount/models/purchase_order.py
```diff
@@ -0,0 +1,142 @@
+# Copyright 2021 Comunitea - Omar Castiñeira
```
ajaniszewska-dev, line 1:
```suggestion
# Copyright 2023 Comunitea - Omar Castiñeira
```

ajaniszewska-dev (approved):
seems good,except it doesnt look like a migration but a new module -> replace `MIG` with `ADD` in  PR title

anmarmo1:
The module is functional but I think there's an incompatibility with the "purchase_order_general_discount" module since, when applying the global discount, the line discount doesn't apply.
There are two similar modules in sales that are compatible.

image
 

I've configured the two discounts in the supplier, but in the purchase order only one of them has been applied.

image

I don't know if it's the expected functionality.

anmarmo1:
@omar7r have you been able to check this?

> The module is functional but I think there's an incompatibility with the "purchase_order_general_discount" module since, when applying the global discount, the line discount doesn't apply. There are two similar modules in sales that are compatible.
> 
> image
> 
> I've configured the two discounts in the supplier, but in the purchase order only one of them has been applied.
> 
> image
> 
> I don't know if it's the expected functionality.

omar7r:
@anmarmo1 They are incompatibles, for this version and at least in the past too, in sales are incompatibles too. I have not reviewed the reason, it was not the objective of this PR, but I have encountered the problem in sales too, some time ago.

anmarmo1 (approved):
Ok, so at a functional level it is correct, thank you very much for your answer

PR #1520, opened 2022-07-06
[UPD] Switch to Github Actions

rousseldenis:


pedrobaeza:
This one is not going to be easy...

rousseldenis:
> This one is not going to be easy...

Not so hard.

Deactivated license mandatory check

rousseldenis:
@LoisRForgeFlow @JordiBForgeFlow purchase_order_approval_block is LGPL but based on purchase_exception which is AGPL. This is forbiddden. IMHO you should go to AGPL

rousseldenis:
@pedrobaeza :green_apple:

pedrobaeza:
Great, I think we should directly switch that modules to AGPL and don't deactivate license check, but you mark the rhythm.

rousseldenis:
> Great, I think we should directly switch that modules to AGPL and don't deactivate license check, but you mark the rhythm.

I usually ping the authors to warn them as I'd prefer to have their feeling about that.

Even if you are right, this check should pass anyway

rousseldenis:
Following this https://github.com/odoo/odoo/commit/49b143e1b7f12ee6cf4d14754904f3e9b64a1011

It seems we don't need anymore purchase_order_line_sequence module.

@damdam-s @BernatPForgeFlow

rousseldenis:
@pedrobaeza What do we do here ? Let license and ignore or change license ?

pedrobaeza:
I would switch the license, as the author are warned and can make any later move, but keep the check for possible new contributions.

rousseldenis:
> I would switch the license, as the author are warned and can make any later move, but keep the check for possible new contributions.

Done

purchase_order_line_sequence/__manifest__.py
```diff
@@ -24,7 +24,7 @@
         "views/report_purchasequotation.xml",
     ],
     "post_init_hook": "post_init_hook",
-    "installable": True,
+    "installable": False,
```
matteoopenf, line 27: why this key became false?
Which is the problem?

matteoopenf: https://github.com/OCA/purchase-workflow/pull/1564

PR #1528, opened 2022-07-12
[13.0][FIX] purchase_blanket_order: Don't compare with translatable string

pedrobaeza:
Currently the check for assigning a new sequence is done over a translatable string, but the initial sequence name is not done being translatable, so we should make both to compare the same string.

TT37792


victoralmau (approved):

CarlosRoca13 (approved):

pedrobaeza:
Cherry-picked to v14 at 10e3029e1150f7ecbfc81edd9222ac0e6888cb2f and to v15 at 57fdf8a9da6e92bdd9fb75aaceda88a0a484144f

PR #1532, opened 2022-07-20
[14.0][FIX] purchase_force_invoiced: Adapt BI report to the check

pedrobaeza:
The purchase BI analysis is not taking into account the "Force invoiced" check for showing their quantity to be billed, so it
gives contradictory information from the search filter and the pivot.

This commit adapts the SQL report for taking this into account.

FW-port of #1531 for v14

@Tecnativa


MiquelRForgeFlow (approved):

CarlosRoca13 (approved):

LudLaf (approved):
Changes approved.

PR #1533, opened 2022-07-22
[IMP] purchase_partner_incoterm: use commercial_partner_id

hparfr:
Incoterms should be on the supplier, not on the contact.

### DIFF
--- a/purchase_partner_incoterm/models/res_partner.py
+++ b/purchase_partner_incoterm/models/res_partner.py
@@ -1,6 +1,6 @@
 # Copyright 2020 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
-from odoo import fields, models
+from odoo import api, fields, models
 
 
 class ResPartner(models.Model):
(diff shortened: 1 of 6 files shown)


TDu:
Make sense, but we may need a migration script for it.

purchase_partner_incoterm/models/res_partner.py
```diff
@@ -16,3 +16,10 @@ class ResPartner(models.Model):
         ondelete="restrict",
         string="Default Purchase Incoterm Address",
     )
+
+    @api.model
+    def _commercial_fields(self):
+        return super(ResPartner, self)._commercial_fields() + [
```
bealdav:
```suggestion
        return super()._commercial_fields() + [
```

HviorForgeFlow (approved):

bealdav (approved):

HviorForgeFlow:
@hparfr I'm not sure if applying major merge will overwrite your changes on version's manifest. Could you undo undo that version change and I apply major merge?

hparfr:
> @hparfr I'm not sure if applying major merge will overwrite your changes on version's manifest. Could you undo undo that version change and I apply major merge?

The hardcoded version change in the manifest is to ensure the naming of the migration script.
I think you can use `ocabot merge nobump`

PR #1542, opened 2022-08-09
[13.0][FIX] purchase_propagate_qty: CI error due to upstream change

pedrobaeza:
Since https://github.com/odoo/odoo/commit/ba4111e7a9202d801e3d51b55cad9659877e514a, tests are failing.

Analyzing the roout cause, this module was testing things that are not done by it, but a side effect of the previous status quo.

As such, we remove the tests that were testing this expected side effects, as well as removing the reference in the README. If the
behavior wants to be restored, it should be explicitly handled in a module with its own tests.

@Tecnativa

### DIFF
--- a/purchase_propagate_qty/README.rst
+++ b/purchase_propagate_qty/README.rst
@@ -31,10 +31,6 @@ quantity is decreased. This module changes the behavior of Odoo and propagates
 the decrease if the new purchased quantity is below the already received
 quantity. The increase is also propagated.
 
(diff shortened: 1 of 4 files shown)


cesar-tecnativa (approved):

victoralmau (approved):

pedrobaeza:
@rousseldenis this is needed for making the branch green and not testing things that are not done by the module itself.

pedrobaeza:
Fast-tracking this for fixing the branch:

/ocabot merge patch

PR #1546, opened 2022-08-23
[IMP] purchase_request: split field "Quantity in RFQ or PO"" in purchased_qty and rfq_qty fields.

augusto-weiss:
Features in this PR:
* Split field purchased_qty (Quantity in RFQ or PO) into twice different fields: purchased_qty and rfq_qty. Also added to views.
* Add to the compute method of the purchase request line state the dependence on quantities (i.e.: A purchase request line is not done if the purchase order quantities are less than de purchase request line quantities)
* Add the description of purchase request line to the purchase order line.


bruno-zanotti (approved):
LGTM

PR #1547, opened 2022-08-23
 [14.0][ADD] purchase_vendor_bill_product_breakdown

geomer198:


rousseldenis:
Hi @geomer198. Thanks for this.

I have tested it. Functionally, I think you need to find a way to fill in the quantities received on Reception picking too (that fill then the quantities in PO components).

IMHO, you could split this module with a first module that defines the products supplier components. Then, the bill breakdown.

elvise:
> IMHO, you could split this module with a first module that defines the products supplier components. Then, the bill breakdown.

@rousseldenis good point! what's the advantage?

rousseldenis:
> > IMHO, you could split this module with a first module that defines the products supplier components. Then, the bill breakdown.
> 
> @rousseldenis good point! what's the advantage?

A better inheritance mechanism without pulling the whole big module. We could imagine another workflow that uses thoses 'breakable' products.

geomer198:
> > > IMHO, you could split this module with a first module that defines the products supplier components. Then, the bill breakdown.
> > 
> > 
> > @rousseldenis good point! what's the advantage?
> 
> A better inheritance mechanism without pulling the whole big module. We could imagine another workflow that uses thoses 'breakable' products.

@rousseldenis Hi!) I start splitting the module. Can you do a technical review? 
First of two module: https://github.com/OCA/purchase-workflow/pull/1554

pedrobaeza:
Isn't this similar to what is done on `product_pack` and `sale_product_pack`?

geomer198:
> Isn't this similar to what is done on `product_pack` and `sale_product_pack`?

Module 'product_pack' uses fixed products sets.
This module uses dynamic product pack which depend on vendor. Moreover same vendor can have same products in the same document one of them is a pack and another one is not.
The primary goal of this module is to handle accounting documents not to track physical products

pedrobaeza:
OK, I won't force any kind of convergence.

francesco-ooops:
@geomer198 please close this one

PR #1556, opened 2022-09-07
[13.0][IMP] purchase_order_uninvoiced_amount: Make field stored to fix the grouped sum values in purchase tree view

sergio-teruel:
cc @Tecnativa TT38874

The amount_uninvoiced field in the purchase order tree view is set with the attribute sum=.... but the field is compute no store so the grouped value is not displayed.

ping @carlosdauden @chienandalu


chienandalu (approved):

pedrobaeza (approved):
The computation of the field on update is going to be a bit huge, but it's needed for summing on groups and sorting

pedrobaeza:
Please fw-port it to v14

PR #1559, opened 2022-09-09
[13.0][FIX] purchase_landed_cost: Avoid error in moves if the user does not have any purchase permission.

victoralmau:
Avoid error in moves if the user does not have any purchase permission.

@Tecnativa TT38904

### DIFF
--- a/purchase_landed_cost/views/account_move_view.xml
+++ b/purchase_landed_cost/views/account_move_view.xml
@@ -3,6 +3,7 @@
     <record id="view_move_form" model="ir.ui.view">
         <field name="model">account.move</field>
         <field name="inherit_id" ref="account.view_move_form" />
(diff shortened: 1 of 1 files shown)


purchase_landed_cost/views/account_move_view.xml
```diff
@@ -3,6 +3,10 @@
     <record id="view_move_form" model="ir.ui.view">
         <field name="model">account.move</field>
         <field name="inherit_id" ref="account.view_move_form" />
+        <field
+            name="groups_id"
+            eval="[(6, 0, [ref('purchase.group_purchase_user')])]"
```
chienandalu:
```suggestion
            eval="[(4, ref('base.group_purchase_user'))]"
```

chienandalu (approved):

pedrobaeza (approved):

PR #1570, opened 2022-09-26
[13.0][FIX] purchase_last_price_info : convert last price based on product purchase unit of measure

benwillig:
Before this PR, if you purchased for example a **dozen** of units at 1200, you would have 1200 as last purchase price. Even if the product purchase UoM was unit.
After this PR, the purchase last price of the product will be 100. As we purchased 12 units for a total of 1200. So the last purchase price is converted between the actual purchase UoM and the product purchase UoM.

I can forward it to 14.0 if accepted.

@LoisRForgeFlow


baimont (approved):
Code review

PR #1578, opened 2022-09-28
[14.0][FIX] sale_purchase_force_vendor: Display the vendor_id field only 1 time so that it is always displayed and apply the correct domain in both use cases

victoralmau:
Display the `vendor_id` field only 1 time so that it is always displayed and apply the correct domain in both use cases

@Tecnativa TT39679

### DIFF
--- a/sale_purchase_force_vendor/models/sale_order_line.py
+++ b/sale_purchase_force_vendor/models/sale_order_line.py
@@ -1,5 +1,7 @@
 # Copyright 2022 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+import json
+
 from odoo import api, fields, models
 
 
@@ -18,24 +20,21 @@ class SaleOrderLine(models.Model):
         comodel_name="res.partner",
         string="Vendor",
     )
-    allowed_vendor_ids = fields.One2many(
-        comodel_name="res.partner",
-        string="Proveedor",
-        compute="_compute_allowed_vendor_ids",
+    vendor_id_domain = fields.Char(
+        compute="_compute_vendor_id_domain",
+        readonly=True,
+        store=False,
     )
 
     @api.depends("product_id")
-    def _compute_allowed_vendor_ids(self):
-        supplierinfo_model = self.env["product.supplierinfo"]
-        self.allowed_vendor_ids = self.allowed_vendor_ids
-        for item in self.filtered(lambda x: x.product_id):
-            item.allowed_vendor_ids = supplierinfo_model.search(
-                [
-                    "|",
-                    ("product_tmpl_id", "=", item.product_id.product_tmpl_id.id),
-                    ("product_id", "=", item.product_id.id),
-                ]
-            ).mapped("name")
+    def _compute_vendor_id_domain(self):
+        for item in self:
+            domain = (
+                [("id", "in", item.product_id.variant_seller_ids.name.ids)]
+                if item.order_id.sale_purchase_force_vendor_restrict
+                else []
+            )
+            item.vendor_id_domain = json.dumps(domain)
 
     def _prepare_procurement_values(self, group_id=False):
         """Inject in the procurement values the preferred vendor if any, and create
--- a/sale_purchase_force_vendor/views/sale_order_view.xml
+++ b/sale_purchase_force_vendor/views/sale_order_view.xml
@@ -12,33 +12,22 @@
                 expr="//field[@name='order_line']/tree/field[@name='price_unit']"
                 position="before"
             >
-                <field name="allowed_vendor_ids" invisible="1" />
+                <field name="vendor_id_domain" invisible="1" />
                 <field
                     name="vendor_id"
-                    attrs="{'column_invisible': [('parent.sale_purchase_force_vendor_restrict', '=', False)]}"
-                    domain="[('id','in', allowed_vendor_ids)]"
+                    domain="vendor_id_domain"
                     options='{"no_create": True}'
-                />
-                <field
-                    name="vendor_id"
-                    attrs="{'column_invisible': [('parent.sale_purchase_force_vendor_restrict', '=', True)]}"
-                    options='{"no_create": True}'
+                    optional="hide"
                 />
            </xpath>
            <xpath
                 expr="//field[@name='order_line']/form/group/group/field[@name='price_unit']"
                 position="before"
             >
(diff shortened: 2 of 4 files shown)


victoralmau:
Ping @chienandalu

sale_purchase_force_vendor/models/sale_order_line.py
```diff
@@ -27,15 +27,23 @@ class SaleOrderLine(models.Model):
     @api.depends("product_id")
     def _compute_allowed_vendor_ids(self):
         supplierinfo_model = self.env["product.supplierinfo"]
+        partner_model = self.env["res.partner"]
         self.allowed_vendor_ids = self.allowed_vendor_ids
-        for item in self.filtered(lambda x: x.product_id):
+        for item in self.filtered(
+            lambda x: x.product_id and x.order_id.sale_purchase_force_vendor_restrict
```
pedrobaeza: If you are here filtering section and notes, you should filter the proper field (`display_type`?)

sale_purchase_force_vendor/models/sale_order_line.py
```diff
@@ -27,15 +27,23 @@ class SaleOrderLine(models.Model):
...
+        ):
             item.allowed_vendor_ids = supplierinfo_model.search(
                 [
                     "|",
                     ("product_tmpl_id", "=", item.product_id.product_tmpl_id.id),
                     ("product_id", "=", item.product_id.id),
                 ]
             ).mapped("name")
+        for item in self.filtered(
```
pedrobaeza: Please don't do two times the same `filtered` for performance reason.

sale_purchase_force_vendor/models/sale_order_line.py
```diff
@@ -27,15 +27,23 @@ class SaleOrderLine(models.Model):
...
+            lambda x: x.product_id
+            and not x.order_id.sale_purchase_force_vendor_restrict
+        ):
+            item.allowed_vendor_ids = partner_model.search([])
```
pedrobaeza: This is totally overkill on a populated database. Isn't there other solution?

pedrobaeza (changes requested):

victoralmau:
I have changed the approach, what do you think now?

victoralmau:
Ping @pedrobaeza

sale_purchase_force_vendor/models/sale_order_line.py
```diff
@@ -18,17 +20,36 @@ class SaleOrderLine(models.Model):
...
+        for item in self:
+            domain = (
+                [("id", "in", item.allowed_vendor_ids.ids)]
+                if item.order_id.sale_purchase_force_vendor_restrict
+                else [(1, "=", 1)]
+            )
+            item.vendor_id_domain = json.dumps(domain)
+
     @api.depends("product_id")
     def _compute_allowed_vendor_ids(self):
```
pedrobaeza: Why keeping this field?

victoralmau: You are right, `allowed_vendor_ids` field is no longer needed, it is already changed.

sale_purchase_force_vendor/views/sale_order_view.xml
```diff
@@ -12,33 +12,22 @@
...
-                    attrs="{'column_invisible': [('parent.sale_purchase_force_vendor_restrict', '=', False)]}"
-                    domain="[('id','in', allowed_vendor_ids)]"
-                    options='{"no_create": True}'
-                />
-                <field
-                    name="vendor_id"
-                    attrs="{'column_invisible': [('parent.sale_purchase_force_vendor_restrict', '=', True)]}"
+                    domain="vendor_id_domain"
                     options='{"no_create": True}'
+                    optional="hide"
```
pedrobaeza, line 20: This one should be invisible.

sale_purchase_force_vendor/tests/common.py
```diff
@@ -8,6 +8,7 @@ class TestSalePurchaseForceVendorBase(common.SavepointCase):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
+        cls.company = cls.env.company
```
pedrobaeza: Do you really need it?

sale_purchase_force_vendor/models/sale_order_line.py
```diff
@@ -18,24 +20,21 @@ class SaleOrderLine(models.Model):
...
-                ]
-            ).mapped("name")
+    def _compute_vendor_id_domain(self):
+        for item in self:
+            domain = (
+                [("id", "in", item.product_id.variant_seller_ids.name.ids)]
+                if item.order_id.sale_purchase_force_vendor_restrict
+                else [(1, "=", 1)]
+            )
+            item.vendor_id_domain = json.dumps(domain)
```
pedrobaeza, line 37: Is json dump needed?

sale_purchase_force_vendor/models/sale_order_line.py
```diff
@@ -18,24 +20,21 @@ class SaleOrderLine(models.Model):
...
-                    ("product_tmpl_id", "=", item.product_id.product_tmpl_id.id),
-                    ("product_id", "=", item.product_id.id),
-                ]
-            ).mapped("name")
+    def _compute_vendor_id_domain(self):
+        for item in self:
+            domain = (
+                [("id", "in", item.product_id.variant_seller_ids.name.ids)]
+                if item.order_id.sale_purchase_force_vendor_restrict
+                else [(1, "=", 1)]
```
pedrobaeza: Isn't `[]` working for empty domain and thus, searching for everything?

sale_purchase_force_vendor/views/sale_order_view.xml
victoralmau: According to previous approach `vendor_id` field to be displayed.

sale_purchase_force_vendor/models/sale_order_line.py
victoralmau: Yes, according to documentation: https://github.com/OCA/web/tree/14.0/web_domain_field

sale_purchase_force_vendor/views/sale_order_view.xml
pedrobaeza: Sorry, I have mixed the field `vendor_id_domain`.

pedrobaeza (approved):

chienandalu (approved):
Nice :+1:

PR #1586, opened 2022-10-04
[14.0] purchase_request consider so origin

mmequignon:
When a `stock.rule` is configured as `mts_else_mto`, we might have
`purchase.request` created with `SOXXXXXX` as origin.
This commit includes those `purchase.request` in the
`stock.warehouse.orderpoint`'s `qty_forecast` compute method.

Also, as we introduced fields in the _compute_qty method, we need to add them
in its `@api.depends`

### DIFF
--- a/purchase_request/models/orderpoint.py
+++ b/purchase_request/models/orderpoint.py
@@ -9,14 +9,15 @@ class Orderpoint(models.Model):
 
     def _quantity_in_progress(self):
         res = super(Orderpoint, self)._quantity_in_progress()
-        for prline in self.env["purchase.request.line"].search(
-            [
-                ("request_id.state", "in", ("draft", "approved", "to_approve")),
-                ("orderpoint_id", "in", self.ids),
-                ("purchase_state", "=", False),
-            ]
-        ):
-            res[prline.orderpoint_id.id] += prline.product_uom_id._compute_quantity(
-                prline.product_qty, prline.orderpoint_id.product_uom, round=False
-            )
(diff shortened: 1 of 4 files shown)


mmequignon:
Note: tests are coming.

purchase_request/models/orderpoint.py
```diff
@@ -1,22 +1,33 @@
...
-                ("request_id.state", "in", ("draft", "approved", "to_approve")),
-                ("orderpoint_id", "in", self.ids),
-                ("purchase_state", "=", False),
-            ]
-        ):
-            res[prline.orderpoint_id.id] += prline.product_uom_id._compute_quantity(
-                prline.product_qty, prline.orderpoint_id.product_uom, round=False
-            )
+        for orderpoint in self:
+            for prline in self.env["purchase.request.line"].search(
```
grindtildeath, line 13: Maybe a loop on the results of a `read_group` (using group by product and UOM) could be more performant than a loop on the `search` results.

grindtildeath (commented):
Looks good, but please add the tests :pray: 
Just some nitpicking, nothing blocking :wink: 

> When a stock.rule is configured as mts_else_mto

Actually it's for `make_to_order` procure method, which is the fallback in case `make_to_stock` cannot be used in the `mts_els_mto` stock rule configuration.

purchase_request/models/orderpoint.py
grindtildeath: Do you have time to address this one?

PR #1589, opened 2022-10-06
[13][ADD] purchase_origin_link: add a link to the Source Document

JudithAforgeFlow:
This module adds a clickable link to the "Source Document" in Purchase Orders.

CC @Forgeflow

### DIFF
--- a/purchase_origin_link/models/purchase.py
+++ b/purchase_origin_link/models/purchase.py
@@ -1,0 +1,27 @@
+# Copyright 2022 ForgeFlow S.L. (https://www.forgeflow.com)
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
+
+from odoo import api, fields, models
+
+
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+
+    origin_reference = fields.Reference(
+        selection="_selection_reference_model",
+        compute="_compute_source_doc_ref",
+        readonly=True,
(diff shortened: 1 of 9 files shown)


purchase_origin_link/models/purchase.py
```diff
@@ -0,0 +1,27 @@
...
+        return []
+
+    @api.model
+    def _get_depends_compute_source_doc_ref(self):
+        return ["origin"]
+
+    @api.depends(lambda x: x._get_depends_compute_source_doc_ref())
+    def _compute_source_doc_ref(self):
+        for rec in self:
+            rec.origin_reference = False
```
MiquelRForgeFlow, line 27:
```suggestion
            if not rec.origin:
                rec.origin_reference = False
            else:
               rec.origin_reference = ...
```
🤔

LoisRForgeFlow: @MiquelRForgeFlow it seems this module does nothing by itself, it needs some submodules to compute links to other documents (sales for instance -> #1590 ).

@JudithAforgeFlow I think it is worth to mention this on the readme to avoid confusion.

JudithAforgeFlow:
@LoisRForgeFlow @MiquelRForgeFlow

MiquelRForgeFlow (approved):

JordiMForgeFlow (approved):
LGTM 👍🏼

PR #1600, opened 2022-10-19
[14.0][FIX]product_form_purchase_link: replace purchased_product_qty by purchase_lines_count

mariadforgeflow:
Fixes same issue as in https://github.com/OCA/sale-workflow/pull/2120/

The purchase_lines_count field calculates the sum of the "product_uom_qty" of that product for each purchase line instead of the number of purchase lines.

Now, the number in the button shows the correct number of purchase lines.

### DIFF
--- a/product_form_purchase_link/__init__.py
+++ b/product_form_purchase_link/__init__.py
@@ -1,0 +1,1 @@
+from . import models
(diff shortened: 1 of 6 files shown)


MiquelRForgeFlow (approved):

LoisRForgeFlow (approved):
LGTM

PR #1185, opened 2021-05-24
 [IMP][12.0] purchase_discount / purchase_triple_discount : add Apply button to mass update supplierinfos

legalsylvain:
**Impact ``purchase_discount`` and ``purchase_triple_discount``**

- add new buttons "Apply" in the partner form view, near default supplierinfo value. (1 / 2 / 3) 

image

- add a warning message, when clicking on them. If confirmed, all the supplierinfo will be updated.

image

- add tests

- update ``usage.rst`` files.

**Merge process** : ``/minor``

CC : @quentinDupont #GRAPOCA (apps/deck/#/board/144/card/1552)

PR #1602, opened 2022-10-19
[16.0][MIG] purchase_discount: Migration to 16.0

ramiadavid:


legalsylvain (approved):
Hi @ramiadavid. Thanks for porting this usefull modules.

Diff review / No test. LGTM.

could you take a look on that pending V12 PR : https://github.com/OCA/purchase-workflow/pull/1185 (and cherry pick if you agree)
When a supplier changes its default discount, it allows to mass apply the value in all the supplierinfo.

purchase_discount/tests/test_product_supplierinfo_discount.py
```diff
@@ -47,32 +47,38 @@ def setUpClass(cls):
...
-        self.assertEqual(
-            self.po_line_1.discount,
-            10,
-            "Incorrect discount for product 6 with partner 3 and qty 1: "
-            "Should be 10%",
-        )
+        purchase_form = Form(self.purchase_order)
+        with purchase_form.order_line.edit(0) as line:
+            line.product_qty = 1.0
+
```
pedrobaeza: Don't add empty lines inside of a method

purchase_discount/models/product_supplierinfo.py
```diff
@@ -10,11 +10,13 @@ class ProductSupplierInfo(models.Model):
 
     discount = fields.Float(string="Discount (%)", digits="Discount")
 
-    @api.onchange("name")
-    def onchange_name(self):
+    @api.onchange("partner_id")
+    def onchange_partner_id(self):
```
pedrobaeza: Instead of this, do `discount` a computed writable field depending on `partner_id`:

```python
    discount = fields.Float(string="Discount (%)", digits="Discount", compute="_compute_discount", store=True, readonly=False)

    @api.depends("partner_id")
    def _compute_discount(self):
        for record in self:
            record.discount = record.partner_id.default_supplierinfo_discount
```

purchase_discount/tests/test_purchase_discount.py
```diff
@@ -22,16 +22,16 @@ def setUpClass(cls):
         currency_rate_model = cls.env["res.currency.rate"]
         # Set the Exchange rate for the currency of the company to 1
         # to avoid issues with rates
-        latest_currency_rate_line = currency_rate_model.search(
+        la_test_currency_rate_line = currency_rate_model.search(
```
pedrobaeza: This variable name has been changed without a reason.

setup/purchase_discount/odoo/addons/purchase_discount
```diff
@@ -0,0 +1 @@
+../../../../account_invoice_triple_discount
```
pedrobaeza: This seems to be not correct.

pedrobaeza (changes requested):

pedrobaeza:
I almost forgot: now on write, the price unit is synchronized with the stock move, so we should:

- Take into account discount for this computation.
- Write it also on discount field change.

purchase_discount/tests/test_product_supplierinfo_discount.py
ramiadavid: done

purchase_discount/models/product_supplierinfo.py
ramiadavid: done

setup/purchase_discount/odoo/addons/purchase_discount
ramiadavid: done

pedrobaeza:
Can you please check my last comment?

ramiadavid:
> Can you please check my last comment?

Yes, i'm checking it out

purchase_discount/tests/test_purchase_discount.py
ramiadavid: done

ramiadavid:
Hi @pedrobaeza 
I am not sure if I have correctly understood what you are requesting.

I used the same method found in `purchase_stock` to sync the `price_unit` with `stock_move` but changed the value to the discounted price.
Is that correct? I have added some tests to check its correct operation.

purchase_discount/models/purchase_order.py
```diff
@@ -0,0 +1,149 @@
...
+    @api.model
+    def _prepare_purchase_order_line_from_seller(self, seller):
+        """Overload this function to prepare other data from seller,
+        like in purchase_triple_discount module"""
+        if not seller:
+            return {}
+        return {"discount": seller.discount}
+
+    def write(self, vals):
+        res = super(PurchaseOrderLine, self).write(vals)
```
pedrobaeza:
```suggestion
        res = super().write(vals)
```

purchase_discount/models/purchase_order.py
```diff
@@ -0,0 +1,149 @@
...
+        if "discount" in vals or "price_unit" in vals:
+            for line in self.filtered(lambda l: l.order_id.state == "purchase"):
+                # Avoid updating kit components' stock.move
+                moves = line.move_ids.filtered(
+                    lambda s: s.state not in ("cancel", "done")
+                    and s.product_id == line.product_id
+                )
+                moves.write({"price_unit": line._get_discounted_price_unit()})
+
```
pedrobaeza: Please don't introduce empty lines inside a method.

pedrobaeza (commented):
The write method seems OK. You mentioned a test. Where is it?

purchase_discount/tests/test_purchase_discount.py
```diff
@@ -0,0 +1,192 @@
...
+        self.assertAlmostEqual(self.po_line_1.price_unit, 10.0)
+        self.assertAlmostEqual(self.po_line_1.discount, 50.0)
+
+    def test_move_price_unit_discount_sync(self):
+        self.purchase_order.button_confirm()
+        picking = self.purchase_order.picking_ids
+        moves = picking.move_ids
+        self.po_line_1.discount = 25
+        self.po_line_2.discount = 50
+        self.po_line_3.discount = 10
```
ramiadavid, line 148: @pedrobaeza this is the test

purchase_discount/models/purchase_order.py
ramiadavid: done

purchase_discount/models/purchase_order.py
ramiadavid: done

pedrobaeza (approved):

pedrobaeza:
Don't get why this is not merged... cc @sbidoul 

Let's try again:

/ocabot merge nobump

setup/purchase_discount/odoo/addons/purchase_discount
```diff
@@ -0,0 +1 @@
+../../../../purchase_discount
```
sbidoul, line 1: This seems to be a regular file, instead of a symlink.
Due to this, the bot can't build a wheel and fails.

Why the bot does not report the error is unknown at this stage and tracked in https://github.com/OCA/oca-github-bot/issues/221

sbidoul: @ramiadavid can you remove this file and run `pre-commit run --all-files` again?

ramiadavid: Hi @sbidoul
I work from Windows and pre-commit fails when it tries to create that file, I usually create it manually, but I don't know if it's correct.

sbidoul: @ramiadavid ha, symlinks on windows may be complicated. I could push the fix to your branch.

ramiadavid: @sbidoul Thanks,
I'll check to see how I can make it work, for other migrations.

ramiadavid:
Hi @pedrobaeza  
the problem is already solved, can you retry the merge, please?

PR #1604, opened 2022-10-24
purchase_request: fix cancel purchase_request_line

mmequignon:
With odoo core, when a purchase order line is cancelled, the `move_dest_ids`'s procure methods are updated, as seen here https://github.com/odoo/odoo/blob/14.0/addons/purchase_stock/models/purchase.py#L109-L110

With purchase_request, we should have the same behavior, if a purchase_request_line related to stock_moves is cancelled, those moves shouldn't be `Waiting for another operation anymore`.

### DIFF
--- a/purchase_request/models/purchase_request_line.py
+++ b/purchase_request/models/purchase_request_line.py
@@ -298,10 +298,20 @@ class PurchaseRequestLine(models.Model):
 
     def do_cancel(self):
         """Actions to perform when cancelling a purchase request line."""
+        # Changing the procure method, since related moves should pick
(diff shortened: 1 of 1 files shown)


mmequignon:
note tests are coming

ivantodorovich (approved):
Makes sense

i-vyshnevska (approved):

simahawk:
@LoisRForgeFlow or @AaronHForgeFlow good for you?

AaronHForgeFlow (approved):
It makes sense IMO.

PR #1612, opened 2022-10-26
[IMP] added UTC time to te scheduled_date

JoelZilli:


JoelZilli:
This PR was created because the request date originated form the purchase request does not match the scheduled date from the purchase order. By adding the "utcnow()" we save the date_planned as UTC so that when you pick it up t takes it in the user's TZ

PR #1588, opened 2022-10-05
[15.0][MIG] purchase_order_approved

MiquelRForgeFlow:
Supersede of https://github.com/OCA/purchase-workflow/pull/1312.

PR #1030, opened 2020-11-27
[14.0][MIG] purchase_order_approved

SurachetS:
standard migration

PR #1616, opened 2022-10-28
[FIX] purchase_order_approved: Allow to receive products when in 'Purchase' state

rousseldenis:
…chase' state

In core module, service products can be received by editing the received quantity on 'purchase' order state. This restablishes that behaviour

### DIFF
--- a/purchase_order_approved/views/purchase_order_view.xml
+++ b/purchase_order_approved/views/purchase_order_view.xml
@@ -26,7 +26,7 @@
             <field name="order_line" position="attributes">
                 <attribute
                     name="attrs"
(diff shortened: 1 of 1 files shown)


rousseldenis:
Related to https://github.com/OCA/purchase-workflow/pull/1030#discussion_r1007750574

@kittiu @HviorForgeFlow IMHO this should be included in https://github.com/OCA/purchase-workflow/pull/1588

rousseldenis:
Note: the line is protected on that state unless product is a service.

HviorForgeFlow (approved):

PR #1618, opened 2022-10-31
[14.0] purchase_request: fix done purchase_request_line

mmequignon:
A purchase request can be set to done even if it hasn't been processed. Meaning that there's no purchase order for any of the request lines move_dest_ids, that might be waiting.
In the end, this is an incomplete process, and the related moves will hang forever for a purchase process that will never end.

This PR also add tests for this PR https://github.com/OCA/purchase-workflow/pull/1604 that has already been merged.

### DIFF
--- a/purchase_request/models/purchase_request.py
+++ b/purchase_request/models/purchase_request.py
@@ -281,6 +281,12 @@ class PurchaseRequest(models.Model):
         return self.write({"state": "rejected"})
 
     def button_done(self):
+        incomplete_lines = self.line_ids.filtered(
+            lambda l: l.purchased_qty < l.product_qty
+        )
+        # Related move lines will hang out forever as waiting for another operation
+        # if requests lines aren't completely purchased.
+        incomplete_lines._set_dest_move_as_mts()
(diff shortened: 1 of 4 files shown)


mmequignon:
@LoisRForgeFlow or @AaronHForgeFlow does it looks ok for you?

JuMiSanAr (approved):

simahawk (approved):

AaronHForgeFlow (approved):
Ok to me!

bunhin:
@mmequignon,  is it having a mis placing of function description?

def _set_dest_move_as_mto(self):
        """Sets related moves as `make_to_order`.
        This should be called when a purchase line is cancelled or rejected,
        so the related moves won't wait for this purchase request to be done.
        """

and 

def _set_dest_move_as_mts(self):
        """Sets related moves as `make_to_stock`.
        This should be called when a PRL is reset, so related moves will
        wait back for it to be processed, instead of picking goods from stock.
        """

but i am sorry if i have mis-understand it.

PR #1624, opened 2022-11-08
Added border of line.discount

altela:
Added border for line.discount.

before : 
Screen Shot 2022-11-08 at 10 53 02

after : 
Screen Shot 2022-11-08 at 10 53 08


legalsylvain (approved):
Thanks.

(Commits should be squashed, imo)

PR #1626, opened 2022-11-09
[FIX] purchase_blanket_order: fix blanket order automatically setting…

JuanyDForgeflow:
@forgeflow

fix blanket order automatically setting to state open after being cancelled and edited on draft

current behavior:
order changes states from draft to open after editing lines on an previously cancelled purchase blanket order

desired behavior:
order stays in draft state after editing lines

### DIFF
--- a/purchase_blanket_order/models/blanket_orders.py
+++ b/purchase_blanket_order/models/blanket_orders.py
@@ -74,6 +74,7 @@ class BlanketOrder(models.Model):
         states={"draft": [("readonly", False)]},
     )
     confirmed = fields.Boolean(copy=False)
+    cancelled = fields.Boolean(copy=False)
     state = fields.Selection(
         selection=[
             ("draft", "Draft"),
@@ -178,6 +179,7 @@ class BlanketOrder(models.Model):
         "line_ids.remaining_uom_qty",
         "validity_date",
         "confirmed",
+        "cancelled",
     )
     def _compute_state(self):
         today = fields.Date.today()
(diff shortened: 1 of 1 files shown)


purchase_blanket_order/models/blanket_orders.py
```diff
@@ -185,9 +186,9 @@ def _compute_state(self):
             "Product Unit of Measure"
         )
         for order in self:
-            if not order.confirmed:
+            if not order.confirmed and not order.cancelled:
```
JordiBForgeFlow, line 190: If you add a field here then you must also add it in the api.depends. Otherwise there will be cases where the field value won't be computed correctly

purchase_blanket_order/models/blanket_orders.py
```diff
@@ -298,7 +299,7 @@ def action_cancel(self):
                                 "Try to cancel them before."
                             )
                         )
-            order.write({"state": "expired"})
+            order.write({"state": "expired", "cancelled": True, "confirmed": False})
```
JordiBForgeFlow: Dies not make sense to write the state 'expired' because the "cancelled" and "confirmed" fields will drive the correct state during the computation

purchase_blanket_order/models/blanket_orders.py
```diff
@@ -269,7 +270,7 @@ def _validate(self):
 
     def set_to_draft(self):
         for order in self:
-            order.write({"state": "draft"})
+            order.write({"state": "draft", "cancelled": False})
```
JordiBForgeFlow: Does not make sense to write the state "draft".

JuanyDForgeflow: @JordiBForgeFlow It's done now

purchase_blanket_order/models/blanket_orders.py
JuanyDForgeflow: @JordiBForgeFlow It's done now

purchase_blanket_order/models/blanket_orders.py
JuanyDForgeflow: @JordiBForgeFlow It's done now

JordiBForgeFlow (approved):

PR #1628, opened 2022-11-10
[13.0][FIX] purchase_open_qty: Quantity to invoice less than 0 has to be to invoice in purchase.order.lines too

CarlosRoca13:
cc @Tecnativa TT40418

ping @pedrobaeza @chienandalu

### DIFF
--- a/purchase_open_qty/models/purchase_order.py
+++ b/purchase_open_qty/models/purchase_order.py
@@ -3,7 +3,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
 from odoo import _, api, fields, models
-from odoo.tools import float_compare, float_is_zero
+from odoo.tools import float_is_zero
 
 
 class PurchaseOrderLine(models.Model):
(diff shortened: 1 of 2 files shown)


purchase_open_qty/views/purchase_view.xml
```diff
@@ -63,7 +63,7 @@
                 <filter
                     name="pending_qty_to_invoice"
                     string="Pending Qty to Bill"
-                    domain="[('qty_to_invoice','>',0.0)]"
+                    domain="[('qty_to_invoice','!=',0.0)]"
```
pedrobaeza, line 66: Shouldn't you use a similar technique creating a computed field with `float_is_zero` (or `float_compare` comparison?

CarlosRoca13: In this case, the field is stored, so we can use the main field to check it, so I think that is not necesary to make another field

pedrobaeza: The problem is not to be stored or not, but the rounding precision. A stored value of 0.0000001 will appear in this filter.

CarlosRoca13: I think, now this problem shouldn't appear

purchase_open_qty/models/purchase_order.py
```diff
@@ -19,12 +19,14 @@ def _compute_qty_to_invoice(self):
         for line in self:
             if line.product_id.purchase_method == "receive":
                 qty = line.qty_received - line.qty_invoiced
-                if float_compare(qty, 0.0, precision_digits=precision):
-                    line.qty_to_invoice = qty
-                else:
-                    line.qty_to_invoice = 0.0
+                if float_is_zero(qty, precision_digits=precision):
```
pedrobaeza, line 24: Put a comment indicating that this is done for avoid floats like 0.00001 that don't match the filter `qty_to_invoice != 0`

CarlosRoca13:
Added

pedrobaeza (approved):

pedrobaeza:
We can fast-track this one to complete the fix:

/ocabot merge minor

PR #1634, opened 2022-11-15
Draft: [ADD] purchase_autolock in v13

AaronHForgeFlow:
This module means to add a cron job that locks PO where no more activity is expected. This is needed in companies with high volume of purchases. Only the ones that cannot be autolocked under some criteria are reviewed by the purchase team.

cc @ForgeFlow


AaronHForgeFlow:
This can be resolved using the base_automation module. Closing for now.

PR #1642, opened 2022-11-18
[14.0][FIX] purchase_order_type tests lint fixing

HviorForgeFlow:
Maybe not the better way to fix it, but it works

CC @ForgeFlow

@dreispt @rousseldenis


HviorForgeFlow:
@pedrobaeza can you take a look here, thanks!

pedrobaeza (approved):
Although I don't know the original problem, it seems correct enough. I wouldn't use any demo data in `purchase_manual_currency`, but it's a bigger change.

HviorForgeFlow:
Branch 14.0 is broken due to this errors:
- https://github.com/OCA/purchase-workflow/actions/runs/3496378949/jobs/5854224561#step:8:39
- https://github.com/OCA/purchase-workflow/actions/runs/3496378949/jobs/5854224561#step:8:225

PR #1653, opened 2022-11-25
[MIG] purchase_request: Migration to 16.0

StefanRijnhart:
* Removed the manual onchange calling as it has been replaced by compute methods in the Odoo `purchase` module.
* Replaced analytic account + tags with analytic distribution (incl. the mixin model). As for the related data, I'm hoping we can rely on the various migration platforms (enterprise, openupgrade) for that.
* Replaced `attrs` like `{'invisible': True}` by `invisible="1"` because the web client crashed on `domain.map is not a function`.
* Removed override of stock.move's `_prepare_merge_move_sort_method` which was removed after Odoo 14 (https://github.com/odoo/odoo/commit/bda3225c6cc27bb2c2933eda3cc68c6f9708daf3)
* Increase test coverage, clarify "split" test
* Fix incorrect field path in stock.move's `_compute_purchase_request_ids` (discovered while increasing test coverage)
* Fix issue with allocation when splitting moves (discovered while increasing test coverage)


dreispt (approved):

Nikul-OSI (approved):
Code Review LGTM :+1: 
Ready to merge :rocket: 

Thank you @StefanRijnhart

StefanRijnhart:
@dreispt @Nikul-OSI For the record, after your reviews I increased test coverage to achieve Codecov's green checkmark, and this uncovered two actual bugs that I was able to fix:

* Incorrect field path in stock.move's _compute_purchase_request_ids
* Incorrect allocation quantities after splitting moves

I left this bit in a separate commit so that it can be backported more easily.

PR #676, opened 2019-04-29
[11][ADD] Purchase Merge module creation

Tonow-c2c:
Add wizard to merge purchase order with criterion's

PR #1679, opened 2022-12-13
[15.0][ADD] Purchase Order Merge

BojanOD:
Merge Purchase Orders

- https://github.com/oca/purchase-workflow/pull/676


ivantodorovich (changes requested):
Pleasse add some unit tests 🙏🏻

purchase_merge/wizard/purchase_merge.py
```diff
@@ -0,0 +1,226 @@
...
+
+        po.message_post(body=body, subject=subject, content_subtype="plaintext")
+
+    def _merge(self, purchases, dst_purchase=None):
+        """private implementation of merge purchase
+        :param purchases : ids of purchase to merge
+        :param dst_purchase : record of destination purchase.order
+
+        # TODO: to improved by using
+        OCA/openupgradelib/openupgrade_merge_records.py
```
ivantodorovich:
```suggestion
```

Seems like it's already implemented

purchase_merge/wizard/purchase_merge.py
```diff
@@ -0,0 +1,226 @@
...
+                start_msg,
+                _("payment terms"),
+                ", ".join([str(x) for x in payment_term_names]),
+            )
+        if len(fiscal_position_names) > 1:
+            error_msg += "{} {}: {}\n".format(
+                start_msg,
+                _("fiscal positions"),
+                ", ".join([str(x) for x in fiscal_position_names]),
+            )
```
ivantodorovich, line 103: This `start_msg` / `error_msg` hack might work on english, but some translations may require a different gramatical structure. So this string concatenation won't work for them. Plus, it'll be a nightmare to translate as separated strings.

I recommend the not lazy approach and translate the full string for each case.
Also, compare records .. not record names.

I'd do something like this:

```py
error_messages = []

currencies = purchase_orders.currency_id
if len(currencies) > 1:
    error_messages.append(
        _(
            "You can't merge purchase orders with different currencies: %s",
            ", ".join(currencies.mapped("name"))
        )
    )

picking_types = purchase_orders.picking_type_id
if len(picking_types) > 1:
    error_messages.append(
        _(
            "You can't merge purchase orders with different picking types: %s",
            ", ".join(picking_types.mapped("name"))
        )
    )

# ... all the other checks ...

if error_messages:
    raise UserError("\n".join(error_messages))
```

purchase_merge/readme/ROADMAP.rst
```diff
@@ -0,0 +1 @@
+* it could be improved by using [https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade_merge_records.py](https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade_merge_records.py)
```
ivantodorovich: Remove this file. It's already implemented

ivantodorovich (changes requested):

imlopes:
@BojanOD 
Can you add the security files for the access to the wizard ?
https://www.odoo.com/pt_BR/forum/ajuda-1/why-access-rights-is-must-for-transient-model-in-odoo-14-181312

Without the security file, only Super User can access the wizard to merge.

purchase_merge/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_purchase_merge_automatic_wizard_user,access_purchase_merge_automatic_wizard_user,model_access_purchase_merge_automatic_wizard,base.group_user,1,1,1,1
```
imlopes: @BojanOD 

Any reason to use base.group_user instead Purchase User group ?

BojanOD: @imlopes 

You are suggesting that I should use `purchase.group_purchase_user` instead?

imlopes: For security reason, I think it's better. Even if a user without purchasing rights is prevented from seeing the wizard menu (merge), it is better to add the purchasing group..

imlopes (approved):
LGTM

ivantodorovich (approved):

yankinmax:
@rvalyi @simahawk can we get it merged pls?

purchase_merge/wizard/purchase_merge.py
```diff
@@ -0,0 +1,225 @@
...
+        """
+        if len(purchases) < 2:
+            return
+        record_ids = purchases - dst_purchase
+        openupgrade_merge_records.merge_records(
+            env=self.env,
+            model_name=self._name,
+            record_ids=record_ids.ids,
+            target_record_id=dst_purchase.id,
+        )
```
quoc-pn, line 183: Hi @BojanOD,
This module is working great—thanks for the work on it.

I do have a question regarding the logic. It looks incorrect to me.
The model_name parameter refers to a wizard model (purchase.merge.automatic.wizard), but we're trying to merge purchase.order records. Both record_ids and target_record_ids belong to purchase.order.

With the current setup, the logic always returns None.
I’m testing this on PR #2544.

Could you take a look?

PR #1688, opened 2022-12-22
[15.0][MIG] purchase_order_general_discount: Migration to version 15.0

pilarvargas-tecnativa:
The general discount field is not positioned correctly, possibly due to the styling of the new tax widget in version 15.

@chienandalu please review the purchase_order_view.xml file to make sure that the general discount field is set correctly.

@Tecnativa TT38331


purchase_order_general_discount/models/purchase_order.py
```diff
@@ -0,0 +1,78 @@
...
+        ),
+    ]
+
+    @api.onchange("partner_id")
+    def onchange_partner_id(self):
+        super().onchange_partner_id()
+        self.general_discount = (
+            self.partner_id.commercial_partner_id.purchase_general_discount
+        )
+        return super().onchange_partner_id()
```
chienandalu: This is not the way to solve the pre-commit warning. You're running super() twice here.

```suggestion
        res = super().onchange_partner_id()
        self.general_discount = (
            self.partner_id.commercial_partner_id.purchase_general_discount
        )
        return res
```

chienandalu (approved):
I don't think we can get much better in a pure xml inheritance approach. We'd need to patch the OWL component and its template to make the perfect fit. Indeed, probably a new module would be needed as a commonground for both purchase and sale general discounts.

chienandalu:
Please review @CarlosRoca13

chienandalu (changes requested):
Sorry I was requesting changes before :)

purchase_order_general_discount/models/purchase_order.py
pilarvargas-tecnativa: You're right, it's already corrected, thanks for your teachings.

pilarvargas-tecnativa:
> Please review @CarlosRoca13

The changes have been made and you can review it @CarlosRoca13  :)

CarlosRoca13 (approved):
LGTM :+1:

chienandalu (approved):
:+1:

PR #1689, opened 2022-12-22
15.0 fix default note purchase order

valentinagiuliani-rapsodoo:
Adding check on function onchange_partner_id on model purchase.order: suppose we have defined a specific default_note on the partner and after deleted it: since the field is html, the <p><br/></p> value remains in the db, which affects the onchange, returning the <p><br/></p> on the purchase.order note.

### DIFF
--- a/purchase_default_terms_conditions/models/purchase.py
+++ b/purchase_default_terms_conditions/models/purchase.py
@@ -2,6 +2,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import api, models
+from odoo.tools import is_html_empty
 
 
 class PurchaseOrder(models.Model):
@@ -10,7 +11,7 @@ class PurchaseOrder(models.Model):
 
     @api.onchange("partner_id", "company_id")
     def onchange_partner_id(self):
-        if self.partner_id.purchase_note:
+        if not is_html_empty(self.partner_id.purchase_note):
             self.notes = self.partner_id.purchase_note
         elif (
             self.env["ir.config_parameter"]


purchase_default_terms_conditions/models/purchase.py
```diff
@@ -10,7 +10,10 @@ class PurchaseOrder(models.Model):
 
     @api.onchange("partner_id", "company_id")
     def onchange_partner_id(self):
-        if self.partner_id.purchase_note:
+        if (
+            self.partner_id.purchase_note
+            and self.partner_id.purchase_note != "<p><br/></p>"
```
quannguyen1560: I think you can try "is_html_empty"  by `from odoo.tools import is_html_empty` to check the purchase_note is empty or not.

quannguyen1560 (changes requested):

purchase_default_terms_conditions/models/purchase.py
valentinagiuliani-rapsodoo: Done @quannguyen1560.. Can you check?

purchase_default_terms_conditions/models/purchase.py
```diff
@@ -10,7 +11,9 @@ class PurchaseOrder(models.Model):
 
     @api.onchange("partner_id", "company_id")
     def onchange_partner_id(self):
-        if self.partner_id.purchase_note:
+        if self.partner_id.purchase_note and not is_html_empty(
```
quannguyen1560: You can remove `self.partner_id.purchase_note` because `is_html_empty` can check for the case the case field is empty or not as well. It would be nice if you can try and check it again.
thank you

quannguyen1560 (changes requested):

purchase_default_terms_conditions/models/purchase.py
valentinagiuliani-rapsodoo: Done @quannguyen156. Can you check please?

quannguyen1560 (approved):
LGTM!

OpenCode (approved):

PR #1726, opened 2023-01-24
[WIP][15.0][ADD]: purchase_receipt_progress

imlopes:
New module to allow users to check the % of received products.
Another feature is to group information by product category :
* Qty purchased
* Qty received
* Qty to receive


purchase_receipt_progress/models/purchase_order.py
```diff
@@ -0,0 +1,89 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class PurchaseOrderReceivedByCategory(models.Model):
+    _name = "purchase.order.received.by.category"
+    _description = "Purchase Order Received by Category"
+
+    order_id = fields.Many2one("purchase.order")
```
SilvioC2C, line 11:
```suggestion
    order_id = fields.Many2one("purchase.order", required=True, ondelete="cascade")
```

purchase_receipt_progress/models/purchase_order.py
```diff
@@ -0,0 +1,89 @@
...
+    category_id = fields.Many2one("product.category")
```
SilvioC2C, line 12: What happens when a category is deleted?

purchase_receipt_progress/models/purchase_order.py
```diff
@@ -0,0 +1,89 @@
...
+    product_qty = fields.Float()
+    qty_received = fields.Float()
+    to_receive = fields.Float()
```
SilvioC2C, line 15: Since we're handling quantities, we should also add the UoM management

purchase_receipt_progress/models/purchase_order.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+
+    purchased_quantity_total = fields.Float(
+        string="Total Quantity", compute="_compute_purchased_quantity_total"
+    )
+    delivery_status = fields.Float(compute="_compute_delivery_status")
```
SilvioC2C, line 24: Same remark about UoM

purchase_receipt_progress/models/purchase_order.py
```diff
@@ -0,0 +1,89 @@
...
+            purchase_order_received_by_category_obj = self.env[
+                "purchase.order.received.by.category"
+            ]
+            order_id = (
+                order._origin.id if isinstance(order.id, models.NewId) else order.id
+            )
+            purchase_order_received_by_category_obj.search(
+                [("order_id", "=", order_id)]
+            ).unlink()
+            order.received_by_category_ids = False
```
SilvioC2C, line 60: If the idea is to remove every `purchase.order.received.by.category` before recomputing everything, it'd be better to add the `ondelete="cascade"` as I suggested above, and add `self.received_by_category_ids = False` at the beginning of the compute method. This way, the field will be emptied on every record at once (instead of emptying the field one-by-one), and the `purchase.order.received.by.category` records will all be deleted (because of the `ondelete="cascade"`).

purchase_receipt_progress/models/purchase_order.py
```diff
@@ -0,0 +1,89 @@
...
+                        "order_id": order_id,
+                        "category_id": category.id,
+                        "product_qty": sum(lines.mapped("product_qty")),
+                        "qty_received": sum(lines.mapped("qty_received")),
+                        "to_receive": sum(lines.mapped("to_receive")),
+                    }
+                    received_by_category_id = (
+                        purchase_order_received_by_category_obj.create(values)
+                    )
+                    received_by_category_ids.append((4, received_by_category_id.id))
```
SilvioC2C, line 77: Instead of creating the record and then adding the ID to the field, you can directly tell Odoo that you want a new record to be created when assigning it to the field:
```suggestion
                    received_by_category_ids.append((0, 0, values))
```
Odoo will see that the first command is `0`, will retrieve the values in third position of the tuple, and create the new record by itself when needed.

purchase_receipt_progress/models/purchase_order.py
```diff
@@ -0,0 +1,89 @@
...
+            order.received_by_category_ids = received_by_category_ids
+
+
+class PurchaseOrderLine(models.Model):
+    _inherit = "purchase.order.line"
+
+    to_receive = fields.Float(compute="_compute_to_receive")
```
SilvioC2C, line 84: Is this field necessary? There already exists the field `qty_received`, introduced by module `purchase` and updated by `purchase_stock` (which makes it compute method take into account the stock movements related to the purchase line).

SilvioC2C (changes requested):
Some remarks here and there.
Also, can you please split the `purchase_order.py` file into multiple files, each one containing a single model? It'd be easier to read.

purchase_receipt_progress/models/purchase_order.py
gurneyalex: it is not required, so presumably nothing (ondelete set null)

purchase_receipt_progress/models/purchase_order.py
gurneyalex: if you need to be able to sum these two fields in pivot tables, consider storing the fields.

purchase_receipt_progress/models/purchase_order.py
```diff
@@ -0,0 +1,89 @@
...
+        "purchase.order.received.by.category",
+        "order_id",
+        string="Received by Category",
+        compute="_compute_received_by_category_ids",
+    )
+
+    @api.depends("order_line.product_qty")
+    def _compute_purchased_quantity_total(self):
+        for order in self:
+            order.purchased_quantity_total = sum(order.mapped("order_line.product_qty"))
```
gurneyalex, line 35: use read_group for performance

purchase_receipt_progress/models/purchase_order.py
```diff
@@ -0,0 +1,89 @@
...
+
+    @api.depends("order_line.product_qty", "order_line.qty_received")
+    def _compute_delivery_status(self):
+        for order in self:
+            received_quantity = sum(order.mapped("order_line.qty_received"))
```
gurneyalex, line 40: consider using read_group outside of the loop for performance reason.

purchase_receipt_progress/models/purchase_order.py
gurneyalex: in any case, avoid calling search() in a loop

purchase_receipt_progress/models/purchase_order.py
```diff
@@ -0,0 +1,89 @@
...
+                if lines:
+                    values = {
+                        "order_id": order_id,
+                        "category_id": category.id,
+                        "product_qty": sum(lines.mapped("product_qty")),
+                        "qty_received": sum(lines.mapped("qty_received")),
+                        "to_receive": sum(lines.mapped("to_receive")),
+                    }
+                    received_by_category_id = (
+                        purchase_order_received_by_category_obj.create(values)
```
gurneyalex, line 75: accumulate the dicts to be passed to create() in a list and call create only once after the loop

purchase_receipt_progress/models/purchase_order.py
imlopes: @SilvioC2C @gurneyalex 
Hmmm.. I don't know !
Here, the objective asked is to sum the quantities by category.
If we need to manage the UoM, we need to to group by category then by UoM because in the same product category maybe we'll have two (or more) different units..

purchase_receipt_progress/models/purchase_order.py
ivantodorovich, line 84:
```suggestion
    qty_to_receive = fields.Float(
        compute="_compute_to_receive",
        digits='Product Unit of Measure',
        store=True,
    )
```

* Naming: `qty_to_receive` is more consistent with existing field `qty_received`.
* Stored: as `qty_received` is stored, let's benefit this for reporting
* Digits: should be set for uom float fields

imlopes: Yes ! This field is the `purchased quantity - received quantity`. Is more direct for end-users to see the remaining quantity to receive.

purchase_receipt_progress/models/purchase_order.py
ivantodorovich: Kind of overlaps with existing module [`purchase_reception_status`](https://github.com/OCA/purchase-workflow/tree/15.0/purchase_reception_status), althuogh that one is computing only a 'state', and not an actual percentage.

In any case, I think `delivery_status` should be renamed as `reception_percentage`, and if really needed, extracted out into a separate module that only does that

purchase_receipt_progress/models/purchase_order.py
imlopes: I didn't get it. Can you show me a example ?

ivantodorovich (commented):
IMO the "reception percentage" and the "reception state by product category" features should be in two different modules.

SilvioC2C:
These PR features have been split into 4 different modules, see:
#1731 
#1732 
#1735 
#1736 

@imlopes can you please review them and, if it's ok for you, close this PR?

PR #1727, opened 2023-01-26
[15.0][IMP] purchase_request - multi currency

Saran440:
Add field currency in purchase request and send it to RFQ

Purchase Request
- Add currency view and can see with multi currency group
Selection_024

RFQ
- Received currency from Purchase Request
Selection_025

PR #1728, opened 2023-01-26
[15.0][ADD] purchase_request_manual_currency

Saran440:
This module allows users to update the currency manual of Purchase Request when it is in draft state.
It is used to keep track of the exchange rate on the day when the purchase request is viewed,
even though the purchase request may not be created on the same day.
And the exchange rate can be used by the users to make decisions.

Selection_023
When users create PR. it will compute estimated cost with company currency by default rate or manual rate

Depends:
- [x] https://github.com/OCA/purchase-workflow/pull/1727
- [x] https://github.com/OCA/purchase-workflow/pull/1765


Taewii23 (approved):
Ok

PR #1731, opened 2023-01-27
[15.0][ADD] purchase_total_ordered_qty

SilvioC2C:


purchase_total_ordered_qty/models/purchase_order.py
```diff
@@ -0,0 +1,66 @@
...
+                ordered_qty_by_uom[ref] += uom._compute_quantity(qty, ref, round=False)
+            order.total_ordered_qty_json = json.dumps(
+                {u.id: q for u, q in ordered_qty_by_uom.items()}
+            )
+
+    @api.depends("total_ordered_qty_json")
+    @api.depends_context("lang")
+    def _compute_total_ordered_qty_text(self):
+        lang = self.env.context.get("lang") or self.env.user.lang
+        uom_browse = self.env["uom.uom"].with_context(lang=lang).browse
```
ivantodorovich, line 57:
```suggestion
```
This isn't needed.
You don't need to add `with_context(lang=lang)` because it would already be there and set, as this method already depends on context's `lang`, you can expect to find the right value in it.

ivantodorovich (approved):

imlopes (approved):

purchase_total_ordered_qty/models/purchase_order.py
SilvioC2C: I recall cases in which the lang is not correctly set in the `context`; that's why I retrieve it from the user and add it manually to the `uom.uom()` context.
Maybe this behavior has been fixed, but better safe than sorry, I guess... :sweat_smile:

PR #1732, opened 2023-01-27
[15.0][ADD] purchase_receipt_percentage

SilvioC2C:


purchase_receipt_percentage/readme/USAGE.rst
```diff
@@ -0,0 +1 @@
+No configuration is required. Install the module, and you are ready to go.
```
ivantodorovich: Then you can drop this file ;)

purchase_receipt_percentage/models/purchase_order.py
```diff
@@ -0,0 +1,39 @@
...
+            perc = sum(lines_percentages) / len(lines_percentages)
+            order.update(
+                {
+                    "receipt_percentage": perc,
+                    # NB: we divide by 100 because we want this field to be in
+                    # [0, 1], the percentage widget in the view will take care
+                    # of displaying it correctly
+                    "receipt_percentage_display": perc / 100,
+                }
+            )
```
ivantodorovich, line 43: You could do a read_group with `receipt_percentage:avg`.
With it you'd let psql do the computation + you'd compute multiple records at once, which would be a huge time saver if the field is shown on a tree view/kanban view

purchase_receipt_percentage/models/purchase_order_line.py
```diff
@@ -0,0 +1,21 @@
...
+
+    @api.depends("qty_received", "product_qty")
+    def _compute_receipt_percentage(self):
+        for line in self:
+            perc = 100  # 100% by default, in case ordered qty is 0
+            if not float_is_zero(
+                line.product_qty, precision_rounding=line.product_uom.rounding
+            ):
+                perc *= line.qty_received / line.product_qty
+            line.receipt_percentage = min(max(perc, 0), 100)  # 0 <= % <= 100
```
ivantodorovich, line 21: Maybe compute this only for lines with product.type in `consu` / `product` ?
I guess it doesn't make sense to compute for services or other kinds, right?
In those cases it should probably default to either 100% or 0%. If it's 0%, then it shouldn't be accounted for in the order's `receipt_percentage` computation

purchase_receipt_percentage/models/purchase_order.py
```diff
@@ -0,0 +1,39 @@
...
+        compute="_compute_receipt_percentage",
+        store=True,
+        help="Receipt percentage between 0% and 100%",
+    )
+    receipt_percentage_display = fields.Float(
+        compute="_compute_receipt_percentage",
+        store=True,
+        help="Technical field to be displayed in view. Its value is between"
+        " 0 and 1, the percentage widget will format the value properly",
+    )
```
imlopes, line 20: Why do you need two fields for the receipt percentage ?

purchase_receipt_percentage/models/purchase_order_line.py
```diff
@@ -0,0 +1,21 @@
...
+
+class PurchaseOrderLine(models.Model):
+    _inherit = "purchase.order.line"
+
+    receipt_percentage = fields.Float(compute="_compute_receipt_percentage", store=True)
+
+    @api.depends("qty_received", "product_qty")
+    def _compute_receipt_percentage(self):
+        for line in self:
+            perc = 100  # 100% by default, in case ordered qty is 0
```
imlopes, line 16: If you have any ordered qty, your receipt percentage is 0. Any reason to use 100% as default ?

imlopes (approved):
LGTM

purchase_receipt_percentage/readme/USAGE.rst
SilvioC2C: Didn't know that, thanks!

purchase_receipt_percentage/models/purchase_order_line.py
SilvioC2C: In some cases, you might wanna know this data for services as well; for example, a company might buy a 40-hrs training course for an employee, and wants to know how many hours have been effectively used for the training. I'd keep it as it is for the moment, and let project-specific modules handle this case if needed.

purchase_receipt_percentage/models/purchase_order.py
SilvioC2C: The first one stores the actual percentage (included between 0 and 100) and can be used in the code directly since it holds the real percentage, the other one should be used only in UI (and its value is included between 0 and 1, the `widget="percentage"` takes care of the proper conversion).

purchase_receipt_percentage/models/purchase_order_line.py
SilvioC2C: Yep, that's the idea: by default, the value is 100%, but if we the ordered qty is different than 0, then we recompute the values by dividing the received qty by the ordered qty.
Result:
- lines with 0 qty ordered => `100%` by default
- lines with X qty ordered and Y qty received =>` Y/X*100` of received percentage

imlopes: In my POV : 

- lines with 0 qty ordered: percentage received = 0% by default (because 100% means that we received anything and it's not the case)
- lines with X qty ordered and Y qty received: percentage received = Y/X*100 of received percentage

purchase_receipt_percentage/models/purchase_order_line.py
ivantodorovich: Ah, indeed. I assumed services wouldn't have `qty_received`.. but I've just checked and in fact the user can set it manually.

ivantodorovich (approved):

purchase_receipt_percentage/models/purchase_order.py
SilvioC2C: @ivantodorovich had to revert back to the old computation without `read_group` to implement the weighted average of lines' receipt percentage, otherwise we were gonna have this situation:
1. line with 99u ordered, 0u received => PO line receipt percentage = 0%
2. line with 1u ordere, 1u received => PO line receipt percentage = 100%

=> PO receipt percentage = 50%

PR #1752, opened 2023-02-07
[15.0][IMP] purchase_discount*: Improve qweb xpath

ao-landoo:
'name' is added to discount column so that discount2 and discount3 xpaths are simpler.


pedrobaeza (approved):
Straight forward, so I fast-track it:

/ocabot merge minor

PR #1764, opened 2023-02-15
[14.0][ADD] purchase_invoice_method

kluna1998:
With this module, you can force the creation of an invoice for a purchase.


purchase_invoice_method/models/purchase_order.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 2023 CreuBlanca
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+
+    invoice_on_ordered = fields.Boolean()
```
LoisRForgeFlow: I would expect this to allow me to change "in both directions", this boolean only allows to force a change from a configured invoice method based on received quantities to a forced method in a specific PO based on ordered quantities. However I cannot do it the other way around (global based on ordered, force a specific PO to be based on received).

LoisRForgeFlow (approved):
LGTM

etobella (approved):

PR #1767, opened 2023-02-18
[13.0][FIX] purchase_stock_price_unit_sync: Don't write on children SVLs

pedrobaeza:
Steps to reproduce:

- Create a purchase order with an storable product and confirm it.
- Validate the reception picking.
- Create a landed cost linked to the reception picking and validate it.
- Now change the unit price in the purchase order.

Result: Both SVLs (stock valuation layers), the one for the reception, and the landed cost one, result overwritten with the new purchase unit price, while only the reception one should.

As the landed cost SVL is put as children (through the field `stock_valuation_layer_id`), we filter out these ones from the write.

@Tecnativa

### DIFF
--- a/purchase_stock_price_unit_sync/models/purchase_order.py
+++ b/purchase_stock_price_unit_sync/models/purchase_order.py
@@ -32,7 +32,10 @@ class PurchaseOrderLine(models.Model):
                 bom_type="phantom",
             ):
                 continue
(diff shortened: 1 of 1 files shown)


rafaelbn (approved):
👍🏼 LGTM

carlosdauden (approved):

PR #1779, opened 2023-02-28
[MIG] purchase_order_approved: Migration to 16.0

rolandojduartem:


rolandojduartem:
Hi, @luisg123v, @desdelinux, @moaadbourhim, @lmignon, @LoisRForgeFlow and @kongrattapong, please, could you review this PR?

purchase_order_approved/readme/HISTORY.rst
```diff
@@ -0,0 +1,25 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~
+
+* [MIG] Migrated to v13.
+
+15.0.1.0.0 (2021-12-05)
+~~~~~~~~~~~~~~~~~~~~~~~
+
+* [MIG] Migrated to v15.
+
+16.0.1.0.0 (2022-02-28)
```
luisg123v: 2023

purchase_order_approved/tests/test_purchase_order.py
```diff
@@ -0,0 +1,120 @@
...
+                "order_id": cls.purchase_order.id,
+                "product_id": product.id,
+                "date_planned": fields.Datetime.now(),
+                "name": "Test",
+                "product_qty": 10.0,
+                "product_uom": product.uom_id.id,
+                "price_unit": 100.0,
+            }
+        )
+        cls.env.invalidate_all()
```
luisg123v: - Are you sure this is necessary? What happens if this is not included?
- If so, are you sure all needs to be invalidated?? (i.e. not enough refreshing the record or model)

luisg123v (changes requested):

purchase_order_approved/tests/test_purchase_order.py
rolandojduartem: - Yes, it is part of [tasks to do for migration](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0#tasks-to-do-in-the-migration) because the method refresh is deprecated.
- If you watch this [method in 16.0](https://github.com/odoo/odoo/blob/3bd28525d9486e2f33b8d338c39675125fa22d7a/odoo/models.py#L5884), they use env.invalidate_all after the warning.

rolandojduartem:
@luisg123v, could you review this MR, again?

purchase_order_approved/tests/test_purchase_order.py
luisg123v: I think I didn't explain myself.

I mean, was this really necessary before? I ask because, if not, we can remove it.

On the other hand, I know before all cache was invalidated. That's why the method was split into small invalidation choices, to prevent the whole cache from being invalidated on every case, because in most of them, only a specific record or model needed invalidation.

rolandojduartem: it can be removed.

rolandojduartem:
@luisg123v friendly reminder.

rolandojduartem:
@luisg123v friendly reminder.

luisg123v (approved):
LGTM :+1: 

@moylop260 could you review, please?

LoisRForgeFlow (approved):
Code review LGTM

rolandojduartem:
Hi, @moylop260, could you review/merge this PR?

PR #1780, opened 2023-03-01
[15.0][FIX] purchase_discount_*: use hook method for supplierinfo

chienandalu:
From version 15 we've got this handy _prepare_supplier_info method so we don't have to do context injection juggling to add the discount info.

please review @victoralmau @sergio-teruel

### DIFF
--- a/purchase_discount/models/product_supplierinfo.py
+++ b/purchase_discount/models/product_supplierinfo.py
@@ -15,24 +15,3 @@ class ProductSupplierInfo(models.Model):
         """Apply the default supplier discount of the selected supplier"""
         for supplierinfo in self.filtered("name"):
             supplierinfo.discount = supplierinfo.name.default_supplierinfo_discount
-
-    @api.model
-    def _get_po_to_supplierinfo_synced_fields(self):
-        """Overwrite this method for adding other fields to be synchronized
-        with product.supplierinfo.
-        """
-        return ["discount"]
-
-    @api.model_create_multi
-    def create(self, vals_list):
-        """Insert discount (or others) from context from purchase.order's
-        _add_supplier_to_product method"""
-        for vals in vals_list:
-            product_tmpl_id = vals.get("product_tmpl_id")
-            po_line_map = self.env.context.get("po_line_map", {})
-            if product_tmpl_id in po_line_map:
-                po_line = po_line_map[product_tmpl_id]
-                for field in self._get_po_to_supplierinfo_synced_fields():
-                    if not vals.get(field):
-                        vals[field] = po_line[field]
-        return super().create(vals_list)
--- a/purchase_discount/models/purchase_order.py
+++ b/purchase_discount/models/purchase_order.py
@@ -10,16 +10,10 @@ from odoo import api, fields, models
 class PurchaseOrder(models.Model):
     _inherit = "purchase.order"
 
-    def _add_supplier_to_product(self):
-        """Insert a mapping of products to PO lines to be picked up
-        in supplierinfo's create()"""
-        self.ensure_one()
-        po_line_map = {
-            line.product_id.product_tmpl_id.id: line for line in self.order_line
-        }
-        return super(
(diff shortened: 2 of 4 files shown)


sergio-teruel (approved):

victoralmau (approved):

pedrobaeza (changes requested):
It's better to not put 2 return in the method.

If there's no `vals["product_tmpl_id"]`, then you should check `vals["product_tmpl_id"]` and extract the template from it.

chienandalu:
> It's better to not put 2 return in the method.
> 
> If there's no `vals["product_tmpl_id"]`, then you should check `vals["product_tmpl_id"]` and extract the template from it.

The thing is supplierinfo records can be created either without any product info or without this context or both.

pedrobaeza:
Correct, but you still need to know the product template through the mechanism I'm talking about.

pedrobaeza:
@chienandalu I have rebased resolving conflicts (due to a new `.get("product_tmpl_id")` instead of `["product_tmpl_id"]`). Can you please check my comment to finish this?

chienandalu:
> If there's no vals["product_tmpl_id"], then you should check vals["product_tmpl_id"] and extract the template from it.

Do you mean from vals["product_id"]?

pedrobaeza:
Yes, sorry.

pedrobaeza:
@chienandalu can you please finish this?

chienandalu:
The main issue was already fixed here: https://github.com/OCA/purchase-workflow/pull/2079

Anyway, I implemented your request, although I'm not completly sure it's needed for the case of this context (the creation of the supplierinfo from the method `_add_supplier_to_product`, which should be quite predictable

purchase_discount/models/product_supplierinfo.py
```diff
@@ -27,9 +27,19 @@ def _get_po_to_supplierinfo_synced_fields(self):
     def create(self, vals_list):
         """Insert discount (or others) from context from purchase.order's
         _add_supplier_to_product method"""
-        for vals in vals_list:
-            product_tmpl_id = vals.get("product_tmpl_id")
-            po_line_map = self.env.context.get("po_line_map", {})
+        po_line_map = self.env.context.get("po_line_map")
+        if not po_line_map:
```
pedrobaeza: I don't understand what this exit condition does.

purchase_discount/models/product_supplierinfo.py
```diff
@@ -27,9 +27,19 @@ def _get_po_to_supplierinfo_synced_fields(self):
...
+            return super().create(vals_list)
+        for vals in filter(
+            lambda x: x.get("product_tmpl_id", x.get("product_id")), vals_list
```
pedrobaeza: This filter seems now incorrect if we fallback to the other value.

purchase_discount/models/product_supplierinfo.py
chienandalu: Well, I was in my way to explain how this is a dict injected in context with the only purpose of passing the discount values on creation when I realized that in this version there's a hook method that makes this engineering unnecessary

pedrobaeza (approved):
This is way better!

/ocabot merge patch

purchase_discount/models/product_supplierinfo.py
chienandalu: Not important now, but it just meant that we could go with that dictionary as we could get the template from the product_id

rousseldenis:
@chienandalu Do you plan to forwardport this in 16 ?

chienandalu:
The thing is that the methods changed completely and now there isn't an obvious hook, so maybe an extra PR to Odoo will be needed to try solve it...

rousseldenis:
> The thing is that the methods changed completely and now there isn't an obvious hook, so maybe an extra PR to Odoo will be needed to try solve it...

Ok. I'll maybe look at it

chienandalu:
Great! ping me if you want to when you're on it :smiley:

PR #1797, opened 2023-03-08
[15.0][FIX]purchase_allowed_product: search product by vendor reference

manuelregidor:
At the moment, when this module is installed, it is not possible to use the vendor product code to find products in purchase order lines when a vendor is selected in the order.

This fix is intended to sort that out. The current version of the module replaces the context applied by the base module purchase on the product_id field in the purchase order form view. With this fix, the original context is kept plus the new features provided by the context parameters restrict_supplier_id and use_only_supplied_product are added.

### DIFF
--- a/purchase_allowed_product/views/purchase_order_view.xml
+++ b/purchase_allowed_product/views/purchase_order_view.xml
@@ -20,19 +20,19 @@
                 expr="//field[@name='order_line']/tree/field[@name='product_id']"
                 position="attributes"
             >
-                <attribute name="context">{
-                    'restrict_supplier_id': parent.partner_id,
-                    'use_only_supplied_product': parent.use_only_supplied_product
-                }</attribute>
+                <attribute name="context" operation="update">
+                    {'restrict_supplier_id': parent.partner_id,
+                    'use_only_supplied_product': parent.use_only_supplied_product}
+                </attribute>
             </xpath>
             <xpath
                 expr="//field[@name='order_line']/form//field[@name='product_id']"
                 position="attributes"
             >
-                <attribute name="context">{
-                    'restrict_supplier_id': parent.partner_id,
-                    'use_only_supplied_product': parent.use_only_supplied_product
-                }</attribute>
+                <attribute name="context" operation="update">
+                    {'restrict_supplier_id': parent.partner_id,
+                    'use_only_supplied_product': parent.use_only_supplied_product}
+                </attribute>
             </xpath>
         </field>
     </record>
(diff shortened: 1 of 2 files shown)


manuelregidor:
@ValentinVinagre @HaraldPanten @luis-ron

HaraldPanten (approved):
Functional review

HaraldPanten:
@pedrobaeza that might be interesting for you, as it fixes a standard Odoo feature for this module.

purchase_allowed_product/views/purchase_order_view.xml
```diff
@@ -22,7 +22,11 @@
             >
                 <attribute name="context">{
                     'restrict_supplier_id': parent.partner_id,
-                    'use_only_supplied_product': parent.use_only_supplied_product
+                    'use_only_supplied_product': parent.use_only_supplied_product,
+                    'partner_id':parent.partner_id,
```
pedrobaeza: Which module adds this context keys?

pedrobaeza (commented):
The proper patch is:

- Depends on `base_view_inheritance_extension` and add the context key cleanly.
- Or overwrite `fields_view_get`, and inject the context key through it.

If not, this is fixing that specific problem, but any extra module that injects other context key will have the same problem.

purchase_allowed_product/views/purchase_order_view.xml
manuelregidor: restrict_supplier_id and use_only_supplied_product are added by purchase _allowed_product
                    
partner_id, quantity, uom, and company_id are added by puchase (purchase_order_form view)

pedrobaeza: OK, this is a "temporary patch", but what do you think about my general comment?

manuelregidor: If a dependecy towards base_view_inheritance_extension is set, wouldn't it be enough to add the context keys this way?

`                <attribute name="context" operation="update">
                    {'restrict_supplier_id': parent.partner_id,
                    'use_only_supplied_product': parent.use_only_supplied_product}
                </attribute>`

Why would overwriting fields_view_get method be necessary?

pedrobaeza: They are 2 different techniques to choose between, not both

manuelregidor: OK, thank you.

ValentinVinagre (approved):
LGTM 👍🏻

pedrobaeza:
Please forward-port it to v16

HaraldPanten:
Hi @pedrobaeza we'll do the FP in a couple of weeks.

PR #1811, opened 2023-03-20
[13.0][FIX][purchase_discount] discount are readonly when invoice state is fully billed

ernesto-garcia-tecnativa:
@tecnativa TT30233

### DIFF
--- a/purchase_discount/views/purchase_discount_view.xml
+++ b/purchase_discount/views/purchase_discount_view.xml
@@ -29,13 +29,19 @@
                 expr="//field[@name='order_line']/tree/field[@name='price_unit']"
                 position="after"
             >
-                <field name="discount" />
+                <field
+                    name="discount"
+                    attrs="{'readonly': [('parent.invoice_status', '=', 'invoiced')]}"
+                />
             </xpath>
             <xpath
                 expr="//field[@name='order_line']/form//field[@name='price_unit']"
                 position="after"
             >
-                <field name="discount" />
+                <field
+                    name="discount"
+                    attrs="{'readonly': [('parent.invoice_status', '=', 'invoiced')]}"
+                />
             </xpath>
         </field>
     </record>


legalsylvain (approved):
Thanks !
Code review / no test.
Note : should be done  on purchase triple discount module also.

ernesto-garcia-tecnativa:
@pedrobaeza @chienandalu  please review

purchase_discount/views/purchase_discount_view.xml
```diff
@@ -16,7 +19,10 @@
         <field name="inherit_id" ref="purchase.purchase_order_line_tree" />
         <field name="arch" type="xml">
             <field name="price_unit" position="after">
-                <field name="discount" />
+                <field
+                    name="discount"
+                    attrs="{'readonly': [('parent.invoice_status', '=', 'invoiced')]}"
```
pedrobaeza: There's no parent in this one AFAIK (being a tree). I doubt this is working.

chienandalu: Indeed this makes the view fail.

ernesto-garcia-tecnativa:
@pedrobaeza please review

purchase_discount/views/purchase_discount_view.xml
```diff
@@ -6,7 +6,10 @@
         <field name="inherit_id" ref="purchase.purchase_order_line_form2" />
         <field name="arch" type="xml">
             <field name="price_unit" position="after">
-                <field name="discount" />
+                <field
+                    name="discount"
+                    attrs="{'readonly': [('parent.invoice_status', '=', 'invoiced')]}"
```
chienandalu: This view has the same issue. `parent` is only available when the view is used inside of a `One2many` `<field>`, otherwise it will crash.

chienandalu (approved):
:smile: :+1:

pedrobaeza (approved):
/ocabot merge minor

PR #1816, opened 2023-03-22
[16.0][ADD] purchase_stock_picking_note

alexeirivera87:


purchase_stock_picking_note/views/purchase_order_view.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+
+    <record id="purchase_order_form_inherit" model="ir.ui.view">
+        <field name="model">purchase.order</field>
+        <field name="inherit_id" ref="purchase.purchase_order_form" />
+        <field name="arch" type="xml">
+            <xpath
+                expr="//page[@name='purchase_delivery_invoice']/group/group"
```
quannguyen1560, line 9: I think we can use  `<group name="other_info" position="inside"> your custom code </group>` 
It will help the other detect location easier.

quannguyen1560 (changes requested):

PR #1388, opened 2022-03-22
[14.0][ADD] purchase_partner_approval

Nikul-OSI:

PR #1819, opened 2023-03-24
[16.0] [MIG] purchase_partner_approval

Nikul-OSI:


francesco-ooops (changes requested):
@Nikul-OSI if you want to proceed with this, please implement a fix for this problem: https://github.com/OCA/purchase-workflow/pull/1388#issuecomment-1428190766

HekkiMelody:
Here's the fix: https://github.com/OCA/purchase-workflow/pull/2090

PR #1831, opened 2023-04-04
[16.0][MIG] purchase_reception_status

Urvisha-OSI:


AungKoKoLin1997:
I think this module doesn't require anymore in v16.0. There is already this behavior https://github.com/odoo/odoo/blob/6ddb77094bb0654d6326c6f9dadbf4049b86f52d/addons/purchase_stock/models/purchase.py#L34.

yostashiro:
I think it's still good to have this module in OCA as the standard logic may not be sufficient.

Here is how the standard field (receipt_status) is computed: https://github.com/qrtl/odoo/blob/e37349618c605977099cfdee04c99157ef02ce31/addons/purchase_stock/models/purchase.py#L64-L74

The standard logic (based on picking states) may fall short on following points:

1. It does not consider purchase returns - even when there is a return, the receipt state can stay as 'Fully Received'.
2. It does not consider services.

IMO, the compute logic that comes in purchase_reception_status (based on quantities) is more appropriate.

AungKoKoLin1997 (approved):
Functional Review and code review: LGTM
IMO, users will confuse seeing both Reception Status(purchase_reception_status) and Receipt Status(odoo default) in the PO form. Do we need to create new custom module to hide odoo default field if we want to use this module or handle in this module? I am not sure what is the best approach.

yostashiro:
> IMO, users will confuse seeing both Reception Status(purchase_reception_status) and Receipt Status(odoo default) in the PO form.

Yes, I think it would be good to make the standard field invisible in form and list views.  @Urvisha-OSI would you consider taking care of this in this PR?

purchase_reception_status/readme/DESCRIPTION.rst
yostashiro, line 1: @Urvisha-OSI Can you please add the information from https://github.com/OCA/purchase-workflow/pull/1831#issuecomment-1548924543?

rinaldifirdaus:
Functional test: LGTM
*As mentioned above, would be good if the default field invisible to prevent confusion.

yostashiro:
> > IMO, users will confuse seeing both Reception Status(purchase_reception_status) and Receipt Status(odoo default) in the PO form.
> 
> Yes, I think it would be good to make the standard field invisible in form and list views. @Urvisha-OSI would you consider taking care of this in this PR?

After all, we created https://github.com/OCA/purchase-workflow/pull/1883 instead to avoid adding the `purchase_stock` dependency to this module.

hparfr (approved):
code read lgtm

PR #1837, opened 2023-04-07
[16.0][mig] purchase_order_qty_change_no_recompute: Migration to 16.0

dsolanki-initos:


dsolanki-initos:
Closing this PR as the same feature is available in Odoo standard v16.

pedrobaeza:
AFAIK, the technique has changed (they don't use anymore an onchange, but computed writable fields):

https://github.com/odoo/odoo/blob/39b4199fe6b25fdf38c1862803f21b1597c1759b/addons/purchase/models/purchase.py#L1205

but the goal is the same: the unit price is changed when you change the quantity, so you have to adapt the code of this module, using another hack for "de-injecting" the dependency on "product_qty".

PR #1844, opened 2023-04-17
[14.0][IMP] purchase_requisition_auto_rfq

anddago78:
With this improvement, in the purchase agreement a functionality has been added to the lines, when they appear in red it means that the product does not have any supplier configured by default.

MT-2584 @moduon @rafaelbn @Shide @yajo @fcvalgar please review :)

### DIFF
--- a/purchase_requisition_auto_rfq/models/purchase_requisition.py
+++ b/purchase_requisition_auto_rfq/models/purchase_requisition.py
@@ -2,7 +2,7 @@
 
 from collections import defaultdict
 
-from odoo import _, models
+from odoo import _, api, fields, models
 
 
 class PurchaseRequisition(models.Model):
@@ -61,3 +61,16 @@ class PurchaseRequisition(models.Model):
         po = self.env["purchase.order"].with_context(**ctx).create(vals)
         po._onchange_requisition_id()
         return po
+
+
+class PurchaseRequisitionLine(models.Model):
+    _inherit = "purchase.requisition.line"
+
+    prod_has_supplier = fields.Boolean(
+        compute="_compute_prod_has_supplier", string="Product without suppliers"
+    )
+
+    @api.depends("product_id.product_tmpl_id.seller_ids")
+    def _compute_prod_has_supplier(self):
+        for rec in self:
+            rec.prod_has_supplier = bool(rec.product_id.product_tmpl_id.seller_ids)
--- a/purchase_requisition_auto_rfq/view/purchase_requisition.xml
+++ b/purchase_requisition_auto_rfq/view/purchase_requisition.xml
@@ -11,9 +11,15 @@
                 <button
                     name="auto_rfq_from_suppliers"
                     type="object"
-                    string="Generate quotations from suppliers"
+                    string="Generate RFQs for suppliers"
                     attrs="{'invisible': ['|', ('line_ids', '=', []), ('state', '!=', 'in_progress')]}"
                 />
+            </xpath>
+            <xpath expr="//field[@name='line_ids']/tree" position="inside">
+                <field name="prod_has_supplier" invisible="1" />
+            </xpath>
+            <xpath expr="//field[@name='line_ids']/tree" position="attributes">
+                <attribute name="decoration-danger">not prod_has_supplier</attribute>
             </xpath>
         </field>
     </record>
--- a/purchase_requisition_auto_rfq/demo/product_product.xml
+++ b/purchase_requisition_auto_rfq/demo/product_product.xml
@@ -2,7 +2,7 @@
 <data noupdate="1">
     <record id="kitchenset" model="product.product">
         <field name="name">Kitchen Set</field>
-        <field name="type">product</field>
+        <field name="type">consu</field>
         <field
                 name="seller_ids"
                 eval="[
@@ -14,7 +14,7 @@
 
     <record id="blankets" model="product.product">
         <field name="name">Blankets</field>
-        <field name="type">product</field>
+        <field name="type">consu</field>
         <field
                 name="seller_ids"
                 eval="[
@@ -26,7 +26,7 @@
 
     <record id="tarpaulin" model="product.product">
         <field name="name">Tarpaulin</field>
-        <field name="type">product</field>
+        <field name="type">consu</field>
     </record>
 </data>
 </odoo>


fcvalgar (approved):
Functional review

LGTM, Thank you @anddago78

purchase_requisition_auto_rfq/models/purchase_requisition.py
```diff
@@ -2,12 +2,21 @@
...
 from collections import defaultdict
 
-from odoo import _, models
+from odoo import _, api, fields, models
 
 
 class PurchaseRequisition(models.Model):
     _inherit = "purchase.requisition"
 
+    partner_rfq_ids = fields.Many2many(
```
yajo: Following English rules, please rename this field, so its name reflects better that we're referring to partners and not to RFQs.

Also add a string:

```suggestion
    rfq_partner_ids = fields.Many2many(
        string="RFQ Partners",
```

purchase_requisition_auto_rfq/models/purchase_requisition.py
```diff
@@ -2,12 +2,21 @@
...
+        comodel_name="res.partner", compute="_compute_rfq_partners"
+    )
+
+    def _compute_rfq_partners(self):
+        rfqs = self.env["purchase.order"].search([("requisition_id", "=", self.id)])
+        for rec in self:
+            rec.partner_rfq_ids = rfqs.mapped("partner_id").ids
```
yajo: It would be more efficient and simple to set the field as related:

```suggestion
        related="purchase_ids.partner_id",
    )
```

purchase_requisition_auto_rfq/models/purchase_requisition.py
```diff
@@ -61,3 +70,70 @@ def _create_purchase_order(self, seller_id):
         po = self.env["purchase.order"].with_context(**ctx).create(vals)
         po._onchange_requisition_id()
         return po
+
+    def action_all_rfq_related_send(self):
+        self.ensure_one()
+        rfqs = self.env["purchase.order"].search([("requisition_id", "=", self.id)])
```
yajo: If apply next suggestion, you'll benefit from Odoo's ORM caching and prefetching. I invite you to see this interesting talk which gives some info about how to benefit from those: https://www.youtube.com/watch?v=5jT5eoCZH6U

For now, let's just say that this is more efficient:

```suggestion
        rfqs = self.purchase_ids
```

purchase_requisition_auto_rfq/__manifest__.py
```diff
@@ -16,6 +16,7 @@
     ],
     "data": [
         "view/purchase_requisition.xml",
+        "wizard/wizard_purchase_mass_mail.xml",
```
yajo: Oops, delete this too please.

```suggestion
```

purchase_requisition_auto_rfq/models/purchase_requisition.py
```diff
@@ -61,3 +70,43 @@ def _create_purchase_order(self, seller_id):
...
+        ir_model_data = self.env["ir.model.data"]
+        try:
+            template_id = ir_model_data.get_object_reference(
+                "purchase", "email_template_edi_purchase"
+            )[1]
+        except ValueError:
+            template_id = False
```
yajo: This is basically the same:

```suggestion
        template_id = self.env.ref("purchase.email_template_edi_purchase", False).id
```

yajo (changes requested):
I have pushed a fixup commit that removes about 120 LOC while keeping the same features. Please review it and we can comment any doubts. The purpose is to squash it with yours.

Apart from that, you have some suggestions below to improve it even more.

Thanks for the work!

purchase_requisition_auto_rfq/models/purchase_requisition.py
```diff
@@ -61,3 +70,43 @@ def _create_purchase_order(self, seller_id):
...
+    _inherit = "purchase.requisition.line"
+
+    prod_has_supplier = fields.Boolean(
+        compute="_compute_prod_has_supplier", string="Product without suppliers"
+    )
+
+    @api.depends("product_id.product_tmpl_id.seller_ids")
+    def _compute_prod_has_supplier(self):
+        for rec in self:
+            rec.prod_has_supplier = bool(rec.product_id.product_tmpl_id.seller_ids)
```
Shide, line 76: You can make this related too.
If you don't make the related field, change the string to "Product has suppliers".

```suggestion
    seller_ids = fields.One2many(
        comodel="product.supplierinfo",
        related="product_id.product_tmpl_id.seller_ids",
    )
```

Shide (changes requested):
Minor changes

anddago78:
All comments are attended, thank you @yajo @Shide @fcvalgar

purchase_requisition_auto_rfq/models/purchase_requisition.py
```diff
@@ -61,3 +72,37 @@ def _create_purchase_order(self, seller_id):
         po = self.env["purchase.order"].with_context(**ctx).create(vals)
         po._onchange_requisition_id()
         return po
+
+    def action_all_rfq_related_send(self):
+        self.ensure_one()
+        rfqs = self.purchase_ids.filtered(lambda x: x.state != "cancel")
+        template_id = self.env.ref("purchase.email_template_edi_purchase", False).id
```
yajo: I just notice this would raise an `AttributeError` if the template isn't found, so please change this:

```suggestion
        template = self.env.ref("purchase.email_template_edi_purchase", False)
```

purchase_requisition_auto_rfq/models/purchase_requisition.py
```diff
@@ -61,3 +72,37 @@ def _create_purchase_order(self, seller_id):
...
+        ctx = dict(
+            self.env.context,
+            default_model="purchase.order",
+            active_model="purchase.order",
+            active_id=rfqs.ids[0],
+            active_ids=rfqs.ids,
+            default_template_id=template_id,
```
yajo: ... and also change this:

```suggestion
            default_template_id=template and template.id,
```

purchase_requisition_auto_rfq/models/purchase_requisition.py
```diff
@@ -2,12 +2,23 @@
...
+        comodel_name="res.partner", compute="_compute_rfq_partners"
+    )
+
+    def _compute_rfq_partners(self):
+        rfqs = self.env["purchase.order"].search(
+            [("requisition_id", "=", self.id), ("state", "!=", "cancel")]
+        )
+        for rec in self:
+            rec.partner_rfq_ids = rfqs.mapped("partner_id").ids
+
```
yajo: If I'm not wrong, you're no longer using this field, right? Then, you can remove it:

```suggestion
```

purchase_requisition_auto_rfq/view/purchase_requisition.xml
```diff
@@ -14,6 +14,18 @@
                     string="Generate quotations from suppliers"
```
yajo: I find confusing this wording. How about this?

```suggestion
                    string="Generate RFQs for suppliers"
```

purchase_requisition_auto_rfq/view/purchase_requisition.xml
```diff
@@ -14,6 +14,18 @@
...
                     attrs="{'invisible': ['|', ('line_ids', '=', []), ('state', '!=', 'in_progress')]}"
                 />
+                <button
+                    name="action_all_rfq_related_send"
+                    type="object"
+                    string="Send quotations from all suppliers"
```
yajo:
```suggestion
                    string="Send RFQs to all suppliers"
```

purchase_requisition_auto_rfq/models/purchase_requisition.py
```diff
@@ -61,3 +72,37 @@ def _create_purchase_order(self, seller_id):
...
+            default_template_id=template_id,
+            default_composition_mode="mass_mail",
+        )
+        return {
+            "type": "ir.actions.act_window",
+            "view_mode": "form",
+            "res_model": "mail.compose.message",
+            "target": "new",
+            "context": ctx,
+        }
```
yajo: Now that the implementation is so much simplified (this is basically just an action), I'm giving it a 2nd thought and I think this is not the place where it should go. At least, not the *main* place.

We're just sending a batch email to a lot of RFQs, so the correct place IMHO would be as a context action item:

imagen

So, from the requisition, you just go to the list of RFQs, select the ones you want to mail, and mail them. The advantage would be that you could select only some RFQs and you could even trigger this action for RFQs that are not related to purchase requisitions.

This makes me think if this feature should be related to this module or if it should be in an independent module with just dependency on `purchase`. 🤔 

WDYT @fcvalgar @Shide @anddago78 @rafaelbn ?

yajo (changes requested):

purchase_requisition_auto_rfq/models/purchase_requisition.py
anddago78: @yajo all suggested changes are resolved except one, I'm waiting @rafaelbn , for me your suggested change is the best solution...

Shide: I'm thinking about the connection between modules, but this module should be able to be used alone (without https://github.com/OCA/sale-workflow/pull/2430 for example).

So I think the @yajo solution is the best approach.
If we need to add a button on the Purchase Requisition in the future, will be easily added if we develop this action first.

rafaelbn: > I'm thinking about the connection between modules, but this module should be able to be used alone (without https://github.com/OCA/sale-workflow/pull/2430 for example).

Correct, both modules are independent @anddago78 

So, what is missing here?

rafaelbn:
Commented with @anddago78 , we close this PR and open a new one for a new module to be used independently

purchase_requisition_auto_rfq/view/purchase_requisition.xml
```diff
@@ -11,10 +11,16 @@
...
+                    string="Generate RFQs for suppliers"
                     attrs="{'invisible': ['|', ('line_ids', '=', []), ('state', '!=', 'in_progress')]}"
                 />
             </xpath>
+            <xpath expr="//field[@name='line_ids']/tree" position="inside">
+                <field name="prod_has_supplier" invisible="1" />
+            </xpath>
+            <xpath expr="//field[@name='line_ids']/tree" position="attributes">
+                <attribute name="decoration-danger">not prod_has_supplier</attribute>
+            </xpath>
```
yajo, line 23: Maybe we don't need to close this PR. This feature is still good here IMHO, isn't it?

anddago78: I have modified this pr and now it only has the improvement mentioned in the description..

yajo (changes requested):
Please, @anddago78 could you amend the commit and push it with a better description explaining the changes? It currently just says `[14.0][IMP] purchase_requisition_auto_rfq`; that's not very helpful.

Remember to follow our internal contribution standards guide.

In my side, that's the only missing part to merge this. Thanks!

rafaelbn (approved):
👍 Great!

@HviorForgeFlow please this is a minor improvement to show in red the lines which don't have supplier! 😄 

Thank you!

HviorForgeFlow:
Pending comment: https://github.com/OCA/purchase-workflow/pull/1844#pullrequestreview-1395289771
@rafaelbn @anddago78

anddago78:
thanks @HviorForgeFlow  all comments are attended

yajo (approved):

HviorForgeFlow (approved):

PR #1846, opened 2023-04-18
[14.0][ADD] product_supplier_search

Kev-Roche:
Add a supplier filter possibility on products.


legalsylvain (changes requested):
Thanks for sharing. 

Could you add a test ?

@quentinDupont could you take a look on this one ?

PR #1847, opened 2023-04-18
[14.0][IMP] improve supplier search and purchase_ok filter

Kev-Roche:
- improve supplier search
- replace purchase_ok filter in context by a domain to be constistant with the purchase order regular view.


sebastienbeau (approved):
LGTM (code review and test)

PR #1168, opened 2021-04-30
 [ADD] 12.0 : new module purchase_no_rfq that remove request for quotation concept in purchase workflow

legalsylvain:
This module extends the odoo purchase module, to remove 'Request for Quotation' state, simplifying
the workflow for the ends users.

Once installed :

* The menu item 'Purchase > Purchase > Requests for Quotation' is hidden. A single menu item 'Purchase Order'
  is available

* The states names of the purchase order is altered. 'RFQ' is replaced by 'Draft' and 'RFQ sent' by 'Sent'.

* The colors in the tree view is correctly set to ``decoration-info`` for 'draft' and 'sent' orders.

purchase_order_tree

* In the form view, all the RFQ names are removed or replaced by 'Purchase Order'.
  The module makes also the 'Print' button allways available and not only on 'draft' and 'sent' status.

* The option 'Print > Request For quotation' is also disabled.

purchase_order_form

CC : @quentinDupont (apps/deck/#/board/144/card/1424 : partie 1)

PR #1849, opened 2023-04-19
[16.0] MIG purchase_no_rfq

legalsylvain:
Quite simple migration. (from 12.0 to 16.0)

- [original V12 PR](https://github.com/OCA/purchase-workflow/pull/1168)
- commit to review : https://github.com/OCA/purchase-workflow/commit/88bc22644e3d09e0511f979eaae35f128c838f41

CC : original reviewers : @quentinDupont, @bealdav, @hparfr 

### Description

This module extends the odoo purchase module, to remove 'Request for Quotation' state, simplifying
the workflow for the end users.

Once installed :

* The menu item 'Purchase > Purchase > Requests for Quotation' is hidden. A single menu item 'Purchase Order' is available

* The states names of the purchase order is altered. 'RFQ' is replaced by 'Draft' and 'RFQ sent' by 'Sent'.

* The colors in the tree view is correctly set to ``decoration-info`` for 'draft' and 'sent' orders.

purchase_order_tree

* In the form view, all the RFQ names are removed or replaced by 'Purchase Order'.
  The module makes also the 'Print' button allways available and not only on 'draft' and 'sent' status.

* The option 'Print > Request For quotation' is also disabled.

purchase_order_form


flalexg (approved):
@legalsylvain LGTM! I would only change the name of the state "Sent" to "Purchase Sent" or someting similar to follow the sale state "Quotation Sent", but it doesn't matter if its only "Sent".

bealdav (approved):

legalsylvain:
> @legalsylvain LGTM! I would only change the name of the state "Sent" to "Purchase Sent" or someting similar to follow the sale state "Quotation Sent", but it doesn't matter if its only "Sent".
Hi. Generally, I like to stick with odoo terms, so your proposal could looks valid, but in the other hand, the module is named purchase_no_rfq, so, there is no "purchase" and "quotation" now only "purchase", so "sent" seems valid now and save place.

hparfr (approved):

PR #1850, opened 2023-04-20
[16.0][FIX] purchase_allowed_product: cannot find the vendor's products.

angelgarciadelachica:
[FP] PR #1797 to find products by the seller's reference.

[FIX] When a quote request is created and the option ''Use only allowed products'' is checked, the supplier's products do not appear in the dropdown when trying to add a new product. 
This is because the "name" field of the product.supplierinfo module in [v15](https://github.com/odoo/odoo/blob/92a81e84d92abee48c333a07055324b47e1f5d76/addons/product/models/product.py#L822) is renamed to "partner_id" in [v16](https://github.com/odoo/odoo/blob/966a7e9a01297c95a7768daef0d9662e7d343f6a/addons/product/models/product_supplierinfo.py#L29).

To reproduce the error:.
1. Go to the Purchasing module -> Configuration -> Vendor price List
2. Create a new one with a product and a Vendor.
3. Create a quote request.
4. Fill in the Vendor field with the partner from the 'Vendor price List' we have just created.
5. Check 'Use only allowed products'.
6. Add a product

You can see that no product appears.

### DIFF
--- a/purchase_allowed_product/__manifest__.py
+++ b/purchase_allowed_product/__manifest__.py
@@ -11,7 +11,7 @@
     "website": "https://github.com/OCA/purchase-workflow",
     "author": "Akretion, Odoo Community Association (OCA)",
     "license": "AGPL-3",
-    "depends": ["purchase"],
+    "depends": ["purchase", "base_view_inheritance_extension"],
     "data": [
         "views/res_partner_views.xml",
         "views/account_move_views.xml",
--- a/purchase_allowed_product/models/product.py
+++ b/purchase_allowed_product/models/product.py
@@ -27,7 +27,7 @@ class ProductProduct(models.Model):
                 .commercial_partner_id
             )
             supplierinfos = self.env["product.supplierinfo"].search(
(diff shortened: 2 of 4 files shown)


angelgarciadelachica:
@HaraldPanten @ValentinVinagre @manuelregidor @luis-ron

HaraldPanten (approved):
Tested in runboat. LGTM 👍

HaraldPanten:
@pedrobaeza 

This use case was not checked properly before merging.

This adaptation is needed so that the products are correctly filtered according to their supplier. `name` needs to be changed to `partner_id`.

On the other hand, a forward port of another commit has been done, as we promised a few weeks ago in https://github.com/OCA/purchase-workflow/pull/1797

Do you mind fast tracking this?

THX!

luis-ron (approved):
Functional review. LGTM 👍🏻

ValentinVinagre (approved):
LGTM 👍🏻

PR #1863, opened 2023-05-04
[16.0][IMP] purchase_request: add UTC time to scheduled_date

augusto-weiss:
This PR was created because the request date originated form the purchase request does not match the scheduled date from the purchase order. By adding the "utcnow()" we save the date_planned as UTC so that when you pick it up t takes it in the user's TZ

### DIFF
--- a/purchase_request/wizard/purchase_request_line_make_purchase_order.py
+++ b/purchase_request/wizard/purchase_request_line_make_purchase_order.py
@@ -2,6 +2,8 @@
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl-3.0).
 from datetime import datetime
 
+import pytz
+
 from odoo import _, api, fields, models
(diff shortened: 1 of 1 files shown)


HviorForgeFlow (approved):

augusto-weiss:
Hi @pedrobaeza, Can this pr be merge? Thanks a lot

pedrobaeza (changes requested):
Datetime in server is always in UTC, so I don't really get the problem. Please expose a complete use case with steps to reproduce the problem.

augusto-weiss:
Oh sorry, about that!!

Reply error: (Make sure you have different timezone)

Create a purchase request and request line
Create a RFQ from Request line (wizard)
Field "Expected arrival" (date_planned) has a wrong date
purchase request

nicomacr (approved):
LGTM

PR #1866, opened 2023-05-08
[16.0][ADD] purchase_order_restriction

AungKoKoLin1997:
The purpose of this module is to limit the access to the purchase order to 1) the buyer and 2) purchase managers.

@qrtl


pedrobaeza:
Isn't this the same as https://github.com/OCA/purchase-workflow/tree/15.0/purchase_security ?

AungKoKoLin1997:
@pedrobaeza These are not the same. `purchase_security` allows only see his own purchase order for Purchase (own orders) group. This module adds new field `is_restricted` in purchase order that can only be changed by its own purchase user or purchase manager. So, if purchase order is restricted by its own purchase or purchase manager, this order can only be seen by its own purchase user or purchase manager.

pedrobaeza:
Ok then

yostashiro (commented):
@AungKoKoLin1997 Can you please add some tests.

purchase_order_restriction/tests/test_purchase_order_restriction.py
yostashiro, line 1: @AungKoKoLin1997 Can you please check if we can inherit the test class from the purchase module.

AungKoKoLin1997: @yostashiro  There is no much effective inheriting test class of purchase module. We can only use vendor value if we inherit.

purchase_order_restriction/tests/test_purchase_order_restriction.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def test_order_restriction(self):
+        # PO user should have access to all of them
+        # before it's restricted
+        self.assertEqual(
+            len(
+                self.env["purchase.order"]
+                .with_user(self.user1)
+                .search([("name", "like", "Purchaser Order")])
+                .ids
```
yostashiro:
```suggestion
```

purchase_order_restriction/tests/test_purchase_order_restriction.py
yostashiro: @AungKoKoLin1997 Then can you try and reduce LOC by creating methods such as _create_user(), _create_order(), etc.?

purchase_order_restriction/tests/test_purchase_order_restriction.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def _create_user(self, name, login, email, group):
+        return (
+            self.env["res.users"]
+            .with_context(no_reset_password=True)
+            .create(
+                {
+                    "name": name,
+                    "login": login,
+                    "email": email,
```
yostashiro: `email` can be fixed as it doesn't matter.

```suggestion
                    "email": "test@yourcompany.com",
```

yostashiro (approved):
Code review and functional test. 👍 

Please squash commits.

navarrorico (approved):
Functional review

AungKoKoLin1997:
@dreispt The merge was failed. Can you please try again?

HviorForgeFlow:
@AungKoKoLin1997 merge will still failing because purchase_order_restriction is not compatible with purchase_security tests.

It will be necessary to create new tests ci entry excluding one each other.

@pedrobaeza any hint?

pedrobaeza:
Why not adding the feature `is_restricted` to `purchase_security` instead?

PR #1886, opened 2023-06-05
[16.0][MIG] purchase_order_qty_change_no_recompute

aliciagaarzo:
Migration of **purchase_order_qty_change_no_recompute** module to version 16


pedrobaeza:
Hi, @aliciagaarzo please check how it was reworked on sales part due to the new computed writable fields:

https://github.com/OCA/sale-workflow/blob/16.0/sale_order_qty_change_no_recompute/monkeypatching.py

hugo-cordoba:
hi @pedrobaeza 

We have reviewed what you say about using the monkeypatch, however, we have also seen that the function that is called in the 'sale.order.line' (_compute_price_unit_and_date_planned_and_name), is also called from the purchase lines.

If you remove 'product_uom_qty' and 'product_uom' from the depends, they will not be taken into account in the function when called from the 'name' and 'date_planned' fields in 'purchase.order.line'.

https://github.com/OCA/OCB/blob/8c5483ebd63e656240db62b96afbe1f2f805b28f/addons/purchase/models/purchase.py#LL934C7-L934C7

What do you think about this, can you think of a solution we can develop?

pedrobaeza:
You can add a condition checking the field name, and don't remove for other cases different of the field `price_unit`.

aliciagaarzo:
> You can add a condition checking the field name, and don't remove for other cases different of the field `price_unit`.

Hello @pedrobaeza , I have reviewed the comments and I have not been able to make the development work, what exactly do you mean?

pedrobaeza:
I'm reviewing my link and the monkeypatching is already checking the field names where you remove the depends in:

https://github.com/OCA/sale-workflow/blob/d124964acbf8ca7f07feea012bc97537562decbb/sale_order_qty_change_no_recompute/monkeypatching.py#L13C22-L13C22

so I don't understand the problem.

Please push here the existing code with the monkeypatching, and expose the problem if any, for being able to check it.

imagen

josep-tecnativa:
Please, include this changes https://github.com/OCA/purchase-workflow/pull/2054/commits/aac02e59be1caf22cab5008e4fb959ffebb117f2 using imported DISABLED_MAIL_CONTEXT variable.

victoralmau:
Superseed by https://github.com/OCA/purchase-workflow/pull/2193

PR #1921, opened 2023-06-23
[15.0][IMP] purchase_security: Add groups to purchase orders and new group …

stefan-tecnativa:
…permission feature

cc @Tecnativa TT44023

@pedrobaeza @victoralmau please review!

What has been done is to add groups to purchase orders, along with a new "purchase group" permission. So users who have this permission and belong to the group associated with the purchase order, will be able to see the purchase orders of other users with whom they share that group.


purchase_security/models/purchase.py
```diff
@@ -0,0 +1,13 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+
+    purchase_group_ids = fields.Many2one(
+        "res.groups",
```
pedrobaeza: This should be a new model: `purchase.team`. And call this field `team_id` (without `s`), as it's a many2one and it's inside `purchase.order`, so `purchase` is redundant.

purchase_security/security/security.xml
```diff
@@ -47,5 +55,13 @@
             >['|',('order_id.user_id','=',user.id),('order_id.user_id','=',False)]</field>
             <field name="groups" eval="[(4,ref('group_purchase_own_orders'))]" />
         </record>
+        <record model="ir.rule" id="purchase_order_user_group_purchases_rule">
+            <field name="name">View purchase orders (Group purchases)</field>
```
pedrobaeza: Follow the nomenclature of the rest of the groups:

Selección_037


```suggestion
            <field name="name">User (team orders)</field>
```

And even more, you have broken the inheritance, as you have to change the "Administrator" to inherit from this new group.

pedrobaeza (changes requested):
You need to add as well a `purchase.order` rule for the new group.

Check `sales_team` for mimicking it on purchases.

purchase_security/security/security.xml
stefan-tecnativa: Thanks for all the suggestions @pedrobaeza , now I have a better understanding of the approach to be taken. I made some modifications, and I am aware that i have not added yet some view for the new model. Review my changes when you have time please.

purchase_security/models/purchase.py
```diff
@@ -0,0 +1,18 @@
...
+
+from odoo import fields, models
+
+
+class PurchaseTeam(models.Model):
+    _name = "purchase.team"
+    _description = "Purchase Team"
+
+    name = fields.Char()
+    user_id = fields.Many2many("res.users", string="Purchase Users")
```
pedrobaeza: Please respect conventions. m2m/o2m always with `_ids`.

```suggestion
    user_ids = fields.Many2many("res.users", string="Purchase Users")
```

purchase_security/models/purchase.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2023 Tecnativa - Stefan Ungureanu
```
pedrobaeza: Only one model per file, unless it's a "children" model, so this file should be called `purchase_team.py`, and only contains the first class. The other should be put in a file `purchase_order.py`. If you also inherit `purchase.order.line`, then that one can fit inside `purchase_order.py`.

purchase_security/models/res_users.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2023 Tecnativa - Stefan Ungureanu
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResUsers(models.Model):
+    _inherit = "res.users"
+
+    purchase_team_id = fields.Many2many("purchase.team", string="Purchase Teams")
```
pedrobaeza: No need of this inverse field.

purchase_security/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+purchase_security.access_purchase_team,access_purchase_team,purchase_security.model_purchase_team,base.group_user,1,1,1,1
```
pedrobaeza: This permission is too broad. That's a bad practice. If this is for purchase, why everything is able to modify it? Put read permission for purchase users, and full access to purchase admins.

purchase_security/security/security.xml
```diff
@@ -30,6 +38,15 @@
             >['|',('user_id','=',user.id),('user_id','=',False)]</field>
             <field name="groups" eval="[(4,ref('group_purchase_own_orders'))]" />
         </record>
+        <record model="ir.rule" id="purchase_order_group_purchase_order_group_orders">
+            <field name="name">View purchase orders (group responsible)</field>
+            <field name="model_id" ref="purchase.model_purchase_order" />
+            <field
+                name="domain_force"
+            >['|', '|', ('user_id', '=', user.id), ('user_id', '=', False), '|', ('team_id', '=', user.purchase_team_id.id),('team_id', '=', False)]
```
pedrobaeza: Why adding the user part? That one is already covered by the inferior record rule, isn't it?

purchase_security/models/purchase.py
```diff
@@ -0,0 +1,18 @@
...
+    _description = "Purchase Team"
+
+    name = fields.Char()
+    user_id = fields.Many2many("res.users", string="Purchase Users")
+
+
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+
+    team_id = fields.Many2one("purchase.team", string="Team")
```
pedrobaeza:
```suggestion
    team_id = fields.Many2one("purchase.team", string="Team", index=True)
```

pedrobaeza (changes requested):
Missing menu entry and view for purchase teams. The commit message shouldn't be `[WIP] ...`.

purchase_security/security/security.xml
stefan-tecnativa: Thanks for all the comments. You are right, I wasn't fully understanding the difference between this rules, thats wh i was checking the user twice. I made the suggested changes.

purchase_security/security/security.xml
```diff
@@ -30,6 +38,15 @@
             >['|',('user_id','=',user.id),('user_id','=',False)]</field>
             <field name="groups" eval="[(4,ref('group_purchase_own_orders'))]" />
         </record>
+        <record model="ir.rule" id="purchase_order_group_purchase_order_group_orders">
+            <field name="name">View purchase orders (group responsible)</field>
```
chienandalu:
```suggestion
            <field name="name">View purchase orders (purchase team member)</field>
```

purchase_security/security/security.xml
```diff
@@ -30,6 +38,15 @@
...
+            <field name="model_id" ref="purchase.model_purchase_order" />
+            <field
+                name="domain_force"
+            >['|', ('team_id', '=', user.purchase_team_id.id),('team_id', '=', False)]
```
chienandalu: That field `purchase_team_id` on `res.users` doesn't even exists. How is this supposed to work?

chienandalu (changes requested):
Some comments:

- The module doesn't work :confused: . You can figure out why in the comments below...
- In that regard, I'm missing some tests for these new features.
- Should team members be subscribed automatically to team POs?
- Something with this module is making the user purchase permissions show up as checkboxes instead of a select:

image

pedrobaeza:
> Should team members be subscribed automatically to team POs?

Out of scope of this PR for now. It can be put in the ROADMAP though.

purchase_security/security/security.xml
stefan-tecnativa: Thanks for the comments, can you check if now it is correct?
I will improve the test coverage after.

purchase_security/security/security.xml
```diff
@@ -47,5 +65,13 @@
             >['|',('order_id.user_id','=',user.id),('order_id.user_id','=',False)]</field>
             <field name="groups" eval="[(4,ref('group_purchase_own_orders'))]" />
         </record>
+        <record model="ir.rule" id="purchase_order_user_group_purchases_rule">
+            <field name="name">View purchase order lines (purchase responsible)</field>
+            <field name="model_id" ref="purchase.model_purchase_order_line" />
+            <field
+                name="domain_force"
+            >['|',(user.id, 'in', 'order_id.team_id.user_ids'),('order_id.team_id', '=',False)]</field>
```
chienandalu: @stefan-tecnativa please at least make the minimum functional tests before asking for reviews. This is wrong

purchase_security/security/security.xml
```diff
@@ -8,10 +8,19 @@
...
+                name="implied_ids"
+                eval="[(4, ref('purchase_security.group_purchase_own_orders'))]"
+            />
+        </record>
         <record id="purchase.group_purchase_manager" model="res.groups">
             <field
                 name="implied_ids"
-                eval="[(3, ref('purchase.group_purchase_user')), (4, ref('group_purchase_own_orders'))]"
+                eval="[(3, ref('purchase.group_purchase_user')),
+                (4, ref('group_purchase_own_orders')),(4, ref('group_purchase_group_orders'))]"
```
chienandalu, line 23: I don't think you need to imply these groups as managers override the access rules

purchase_security/views/purchase_team_views.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="purchase_team_form" model="ir.ui.view">
+            <field name="name">purchase.team.form</field>
+            <field name="model">purchase.team</field>
+            <field name="arch" type="xml">
+                <form string="Purchase Team">
+                    <group name="team">
+                        <field name="name" />
+                        <field name="user_ids" />
```
chienandalu: Maybe you can customize the tree a little (or use a different one) so we don't have much columns or you can copy from the `sales_team.crm_team_view_form` view that uses a kanban which is more visually pleasant.

purchase_security/security/security.xml
```diff
@@ -47,5 +65,13 @@
             >['|',('order_id.user_id','=',user.id),('order_id.user_id','=',False)]</field>
             <field name="groups" eval="[(4,ref('group_purchase_own_orders'))]" />
         </record>
+        <record model="ir.rule" id="purchase_order_user_group_purchases_rule">
+            <field name="name">View purchase order lines (purchase responsible)</field>
+            <field name="model_id" ref="purchase.model_purchase_order_line" />
+            <field
+                name="domain_force"
+            >['|', ('order_id.team_id.user_ids', 'in', [user.id]), ('order_id.team_id', '=', False)]</field>
```
chienandalu, line 73: This is not going to be very performant. What do you think @pedrobaeza ?

chienandalu (commented):
Now it seems to work. Some comments:

purchase_security/security/security.xml
pedrobaeza: We can put `auto_join=True` for having less penalty, and it's compatible with security.

stefan-tecnativa: Thanks for all the suggestions!

purchase_security/tests/test_access_rights.py
```diff
@@ -14,15 +14,56 @@ def setUpClass(cls):
         super(TestPurchaseOrderSecurity, cls).setUpClass()
         # Users
         users = cls.env["res.users"].with_context(no_reset_password=True)
-        group_name = "group_purchase_own_orders"
+        cls.team1 = cls.env["purchase.team"].create({"name": "Team1"})
+        cls.team2 = cls.env["purchase.team"].create({"name": "Team2"})
+        group_name_own = "group_purchase_own_orders"
+        group_name_team = "group_purchase_group_orders"
```
chienandalu: Don't use string formating for this. It isn't very useful and removes clarity to the case (I know it was like that before, but you can fix it now)

purchase_security/models/purchase_team.py
```diff
@@ -0,0 +1,12 @@
...
+
+from odoo import fields, models
+
+
+class PurchaseTeam(models.Model):
+    _name = "purchase.team"
+    _description = "Purchase Team"
+
+    name = fields.Char()
+    user_ids = fields.Many2many("res.users", string="Purchase Users", auto_join=True)
```
chienandalu: Do Many2many have auto_join?

pedrobaeza: No, they are automatically auto-joined in ORM. The auto join was for `team_id` many2one (and not sure right now if with `-`, `_`, or nothing).

stefan-tecnativa: Now i got it. I haven't found any other kind of declaration for this parameter than the one you provided to me.

chienandalu: I case of doubt, a natural step is to go to the docs: https://www.odoo.com/documentation/15.0/developer/reference/backend/orm.html?#relational-fields

pedrobaeza (approved):
/ocabot merge major

PR #1922, opened 2023-06-23
[16.0][IMP] purchase update proposal group backend

FranzPoize:
Adds a functionality to purchase_update_proposal using server-backend/group_base_backend to allow backend users to make proposal on purchases


purchase_update_proposal/models/purchase_order.py
```diff
@@ -0,0 +1,294 @@
...
+                # We have to reset order because of changed qty in confirmed order
+                initial_state = self.state
+                self.action_cancel()
+                self.action_cancel_draft()
+            self._update_proposal_to_purchase_line(data, body)
+            self.message_post(body="\n".join(body))
+        if initial_state in ("approved", "confirmed"):
+            # altered quantity implied to reset order, then we approve them again
+            self.signal_workflow("purchase_confirm")
+            self.signal_workflow("purchase_approve")
```
bealdav: These 2 lines above shouldn't exists because it's 8.0 syntax.

Also in 16.0 we shouldn't cancel an order to update it. It should impact many lines of code

PR #1924, opened 2023-06-27
[13.0][IMP] purchase_order_line_sequence: use correct sequence in moves when section or notes in purchase order

OriolMForgeFlow:
Assign sequence correctly to the move associated with the purchase order line it references when there are sections or notes in the purchase order.


LoisRForgeFlow (approved):
LGTM, thanks for the test cases

HviorForgeFlow (approved):
Code review :+1:

LoisRForgeFlow:
@OriolMForgeFlow could you forward port this change to v14?

PR #1926, opened 2023-06-29
[14.0][IMP] purchase_order_line_sequence: use correct sequence in moves when section or notes in purchase order

OriolMForgeFlow:
Forwardport #1924


HviorForgeFlow (approved):
Fast Forwarding

PR #1933, opened 2023-07-06
[14.0][IMP] purchase_order_line_sequence: module refactor, new sequence fields for orders and invoices

OriolMForgeFlow:
Module Refactor Update:
- A new field has been introduced to compute its value independently, eliminating the reliance on the standard sequence field.
- The sequence is now accurately propagated to both the stock.move and the account.move.


purchase_order_line_sequence/models/account_move_line.py
```diff
@@ -0,0 +1,28 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+
+
+class AccountMoveLine(models.Model):
+    _inherit = "account.move.line"
+
+    related_po_sequence = fields.Char(
+        string="Sequence Related PO",
```
LoisRForgeFlow: PO Line Number

purchase_order_line_sequence/models/purchase_line.py
```diff
@@ -3,15 +3,31 @@
...
-        res["sequence"] = self.sequence
+    sequence = fields.Integer(
+        "Hidden Sequence",
+        help="Gives the sequence of the line when " "displaying the purchase order.",
+        default=9999,
+    )
 
-        return res
+    visible_sequence = fields.Integer(
+        "Sequence",
```
LoisRForgeFlow: Line Number

purchase_order_line_sequence/models/account_move_line.py
```diff
@@ -0,0 +1,28 @@
...
+        compute="_compute_related_po_sequence",
+    )
+
+    @api.depends("purchase_line_id")
+    def _compute_related_po_sequence(self):
+        for rec in self:
+            if not rec.purchase_line_id or rec.purchase_line_id.display_type:
+                rec.related_po_sequence = False
+                continue
+            if len(rec.move_id.invoice_origin.split(", ")) > 1:
```
LoisRForgeFlow: The origin is an string and it is not the most reliable way to get related POs. Better use something like `rec.move_id.mapped("line_ids.purchase_order_id")`

LoisRForgeFlow (changes requested):
Some comments

AaronHForgeFlow (commented):
This refactor is awesome. I tested in a different ways, mixing POs in a single vendor bills. It worked correctly.

I think it should change the delivery slip report in the picking. Because if I play with the PO line number in the PO after PO confirmation and I print the delivery slip I get this:
delivery_slip

AaronHForgeFlow (commented):
:+1:  Agree to remove the dependency on the stock_picking_line_sequence as long as the master data is in the purchase order now.

One test failed in runbot: http://oca-purchase-workflow-14-0-pr1933-394ad914cdd0.runboat.odoo-community.org/web#id=9&action=483&model=purchase.order&view_type=form&cids=&menu_id=322

If I create the vendor bill and then I reorder the lines in the PO, the changes are not propagated to the existing vendor bill. Can you check?

LoisRForgeFlow (commented):
Functional review, almost perfect.

I have seen an issue doing the folling:
1. Create a bill from PO.
2. Auto complete that same bill with another PO

image

The PO line number of the lines of the first PO did not update

OriolMForgeFlow:
Feedback applied. Now the field "related_po_sequence" is no longer set to "store=True." 

Currently, when you reorder the lines in the purchase order (PO), the changes are reflected in the invoice, but the moves are not reordered. I'm not sure if this is valid to you, @AaronHForgeFlow.

AaronHForgeFlow:
> Feedback applied. Now the field "related_po_sequence" is no longer set to "store=True."
> 
> Currently, when you reorder the lines in the purchase order (PO), the changes are reflected in the invoice, but the moves are not reordered. I'm not sure if this is valid to you, @AaronHForgeFlow.

Yes, that is valid to me. The important thing is the sequence to be the same in both.

AaronHForgeFlow (approved):
Functional tests. Thanks for the work!

purchase_order_line_sequence/views/report_purchaseorder.xml
```diff
@@ -5,12 +5,12 @@
         inherit_id="purchase.report_purchaseorder_document"
     >
         <xpath expr="//table/thead/tr/th[1]" position="before">
-            <th><strong>Sequence</strong></th>
+            <th><strong>Line Number</strong></th>
```
yostashiro, line 8: Would you consider just making it `No.` to keep it short?

OriolMForgeFlow: Hi @yostashiro,
We have chosen this name for the field to follow the criteria used by Odoo in other standard fields, such as Serial Number. They do not employ abbreviations like No. or Nº for labels.

yostashiro: I see. I thought this case could be different to keep a good balance between the label and the content - the content is usually only 1 or 2 digits.  But I'm OK if other people think otherwise.  Thank you very much for this improvement.

AaronHForgeFlow (approved):
Last changes of not letting the sequence to be updated from the picking are nice. :+1:  LGTM

AungKoKoLin1997 (approved):

LoisRForgeFlow (approved):
Functional and code review, this one is ready for me :+1:

PR #1936, opened 2023-07-09
[MIG] purchase_rfq_number: Migration to 16.0

hassan510-cmd:


dariodelzozzo (approved):

purchase_rfq_number/tests/test_purchase_order.py
```diff
@@ -68,10 +68,7 @@ def test_error_confirmation_sequence(self):
             ]
         )
         next_name = sequence_id.get_next_char(sequence_id.number_next_actual)
-        try:
-            order.button_confirm()
-        except UserError:
-            pass
+        order.button_confirm()
```
StefanRijnhart, line 71: What was the error before? Given the name of the method, it seems kind of important that the error is reproduced.

purchase_rfq_number/__manifest__.py
```diff
@@ -16,7 +16,6 @@
         "views/res_config_settings_views.xml",
         "views/purchase_views.xml",
     ],
-    "maintainer": ["prapassornS"],
```
StefanRijnhart: Is the removal by request of the maintainer?

purchase_rfq_number/tests/test_purchase_order.py
StefanRijnhart: I think the error hasn't been occurring for some time now. I've tried to clarify the test according to my own interpretation in the 17.0 migration of this module at https://github.com/OCA/purchase-workflow/pull/2349

PR #1939, opened 2023-07-13
[15.0][ENH] purchase_deposit, add advance return option

kittiu:
Currently, for PO with Deposit. 1st Invoice is Deposit, and 2nd Invoice will return "full" deposit regard less if it is a partial invoice.
This enhancement, allow user to choose how to return deposit in invoices that follows 1) Proportional 2) Full (standard)

Clicking Create Bill on PO with Deposit will show this,

image

If user choose to return in proportion,

image

### DIFF
--- a/purchase_deposit/README.rst
+++ b/purchase_deposit/README.rst
@@ -41,6 +41,7 @@ Normally, deposit will be used to create the 1st bill (as deposit).
 #. On confirmed purchase order, click Register Deposit button, wizard will open
 #. 2 type of deposit bill can be create 1) Down Payment (percentage) 2) Deposit Payment (fixed amount)
 #. Fill in the value and click Create bill or just create deposit.
+#. When create bill, there are 2 options to return advance amount, by proportional and in full.
 
 As deposit is created, when user click button "Create Bill" again in purchase order,
 the Vendor Bill will be created with deposit amount deducted.
--- a/purchase_deposit/__manifest__.py
+++ b/purchase_deposit/__manifest__.py
@@ -14,6 +14,7 @@
     "data": [
         "security/ir.model.access.csv",
         "wizard/purchase_make_invoice_advance_views.xml",
+        "wizard/purchase_advance_deduct_option_views.xml",
         "views/res_config_settings_views.xml",
         "views/purchase_view.xml",
     ],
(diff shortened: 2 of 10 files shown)


ps-tubtim (approved):
Functional Test: Ok.

ps-tubtim:
@kittiu Do you have plans to update the test to cover proportional return cases?

kittiu:
> @kittiu Do you have plans to update the test to cover proportional return cases?

Test added

kittiu:
@AungKoKoLin1997, are you interested to review this feature?

purchase_deposit/wizard/purchase_advance_return_option.py
```diff
@@ -0,0 +1,31 @@
...
+from odoo import fields, models
+
+
+class PurchaseAdvanceReturnOption(models.TransientModel):
+    _name = "purchase.advance.return.option"
+    _description = "Select how to return advance in vendor bill"
+
+    advance_return_option = fields.Selection(
+        [
+            ("proportional", "Propotional Return"),
```
Saran440: I think `Clear` is a simpler term to understand than `Return`

What do you think?

Saran440 (commented):
Functional Test :+1:

AungKoKoLin1997 (commented):
As suggested above, I think it may confuse to the users. How about using deduct instead of return?
```
advance_return_option = fields.Selection(
        [
            ("proportional", "Deduct Deposit Proportionally"),
            ("full", "Deduct Full Deposit (Standard)"),
        ],
        string="Advance/Deposit Deduction Option",
        default="proportional",
        required=True,
    )
```

purchase_deposit/wizard/purchase_advance_return_option.py
kittiu: I think return is also good. May I keep it this way?

kittiu:
> As suggested above, I think it may confuse to the users. How about using deduct instead of return?
> 
> ```
> advance_return_option = fields.Selection(
>         [
>             ("proportional", "Deduct Deposit Proportionally"),
>             ("full", "Deduct Full Deposit (Standard)"),
>         ],
>         string="Advance/Deposit Deduction Option",
>         default="proportional",
>         required=True,
>     )
> ```

Changed to Deduction

AungKoKoLin1997 (approved):
LGTM

PR #1942, opened 2023-07-14
[FIX] purchase_request: fix datetime from request to order 

alan196:
This PR fixes an issue that pass an incorrect date to the deadline to the purchase order.

Steps to reproduce:

Configure Odoo and your OS in a timezone different from UTC. For example, I'm in México (UTC-6).

Create a purchase request and define required date at 30-07-2023.

Then create a purchase order from the purchase request.

The scheduled date for the purchase order line is 29-07-2023 18:00:00 

With this change, the date now will be 30-07-2023 00:00:00


HviorForgeFlow (approved):

YahairaCVJarsa (approved):

HviorForgeFlow:
@alan196 is it necessary in v16?

PR #1944, opened 2023-07-19
[15.0] [FIX] purchase_delivery_split_date: Deadline and Scheduled Date

florentx:


simahawk (approved):
Seems legit. Can you please shorten the commit msg?
"fix is enough".
FTR https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

grindtildeath (approved):
IMO would be great if you could explain a bit more in the commit message what the fix is about, because it doesn't tell much otherwise

PR #1945, opened 2023-07-19
[15.0] [FIX] purchase_request: invalid field

florentx:


ajaniszewska-dev (approved):

purchase_request/models/stock_move.py
```diff
@@ -75,7 +75,7 @@ def _action_cancel(self):
     def _compute_purchase_request_ids(self):
         for rec in self:
             rec.purchase_request_ids = rec.purchase_request_allocation_ids.mapped(
```
simahawk: you could also drop "mapped"

simahawk (approved):

purchase_request/models/stock_move.py
florentx: how?

simahawk: just remove it: `rec.purchase_request_allocation_ids.purchase_request_line_id.request_id`.
Works fine w/ relations.

florentx: I didn't know that this works for o2m and m2m fields.
I still wonder if `mapped` has a benefit to remove duplicates before returning the recordset?

PR #1952, opened 2023-07-24
[14.0][MIG] subcontracted_service

LoisRForgeFlow:
* Depends on https://github.com/OCA/stock-logistics-warehouse/pull/1801.

@ForgeFlow


AaronHForgeFlow (approved):
LGTM, it creates RFQ when confirming sales order with kits with products that are subcontracted.

In v10 when confirming a sales order with a subcobtracted product it created an RFW as well, like it were a MTO for services. I think that makes no sense because the subcobtracted service must come with a storable product the company is selling. In any case that is how the module works in v10 (not in newer versions).

Then I guess some integration is needed MRP in order to create RFQ upon MOs that include subcontracted services, but that should be done in other module if that is the case.

:+1:  ok to refactor.

LoisRForgeFlow:
> LGTM, it creates RFQ when confirming sales order with kits with products that are subcontracted.
> 
> In v10 when confirming a sales order with a subcobtracted product it created an RFW as well, like it were a MTO for services. I think that makes no sense because the subcobtracted service must come with a storable product the company is selling. In any case that is how the module works in v10 (not in newer versions).
> 
> Then I guess some integration is needed MRP in order to create RFQ upon MOs that include subcontracted services, but that should be done in other module if that is the case.
> 
> +1 ok to refactor.

Yes, indeed. Once this base module get merged, I'll propose a submodule to cover those usecases!

DavidJForgeFlow (approved):
LGTM!

HviorForgeFlow (approved):
LGTM

AaronHForgeFlow:
> > LGTM, it creates RFQ when confirming sales order with kits with products that are subcontracted.
> > In v10 when confirming a sales order with a subcobtracted product it created an RFW as well, like it were a MTO for services. I think that makes no sense because the subcobtracted service must come with a storable product the company is selling. In any case that is how the module works in v10 (not in newer versions).
> > Then I guess some integration is needed MRP in order to create RFQ upon MOs that include subcontracted services, but that should be done in other module if that is the case.
> > +1 ok to refactor.
> 
> Yes, indeed. Once this base module get merged, I'll propose a submodule to cover those usecases!

I cover the case of procuring a service upon the procurement of a storable product here: https://github.com/OCA/purchase-workflow/pull/2030

PR #1954, opened 2023-07-26
[15.0][FIX] purchase_stock_price_unit_sync: _bom_find has no keyword argument 'product'

CarlosRoca13:
cc @Tecnativa TT40230

ping @pedrobaeza @carlosdauden


pedrobaeza (approved):
Checked with https://github.com/odoo/odoo/blob/a97ae6ba6d1fcf7f5a3e0cef12c9810617d235a6/addons/mrp/models/mrp_bom.py#L234, so we can directly fast-track this to solve the problem:

/ocabot merge patch

PR #1958, opened 2023-07-28
Fix sequence usage considering company

simoneversienti:
This PR is to fix a not completely correct behavior of this module in a multicompany context. 

Example: we have three companies active in the drop-down menu on the right (A, B and C) and the active one is A. So,  `self.env.company` is A. 
Purchase --> Request for quotations --> Create. 

I can create a RFQ with company_id = B. 
When i save, the module calculates the sequence calling `next_by_code` but, if not explicitly stated, the company considered is `self.env.company`, so A. 

In this case, i have a RFQ with company B but with the sequence of company A, which is wrong. 

The same happens when I confirm a RFQ of company B.

### DIFF
--- a/purchase_rfq_number/models/purchase_order.py
+++ b/purchase_rfq_number/models/purchase_order.py
@@ -33,7 +33,13 @@ class PurchaseOrder(models.Model):
             keep_name_po = self.env.company.keep_name_po
 
         if not keep_name_po and vals.get("name", "New") == "New":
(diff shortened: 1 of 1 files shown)


dariodelzozzo (approved):

valeriobelcastro (approved):

HviorForgeFlow (approved):
Code review :+1:

simoneversienti:
@HviorForgeFlow thanks for the merge! FYI here you can find the same changes, but for `sale-workflow`: https://github.com/OCA/sale-workflow/pull/2625

PR #1962, opened 2023-08-01
[13.0][FIX]purchase_origin_link_sale: properly get the SO from Source Document

GuillemCForgeFlow:
Before the change, the obtaining of the Sales Order was done in a way that could return several, the Source Document may only be clickable if it points to a single Sales Order.

cc @ForgeFlow

### DIFF
--- a/purchase_origin_link_sale/models/purchase.py
+++ b/purchase_origin_link_sale/models/purchase.py
@@ -3,18 +3,23 @@
 
 from odoo import api, models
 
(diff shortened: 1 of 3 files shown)


JordiMForgeFlow (approved):
LGTM 👍🏼

mariadforgeflow (approved):
LGTM!

BernatPForgeFlow (approved):

PR #1979, opened 2023-08-18
[IMP] purchase_order_weight_volume: added license

ilyasProgrammer:
CP from 14.0


ilyasProgrammer:
@simahawk here you go

PR #1983, opened 2023-08-23
[16.0][ADD] purchase_sale_link_by_origin

hoangphuc1410:
This module also link them by the PO's Origin field, to cover more cases. For example:
- If a user cancels a PO, by default the link would have been broken; now it won't;
- Or if a user manually defines or updates the Origin field of a PO, it will be taken into account.


simahawk (approved):
LG

ajaniszewska-dev (approved):

PR #1985, opened 2023-08-23
[16.0][MIG] purchase_manual_delivery: Migration to 16.0

PauBForgeFlow:


lfreeke (commented):
I notice that the wrong quantity is set when creating an incoming shipment. 

Steps to reproduce: 
1. Create Purchase order
2. Add a product to the purchase order lines with for example quantity 50
3. Check on 'Purchase manual delivery?'
4. Click on Confirm Order
5. Click on 'Create Incoming Shipment'
6. Set quantity 20 on the product order line in the wizard

Result an incoming transfer is created with Demand 100 instead of the 20 added in 'Create Incoming Shipment' wizard

PauBForgeFlow:
> I notice that the wrong quantity is set when creating an incoming shipment.
> 
> Steps to reproduce:
> 
> 1. Create Purchase order
> 2. Add a product to the purchase order lines with for example quantity 50
> 3. Check on 'Purchase manual delivery?'
> 4. Click on Confirm Order
> 5. Click on 'Create Incoming Shipment'
> 6. Set quantity 20 on the product order line in the wizard
> 
> Result an incoming transfer is created with Demand 100 instead of the 20 added in 'Create Incoming Shipment' wizard

Thank you for this review. I think it is wrong in 14v too, so I will fix there.

yung-wang (approved):
Seems to work fine

HviorForgeFlow (approved):

PR #1994, opened 2023-09-06
[IMP] purchase_default_terms_conditons : use parent terms if present

xavier-bouquiaux:
Improvethe module to take the note from the parent if the partner has no note


HviorForgeFlow:
It could be considered to use commercial_partner_id instead of parent_id

acsonefho (approved):

fgi-inoia:
> It could be considered to use commercial_partner_id instead of parent_id

What does it change? The commercial_partner_id being the parent. If no parent, the commercial_partner_id is the current record.

PR #2000, opened 2023-09-13
[14.0][FIX] purchase_stock_price_unit_sync: Change to the appropriate context used in purchase_discount v14

victoralmau:
Change to the appropriate context used in `purchase_discount` v14

This is only necessary in v14, in v13 the context was `purchase_discount` (https://github.com/OCA/purchase-workflow/blob/13.0/purchase_discount/models/purchase_order.py#L72C31-L72C48) and in v15 it is `skip_update_price_unit` (https://github.com/OCA/purchase-workflow/blob/15.0/purchase_discount/models/purchase_order.py#L72C34-L72C56).

@Tecnativa


pedrobaeza (approved):
Fast-tracking it:

/ocabot merge patch

Why didn't you check it before blindly forward-porting the commits?

PR #2001, opened 2023-09-13
[16.0] update dotfiles

sbidoul:
The pylint-odoo version we got with the latest template is complaining about a translation string, which I fix too.


pedrobaeza (approved):
/ocabot merge patch

PR #2011, opened 2023-09-20
[IMP] purchase_request: transfer requested by user to purchase request

alan196:
This PR improves the Purchase Request.

If another document is created and the product is configured as MTO and to create a purchase request.

It will create a purchase request, with odoobot as requested by.

With this change, the user will be passed to the Purchase Request


alan196:
@hectormerazchavez

Could you help me to review?

PR #2012, opened 2023-09-23
[16.0][FIX] purchase_discount: Apply always seller discount

### LINKED ISSUE #1996: [16] Purchase_discount - Supplier discount not applied on variants
Supplier discount not applied on variants

## Module
Purchase_discount 

## Describe the bug
When adding a product with variants in a purchase order, the variant selection wizard allow variant selection.
When validating, selected products are added to purchase order but for these products, supplier discount is not applied, even if correctly declared in product purchase line info

## To Reproduce
odoo v16

1 - Create supplier S1 with default discount 10%
2 - Create a product P1 with 2 variants
3 - Filling purchase lines for supplier S1 for each variant. At this step, default S1 discount is well applied on lines
4 - Create a purchase PO001 order with supplier S1
5 - Add a purchase line with P1
6 - The variant selection wizard openned
7 - select qty 1 for first variant
8 - Click on Confirm button

**Expected behavior**
The new purchase line for P1 is correctly added to PO001, but the discount value is 0
This discount value should be 10%

**Additional context**
the _onchange_quantity of PurchaseOrderLine class is not triggered when coming from the variant selection wizard
I tried some modifications, like overriding create method of PurchaseOrderLine class but it seams that nothing is triggered when coming back from the wizard (witch I supposed is js managed)

pedrobaeza:
When using the matrix input, the previously used onchange method doesn't serve, so we need to use the computed field that is now for this purpose.

Fixes #1996

@Tecnativa

### DIFF
--- a/purchase_discount/models/purchase_order.py
+++ b/purchase_discount/models/purchase_order.py
@@ -82,24 +82,24 @@ class PurchaseOrderLine(models.Model):
             self.with_context(skip_update_price_unit=True).price_unit = price_unit
         return price
 
-    @api.onchange("product_qty", "product_uom")
-    def _onchange_quantity(self):
+    def _compute_price_unit_and_date_planned_and_name(self):
+        """Get also the discount from the seller. Unfortunately, this requires to
+        select again the seller to be used, as there isn't any hook to use the already
+        selected one.
         """
-        Check if a discount is defined into the supplier info and if so then
-        apply it to the current purchase order line
-        """
-        if self.product_id:
-            date = None
-            if self.order_id.date_order:
-                date = self.order_id.date_order.date()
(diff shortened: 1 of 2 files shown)


legalsylvain (approved):
Thanks for the fix, pedro ! 

Cascade of bugs : CI is red because other module ``purchase_order_product_recommendation`` also overload the non existing function ``_onchange_quantity``. https://github.com/OCA/purchase-workflow/actions/runs/6284384018/job/17065702775?pr=2012#step:8:560

chienandalu (approved):

pedrobaeza:
Fixed also the one in the other module, so let's go!:

/ocabot merge patch

PR #2015, opened 2023-09-25
[16.0][FIX] purchase_discount: Don't call _select_seller without product

rousseldenis:
An example of failed job :

https://github.com/OCA/account-invoicing/actions/runs/6297612620/job/17094885728?pr=1562


rousseldenis:
@pedrobaeza This is a fix after the change you did there https://github.com/OCA/purchase-workflow/pull/2012

pedrobaeza (approved):
/ocabot merge patch

PR #2021, opened 2023-10-04
[14.0][FIX] purchase_discount: ocb compatibility

chienandalu:
Depends on:

- [x] https://github.com/OCA/OCB/pull/1193

Since oca/ocb@21a01c303347b8b80d0ffa555ed72e680c251e76 we already have proper hooks in OCB so we don't need to do nasty hacks, but we need to keep compatibility with both Odoo and OCB.

cc @Tecnativa TT45385

please review @pedrobaeza @victoralmau

### DIFF
--- a/purchase_discount/models/purchase_order.py
+++ b/purchase_discount/models/purchase_order.py
@@ -65,6 +65,8 @@ class PurchaseOrderLine(models.Model):
         HACK: This is needed while https://github.com/odoo/odoo/pull/29983
         is not merged.
         """
+        if hasattr(self.env, "ocb"):
+            return super()._get_stock_move_price_unit()
         bypass_price_unit = self.env.context.get("bypass_override_price_unit")
         price_unit = False
         price = self._get_discounted_price_unit()
--- a/purchase_discount/models/stock_move.py
+++ b/purchase_discount/models/stock_move.py
@@ -15,6 +15,8 @@ class StockMove(models.Model):
         HACK: This is needed while https://github.com/odoo/odoo/pull/29983
         is not merged.
         """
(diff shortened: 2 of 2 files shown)


pedrobaeza (approved):

purchase_discount/models/purchase_order.py
```diff
@@ -65,6 +65,8 @@ def _get_stock_move_price_unit(self):
         HACK: This is needed while https://github.com/odoo/odoo/pull/29983
         is not merged.
         """
+        if hasattr(self.env, "ocb"):
```
legalsylvain, line 68: naive question : I didn't know ocb adds an extra val in self.env. is it documented somewhere ?

legalsylvain (approved):
code review / no test.

purchase_discount/models/purchase_order.py
pedrobaeza: Check OCA/OCB#1193, which is flagged on the header :wink:

legalsylvain: thanks !

victoralmau (approved):

chienandalu:
Is there a way to force a new OCB package for the CI build? (https://github.com/OCA/oca-ci/pkgs/container/oca-ci%2Fpy3.6-ocb14.0)

pedrobaeza:
We have to wait till tomorrow...

pedrobaeza:
@chienandalu why is it still failing?

chienandalu:
> @chienandalu why is it still failing?

The package was updated 1 hour ago. Now is green :)

PR #2025, opened 2023-10-05
[DOC] Add image Test

TumbaoJu:
This a test


TumbaoJu:
Adding an image to the Read me from the repository forked doesn't work because the image is stored in the user repository, not in OCA repository.
@francesco-ooops 
@sbidoul

legalsylvain:
hi. you should add the image in the static folder and make a reference in the md / rst file with a relative path.

PR #2030, opened 2023-10-11
[15.0][ADD] subcontracted_service_product

AaronHForgeFlow:
This module allows to automatically procure (purchase) a service upon the procurement of a storable product.
For example, we sell a fridge and that triggers a purchase of the service for the installation, that is external to the company

cc @ForgeFlow


subcontracted_service_product/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,4 @@
+This module allows to automatically purchase a service upon the procurement of
+a storable product.
+For example, we sell a fridge and that triggers a purchase of the service for
```
JordiMForgeFlow, line 3:
```suggestion
For example, we purchase a fridge and that triggers a purchase of the service for
```

subcontracted_service_product/models/procurement_group.py
```diff
@@ -0,0 +1,64 @@
...
+
+    @api.model
+    def _is_subcontracted_service_procurement(self, product_id):
+        return (
+            product_id.type == "service"
+            and len(
+                self.env["product.template"].search(
+                    [("subcontracted_product_id", "=", product_id.id)]
+                )
+            )
```
JordiMForgeFlow: you could use search_count

subcontracted_service_product/models/procurement_group.py
```diff
@@ -0,0 +1,64 @@
...
+            > 0
+        )
+
+    @api.model
+    def _get_rule(self, product_id, location_id, values):
+        res = super()._get_rule(product_id, location_id, values)
+        if self._has_subcontracted_service_procurement(product_id):
+            self.trigger_service_procurement(product_id, location_id, values)
+        return res
```
JordiMForgeFlow, line 37: wouldn't it be more correct to do it on the _run_buy method?

subcontracted_service_product/readme/DESCRIPTION.rst
AaronHForgeFlow: Actually, the use case I have is a sales order. I sell the fridge and the customer want us to install it. But we contract an external service for doing that.

subcontracted_service_product/models/procurement_group.py
AaronHForgeFlow: good point!

subcontracted_service_product/models/procurement_group.py
AaronHForgeFlow: Because this is needed when creating a MTS sales order it has to be that way.

AaronHForgeFlow:
Stale + not needed. Closing.

PR #2034, opened 2023-10-12
[16.0][ADD] purchase_product_packaging_container_deposit

santostelmo:
Add purchase_product_packaging_container_deposit

- https://github.com/OCA/product-attribute/pull/1418


cyrilmanuel (approved):

cyrilmanuel (approved):

cyrilmanuel:
Missing unit price for Products: Product A,  in unitest

santostelmo:
> Missing unit price for Products: Product A, in unitest

Yes. That was a requirements from another module: purchase_order_no_zero_price

cyrilmanuel (approved):

Ricardoalso (approved):

purchase_product_packaging_container_deposit/models/purchase_order_line.py
```diff
@@ -0,0 +1,20 @@
...
+
+    def _get_product_qty_field(self):
+        return "product_qty"
+
+    def _get_product_qty_delivered_received_field(self):
+        return "qty_received"
+
+    def _compute_qty_received(self):
+        res = super()._compute_qty_received()
+        self.mapped("order_id").update_order_container_deposit_quantity()
```
simahawk: I'm looking again at this and I don't think it's a good idea to trigger other writes in a computed field.
It can be dangerous and trigger side effects you won't expect.
I know that the same code exists in https://github.com/OCA/sale-workflow/pull/2692 but I think I overlooked there.
I'd rather  override `write` or use an inverse method.
In fact, if you override the write in the mixin you can avoid this override here.
You can then check if this field is written, if yes, you call the update.

simahawk (changes requested):

purchase_product_packaging_container_deposit/models/purchase_order_line.py
santostelmo: While testing `test_purchase_product_packaging_container_deposit_quantities_case7`
I tried to execute `update_order_container_deposit_quantity` inside `write` or in method `inverse_qty_received`. But this methods does not get triggered when `qty_received` is updated. Therefore I don't see how I can implement your request.

simahawk: I doubt write is not called when you change `qty_received`. We can have a look together.

santostelmo: 🙏 Thanks let's check together.

santostelmo:
@simahawk I updated code with fix done in sale-workflow https://github.com/OCA/sale-workflow/pull/2833

purchase_product_packaging_container_deposit/tests/test_purchase_product_packaging_container_deposit.py
```diff
@@ -0,0 +1,283 @@
...
+            lambda ol: ol.product_id == self.box
+        )
+
+        (pallet_line | box_line).invalidate_cache()
+        self.assertEqual(pallet_line.qty_received, 0)
+        self.assertEqual(box_line.qty_received, 5)
+
+        self.purchase_order.order_line[0].qty_received = 200
+        self.assertEqual(pallet_line.qty_received, 0)
+        self.assertEqual(box_line.qty_received, 8)
```
simahawk, line 283: You are not testing the form here. Can you add the same test as in the sale module?

simahawk (commented):
LG, just one thing more

purchase_product_packaging_container_deposit/tests/test_purchase_product_packaging_container_deposit.py
santostelmo: The Form unit test was added.

simahawk (approved):

PR #2039, opened 2023-10-17
[ADD] purchase_transport_mode

santostelmo:
Purchases can be received by different modes of transport. For example, we can have a truck capable of transporting X tonnes and another Y tonnes.
With this module it is possible to validate that the purchase complies with the requirements of a particular mode of transport.

To use this module, you need to:
- In Purchase settings enable purchase transport mode validation
- Create a transport modes with constraints applied on PO  
- Select transport mode on PO

When you click on Confirm order, if the transport requirements are not met, a message in yellow will be displayed at the top of the form.


purchase_transport_mode/models/purchase_order.py
```diff
@@ -0,0 +1,65 @@
...
+        for rec in self:
+            rec.transport_mode_status = rec._get_transport_mode_validation_status()
+            rec.transport_mode_status_display = (
+                rec._get_transport_mode_validation_status_display()
+            )
+            rec.transport_mode_status_ok = rec.transport_mode_status == {}
+
+    @api.onchange("partner_id")
+    def onchange_partner_id(self):
+        if self.partner_id and not self.transport_mode_id:
```
simahawk:
```suggestion
        if self.partner_id:
```
otherwise if you switch from a partner w/ t-mode to another w/o you won't update the t-mode

purchase_transport_mode/tests/test_purchase_transport_mode.py
```diff
@@ -0,0 +1,67 @@
...
+                "description": "The total amount of the order must be higher than 5000",
+                "purchase_transport_mode_id": cls.transport_mode_a.id,
+                "purchase_domain": [("amount_total", ">", 5000)],
+            }
+        )
+
+    def enable_purchase_transaport_mode_validation(self):
+        self.settings = self.env["res.config.settings"].create({})
+        self.settings.check_purchase_transport_mode_contraints = True
+        self.settings.execute()
```
simahawk: never use execute because it does too many things: always use `set_values`

purchase_transport_mode/models/res_company.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2023 Camptocamp (<https://www.camptocamp.com>).
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    check_purchase_transport_mode_contraints = fields.Boolean()
```
simahawk:
```suggestion
    purchase_transport_mode_contraint_enabled = fields.Boolean()
```

purchase_transport_mode/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,3 @@
+* Telmo Santos <telmo.santos@camptocamp.com>
+* Cyril Jeanneret <cyril.jeanneret@camptocamp.com>
+* Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
```
simahawk, line 3:
```suggestion
* Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
* Simone Orsi <simone.orsi@camptocamp.com>
```

purchase_transport_mode/tests/test_purchase_transport_mode.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+from odoo.tests import TransactionCase
+
+
+class TestPurchaseTransportMode(TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
```
simahawk: just found out that in v16 we can use `from odoo.addons.base.tests.common import DISABLED_MAIL_CONTEXT` to set all mail tracking keys

purchase_transport_mode/models/purchase_order.py
```diff
@@ -0,0 +1,65 @@
...
+    )
+
+    @api.depends("transport_mode_id")
+    def _compute_transport_mode_validation_status(self):
+        for rec in self:
+            rec.transport_mode_status = rec._get_transport_mode_validation_status()
+            rec.transport_mode_status_display = (
+                rec._get_transport_mode_validation_status_display()
+            )
+            rec.transport_mode_status_ok = rec.transport_mode_status == {}
```
simahawk:
```suggestion
            rec.transport_mode_status_ok = not rec.transport_mode_status.get("errors")
```

purchase_transport_mode/readme/USAGE.rst
```diff
@@ -0,0 +1,7 @@
+To use this module, you need to:
+
+#. In Purchase settings enable purchase transport mode validation
+#. Create a transport modes with contraints applied on PO  
+#. Select transport mode on PO
+
+When you click on *Confirm order*, if the transport requirements are not met, a message in yellow will be displayed at the top of the form.
```
simahawk: actually you'll get the warning immediately

purchase_transport_mode/views/purchase_order_view.xml
```diff
@@ -0,0 +1,22 @@
...
+    <record id="purchase_order_view_form_inherit" model="ir.ui.view">
+        <field name="name">purchase.order.form.inherit</field>
+        <field name="model">purchase.order</field>
+        <field name="inherit_id" ref="purchase.purchase_order_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='currency_id']" position="after">
+                <field name="transport_mode_id" />
+            </xpath>
+            <xpath expr="//sheet" position="before">
+                <field name="transport_mode_status_display" nolabel="1" />
```
simahawk: I'm not sure the position is correct... it would be better before lines IMO and wrapped into an alert maybe.
Also, the field should be protected like this:
```suggestion
                <field name="transport_mode_status_ok" invisible="1" />
                <field name="transport_mode_status_display" nolabel="1"  attrs="{'invisible': [('transport_mode_status_ok', '=', False)]}" />
```

simahawk (changes requested):
LG overall

purchase_transport_mode/models/purchase_order.py
simahawk: Also, maybe worth to lookup for the t-mode on commercial partner?

purchase_transport_mode/tests/test_purchase_transport_mode.py
santostelmo: thank you ! 
indeed it will disable the following:
```
DISABLED_MAIL_CONTEXT = {
    'tracking_disable': True,
    'mail_create_nolog': True,
    'mail_create_nosubscribe': True,
    'mail_notrack': True,
    'no_reset_password': True,
}
```

purchase_transport_mode/tests/__init__.py
```diff
@@ -0,0 +1 @@
+# from . import test_purchase_transport_mode
```
jbaudoux: remove commented line

jbaudoux (approved):
LG, not tested

purchase_transport_mode/tests/__init__.py
santostelmo: Removed commented line.

santostelmo:
@simahawk anythingesle blocking ?

cyrilmanuel (approved):

simahawk:
tests to be fixed

santostelmo:
@simahawk tests are falling in `purchase_order_type` and I don't see any relation with `purchase_transport_move` 😕

simahawk:
it means there's a flappy test in that module... to be checked

santostelmo:
@simahawk tests are fixed and test coverage improved. Can you please check again ?

purchase_transport_mode/models/purchase_order.py
```diff
@@ -0,0 +1,71 @@
...
+
+from odoo.addons.base_sparse_field.models.fields import Serialized
+
+
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+    _name = "purchase.order"
+
+    transport_mode_id = fields.Many2one("purchase.transport.mode")
+    transport_mode_status = Serialized(
```
jbaudoux: @simahawk @gurneyalex Shouldn't we use `Json` field instead ? What's the pro/con between the 2 fields type?

simahawk: @jbaudoux you are right. I forgot we have that type now.

simahawk: @santostelmo ping :)

simahawk (approved):
beside the Json field looks good. pre-approving

purchase_transport_mode/models/purchase_order.py
santostelmo: I have converted to `Json` field

PR #1326, opened 2022-01-06
[14.0] Refactor purchase_delivery_split_date (support new PO line or increased qty on a confirmed PO)

sebalix:
Port of:
- https://github.com/OCA/purchase-workflow/pull/1246

PR #2049, opened 2023-10-19
[IMP] purchase_delivery_split_date: Split improvements

rousseldenis:
In order to not browsing the purchase order when splitting, do the split directly on purchase order lines as we have the modified recordset (for date_planned field).

Use partition() function in order to group lines per date then per picking to improve readability.

Use the standard move picking assignation (_assign_picking()) and add the grouping date in the domain.


rousseldenis:
@TDu @LoisRForgeFlow @JordiBForgeFlow

jbaudoux (approved):

jbaudoux (changes requested):
There are several issues that I fixed here: https://github.com/OCA/purchase-workflow/pull/1326

PR #2050, opened 2023-10-20
[15.0][FIX] purchase_request: activity assigned user upon procurement cancellation

AaronHForgeFlow:
Create a MTO sales order with a product that has purchase request activated in the routes section and has no responsible. Confirm the SO. Cancel the SO. The system fails because it tries to create a mail activity for the responsible of the product.
mail_activity_error

I keep the criteria in case there is a responsible (that also matches Odoo standard activity for purchases), then I choose the responsible of the purchase request, otherwise I choose the user that cancelled the sales order.

cc @ForgeFlow

### DIFF
--- a/purchase_request/models/stock_move.py
+++ b/purchase_request/models/stock_move.py
@@ -54,6 +54,14 @@ class StockMove(models.Model):
                 except ValueError:
                     activity_type_id = False
                 pr_line = move.created_purchase_request_line_id
+                if pr_line.product_id.responsible_id:
+                    activity_user = pr_line.product_id.responsible_id
+                elif pr_line.request_id.assigned_to:
+                    activity_user = pr_line.request_id.assigned_to
+                elif move.picking_id.user_id:
+                    activity_user = move.picking_id.user_id
+                else:
+                    activity_user = self.env.user
                 self.env["mail.activity"].sudo().create(
                     {
                         "activity_type_id": activity_type_id,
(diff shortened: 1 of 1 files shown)


LoisRForgeFlow (approved):
LGTM

JordiBForgeFlow (approved):

purchase_request/models/stock_move.py
```diff
@@ -54,6 +54,12 @@ def _action_cancel(self):
                 except ValueError:
                     activity_type_id = False
                 pr_line = move.created_purchase_request_line_id
+                if pr_line.product_id.responsible_id:
+                    activity_user = pr_line.product_id.responsible_id
+                elif pr_line.request_id.assigned_to:
+                    activity_user = pr_line.request_id.assigned_to
+                else:
```
MiquelRForgeFlow, line 63:
```suggestion
                elif move.picking_id.user_id:
                    activity_user = move.picking_id.user_id
                else:
```
does this make sense?

AaronHForgeFlow: Well, if there is a responsible in the picking, then it is ok to me that is notified with preference to the user that triggers the action. The preference is then:

* responsible in the product: in order to match Odoo standard notifications for purchases
* responsible of the purchase request
* responsible of the picking
* the user that triggers the action

I just did that.

MiquelRForgeFlow (approved):

HviorForgeFlow (approved):

PR #2053, opened 2023-10-23
[15.0][FIX] sale_purchase_secondary_unit: When a so is cancelled the secondary_uom_qty on po lines is not updated

sergio-teruel:
…ary_uom_qty on po lines is not updated

cc @Tecnativa TT45601

ping @carlosdauden @CarlosRoca13


carlosdauden (approved):

CarlosRoca13 (approved):
LGTM

PR #2060, opened 2023-10-26
[IMP] purchase_order_weight_volume store total weight volume store

santostelmo:
### Context

Often we need to search or apply a domain on the fields total weight and volume. To do it we need to store  the field.

Also if we store those values we keep the right value in the PO even if the configuration of the product change in time.


jbaudoux (approved):

cyrilmanuel (approved):

santostelmo:
@ilyasProgrammer as you're the maintainer can you please have a look here ? PR is ready to merge

ilyasProgrammer (approved):
LGTM

santostelmo:
Thank you @ilyasProgrammer !

PR #2070, opened 2023-10-30
[IMP] purchase_discount: checkbox to use real purchase supplier discount

angelgarciadelachica:
When a product quantity is lowered below the supplier's minimum quantity the discount is not updated to 0.00.

**To reproduce the error:**

1. Create a pricelist for a product with minimum quantity greater than one and discount greater than zero.
2. Create a purchase
3. Select the same partner of the pricelist created in step 1.
4. Enter the same product of the pricelist created in step 1.
5. Change the quantity to 1.

It is noted that the discount pricelist is maintained and it is not set to 0.00.

In addition, we have added a test to check this case.

PR #2071, opened 2023-10-30
[IMP] purchase_triple_discount: checkbox to use real purchase supplier discounts

angelgarciadelachica:
Similar error than FIX https://github.com/OCA/purchase-workflow/pull/2070
When a product quantity is lowered below the supplier's minimum quantity the discounts is not updated to 0.00.

**To reproduce the error:**
1. Create a pricelist for a product with minimum quantity greater than one and discount1 and discount2 greater than zero.
2. Create a purchase
3. Select the same partner of the pricelist created in step 1.
4. Enter the same product of the pricelist created in step 1.
5. Change the quantity to 1.

It is noted that the discounts pricelist is maintained and it is not set at 0.00.

In addition, we have added a test to check this case.


angelgarciadelachica:

HaraldPanten (approved):

ValentinVinagre (approved):
LGTM 👍🏻

legalsylvain (approved):
Thanks ! 
could you port this patch on 16.0 also ? 

/ocabot merge patch

pedrobaeza (changes requested):
I don't agree with this change, and even if so, the implementation assigning a variable that is only used once is not the best one for clarity and lines length.

pedrobaeza:
See #2070 for extended explanation.

legalsylvain:
> I don't agree with this change, and even if so, the implementation assigning a variable that is only used once is not the best one for clarity and lines length.


Thanks for reporting. so let's wait for the discussion to be resolved in #2070.

cyrilmanuel (approved):

PR #2073, opened 2023-10-30
[FIX] purchase_fop_shipping: purchase fop should not block negative PO

cyrilmanuel:
The exception that the PO cannot be below the FOP should not raise when the PO is negative as this is allowed.


santostelmo (approved):

Camille0907 (approved):

cyrilmanuel:
Hello @pedrobaeza is it good for you ? can we merge this PR ?

pedrobaeza:
I don't know anything about this module, but based on the existing reviews:

/ocabot merge patch

PR #2074, opened 2023-10-30
[16.0][FIX] purchase_allowed_product: use correct field name

JordiMForgeFlow:
In V16 the field is renamed to partner_id in Odoo (https://github.com/odoo/odoo/blob/5941629609977512829e30b3ccd7ad376bc2b71b/addons/product/models/product_supplierinfo.py#L29)

CC @pedrobaeza @HaraldPanten @angelgarciadelachica

### DIFF
--- a/purchase_allowed_product/models/product_supplierinfo.py
+++ b/purchase_allowed_product/models/product_supplierinfo.py
@@ -7,4 +7,5 @@ from odoo import fields, models
 class ProductSupplierinfo(models.Model):
     _inherit = "product.supplierinfo"
 
-    name = fields.Many2one("res.partner", index=True)
+    # Index partner_id
+    partner_id = fields.Many2one(index=True)


HaraldPanten (approved):
Makes sense 👍

ValentinVinagre (changes requested):
If the field changes name, a migration script is required to move the data.

JordiMForgeFlow:
> If the field changes name, a migration script is required to move the data.

I wasn't sure about that, as it is not meant to be a new field but to index the base one. The renaming should already be taken care of in the migration scripts (either via Odoo enterprise or Openupgrade). I think that users using this module in 16.0 simply have no data in the _name_ field.

ValentinVinagre:
> > If the field changes name, a migration script is required to move the data.
> 
> I wasn't sure about that, as it is not meant to be a new field but to index the base one. The renaming should already be taken care of in the migration scripts (either via Odoo enterprise or Openupgrade). I think that users using this module in 16.0 simply have no data in the _name_ field.

In this case LGTM 👍🏻  .

ValentinVinagre (approved):
LGTM 👍🏻

purchase_allowed_product/models/product_supplierinfo.py
```diff
@@ -7,4 +7,4 @@
 class ProductSupplierinfo(models.Model):
     _inherit = "product.supplierinfo"
 
-    name = fields.Many2one("res.partner", index=True)
+    partner_id = fields.Many2one("res.partner", index=True)
```
pedrobaeza: No need of the comodel name and put a comment to know what is happening.

```suggestion
    # Index this field
    partner_id = fields.Many2one(index=True)
```

JordiMForgeFlow: done! thank you @pedrobaeza

pedrobaeza (approved):
/ocabot merge patch

PR #2078, opened 2023-11-02
[16.0][MIG] purchase_invoice_new_picking_line

carmenbianca:
Internal task: https://gestion.coopiteasy.be/web#id=10753&action=479&model=project.task&view_type=form&menu_id=536


purchase_invoice_new_picking_line/models/purchase_order.py
```diff
@@ -0,0 +1,80 @@
...
+                partner_id=self.partner_id,
+                quantity=qty,
+                date=date,
+                uom_id=stock_move_id.product_uom,
+                params={"order_id": self},
+            )
+        if supplierinfo:
+            price_unit = supplierinfo.price
+        else:
+            price_unit = stock_move_id.product_id.lst_price
```
carmenbianca: I am not entirely sure about this. Previously, the `price_unit` came from the identically named field on `stock.move`. This field still exists in 16, but only contains useful data in some corner-cases.

I wrote the quoted snippet under the impression that it should be sufficient. However, upstream in `purchase`, the code to get the unit price for a `purchase.order.line` is much much much more complicated. See [this permalink](https://github.com/odoo/odoo/blob/b5b23c5c2f79678c8de9f1ebf6435246cb86fa23/addons/purchase/models/purchase.py#L1244). There is no way to easily re-use this code except copy-pasting and adjusting it, but I do not frankly understand some of what is going on there.

I am hopeful that the above will suffice. If not, I see two options:

1. Copy-paste the complicated linked code. Not great.
2. Pass the parcel. Instead of creating invoice lines from `stock.move`s, create `purchase.order.line`s from `stock.move`s upon validating `stock.picking`s. This seems doable, but I'm not sure if it's semantically correct, because these products were never _ordered_ as such.

purchase_invoice_new_picking_line/models/purchase_order.py
```diff
@@ -0,0 +1,81 @@
...
+                self.company_id,
+                date,
+                round=False,
+            ),
+            "tax_ids": [Command.set(tax_ids.ids)],
+        }
+
+        return data
+
+    def add_missing_picking_products_to_account_move(self, account_move_id):
```
carmenbianca, line 64: This could theoretically be called from `purchase_order_change` in `account.move`, which has changed in semantics since Odoo 12. `purchase_order` is not stored on `account.move`. Instead, if you add a purchase order in the interface, it 'loads' data from the purchase order. I don't know if `purchase_order_change` is idempotent—this method is not.

carmenbianca (commented):
still need to write some tests

carmenbianca:
Wrote the test. Ready to review, but also still needs some functional testing.

victor-champonnois (commented):
Functional Review OK, thanks !

victor-champonnois (approved):

remytms (approved):
LGTM

HviorForgeFlow (approved):

PR #2079, opened 2023-11-02
[15.0][FIX] purchase_discount: allow creating without product_tmpl_id

yajo:
Previous code was expecting a non-required field to be passed always. Now it's more resilient.

@moduon MT-1075


HviorForgeFlow (approved):
Code review :+1:

MiquelRForgeFlow (approved):

PR #2080, opened 2023-11-02
[FIX] purchase_discount: allow creating without product_tmpl_id

yajo:
Previous code was expecting a non-required field to be passed always. Now it's more resilient.

@moduon MT-1075 fwport-of https://github.com/OCA/purchase-workflow/pull/2079

### DIFF
--- a/purchase_discount/tests/test_product_supplierinfo_discount.py
+++ b/purchase_discount/tests/test_product_supplierinfo_discount.py
@@ -30,10 +30,12 @@ class TestProductSupplierinfoDiscount(TransactionCase):
             {
                 "min_qty": 10.0,
                 "partner_id": cls.partner_3.id,
-                "product_tmpl_id": cls.product.product_tmpl_id.id,
                 "discount": 20,
             }
         )
+        # Assign it later this value for testing that the record
+        # creation is not failing without this field
+        cls.supplierinfo2["product_tmpl_id"] = cls.product.product_tmpl_id
         cls.purchase_order = cls.env["purchase.order"].create(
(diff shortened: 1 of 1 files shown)


purchase_discount/tests/test_product_supplierinfo_discount.py
```diff
@@ -30,10 +30,10 @@ def setUpClass(cls):
             {
                 "min_qty": 10.0,
                 "partner_id": cls.partner_3.id,
-                "product_tmpl_id": cls.product.product_tmpl_id.id,
                 "discount": 20,
             }
         )
+        cls.supplierinfo2["product_tmpl_id"] = cls.product.product_tmpl_id
```
pedrobaeza, line 38: Put a comment why this weird thing:

```suggestion
        # Assign it later this value for testing that the record
        # creation is not failing without this field
        cls.supplierinfo2.product_tmpl_id = cls.product.product_tmpl_id
```

chienandalu (changes requested):
Take a look at https://github.com/OCA/purchase-workflow/pull/1780 :smile:

purchase_discount/tests/test_product_supplierinfo_discount.py
yajo: Thanks, fixed.

yajo:
> Take a look at #1780 😄

I see you included the fix from https://github.com/OCA/purchase-workflow/pull/2079/files#diff-d295e248990e2d37610dbda37a9460c56705b72005b4d1c7144af8c1c829a1c9 so I guess this PR is just a little test enhancement now.

PR #2081, opened 2023-11-02
[16.0][MIG]purchase_picking_state

manuelregidor:
Supersedes https://github.com/OCA/purchase-workflow/pull/1832


manuelregidor:

manuelregidor:
@sbidoul @MiquelRForgeFlow @pedrobaeza @HaraldPanten  Since the module purchase_stock in v16 adds the field receipt_status to purchase.order, is it necessary to migrate this module to v16? The main difference is that purchase_stock only includes three states (purchase_picking_state adds 5). You can see the changes in purchase_stock in v16 here: https://github.com/OCA/OpenUpgrade/blob/16.0/openupgrade_scripts/scripts/purchase_stock/16.0.1.2/upgrade_analysis.txt
In case migration is not needed, should we add a migration script?
Thank you

pedrobaeza:
Yes, I think it's perfectly replaceable. The cancel status is the only one missing, as draft is a very weird one.

manuelregidor:
In that case, do you think we should keep purchase_picking_state in v16 to add the receipt_status field from module purchase_stock to the tree and search views?

pedrobaeza:
I would first check what Odoo does with cancel stock.move.

ValentinVinagre:
> I would first check what Odoo does with cancel stock.move.

If all pickings are in state "cancel" the value is false.
https://github.com/odoo/odoo/blob/d9a3740ee7578f51f8f77fe5b1176bb6792364ea/addons/purchase_stock/models/purchase.py#L67

SkiBY (approved):
Sorry. Wrong PR

ValentinVinagre:
@pedrobaeza Do you think it is correct to change the module based on what was discussed previously?

> In that case, do you think we should keep purchase_picking_state in v16 to add the receipt_status field from module purchase_stock to the tree and search views?

pedrobaeza:
With your last comment, I see this module as non needed.

ValentinVinagre:
> With your last comment, I see this module as non needed.

Wouldn't it be necessary to show the field in the shopping tree view and in the group view? Odoo only shows this field in the form view.
I don't see adding the "Cancelled" option as a bad idea to complement what Odoo has with what this module previously provided.
Should we let it die or adapt it to improve odoo?

SkiBY (changes requested):
revoken, sorry

pedrobaeza:
> Wouldn't it be necessary to show the field in the shopping tree view and in the group view? Odoo only shows this field in the form view.

Well, that's a desirable one, but I think a module just for that isn't worth.

> I don't see adding the "Cancelled" option as a bad idea to complement what Odoo has with what this module previously provided. Should we let it die or adapt it to improve odoo?

In fact, I see better the Odoo approach to not put anything as value in this case.

ValentinVinagre:
> > Wouldn't it be necessary to show the field in the shopping tree view and in the group view? Odoo only shows this field in the form view.
> 
> Well, that's a desirable one, but I think a module just for that isn't worth.
> 
> > I don't see adding the "Cancelled" option as a bad idea to complement what Odoo has with what this module previously provided. Should we let it die or adapt it to improve odoo?
> 
> In fact, I see better the Odoo approach to not put anything as value in this case.

It's okay. Could you indicate in the migration issue that this module is discontinued? For our part we will close the PR.
Thanks Pedro.

PR #2085, opened 2023-11-03
[IMP] purchase_discount: allow hiding discount column

### LINKED ISSUE #2067: [16] purchase_discount - field discount optional in purchase.order.line
**Is your feature request related to a problem?**
Sometimes there are many fields in the purchase order line and the discount field cannot be hidden.
It would be nice to be able to choose whether to show it or not.  

**Describe the solution you'd like**
Make the visibility of the `discount` field in the purchase order line optional. 

**Describe alternatives you've considered**
Sometimes there are many fields and you have to scroll. I think it would be useful if the visibility of the discount `field` was optional and could be hidden if desired.
image

ping @ramiadavid 
@moduom MT-4081

yajo:
Still displayed by default, but it can be hidden now:

flameshot_2023-11-03_11-09

Fixes https://github.com/OCA/purchase-workflow/issues/2067.

@moduon MT-4081

### DIFF
--- a/purchase_discount/views/purchase_discount_view.xml
+++ b/purchase_discount/views/purchase_discount_view.xml
@@ -6,7 +6,7 @@
         <field name="inherit_id" ref="purchase.purchase_order_line_form2" />
         <field name="arch" type="xml">
             <field name="price_unit" position="after">
-                <field name="discount" />
+                <field name="discount" optional="show" />
             </field>
         </field>
     </record>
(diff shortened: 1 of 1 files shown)


rafaelbn (approved):
👍🏼 

Thank you!

Gelojr (approved):
LGTM https://www.loom.com/share/87424acad0a34ec98b7d442f45892edf?sid=a50410fe-4825-4bda-8a0a-d97d51984b1b
Thanks @yajo

PR #2088, opened 2023-11-08
[14.0] [IMP] purchase_allowed_product: global setting and access rights

renda-dev:
- Add a global setting in purchase: "**Use only allowed products** by default"
    - When creating a new PO, field "**Use only allowed products**" is enabled by default
- Add a configuration in user: "Hide "**Use only allowed products**" in PO and invoice"
    - If this field is enabled in user, field "**Use only allowed products**" is not visible for this user.


pedrobaeza:
Your can do that setting a default, and the configuration settings have a direct option for that. Check other examples in Odoo core.

renda-dev:
Hi @pedrobaeza, that's awesome, thank you so much for letting me know :)

francesco-ooops:
@renda-dev there is something strange in settings as I don't see the setting and the "Purchase" section is duplicated

image

francesco-ooops (approved):
Functional ok!

the error above is due to some other module (will try to find the culprit)

geomer198 (approved):
LGTM

francesco-ooops:
@pedrobaeza good for merge?

purchase_allowed_product/models/supplied_product_mixin.py
```diff
@@ -8,15 +8,27 @@ class SuppliedProductMixin(models.AbstractModel):
...
+        'Hide "Use only allowed products" in PO and invoice',
+        default=_default_hide_use_only_supplied_product,
+        readonly=True,
+    )
+
     @api.onchange("partner_id")
     def _onchange_partner_id_supplied_product(self):
         self.use_only_supplied_product = (
-            self.partner_id.use_only_supplied_product
+            self.use_only_supplied_product
```
pedrobaeza: With this, you are avoiding the option to disable the check automatically if previously was set. This is an undesired behavior.

purchase_allowed_product/models/supplied_product_mixin.py
```diff
@@ -8,15 +8,27 @@ class SuppliedProductMixin(models.AbstractModel):
     _name = "supplied.product.mixin"
     _description = "Mixin for documents that want to restrict products"
 
+    def _default_hide_use_only_supplied_product(self):
```
pedrobaeza: Instead of doing this, just put the corresponding `groups` in the view.

pedrobaeza (changes requested):

francesco-ooops (approved):
Functionally still good, @pedrobaeza what do you say?

francesco-ooops:
@renda-dev rebase please

francesco-ooops:
@pedrobaeza shall we go? :)

pedrobaeza (approved):
/ocabot merge minor

PR #2090, opened 2023-11-08
[14.0][FIX] purchase_partner_approval: pre-approve existing partners

HekkiMelody:
Before this PR, only the first partner in alphabetical order was pre-approved for purchase.

I've also moved the logic to a post_init_hook , which is more appropriate for this than a data file with noupdate.


francesco-ooops (approved):
Functional ok! Solves issue reported in https://github.com/OCA/purchase-workflow/pull/1388#issuecomment-1428190766

@dreispt good for you?

geomer198 (approved):
Code Review ok!

francesco-ooops:
@pedrobaeza runboat bugfix

purchase_partner_approval/hooks.py
```diff
@@ -0,0 +1,16 @@
+from odoo import SUPERUSER_ID, api
+
+
+def approve_existing_partners_for_purchase(env):
+    """
+    To not cause disruption upon install,
+    set pre-existing Partners as Can Purchase
+    """
+    env["res.partner"].search([]).write(
```
pedrobaeza: Do this with SQL:

```
UPDATE res_partner SET purchase_ok=True, candidate_purchase=True
```

although you should value if putting such defaults on the model. I think it's more frequent to allow by default than to block by default.

dreispt: Why is it bad to do this through the ORM?

dreispt (approved):

purchase_partner_approval/hooks.py
HekkiMelody: > although you should value if putting such defaults on the model. I think it's more frequent to allow by default than to block by default.

I'm not against it, but I didn't want to change the logic of the original module and wanted to get this fixed asap with as little fuss as possible.

Changed to SQL, which is much better performance-wise.

francesco-ooops:
@pedrobaeza better now?

purchase_partner_approval/hooks.py
pedrobaeza: I understand, but thinking twice, any new partner created in tests won't have these flags, and the behavior in this module will interfere in the tests, so I think it's saner to put the default and don't need the hook at all.

HekkiMelody: done

HekkiMelody: Though I'm not convinced this is the correct way to go about it, I feel like the module loses its purpose if you have to go out of your way to disapprove new contacts, as opposed to having to approve them.

dreispt: This is for data init only.
When you install, existing contacts are assumed as approved.
NEW contacts from there on would be subject to approval.

HekkiMelody:
As I pointed out here https://github.com/OCA/purchase-workflow/pull/2090#discussion_r1395345478  I don't think this is the right way to go about it with this module, in fact it almost makes the module pointless. I made the change because I was asked to, and then there was no further discussion.

Could also just close the PR, especially now that we have the runboat with no modules installed.

francesco-ooops:
Yes, the fix was requested with the only goal of restoring base runboat functionalities (eg: confirming a PO) when runboat had all modules in repo automatically installed

PR #2094, opened 2023-11-15
[16.0][FIX] purchase_order_weight_volume: disable display on PO if settings display = False

cyrilmanuel:
This fix ensure that we don't show the fields and labels if the display settings is set to false

### DIFF
--- a/purchase_order_weight_volume/models/purchase_order.py
+++ b/purchase_order_weight_volume/models/purchase_order.py
@@ -50,6 +50,25 @@ class PurchaseOrder(models.Model):
         "Display Volume in Report", default=True
     )
 
+    display_order_weight_in_po = fields.Boolean(
+        "Display Order Weight in PO",
+        compute="_compute_display_order",
+    )
+    display_order_volume_in_po = fields.Boolean(
+        "Display Order Volume in PO",
+        compute="_compute_display_order",
+    )
+
+    @api.depends("company_id")
+    def _compute_display_order(self):
+        for purchase in self:
+            self.display_order_weight_in_po = (
+                purchase.company_id.display_order_weight_in_po
+            )
+            self.display_order_volume_in_po = (
+                purchase.company_id.display_order_volume_in_po
+            )
+
     @api.depends("order_line.product_uom_qty", "order_line.product_id")
     def _compute_total_physical_properties(self):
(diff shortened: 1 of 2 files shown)


purchase_order_weight_volume/models/purchase_order.py
```diff
@@ -26,15 +26,22 @@ class PurchaseOrder(models.Model):
     total_weight = fields.Float(
         compute="_compute_total_physical_properties",
         digits="Stock Weight",
+        store=True,
```
santostelmo: Already added in the dependency PR: https://github.com/OCA/purchase-workflow/blob/ce3c13c7057fce8af43d79ecd34ab97b9172ca2f/purchase_order_weight_volume/models/purchase_order.py#L29

https://github.com/OCA/purchase-workflow/pull/2060

santostelmo (approved):
LG

purchase_order_weight_volume/models/purchase_order.py
```diff
@@ -50,6 +50,25 @@ class PurchaseOrder(models.Model):
...
+    display_order_weight_in_po = fields.Boolean(
+        "Display Order Weight in PO",
+        compute="_compute_display_order",
+    )
+    display_order_volume_in_po = fields.Boolean(
+        "Display Order Volume in PO",
+        compute="_compute_display_order",
+    )
+
+    @api.depends("company_id")
```
Ricardoalso, line 62: AFAIK, @api.depends_context("company_id") should be used instead of @api.depends :thinking:

santostelmo: My understanding is that when the user.company changes `_compute_display_order` is triggered with `@api.depends_context`
But company id is a stored field. So I think is okay to use  `@api.depends`

Ricardoalso: If display_order_weight_in_po and display_order_volume_in_po are defined at the purchase level, api.depends seems indeed correct

Ricardoalso (approved):

cyrilmanuel:
hello @pedrobaeza can you merge the pr ( take a look to be sure that it's good for you )  ?

pedrobaeza:
Merging based on reviews:

/ocabot merge minor

PR #2097, opened 2023-11-21
[ADD] purchase_order_cost_update

victor-champonnois:


pedrobaeza:
IMO this is not a good approach, as your product cost is not your last purchased cost (unless your price method is FIFO, but let the system do it). You have https://github.com/OCA/purchase-workflow/tree/15.0/purchase_stock_price_unit_sync for synchronizing the purchase order line price to the move price (before validating the picking), or go to something more advanced with https://github.com/OCA/stock-logistics-workflow/tree/15.0/product_cost_price_avco_sync

victor-champonnois:
@pedrobaeza Thank you for your advices. Our clients want the sale price to be computed from the last purchase price, so average cost methods are not an option (they won't change their mind in this regard). Also they cannot use FIFO because they don't manage stocks precisely enough (especially with weighted food products). 
So the solution to set the cost as the last purchase price from the PO seemed like the "least bad" option. What do you think ?
If this makes this code non-suitable for OCA I understand, just wanted to be sure that I didn't miss something.

pedrobaeza:
If you want to do that, I recommend you to develop instead a new option in product.pricelist saying so, and on the involved code, fetch it on "real time" (a search on `purchase.order.line` for that product order by date reverse and getting the first one). It's less error-prone, and you are not using something for which the system is not designed for.

robinkeunen:
@victor-champonnois can you close this in favor of https://github.com/coopiteasy/cie-purchase/pull/1 ?

PR #2104, opened 2023-11-28
[16.0][FIX] purchase_manual_delivery, fix compute fields make both store True

ChrisOForgeFlow:
@ForgeFlow


MiquelRForgeFlow (approved):

HviorForgeFlow (approved):
it makes sense

PR #2114, opened 2023-12-08
[17.0][MIG] purchase_discount: Migration to 17.0

ramiadavid:


pedrobaeza:
AFAIK, this version includes discount natively in purchases.

ramiadavid:
@pedrobaeza Yes, but some features of the module are not the same, I have tried to leave only what is different

pedrobaeza:
Which features? I don't see any change in the readme files, and maybe the module name should be changed.

ramiadavid:
No, the readmes have not been modified yet, waiting to see what other people think about the future of this module.

While I was migrating it I saw that discounts on purchase orders are already included natively in Odoo 17 but there are functions of this module that are not.

- The default_supplierinfo_discount field and its respective functions are not in the standard version.

- Adding the discount to the supplierinfo when created from a purchase order is also not done by standard.

Basically I have left everything necessary so that the tests do not give errors.

I personally don't use those functions and I don't care whether they are there or not, but I don't know who has to decide what to do with the module.

The standard function works perfectly for me, but I need to have a solution with this module to be able to migrate the purchase_triple_discount module that depends on it and is the one I really need

pedrobaeza:
> Adding the discount to the supplierinfo when created from a purchase order is also not done by standard.

This can be considered a bug to be reported to Odoo.

gurneyalex:
@ramiadavid Please update the readme/* files

pedrobaeza:
As stated, I don't think this module should be continued as is.

gurneyalex (approved):

pedrobaeza (changes requested):
This shouldn't be merged as is, as stated in the comments.

ramiadavid:
@pedrobaeza I know, what do you propose we do, I'm waiting for instructions, I don't think it's something I have to decide.

BT-jmichaud:
Integrated in standard, not needed anymore

StefanRijnhart (changes requested):
I understand that there are small usability improvements that could have been preserved from the old purchase_discount module before Odoo implemented their own discount on supplier info and purchase order line models, but in all honesty, the current state of this PR does not make a good enough case for it, and after 14 months it's probably safe to say judging by the comments on here that there isn't enough interest in it.

PR #2115, opened 2023-12-12
[14.0][FIX] Consider procurement group for purchase request even if empty

grindtildeath:
In case a purchase request already exists with a defined procurement group, if a new procurement is being executed with an empty procurement group, it  must not be merged with the existing one having a procurement group defined.

### DIFF
--- a/purchase_request/models/stock_rule.py
+++ b/purchase_request/models/stock_rule.py
@@ -51,19 +51,18 @@ class StockRule(models.Model):
         extended.
         :return: False
         """
-        domain = (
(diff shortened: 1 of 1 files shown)


grindtildeath:
ping @mmequignon @JordiBForgeFlow @LoisRForgeFlow

LoisRForgeFlow (approved):
Makes sense to me

HviorForgeFlow (approved):

HviorForgeFlow:
@grindtildeath could you forward port it to more recent version? I will review and move it forward asap

PR #2116, opened 2023-12-14
[16.0][ADD] purchase_order_supplier_return

santostelmo:
Currently when we validate a PO with negative quantity on order line, a return to supplier picking is generate. If we validate that picking nothing happens in the quantity received.

That module allow to return products to supplier. On purchase orders validated and shipped with negative product quantity, received quantity is updated.


purchase_order_supplier_return/models/purchase_order.py
```diff
@@ -0,0 +1,18 @@
...
+from odoo import api, models
+
+
+class PurchaseOrderLine(models.Model):
+    _inherit = "purchase.order.line"
+
+    @api.depends("move_ids.state", "move_ids.product_uom_qty", "move_ids.product_uom")
+    def _compute_qty_received(self):
+        for line in self.filtered(
+            lambda line: line.qty_received_method == "stock_moves"
```
simahawk: Rule of thumb: never use filtered if you are about to loop on the same records, otherwise you loop twice.
Use it only if you have to define a var to reuse.

purchase_order_supplier_return/tests/test_purchase_order_supplier_return.py
```diff
@@ -0,0 +1,59 @@
+# Copyright 2023 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from odoo.tests.common import TransactionCase
+
+
+class TestPurchaseOrderSupplierReturn(TransactionCase):
+    def setUp(self):
```
simahawk: pls use setUpClass and disable tracking in ctx

purchase_order_supplier_return/tests/test_purchase_order_supplier_return.py
```diff
@@ -0,0 +1,59 @@
...
+                "product_id": self.product.id,
+                "product_qty": 1,
+            }
+        )
+
+    def test_negative_purchase_order_line(self):
+        """Test that the negative purchase_order_line is correctly received
+        with the negative quantity when the related picking is validated."""
+        self.purchase_order.button_confirm()
+        assert self.purchase_order_line_1.qty_received == 0
```
simahawk: use self.assert*

purchase_order_supplier_return/tests/test_purchase_order_supplier_return.py
```diff
@@ -0,0 +1,59 @@
...
+        assert self.purchase_order_line_1.qty_received == 0
+
+        # Validate the pickings related to the purchase order
+        for picking in self.purchase_order.picking_ids:
+            picking.action_set_quantities_to_reservation()
+            picking.button_validate()
+
+        self.purchase_order.order_line._compute_qty_received()
```
simahawk: if you need to call compute here there might be an issue... what is the reason?

simahawk (changes requested):

purchase_order_supplier_return/tests/test_purchase_order_supplier_return.py
santostelmo: To be honest I don't know why method ` _compute_qty_received` is not called on the unit test.
I refactored the code to update the quantity received in `move._action_done`.

purchase_order_supplier_return/models/stock_move.py
```diff
@@ -0,0 +1,22 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import models
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    def _action_done(self, cancel_backorder=False):
+        todo_moves = super()._action_done(cancel_backorder=cancel_backorder)
```
simahawk:
```suggestion
        done_moves = super()._action_done(cancel_backorder=cancel_backorder)
```
I follow the remark by JE on one of my PRs :wink:

purchase_order_supplier_return/tests/test_purchase_order_supplier_return.py
```diff
@@ -0,0 +1,63 @@
...
+        """Test that the negative purchase_order_line is correctly received
+        with the negative quantity when the related picking is validated."""
+        self.purchase_order.button_confirm()
+        self.assertEqual(self.purchase_order.order_line[0].qty_received, 0)
+
+        # Validate the pickings related to the purchase order
+        for picking in self.purchase_order.picking_ids:
+            picking.action_set_quantities_to_reservation()
+            picking.button_validate()
+        # self.purchase_order.order_line._compute_qty_received()
```
simahawk: you can drop the comment

simahawk (approved):
LG, minor remarks. Pre-approving.

cyrilmanuel (approved):

vvrossem (approved):

cyrilmanuel:
@simahawk can you merge this PR

PR #2120, opened 2023-12-17
[17.0][MIG] purchase_force_invoiced

celm1990:


cormaza (approved):
LGTM

gurneyalex (approved):

dalonsod:
@celm1990 I've notice that you've actually changed `readonly` condition for `force_invoiced` flag. Was it an error or an intended change? Otherwise, this could be restored to its original behavior IMO

cc @cormaza @gurneyalex

PR #1264, opened 2021-09-13
[13.0][ADD] product_supplierinfo_qty_multiplier: Add Multiplier Qty to vendor pricelist in products to auto-set correct qty in purchase order lines

victoralmau:
Add Multiplier Qty to vendor pricelist in products to auto-set correct qty in purchase order lines.

Please @pedrobaeza and @joao-p-marques can you review it?

@Tecnativa TT30744

PR #2122, opened 2023-12-18
[16.0][MIG] product_supplierinfo_qty_multiplier (from 12.0)

legalsylvain:
### Description

This module allows to define multiplier quantities at the supplier level in the products to apply it in the purchase order lines.

### Configure

* Go to Purchase > Products > Products.
* Select a product.
* Go to the 'Purchase' tab and edit a vendor pricelist line to set 'Multiplier Qty' to 6.0

(If you can't find that column, you can activate it by clicking the three dots at the corner of the table and enabling it).

product_product_form

### Usage

* Create a new purchase order.
* Assign it to the partner where the multiplier quantity was previously set.
* Add a product line with the previously edited product.
* Quantity is set to 6.0.
* If you set 9.0 in the quantity field, the quantity will be rounded up to 12.0

purchase_order_form

CC : @quentinDupont


legalsylvain:
Migrated module introduced here in V13 https://github.com/OCA/purchase-workflow/pull/1264

CC original reviewers / contributors : @victoralmau,  @pedrobaeza, @joao-p-marques

product_supplierinfo_qty_multiplier/demo/product_product.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8" ?>
```
pedrobaeza: The file should be named by convention `product_product_demo.xml`. And the same of the others.

product_supplierinfo_qty_multiplier/models/purchase_order_line.py
```diff
@@ -4,47 +4,41 @@
...
-        # Change context in a compatible way with onchange usage
-        # See : https://github.com/odoo/odoo/issues/7472#issuecomment-119503916
-        self.env.context = frozendict(self.env.context, multiplier_qty_message=False)
-        res = super().onchange_product_id()
-        self.env.context = frozendict(self.env.context, multiplier_qty_message=True)
-        return res
-
     @api.onchange("product_qty")
-    def _onchange_quantity(self):
+    def onchange_product_qty_multiplier_qty(self):
```
pedrobaeza: Please follow naming conventions:
```suggestion
    def _onchange_product_qty_multiplier_qty(self):
```

legalsylvain: Is it a convention ? 
I see both results in odoo. (Indeed more private function than public function)

- def onchange : 50 results in odoo/odoo/addons
- def _onchange : over 300 results in odoo/odoo/addons

nevermind, changed.

product_supplierinfo_qty_multiplier/demo/product_product.xml
legalsylvain: done. Thanks.

product_supplierinfo_qty_multiplier/models/purchase_order_line.py
pedrobaeza: Odoo doesn't serve for guidelines. Follow OCA guidelines:

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#37models

pedrobaeza: As corollary, I wanted to add some notes for this case where the onchange is not declared on the module that adds the field, and my take was to put `_onchange_<field_name>_<module_name>` as method name, for avoiding unintentional method overrides on 2 non-dependent modules, but I don't remember where this proposal goes (but definitively not merged).

quentinDupont (approved):
Tested on Runboat, works like charm :rose: 
And code reviewed :heavy_check_mark:

product_supplierinfo_qty_multiplier/models/purchase_order_line.py
legalsylvain: Hi @pedrobaeza. I applied your suggestion. (_onchange_product_qty_product_supplierinfo_qty_multiplier)
Is it OK for you ? thanks !

PR #2124, opened 2023-12-18
[13.0][ADD]purchase_order_line_seller_used

GuillemCForgeFlow:
New module to set the Vendor Pricelist used for the Purchase Order Line every time the Product Quantity gets updated.

cc @ForgeFlow

### DIFF
--- a/purchase_order_line_seller_used/models/purchase_order_line.py
+++ b/purchase_order_line_seller_used/models/purchase_order_line.py
@@ -1,0 +1,66 @@
+# Copyright 2023 ForgeFlow S.L. (https://www.forgeflow.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from odoo import api, fields, models
+
+SELLER_USED_TRIGGER_FIELDS = ["product_qty", "product_uom"]
+
+
+class PurchaseOrderLine(models.Model):
+    _inherit = "purchase.order.line"
+
+    seller_used_id = fields.Many2one(
+        comodel_name="product.supplierinfo",
+        string="Vendor Pricelist Used",
+        readonly=True,
+        copy=False,
+    )
+
+    def _get_seller_from_line(self, vals=False):
+        self.ensure_one()
+        product_qty = self.product_qty
+        product_uom = self.product_uom
+        if vals:
+            if vals.get("product_qty", False):
+                product_qty = vals["product_qty"]
+            if vals.get("product_uom", False):
+                product_uom_id = vals["product_uom"]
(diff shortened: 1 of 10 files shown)


purchase_order_line_seller_used/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,6 @@
+This module adds the Vendor Pricelist used in the Purchase Order Line every
+time it gets updated in the Odoo standard flow.
+
+In Odoo, every time the `Product Quantity` is changed, the new Vendor Pricelist
+is also searched. This Vendor Pricelist is then assigned to the Purchase Order
+Line to be then used as needed.
```
legalsylvain, line 6: Naive question : In which module / use case do you use this new field ?

purchase_order_line_seller_used/models/purchase_order_line.py
```diff
@@ -0,0 +1,31 @@
...
+        params = {"order_id": self.order_id}
+        seller = self.product_id._select_seller(
+            partner_id=self.partner_id,
+            quantity=self.product_qty,
+            date=self.order_id.date_order and self.order_id.date_order.date(),
+            uom_id=self.product_uom,
+            params=params,
+        )
+        if seller:
+            self.seller_used_id = seller
```
legalsylvain: If seller is not found, seller_used_id should be set to False, dont you think ?

purchase_order_line_seller_used/readme/DESCRIPTION.rst
GuillemCForgeFlow: My need is for a custom module to restrict the PO confirmation if there is not a Vendor Pricelist used for any of the lines. But I have set it up here in this module as it can be used for other use cases, such as having the reverse relation, for example.
Perhaps, my use case could also be implemented as an OCA module. What do you think?

purchase_order_line_seller_used/models/purchase_order_line.py
GuillemCForgeFlow: I think it makes more sense, I will change. Thank you 😄

purchase_order_line_seller_used/readme/DESCRIPTION.rst
legalsylvain: >  Perhaps, my use case could also be implemented as an OCA module. What do you think?

Not a clear point of view. I personaly don't use this second feature. In fact, I don't understand why restrict confirmation in such situation.
But well, maybe it can be interesting for other OCA users.

PR #2103, opened 2023-11-24
[15.0][FIX] purchase_stock_secondary_unit: Tests. Secondary_uom_qty is like as qty_done in sml's

sergio-teruel:
cc @Tecnativa 

ping @carlosdauden @CarlosRoca13 

Depends on:

- [x] https://github.com/OCA/stock-logistics-warehouse/pull/1606

PR #2125, opened 2023-12-18
[15.0][FIX] purchase_stock_secondary_unit: fix tests

DavidJForgeFlow:
When trying to check if the stock.move.line has the quantities to match the new demand, these moves hasn't been created yet. If we want to check if the stock.move.line gets the values correctly from the stock.move we should firstly assign the quantity.

Supersedes #2103

@sergio-teruel @ForgeFlow


sergio-teruel:
Sorry @DavidJForgeFlow this PR is not correct.. the default value for secondary_qty in sml would be the same that qty_done...

pedrobaeza:
Coming back to #2103, as the CI problem with it was the non merged dependency.

PR #2127, opened 2023-12-18
[15.0][FIX] purchase_last_price_info: Adapt tests according to company currency

pedrobaeza:
Forward-port of #1275

Depending on the installed modules, the company currency may be USD or EUR, affecting the result of these tests. With this PR, the results are adjusted for both cases.

@Tecnativa

PR #2129, opened 2023-12-19
[16.0][FIX] purchase_last_price_info: Adapt tests according to company currency

pedrobaeza:
Forward-port of #1275 and #2127 

Depending on the installed modules, the company currency may be USD or EUR, affecting the result of these tests. With this PR, the results are adjusted for both cases.

@Tecnativa


pedrobaeza:
Fast-tracking for completing the fw-port chain:

/ocabot merge patch

PR #2133, opened 2023-12-26
[16.0][ADD] purchase_order_line_receipt_status

imlopes:


purchase_order_line_receipt_status/models/purchase_order.py
```diff
@@ -0,0 +1,22 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, models
+
+
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+
+    @api.depends("order_line")
+    def _compute_receipt_status(self):
```
ivantodorovich: There's no need to completely overwrite [the core method](https://github.com/odoo/odoo/blob/94b5ae780432569e0a6e8367cb46b4a099823737/addons/purchase_stock/models/purchase.py#L65), is there?

IMO we can keep the order-level method as-is, and simply focus on adding the order line one.
The result, even if the implementation is different, should be the same.. right?

ivantodorovich (commented):
I'd be cool to add some unit tests

purchase_order_line_receipt_status/models/purchase_order.py
ivantodorovich: @imlopes ?

imlopes: Hmmm ! I think you're right.
I will remove this block.

ivantodorovich: Thanks!

ivantodorovich (approved):

Camille0907 (approved):
LGTM, thank you

PR #2113, opened 2023-12-07
[13.0][FIX] purchase_order_type: sequence depending on date ranges

dalonsod:
Custom sequence for an order type wasn't taken in account date ranges during order creation, this fix it.

PR #2134, opened 2023-12-27
[14.0][FIX] purchase_order_type: sequence depending on date ranges

dalonsod:
Custom sequence for an order type wasn't take in account date ranges during order creation, this fix it.

Fw-port of #2113


pedrobaeza (approved):
Pretty straight-forward and already merged in 13.0, so fast-tracking it:

/ocabot merge patch

PR #2137, opened 2023-12-28
[16.0][MIG] purchase_stock_price_unit_sync: Migration to 16.0

AungKoKoLin1997:
Standard Migration.

@qrtl


AaronHForgeFlow:
@AungKoKoLin1997  Thank you for you explanation in here: https://github.com/OCA/stock-logistics-workflow/pull/1465

I created a product and set it as AVG cost with automated valuation. Then I received one @ 10USD, then I receive another @ 10 USD. Then change the price on the first PO to be 20 USD. The layer unit price was updated but not the total value:

svg_cost_sync

The cost in the product card was still 10 USD, shouldn't it be updated to 15 USD?

AaronHForgeFlow:
@AungKoKoLin1997 Just did the same in v15 and the behavior is the same, I see that the module product_cost_price_avco_sync is not dependency here, so please disregard my previous comment.

victoralmau (approved):
Code review OK.

Ping @carlosdauden @sergio-teruel 

TT47857

pedrobaeza (approved):
/ocabot merge nobump

PR #2147, opened 2024-01-11
[15.0][MIG] purchase_order_archive: Migration to 15.0.

niboo-sam:


rvalyi (changes requested):
great you got the test working. However, the pre-commit tests aren't passing and more importantly, you should enforce the git process described here to preserve the module git history when doing a migration: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-15.0

Could you please update your PR following this process?

PR #2153, opened 2024-01-22
[15.0][IMP] purchase_security: Restrict partners according to purchase groups

victoralmau:
Restrict partners according to purchase groups

Similar to sales_team_security, partners are restricted according to purchase permissions:
- **User (own orders)**: If you are a follower of the partner or there is no user or you are the partner's user.
- **User (team orders)**: If you are a follower of the partner or there is no user or you are a user of your purchasing team.
- **Administrator**: Without the restrictions of the other groups.

@Tecnativa TT47237


victoralmau:
Ping @pilarvargas-tecnativa and @pedrobaeza

purchase_security/models/res_partner.py
```diff
@@ -0,0 +1,51 @@
...
+    def _remove_key_followers(self, partner):
+        for record in self.mapped("commercial_partner_id"):
+            # Look for delivery and invoice addresses
+            childrens = record.child_ids.filtered(
+                lambda x: x.type in {"invoice", "delivery"}
+            )
+            (childrens + record).message_unsubscribe(partner_ids=partner.ids)
+
+    def _add_followers_from_salesmans(self):
+        """Sync followers in commercial partner + delivery/invoice contacts."""
```
pedrobaeza: There's no delivery/invoice concept in purchases.

purchase_security/models/res_users.py
```diff
@@ -0,0 +1,31 @@
...
+        comodel_name="purchase.team",
+        relation="purchase_team_res_users_rel",
+        column1="res_users_id",
+        column2="purchase_team_id",
+        string="Purchases Teams",
+        check_company=True,
+        copy=False,
+        readonly=True,
+    )
+    purchase_team_id = fields.Many2one(
```
pedrobaeza: Why do you need this field instead using `purchase_team_ids[:1]`?

purchase_security/models/res_partner.py
victoralmau: You are right, it does not exist, but it is possible to select any partner in a purchase (Company > Delivery partner), that's why the followers are Sync in all the "children".

purchase_security/models/res_users.py
victoralmau: This field is required to use it in https://github.com/OCA/purchase-workflow/pull/2153/files#diff-dfb94a4f11b575139dd2bfa579f9a012280cdc3ad65e43aa10e222113209ae2bR42 do you think it could be done in another way?

pedrobaeza: My suggestion about using `purchase_team_ids[:1]`. Anyway, I don't quite understand the rule adding all the users of the first team of the user.

purchase_security/models/res_partner.py
pedrobaeza: No children syncing should be done in this case.

purchase_security/models/res_users.py
victoralmau: The "_User (team orders)_" permission added by this module will allow to see the partners that have set the user's purchase team. 

There is no `purchase_team_id` field in the partners to be able to filter by that field, so the rule adds the domain to be able to see the partners linked to users of the same purchase team of the user.

purchase_security/models/purchase_team.py
```diff
@@ -12,4 +12,10 @@ class PurchaseTeam(models.Model):
 
     name = fields.Char(required=True)
     sequence = fields.Integer(default=10)
-    user_ids = fields.Many2many("res.users", string="Purchase Users")
+    user_ids = fields.Many2many(
+        comodel_name="res.users",
+        relation="purchase_team_res_users_rel",
```
pedrobaeza, line 17: Is this changing the name of the relation?

purchase_security/models/res_users.py
pedrobaeza: I'm still talking about the user, not the partner. Just use `user.purchase_team_ids[:1]`

purchase_security/models/res_partner.py
```diff
@@ -0,0 +1,40 @@
...
+    _inherit = "res.partner"
+
+    # add indexes for better performance on record rules
+    user_id = fields.Many2one(index=True)
+
+    def _remove_key_followers(self, partner):
+        for record in self.mapped("commercial_partner_id"):
+            record.message_unsubscribe(partner_ids=partner.ids)
+
+    def _add_followers_from_salesmans(self):
```
pedrobaeza: I don't think you should propagate followers, unless you add a special new field in `res.partner` being "Purchase representative", distinct from the salesman. If not, you are mixing boundaries using the same field for both purchase/sales.

pedrobaeza (changes requested):

purchase_security/models/purchase_team.py
victoralmau: No, table name and columns are the same.

victoralmau:
Changes done.

pedrobaeza:
Thinking twice, we should add the "Purchase representative" field in `res.partner` and base security on it, as if not, it will be very limiting and conflicting with sales. This way, as the field is going to be empty by default, we assure compatibility with previous version.

victoralmau:
> Thinking twice, we should add the "Purchase representative" field in `res.partner` and base security on it, as if not, it will be very limiting and conflicting with sales. This way, as the field is going to be empty by default, we assure compatibility with previous version.

Changes done.

Some other issues:
- Should the `purchase_user_id` field in `res.partner` be visible to everyone or only to users with purchase permissions?
- The onchange of `partner_id` in purchase orders should be changed to define the "correct" `user_id` value (from partner or user).
- Maybe could also add `purchase_team_id` field in `res.partner` and improve the rules with that field. This new field would cause the "correct" `team_id` value (from partner or user) to be defined.

pedrobaeza:
OK to all. Thanks for taking that into account.

victoralmau:
All changes done.

purchase_security/models/purchase_order.py
```diff
@@ -8,28 +9,36 @@
 class PurchaseOrder(models.Model):
     _inherit = "purchase.order"
 
+    @api.model
+    def _get_default_team(self):
+        """Get team from default, from user or first team."""
```
pedrobaeza: Please don't change the computed writable. Why do you need this? You can model it inside the compute.

purchase_security/views/res_partner_views.xml
```diff
@@ -0,0 +1,50 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="view_res_partner_filter" model="ir.ui.view">
+        <field name="name">res.partner.select</field>
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="base.view_res_partner_filter" />
+        <field name="groups_id" eval="[(4, ref('group_purchase_own_orders'))]" />
+        <field name="arch" type="xml">
+            <field name="user_id" position="after">
```
pedrobaeza, line 9: This should go better on the purchase part of the page.

purchase_security/models/__init__.py
```diff
@@ -1,2 +1,7 @@
-from . import purchase_order
+from . import ir_rule
 from . import purchase_team
+from . import res_partner
+from . import res_users
+
+# Set at the end to avoid access error to the res.users field.
+from . import purchase_order
```
pedrobaeza: What is the error? Reproducible here on runboat? Don't let anyway any empty line

purchase_security/views/res_partner_views.xml
victoralmau: In the form view everything is added in the "Purchase" part, but the `purchase_user_id` field is also added in the tree and search views.

purchase_security/models/purchase_order.py
victoralmau: Ok, compute restored (no need to add anything extra finally).

pedrobaeza (approved):

pilarvargas-tecnativa (approved):

PR #2157, opened 2024-01-24
[16.0][IMP] purchase_partner_selectable_option

AungKoKoLin1997:
This PR improves the process of choosing the default value for purchase_selectable by utilizing a system parameter instead of hardcoding it.

@qrtl

### DIFF
--- a/purchase_partner_selectable_option/models/res_partner.py
+++ b/purchase_partner_selectable_option/models/res_partner.py
@@ -1,12 +1,22 @@
 # Copyright 2021 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
-from odoo import fields, models
+from odoo import api, fields, models
 
 
 class Partner(models.Model):
     _inherit = "res.partner"
 
     purchase_selectable = fields.Boolean(
-        string="Selectable in purchase orders", default=True
+        string="Selectable in purchase orders",
+        default=lambda self: self._default_purchase_selectable(),
     )
+
+    @api.model
+    def _default_purchase_selectable(self):
+        param = (
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param("purchase_partner_selectable_option.default_purchase_selectable")
+        )
+        return False if param == "0" else True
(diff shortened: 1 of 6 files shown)


purchase_partner_selectable_option/models/res_partner.py
```diff
@@ -1,12 +1,24 @@
...
+    @api.model
+    def _default_purchase_selectable(self):
+        param = (
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param("purchase_partner_selectable_option.default_purchase_selectable")
+        )
+        if param == "0":
+            return False
+        return True
```
victoralmau:
```suggestion
        return True if param == "1" else False
```

victoralmau (changes requested):

purchase_partner_selectable_option/readme/DESCRIPTION.rst
```diff
@@ -1 +1,3 @@
 This module limits the partners to be selected in the purchases orders according to the "Selectable in orders" field.
+By default, this field is set to True, allowing all partners to be selectable. The default setting can be modified
+by adjusting the system parameter with the key 'purchase_partner_selectable_option.default_purchase_selectable'.
```
yostashiro: I suggest we put this in CONFIGURE.rst.

yostashiro: I guess it's non-blocking as there seems to be no comment on this in the PR for the sales side https://github.com/OCA/sale-workflow/pull/2922.

yostashiro: On second thought, we should clearly mention that the param should be `"0"` to make the default false.

yostashiro (approved):
Looks good. Thanks!

victoralmau (approved):

rousseldenis (approved):
Code review. Fair

AungKoKoLin1997:
@OCA-purchase-workflow-maintainters
Can we get this PR merged?

victoralmau (approved):

PR #2161, opened 2024-01-29
[15.0][IMP] `purchase_tag`: add parent tags

SilvioC2C:


ivantodorovich (approved):
I'm not against the changes around the `color` field, though.. are they intended?

SilvioC2C:
@ivantodorovich Yep, they are:
1. I changed the `default` attribute in the field declaration to make it possible to override the default method if necessary
2. regarding the changes to the field declaration in the form view, I added the widget to display the color itself instead of the `Integer` value

gurneyalex (approved):

yankinmax (approved):

PR #2166, opened 2024-01-31
[13.0] purchase_discount: use force company to select the seller

adrienpeiffer:
Currently, the selection of the seller don't take to correct company

https://github.com/odoo/odoo/blob/13.0/addons/purchase_stock/models/stock_rule.py#L216


pedrobaeza (approved):

rousseldenis (approved):
A little test would be great

PR #2167, opened 2024-02-06
[16.0][ADD] purchase_stock_move_line_cancel

rousseldenis:


purchase_order_line_stock_move_cancel/models/purchase_order_line.py
```diff
@@ -0,0 +1,68 @@
...
+            body += "<ul>"
+            body += "<li>".join([line.display_name + "</li>" for line in lines])
+            body += "</ul>"
+            purchase.message_post(body=body)
+
+    def _cancel_moves(self) -> bool:
+        """
+        Allows to cancel stock moves from purchase order lines
+
+        :return: _description_
```
lmignon, line 41:
```suggestion

```

luisDIXMIT (commented):
I tested the functionality in Runboat. It looks good, but when I tried to cancel the stock move and then canceled the purchase order and confirmed it again, another stock picking was generated with another stock move for the product in the purchase order line. However, in this case, I couldn't cancel the new stock move again. Should this be the expected behavior?

purchase_order_line_stock_move_cancel/models/purchase_order_line.py
```diff
@@ -0,0 +1,68 @@
...
+    )
+
+    def _compute_can_cancel_moves(self):
+        """
+        All moves from a purchase order line should not be cancelled or done
+        """
+        for line in self:
+            line.can_cancel_moves = bool(
+                line.move_ids
+                and all(move.state not in ("done", "cancel") for move in line.move_ids)
```
etobella, line 23: @luisDIXMIT This might happen because we should use any, not all

etobella: @rousseldenis Can you review?

rousseldenis: Yes I will try tomorrow. On other things right now.

PR #2168, opened 2024-02-08
[16.0][IMP] purchase_packaging_level_qty store total_transport_qty

santostelmo:
We have a requirement where POs needs to be validated or not based on the total transport qty per packaging level.
=> For that we need to store the filed  `total_transport_qty`.

### DIFF
--- a/purchase_packaging_level_qty/models/purchase_order.py
+++ b/purchase_packaging_level_qty/models/purchase_order.py
@@ -4,12 +4,18 @@ from odoo import api, fields, models
 class PurchaseOrder(models.Model):
     _inherit = "purchase.order"
 
-    total_transport_qty = fields.Char(
+    total_transport_qty = fields.Float(
         compute="_compute_total_transport_packaging_qty",
+        store=True,
     )
 
     transport_packaging_level_id = fields.Many2one(
-        "product.packaging.level", compute="_compute_purchase_transport_packaging_level"
+        "product.packaging.level",
+        compute="_compute_purchase_transport_packaging_level",
+    )
+
+    transport_packaging_level_total_qty = fields.Char(
+        compute="_compute_transport_packaging_level_total_qty",
     )
 
     def _compute_purchase_transport_packaging_level(self):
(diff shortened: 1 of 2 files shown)


purchase_packaging_level_qty/models/purchase_order.py
```diff
@@ -26,9 +32,20 @@ def _compute_purchase_transport_packaging_level(self):
     def _compute_total_transport_packaging_qty(self):
```
Ricardoalso, line 32: @api.depends as it is now stored ?

Ricardoalso (approved):

cyrilmanuel (approved):

purchase_packaging_level_qty/models/purchase_order.py
```diff
@@ -26,9 +32,20 @@ def _compute_purchase_transport_packaging_level(self):
...
         for order in self:
             if order.transport_packaging_level_id:
-                total_transport_qty = sum(order.mapped("order_line.transport_qty"))
-                order.total_transport_qty = "{} {}".format(
-                    total_transport_qty,
+                order.total_transport_qty = sum(
+                    order.mapped("order_line.transport_qty")
```
simahawk, line 36: Could we take the chance to improve this?
We should use `read_group` + `sum` operator instead.

purchase_packaging_level_qty/models/purchase_order.py
```diff
@@ -26,9 +32,20 @@ def _compute_purchase_transport_packaging_level(self):
...
+                )
+            else:
+                order.total_transport_qty = False
+
+    @api.depends(
+        "total_transport_qty",
```
simahawk:
```suggestion
        "total_transport_qty", "transport_packaging_level_id",
```

simahawk (changes requested):

purchase_packaging_level_qty/models/purchase_order.py
santostelmo: I did not added dependency on  `transport_packaging_level_id` because `total_transport_qty ` is already dependent on it.
But add it here as well to be more explicit.

purchase_packaging_level_qty/models/purchase_order.py
santostelmo: It is more performant ?
Updating code...

simahawk: >  We should use read_group + sum operator instead.

Of course. For 2 reasons: just one read + the sum is delegated to the db.

cyrilmanuel:
Hi @santostelmo  little ping for pre-commit :)

purchase_packaging_level_qty/models/purchase_order.py
santostelmo: `transport_qty` is not a stored field that can be used to sql sum

santostelmo:
For the tests falling I don't see the relation with this improvement.

HviorForgeFlow (approved):

PR #2184, opened 2024-02-26
[IMP] purch_stock_pack: add test for multi packagings

bealdav:
Scenario that I want implement.

When 2 kind of packaging are used in sale line (mimic by procurement group here), purchase should be set with the same packagings

### DIFF
--- a/purchase_stock_packaging/models/__init__.py
+++ b/purchase_stock_packaging/models/__init__.py
@@ -1,1 +1,2 @@
 from . import purchase_order_line
+from . import stock_rule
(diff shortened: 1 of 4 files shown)


rousseldenis:
@bealdav IMHO, 'product_packaging_id' should maybe added in _get_procurements_to_merge_groupby() method in stock.rule model

rousseldenis:
And this will only work for MTO rule use.

bealdav:
@rousseldenis  You're right , that's the purpose of the second commit which 'll come. Thanks.

PR #2193, opened 2024-02-29
[16.0][MIG] purchase_order_qty_change_no_recompute: Migration to 16.0

victoralmau:
Superseed https://github.com/OCA/purchase-workflow/pull/1886

Changes done:
- [x] Make it work correctly taking into account that the `price_unit`, `name` and `date_planned` fields use the same compute method (and therefore the same depends).

@Tecnativa TT46595


victoralmau:
After changing the approach of `purchase_blanket_order` and `purchase_only_by_packaging` to make them all compatible, I am now running into `purchase_triple_discount` issues.

**Problem**: Previously (v15) there was a context that allowed (in the tests) to omit onchange the relative to `purchase_order_qty_change_no_recompute`, so it "seemed" to have the standard operation and therefore the tests worked correctly.

`Now`: Using monkeypatching of `get_depends()` we can not perform test.

How could we add a context (or similar) in the `purchase_order_qty_change_no_recompute` tests so that in that case the test DOES change the behavior but in the rest it doesn't? (I thought about creating an ir.config.parameter with a value to re-set it when running the module tests, but I think that wouldn't work either). If we got to do it (in one way or another) the rest of changes in other modules will NOT be necessary.

Ping @pedrobaeza

pedrobaeza:
In this case, there's no other solution except put it as rebel module. I have push an update branch with the following changes:

- No dependency on `purchase_discount`. This was added on the previous migration work and it's totally incorrect, and at first sight, unneeded. If something needs to be done, it should be done on the module itself or in a new glue module.
- Improve documentation and test.
- Proper fields in the monkeypatching.
- Remove other modules commits, as it's not needed.

pedrobaeza (approved):

purchase_order_qty_change_no_recompute/monkeypatching.py
```diff
@@ -0,0 +1,30 @@
...
+        if "product_uom" in depends:
+            depends.remove("product_uom")
+        # We need to add product_id to the depends to get the initial values
+        if "product_id" not in depends:
+            depends.append("product_id")
+    return depends, depends_context
+
+
+# Monkey-patching of the method
+Field.get_depends = get_depends
```
chienandalu, line 30: Did you study the possibility to hook into the model [`_setup_fields`](https://github.com/OCA/OCB/blob/def5fa57fc0ada6805e02a37fae960ae792708dc/odoo/models.py#L2812) method and use a similar estrategy to the one in [`_pop_field`](https://github.com/OCA/OCB/blob/def5fa57fc0ada6805e02a37fae960ae792708dc/odoo/models.py#L567)?

pedrobaeza: I already did. It doesn't serve, as the depends is something dynamic that can be expanded through other modules.

chienandalu: I see, the main issue here is that this technique can be use only once? Am I wrong?

pedrobaeza: Yes I'm afraid

pedrobaeza: Although maybe subsequent monkey patches will work over the previous replaced method...

chienandalu (approved):

PR #2196, opened 2024-03-05
[15.0][FIX] purchase_requisition_grouped_by_procurement: Do not grouped from order point rules

sergio-teruel:
cc @Tecnativa TT47617

ping @victoralmau @carlosdauden


sergio-teruel:
Ready for review @victoralmau 

Removed code not needed. The group_id field in not needed because already exists "procurement_group_id" field.

carlosdauden (approved):

victoralmau (approved):
I think commit message and PR title should be changed to really indicate the reason for this change: In `purchase_requisition_stock` v15 the `procurement_group_id` field is added in `purchase.requisition` and `prepare_tender_values()` takes it into account it https://github.com/odoo/odoo/blob/15.0/addons/purchase_requisition_stock/models/purchase_requisition.py#L30.

marcelsavegnago (approved):

PR #2201, opened 2024-03-12
[17.0][MIG] purchase_reception_status

Rad0van:
Standard migration, only view changes (states attribute) were required.


Rad0van:
Now I have found that there is a field _purchase_receipt_status_  in _purchase_stock_ module that does something very similar but based on associated picking ids. It exists in 17.0 as well as in 16.0 Reading Relase Notes for Odoo 16 it has been added there. So migration of this module to 16.0 was probably an omission. So this PR is useless as well probably.

yostashiro:
> So migration of this module to 16.0 was probably an omission. So this PR is useless as well probably.

@Rad0van It should still be good to keep the module in OCA. Please read https://github.com/OCA/purchase-workflow/pull/1831#issuecomment-1548924543, which I suppose should still be valid in 17.0.

And the sales counterpart has gone through a refactoring to use the Odoo standard field in the 16.0 migration here https://github.com/OCA/sale-workflow/pull/2770.  Would you consider adopting the same approach in this migration?

purchase_reception_status/models/purchase_order.py
```diff
@@ -0,0 +1,57 @@
...
+            ("no", "Nothing Received"),
+            ("partial", "Partially Received"),
+            ("received", "Fully Received"),
+        ],
+        compute="_compute_reception_status",
+        store=True,
+    )
+    force_received = fields.Boolean(
+        readonly=True,
+        states={"done": [("readonly", False)]},
```
CasVissers-360ERP: Since Odoo 17, property purchase.order.force_received.states is no longer supported.

CasVissers-360ERP: I think it can be removed and XML can updated with readonly="state != 'done'"

StefanRijnhart: @Rad0van Can you please address this comment?

Rad0van: Done

astirpe (approved):

StefanRijnhart (commented):
Can you at least mention in the readme your own observation that Odoo now provides a `receipt_status` in its purchase_stock module? And please add @yostashiro's suggestion to refactor according to https://github.com/OCA/sale-workflow/pull/2770 in a `ROADMAP.rst`.

purchase_reception_status/views/purchase_order.xml
```diff
@@ -0,0 +1,84 @@
...
+                <field
+                    name="force_received"
+                    groups="purchase.group_purchase_manager"
+                    invisible="state not in ['purchase', 'done']"
+                    readonly="state != 'done'"
+                />
+            </field>
+        </field>
+    </record>
+    <record id="purchase_order_tree" model="ir.ui.view">
```
GuillemCForgeFlow, line 27: Perhaps a good improvement would be to have the field also enabled on the default RFQ tree view: `purchase.purchase_order_kpis_tree`. WDYT? 🙂

GuillemCForgeFlow (commented):
IMHO, I agree on the discussion above, also referencing the explanation for the reason we should keep this one in the OCA: https://github.com/OCA/purchase-workflow/pull/1831#issuecomment-1548924543.

I would just consider adding the improvement, apart from that LGTM 👍🏿

purchase_reception_status/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2020 Akretion France (http://www.akretion.com/)
+# @author: Alexis de Lattre <alexis.delattre@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Purchase Reception Status",
+    "version": "17.0.1.0.1",
```
GuillemCForgeFlow: Also, update the manifest version 👀

purchase_reception_status/views/purchase_order.xml
GuillemCForgeFlow: Disregard this comment, I believe that it should be kept as is, the `reception_status` is only shown when in the `purchase` state already.

GuillemCForgeFlow:
@Rad0van do you plan to continue with the changes so it can be merged? thank you in advance 🙂

purchase_reception_status/__manifest__.py
```diff
@@ -4,7 +4,7 @@
 
 {
     "name": "Purchase Reception Status",
-    "version": "16.0.1.0.1",
+    "version": "17.0.1.0.1",
```
twalter-c2c:
```suggestion
    "version": "17.0.1.0.0",
```

twalter-c2c (approved):
LGTM

sebalix:
@StefanRijnhart @GuillemCForgeFlow is the update of ROADMAP file a blocking point on your side to merge this one?

GuillemCForgeFlow:
> @StefanRijnhart @GuillemCForgeFlow is the update of ROADMAP file a blocking point on your side to merge this one?

IMHO, I would add it before merging if @Rad0van can do it.

LoisRForgeFlow:
> > So migration of this module to 16.0 was probably an omission. So this PR is useless as well probably.
> 
> @Rad0van It should still be good to keep the module in OCA. Please read [#1831 (comment)](https://github.com/OCA/purchase-workflow/pull/1831#issuecomment-1548924543), which I suppose should still be valid in 17.0.
> 
> And the sales counterpart has gone through a refactoring to use the Odoo standard field in the 16.0 migration here [OCA/sale-workflow#2770](https://github.com/OCA/sale-workflow/pull/2770). Would you consider adopting the same approach in this migration?

@Rad0van I do not see that anyone replied to this suggestion by @yostashiro. I think it is the time to do it, otherwise the problem will be dragged though versions... Unless anyone provide a good reason to not doing it I think I cannot agree with this migration.

LoisRForgeFlow (changes requested):
https://github.com/OCA/purchase-workflow/pull/2201#issuecomment-2897031126

Rad0van:
> [#2201 (comment)](https://github.com/OCA/purchase-workflow/pull/2201#issuecomment-2897031126)

Do you mean following the loogic of https://github.com/OCA/sale-workflow/pull/2770 and splitting the modules or you mean updating the README?

JoanSForgeFlow:
Hi @Rad0van,

Our proposal is follow the same pattern as in OCA’s sale_delivery_ state PR https://github.com/OCA/sale-workflow/pull/2770 Here’s our proposed split:

1. **purchase_reception_status**  
   - Base module, depends **only** on purchase (can be installed alone)  
   - Adds in purchase.order the field reception_status but with the compute='_compute_oca_reception_status' in order to:
   - Use OCA logic (compare `qty_received` vs `product_qty` line by line)  
   - Add a `force_received` flag to force the status to “full”
   - Add `reception_status` into views. 

2. **purchase_stock_reception_status**  
   - “Glue” module, depends on both purchase_reception_status **and** purchase_stock.
   - When purchase_stock is installed, override its native compute to activate our `_compute_oca_reception_status` 
   - Hide the places where Odoo displays it (if we consider them redundant or unnecessary).

This way you can install purchase_reception_status by itself. If you also install purchase_stock, add purchase_stock_reception_stat¡us to hook into compute.  

What do you think?

HviorForgeFlow:
Already merged

PR #2211, opened 2024-03-22
[FIX] CI tests: Add specific caompany to purchase order

HviorForgeFlow:
Fixing CI


MiquelRForgeFlow (approved):

LoisRForgeFlow (approved):
For reference, the issue was caused by the introduction of this change upstream https://github.com/odoo/odoo/pull/157861

PR #2213, opened 2024-03-25
[17.0][IMP] purchase_request: split field 'Quantity in RFQ or PO' into purchased_qty and rfq_qty fields

augusto-weiss:
Features in this PR:

- Split field purchased_qty (Quantity in RFQ or PO) into purchased_qty and rfq_qty. And add it to views.

- Improve state of purchase request line to have more specific info. 
- Add to the compute method of the purchase request line state the dependence on quantities (i.e.: A purchase request line is not done if the purchase order quantities are less than de purchase request line quantities)

### DIFF
--- a/purchase_request/models/purchase_request_line.py
+++ b/purchase_request/models/purchase_request_line.py
@@ -5,12 +5,14 @@ from odoo import _, api, fields, models
 from odoo.exceptions import UserError
 
 _STATES = [
-    ("draft", "Draft"),
-    ("to_approve", "To be approved"),
+    ("draft", "RFQ"),
+    ("sent", "RFQ Sent"),
+    ("to_approve", "To Approve"),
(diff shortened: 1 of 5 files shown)


augusto-weiss:
@sonhd91 @gurneyalex Hello guys! Would you like to review this IMP pr ? I hope some feedback

florenciafrigieri2 (approved):
Functional test, is working ok.

florenciafrigieri2:
Hi @pedrobaeza. Can you please check this PR? And also the 2237 which affects the same module.

Thanks in advance!

pedrobaeza (changes requested):
You are modifying things that shouldn't be done in stable. Better that you add the split feature in a new module. The other option is to add migration scripts.

PR #2214, opened 2024-03-26
[FIX]purchase_discount:Bad behaviour with currency

jcadhoc:
I made this PR because when purchase order has a different currency than company, when you edit the purchase order line price (with order in state "purchase"), in picking do not refresh correctly price_unit, because the older behavoiur didn`t contemplate currency.

Besides of this, i think the super call of method _get_stock_move_price_unit don`t  have sense any more, in fact, it has this comment:
HACK: This is needed while https://github.com/odoo/odoo/pull/29983 is not merged.
That PR is merged since 5 years.

Thank you!

### DIFF
--- a/purchase_discount/__manifest__.py
+++ b/purchase_discount/__manifest__.py
@@ -24,4 +24,3 @@
     "installable": True,
     "images": ["images/purchase_discount.png"],
 }
-
--- a/purchase_discount/__manifest__.py
+++ b/purchase_discount/__manifest__.py
@@ -24,3 +24,4 @@
     "installable": True,
     "images": ["images/purchase_discount.png"],
 }
+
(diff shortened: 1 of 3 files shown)


pedrobaeza:
That Odoo PR is not merged. Why do you say so?

legalsylvain (commented):
Hi. Thanks for the fix and the clean.
Is it possible to add a multicurrency test ?

jcadhoc:
Sorry @pedrobaeza, i saw the pr Closed so i think it was merged. Anyways, that code i think is not neccesary any longer, with my change the module works well i think.

I made this because purchase_discount wasn´t working with multicurrency.

pedrobaeza:
Your change is not correct, as you can see in the corresponding tests.

PR #2223, opened 2024-04-02
[13.0][FIX] purchase_landed_cost: Remove ondelete from one2many fields

carolinafernandez-tecnativa:
- Remove ondelete from one2many fields as are not used for these fields.
- In future versions, as it has set ondelete on one2many fields, appears warning in log message

@Tecnativa
TT46594
@pedrobaeza @chienandalu


pedrobaeza (approved):
Fast-tracking it as pretty obvious

/ocabot merge patch

Please don't include the version number on the commit message (but on PR title).

chienandalu (approved):

PR #2227, opened 2024-04-04
[FIX]_request: Add products only in draft state

jcadhoc:
Hi!
I made this pr because in v16 its possible to add productos in approved state, thats functionally bad because if Request it's approved, is because a person approved the products aggregated before, if you change lines, the logical thing would be to approve again.

In v13 this works more or less good, because it didn't let you add productos in approved state.

I made only editable lines when is in "draft" state, maybe it's too prohibitive, what do you think?

### DIFF
--- a/purchase_request/views/purchase_request_view.xml
+++ b/purchase_request/views/purchase_request_view.xml
@@ -156,7 +156,7 @@
                     </group>
                     <notebook>
                         <page string="Products">
-                            <field name="line_ids">
+                            <field name="line_ids" attrs="{'readonly': [('state', '!=', 'draft')]}">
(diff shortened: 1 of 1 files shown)


florenciafrigieri2 (approved):
Functional test ok.
To not allow to add or edit lines once the PR was approved is the current behaviour of the module also in version 15.

florenciafrigieri2:
Hi @pedrobaeza! Can you please take a look at this PR? There was a behaviour in purchase_request that is missing in version 16.
Thanks in advance!

pedrobaeza (approved):
Seems OK

/ocabot merge minor

PR #2231, opened 2024-04-08
[16.0][FIX] purchase_order_owner

AungKoKoLin1997:
This fix PR addresses the issue in the downstream component where available quantities are sometimes misjudged due to ownership considerations in subcontracting.

@qrtl

### DIFF
--- a/purchase_order_owner/models/purchase_order.py
+++ b/purchase_order_owner/models/purchase_order.py
@@ -18,6 +18,17 @@ class PurchaseOrder(models.Model):
         "incoming stock picking.",
     )
 
+    def button_confirm(self):
+        """For subcontracting orders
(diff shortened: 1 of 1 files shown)


purchase_order_owner/models/purchase_order.py
```diff
@@ -18,6 +18,17 @@ class PurchaseOrder(models.Model):
         "incoming stock picking.",
     )
 
+    def button_confirm(self):
+        """For subcontracting orders
+        Without this, the downstream component demands may misjudge the available
+        quantities with owner consideration.
+        """
+        # TODO: Double-check if the logic is optimal
```
yostashiro, line 26: @AungKoKoLin1997 Can you please double-check the validity of this approach, and remove this comment once confirmed?

PR #2236, opened 2024-04-10
[16.0][ADD] purchase_reorder_control

trisdoan:


purchase_reorder_control/tests/test_unpurchaseable_product.py
```diff
@@ -0,0 +1,102 @@
...
+from odoo.exceptions import UserError
+from odoo.fields import Command
+
+from odoo.addons.product.tests.common import ProductAttributesCommon
+
+
+class TestUnpurchaseableProduct(ProductAttributesCommon):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
```
santostelmo, line 13:
```suggestion
        super().setUpClass()
        cls.env = cls.env["base"].with_context(**DISABLED_MAIL_CONTEXT).env
```

santostelmo (approved):

purchase_reorder_control/tests/test_unpurchaseable_product.py
nilshamerlinck: I believe this is already done in odoo's [BaseCommon](https://github.com/odoo/odoo/blob/16.0/odoo/addons/base/tests/common.py#L31) and we have `ProductAttributesCommon < ProductCommon < BaseCommon`

santostelmo: indeed it is done already in `BaseCommon`

cyrilmanuel (approved):

HviorForgeFlow:
Just a comment. Have you considr archieving the Orderpoint instead of unlinking? Maybe it can become active again when purchase_ok become True again (or not), to avoid possible forgotten orderpoints to be active.

trisdoan:
> Just a comment. Have you considr archieving the Orderpoint instead of unlinking? Maybe it can become active again when purchase_ok become True again (or not), to avoid possible forgotten orderpoints to be active.

Hi @HviorForgeFlow, it's a good idea.

I updated as you suggest :pray:

HviorForgeFlow (approved):

PR #2247, opened 2024-04-17
[IMP] purchase_lot : lot_id is computed from stock moves so that it c…

metaminux:
…an be updated after confirmation and purchase order lines aren't merged even if lots aren't defined at creation

As sale_order_lot_selection allows to change lot after confirmation, we think stock moves and related objects should propagate those changes (here purchase order lines)

This PR is linked to these others PRs :

- OCA/stock-logistics-workflow#1560
- OCA/sale-workflow#3057


purchase_lot/tests/test_purchase_lot.py
```diff
@@ -88,3 +108,48 @@ def test_purchase_lot(self):
...
+        sale_order_drp_shpng = so_form.save()
+        sale_order_drp_shpng.order_line.lot_id = self.env["stock.lot"].create(
+            {
+                "name": "Seq test DS pdt",
+                "product_id": self.external_serial_product.id,
+            }
+        )
+        initial_lot = sale_order_drp_shpng.order_line.lot_id
+        # Confirm sales order
+        sale_order_drp_shpng.action_confirm()
```
Kev-Roche: It is confusing to speak about dropshipping here. 
```suggestion
        # Create a sales order with a line of 200 PCE incoming shipment,
        # with route_id buy
        so_form = Form(self.env["sale.order"])
        so_form.partner_id = self.customer
        so_form.payment_term_id = self.env.ref(
            "account.account_payment_term_end_following_month"
        )
        with mute_logger("odoo.tests.common.onchange"):
            # otherwise complains that there's not enough inventory and
            # apparently that's normal according to @jco and @sle
            with so_form.order_line.new() as line:
                line.product_id = self.external_serial_product
                line.product_uom_qty = 200
                line.price_unit = 1.00
                line.route_id = self.env.ref("purchase_stock.route_warehouse0_buy")

        sale_order = so_form.save()
        sale_order.order_line.lot_id = self.env["stock.lot"].create(
            {
                "name": "Seq test DS pdt",
                "product_id": self.external_serial_product.id,
            }
        )
        initial_lot = sale_order.order_line.lot_id
        # Confirm sales order
        sale_order.action_confirm()
```

Kev-Roche (approved):

metaminux:
Hi @florian-dacosta 

Now that OCA/stock-logistics-workflow#1560 has been merged, can we process this one too ?

Regards,

florian-dacosta:
@metaminux I don't really like this because the lot on the purchase line could change anytime silently, no matter the state of the purchase order.

In most cases (or at least in some cases), we want the lot on the purchase so we can send it (+ eventually information stored on the lot) to the supplier, so that it can send a product accordingly.
If the purchase order is already validated or even "sent", and the lot change, you probably want to inform your supplier (maybe by EDI, mail, phone or whatever). 
It does not help to change the lot in Odoo if we do not inform the supplier.

I would be ok to change the lot if the purchase is still at draft state, but otherwise, I'd rather not.

metaminux:
Hi @florian-dacosta 

You're right, maybe I'm too focused on my specific use case which is inter company workflow...
It seems to me that selling a specific lot that we need to purchase implies tight communication with the supplier.

I can only imagine 2 use cases here :
- you are defining the lots/serial numbers your supplier will put on the products you are buying
- your supplier give you the list of lots/serial number he will send to you at early stage and you are dispatching those numbers in your sale orders

In first case, I suppose you are generating lots before confirming the sale order and you don't want to change it after (which is the behaviour we got for the moment)
In second case, your supplier gives you lots after the purchase has been confirmed, so we need to propagate changes from sale to purchase.

Now, in first case, if we change the lots after confirmation, it can be just switching between sales or you need to inform your supplier you want some new lots...

Of course, I could be wrong about use cases here, so feel free to correct me.

emiliesoutiras (approved):

HviorForgeFlow:
Any concern about https://github.com/OCA/purchase-workflow/pull/2247#issuecomment-2451764975 @florian-dacosta

florian-dacosta:
Hello @metaminux Sorry for my so late answer...
I still am uncomfortable with the idea of having the lot changed silently on the PO, but let's say it is not worse than having somthing inconsistent (between moves/sale order line).

There is still one thing, this PR is not compatible with this new one https://github.com/OCA/purchase-workflow/pull/2646

I do think we should have a purchase_lot with no dependency on sale_order_lot_selection. Indeed this module could be usefull without any link with sale order.
Just wanting to tell your supplier which lot he has to send somehow.

What would have to change here I guess : 
Set readonly=False on the lot_id field.
Remove sale_line_id from the computed field
=> This is quite annoying because that would require to have a very small glue module to manage it... 

I have no good solution to propose, but I think it is important to not depend of sale_order_lot_selection for a wider use of this module.

metaminux:
Hello @florian-dacosta and @Tisho99

I'm answering on both #2247 and #2646 to be sure everyone is OK.

I totally agree with @Tisho99's vision.

The underlying concept is to have control on lot / serial numbers in stock moves and that's what **stock_restrict_lot** is offering.
AFAIK, for the moment, only **sale_order_lot_selection** is triggering the restriction on moves (setting the `restrict_lot_id` field)
From the beginning, **purchase_lot** is aiming to get the lot from the sale order lines that created the purchase... maybe it's not enough :)

We could imagine a module being the equivalent of **sale_order_lot_selection** on purchase side (maybe the need is not so common, but why not)
Let's call it **purchase_order_lot_selection**, it would set restrictions on stock moves from the purchase order lines.
The same way **sale_order_lot_selection** is depending only on **sale_stock** and **stock_restrict_lot**, **purchase_order_lot_selection** should depend on **purchase_stock** and **stock_restrict_lot**.

Then, to get the link between sales and purchases (provided by native Odoo glue modules **sale_purchase** and **sale_purchase_stock**), we can have a glue module called **sale_purchase_lot_selection** or whatever.
This glue module would be what **purchase_lot** is doing nowadays, but would depend on **sale_order_lot_selection** and **purchase_order_lot_selection** with `'auto_install': True.`

In that case, `lot_id` fields in sale and purchase order lines would become computed writable fields that should be kept in sync with `restrict_lot_id` fields on related stock moves.

Two questions stay in my mind :
- how do we deal with updates occurring after confirmation of one or more objects involved (sale, purchase) ?
- where should we store the glue module ? should it stay in **OCA/purchase-workflow** as **purchase_lot** ? 

For the first question, the idea of opened PRs for "lot propagation" is to explicitly allow propagation or not and then if allowed, propagate on every objects as long as no stock move in the whole "chain of moves" is done.
For the second one, keeping it in **OCA/purchase-workflow** and only renaming it seems good to me.

What do you think ? @emiliesoutiras @Kev-Roche

metaminux:
Hello @florian-dacosta,

#2646 is now merged and this PR has been updated to be compatible.

Can we move forward ?

PR #2255, opened 2024-04-25
[16.0][IMP] supplier_calendar: robustness

LoisRForgeFlow:
when delta argument is a float `plan_days` method can return unexpected False result. This can lead to silent errors. Therefore we ensure that delta is a integer in the helper.


DavidJForgeFlow (approved):
LGTM!

ThiagoMForgeFlow (approved):
LGTM :+1:

PR #2256, opened 2024-04-26
[15.0][IMP] supplier_calendar: robustness

LoisRForgeFlow:
when delta argument is a float `plan_days` method can return unexpected False result. This can lead to silent errors. Therefore we ensure that delta is a integer in the helper.

Backport of #2255


LoisRForgeFlow:
It is now green as the problem in the other module has been fixed.

/ocabot merge patch

JoanSForgeFlow (approved):

JordiMForgeFlow (approved):

PR #2258, opened 2024-04-29
[FIX] purchase_manual_currency: Calculate Total (Currency Name) with order date

newtratip:
This PR fixed calculate Subtotal and Total with order date
Selection_3716


TheerayutEncoder:
@Saran440  What do you think about this PR.

I agree with @newtratip — we should calculate the total in the company currency using the exchange rate based on the order date, not today's date.

Saran440:
Could you check standard logic when purchase with multi-currency. It will use today or order date?
I think it should be use logic following standard.

PR #2269, opened 2024-05-06
[14.0][ADD] purchase_order_sending_warning

ajaniszewska-dev:


purchase_order_sending_warning/models/mail_mail.py
```diff
@@ -0,0 +1,17 @@
...
+
+from odoo import models
+
+
+class MailMail(models.Model):
+    _inherit = "mail.mail"
+
+    def write(self, vals):
+        res = super(MailMail, self).write(vals)
+        if vals.get("state") == "exception":
```
ajaniszewska-dev, line 12: todo: when success, remove flag?

ajaniszewska-dev: + to add filter in search view for that

purchase_order_sending_warning/views/purchase_order.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+
+    <record id="purchase_order_view_tree" model="ir.ui.view">
+        <field name="name">purchase.order.view.tree</field>
+        <field name="model">purchase.order</field>
+        <field name="inherit_id" ref="purchase.purchase_order_view_tree" />
+        <field name="arch" type="xml">
+            <field name="state" position="after">
+                <field name="sending_warning" readonly="1" />
```
ajaniszewska-dev: to remove since there is already boolean to use, but maybe we want to see it somewhere else?

purchase_order_sending_warning/models/mail_mail.py
```diff
@@ -0,0 +1,22 @@
...
+            for mail in self:
+                if mail.model == "purchase.order":
+                    order = self.env["purchase.order"].browse(mail.res_id)
+                    order.sending_warning = "email_not_sent"
+        else:
+            for mail in self:
+                if mail.model == "purchase.order":
+                    order = self.env["purchase.order"].browse(mail.res_id)
+                    order.sending_warning = False
```
florentx: Maybe refactor like this:
```suggestion
        res = super().write(vals)
        po_mails = res.filtered_domain([("model", "=", "purchase.order")])
        if po_mails:
            error = vals.get("state") == "exception"
            records = self.env["purchase.order"].browse(po_mails.mapped("res_id"))
            records.write({"sending_warning": "email_not_sent" if error else False})
```

purchase_order_sending_warning/models/purchase_order.py
```diff
@@ -0,0 +1,21 @@
...
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+
+    sending_warning = fields.Selection(selection=[("email_not_sent", "Email not sent")])
+
+    error_in_sending = fields.Boolean(
+        store=True,
+        compute="_compute_error_in_sending",
+        string="Error in sending",
+    )
```
florentx: I wonder if we need 2 fields here. Maybe the Selection field is enough.


For the name of the field, I would prefer `transmission_error` or `sending_error` since it is more important than a warning.

ajaniszewska-dev: @florentx 2 fields are there because selection is going to be extended (for example by edi) and user would like to know POs which failed (whatever they failed - edi or email), so computed field will be set whenever mail/edi sending fails. I'll rename.

PR #2270, opened 2024-05-07
[16.0] [ADD] purchase_sign

ByteMeAsap:
This module allows to take online signatures from vendors to confirm purchase orders similar to how it is currently for sale orders.


ByteMeAsap:
@rafaelbn , @yajo , @EmilioPascual , can you please review this

EmilioPascual (commented):
Thank you @ByteMeAsap for your work.
I've test module in runboat but I can't get it to work.
I have made a purchase order with a supplier using a portal user.
image
But no purchase order is displayed on the portal.
image
Also, does it make sense for suppliers to access a portal to confirm their sales? Normally it would be the other way around, the buyer would have to access the supplier's platform.
Perhaps you could add context to the pr by commenting on the situation in which this flow could be used.

ByteMeAsap:
> But no purchase order is displayed on the portal.

@EmilioPascual , only the RFQs in sent state would show up under Request for Quotation(as that's how its setted up in Odoo base). Once the RFQ is sent (by email) to the vendor, it would show up on the portal.

> Also, does it make sense for suppliers to access a portal to confirm their sales?

Yes, it does make sense in certain scenarios , for example we can define our rates for products to be purchased and ask suppliers for confirming the same or add their feedback through the communication section.

EmilioPascual (approved):

Jan-Onestein (approved):
thanks for the great work @ByteMeAsap

ByteMeAsap:
@rafaelbn , @yajo , @EmilioPascual , can we have this merged?

purchase_sign/controllers/main.py
```diff
@@ -0,0 +1,76 @@
...
+
+        try:
+            order_sudo.write(
+                {
+                    "signed_by": name,
+                    "signed_on": fields.Datetime.now(),
+                    "signature": signature,
+                }
+            )
+            request.env.cr.commit()
```
yajo: issue: this is a [discouraged behavior](https://www.odoo.com/documentation/17.0/contributing/development/coding_guidelines.html#never-commit-the-transaction), please remove the commit.

yajo (changes requested):

purchase_sign/controllers/main.py
ByteMeAsap: @yajo , I have removed the commit

yajo (approved):
/ocabot merge nobump

Jan-Onestein:
@rafaelbn , @yajo , @EmilioPascual , can we please have this merged?

PR #1322, opened 2021-12-27
[14.0] [MIG] purchase_discount: Port PR from 12.0 to 14.0

sonhd91:
PR Port:
* [PR #1047 [12.0][FIX] purchase_discount](https://github.com/OCA/purchase-workflow/pull/1047)
* [PR #1091 [12.0][IMP] purchase_discount: Add tracking to default_supplierinfo_discount on partners](https://github.com/OCA/purchase-workflow/pull/1091)
* [PR #1189 [FIX] purchase_discount: access error](https://github.com/OCA/purchase-workflow/pull/1189)
* [PR #1222 [FIX] purchase_discount: supplierinfo with no template](https://github.com/OCA/purchase-workflow/pull/1222)

PR #2272, opened 2024-05-10
[16.0][FIX] purchase_discount: supplierinfo with no template

ferran-S73:
Cherry-pick: https://github.com/OCA/purchase-workflow/pull/1322/commits/26a18a2e6e8d90d2623af0aa37105bf73b625154

Actually, product_tmpl_id isn't a required field, so in theory a dummy record could be created prior to assign it to a product template.

TT23732


pedrobaeza (approved):
Fast-tracking as already merged in previous version:

/ocabot merge patch

PR #2280, opened 2024-05-21
[15.0][FIX] purchase_quick: Fix error test CI

carolinafernandez-tecnativa:


HviorForgeFlow:
Hello @carolinafernandez-tecnativa I think you should force the purchase order and purchase order line to belong to a company.
Something similar done here: https://github.com/OCA/purchase-workflow/pull/2211/files

Thanks for taking care of the CI :)

carolinafernandez-tecnativa:
> Hello @carolinafernandez-tecnativa I think you should force the purchase order and purchase order line to belong to a company. Something similar done here: https://github.com/OCA/purchase-workflow/pull/2211/files
> 
> Thanks for taking care of the CI :)

Thanks for the suggestion! I added but still not working.
Keep working to solve CI.

What I found is that the error belongs to access rule purchase.order.line multi-company

HviorForgeFlow:
@carolinafernandez-tecnativa I cannot create a PR on your repository so please you can take a look here to see if it solves the remaining issue

https://github.com/OCA/purchase-workflow/commit/e5454f1cc483afc89a457c8dd23aa7736c4ecfad

carolinafernandez-tecnativa:
> @carolinafernandez-tecnativa I cannot create a PR on your repository so please you can take a look here to see if it solves the remaining issue
> 
> [e5454f1](https://github.com/OCA/purchase-workflow/commit/e5454f1cc483afc89a457c8dd23aa7736c4ecfad)

I have tried what you said in local but error persists. I also added on this PR so you can check it too.

But the issue is not only with test code, check functionality of module purchase_quick in multi-company environment and you will see does not work. I suspect that module is not supported by multi-company and others that depends on it too.

HviorForgeFlow:
Seems that the play onchanges removes the company_id if it is set previously. A workaround could be to set the company_id after   calling the play onchanges. You can take a look on #2286 and adapt the code. Maybe with some more time I would add an new method on base_product_mass_edition to manage the modifications of the dicts after the play onchanges.

pedrobaeza:
Superseded by #2286

PR #2218, opened 2024-03-29
[17.0][MIG] purchase_last_price_info

mostafabarmshory:

PR #2281, opened 2024-05-22
[17.0][MIG] purchase_last_price_info

anikeenko-viktor:
This PR based on https://github.com/OCA/purchase-workflow/pull/2218.
It fixes access right error that happens on purchase order line creation.

On purchase.order.line creation, `_update_inverses` is called before the value of `purchase.order.line.company_id` is computed.
`_update_inverses` causes Odoo to filter `purchase.order.line` to find the correct lines to put into the cache, which in turn
causes access rights check. And because `company_id` is not yet computed it fails.
Fixing by removing domain from `last_purchase_line_ids` to avoid filtering and filter it later manually.

image


HviorForgeFlow (approved):
Functional review :+1:

MateuGForgeFlow (approved):
Functional LGTM!

PR #2285, opened 2024-05-30
[14.0][ADD] purchase_cancel_remaining_delivery

TDu:
When a purchase order is partially received, this module allows to cancel the remaining deliveries.
This can be done with the new button `Cancel Remaining Delivery`.

The module depends on `purchase_picking_state` which keeps a state of the pickings related to a purchase order.


purchase_cancel_remaining_delivery/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+{
+    "name": "Purchase Cancel Remaining Delivery",
+    "version": "14.0.1.0.0",
+    "author": "Camptocamp SA,Odoo Community Association (OCA)",
+    "category": "Purchase",
+    "license": "AGPL-3",
+    "website": "https://github.com/OCA/purchase-workflow",
+    "depends": ["purchase_stock", "purchase_picking_state"],
```
mt-software-de: `purchase_stock` is already a dependency of `purchase_picking_state`
But why is the dependency to `purchase_picking_state` needed?
I don't think this is needed, we should keep the dependencies small. 

If someone wants to install `purchase_picking_state` it can be installed.

mt-software-de (changes requested):

purchase_cancel_remaining_delivery/__manifest__.py
TDu: `purchase_picking_state` is used to check if the button for this module needs to be displayed or not.
I could not include it but then I would need to re-implement the same computation in this module.

As that module is small and implements only the functionality that is needed here. It makes more sense to me to use it as dependency.

mt-software-de: Yes you are right i have not seen the invisible.
Could you please remove purchase_stock?

purchase_cancel_remaining_delivery/views/purchase_order.xml
```diff
@@ -0,0 +1,19 @@
...
+        <field name="name">purchase.order.form</field>
+        <field name="model">purchase.order</field>
+        <field name="inherit_id" ref="purchase.purchase_order_form" />
+        <field name="arch" type="xml">
+            <button name="button_done" position="after">
+                <button
+                    name="button_cancel_remaining_delivery"
+                    string="Cancel Remaining Delivery"
+                    type="object"
+                    groups="account.group_account_invoice"
```
mt-software-de: Why is this button not shown for all purchase users, or the manager, why is it only shown for the invoicing group?

purchase_cancel_remaining_delivery/__manifest__.py
TDu: Done

purchase_cancel_remaining_delivery/views/purchase_order.xml
TDu: Mistake :face_in_clouds: I removed it.
Thanks for your review

mt-software-de (approved):

vuwnevska (approved):

PR #2288, opened 2024-05-30
[16.0][FIX] purchase_return: fix create method

pilarvargas-tecnativa:
Fix purchase_return_order is not overriding the create method in batch
cc @Tecnativa

@carolinafernandez-tecnativa @pedrobaeza please review

### DIFF
--- a/purchase_return/models/purchase_return_order.py
+++ b/purchase_return/models/purchase_return_order.py
@@ -334,26 +334,27 @@ class PurchaseOrderReturn(models.Model):
                 lambda line: not line.display_type
             ).date_planned = self.date_planned
 
-    @api.model
-    def create(self, vals):
-        company_id = vals.get(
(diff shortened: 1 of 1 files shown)


purchase_return/models/purchase_return_order.py
```diff
@@ -334,26 +334,27 @@ def onchange_date_planned(self):
...
+                    seq_date = fields.Datetime.context_timestamp(
+                        self, fields.Datetime.to_datetime(vals["date_order"])
+                    )
+                vals["name"] = (
+                    self_comp.env["ir.sequence"].next_by_code(
+                        "purchase.return.order", sequence_date=seq_date
+                    )
+                    or "/"
+                )
+        return super(PurchaseOrderReturn, self_comp).create(vals_list)
```
pedrobaeza, line 357: This is not valid, as you can't pass a self with the company of just one of the records. This can't simply be done in batch.

PR #2290, opened 2024-05-31
[ADD] purchase_supplierinfo_editable_tree

victor-champonnois:
This module sets the supplierinfo tree view as editable. This allows to quickly edit purchase prices.
This is useful when we have to update the prices from a supplier, without having to do an export/import of supplierinfos.


purchase_supplierinfo_editable_tree/tests/test_purchase_order_supplierinfo_update.py
```diff
@@ -0,0 +1,132 @@
...
+        )
+        cls.supplierinfo_2 = cls.env["product.supplierinfo"].create(
+            {
+                "partner_id": cls.supplier.id,
+                "product_tmpl_id": cls.product_2.product_tmpl_id.id,
+                "price": 10,
+            }
+        )
+
+    def test_confirn_purchase_order(self):
```
pedrobaeza: These tests don't belong here, as you are testing behavior from other modules, not what this module changes.

pedrobaeza (commented):
Thanks for the module. Comment inline

purchase_supplierinfo_editable_tree/tests/test_purchase_order_supplierinfo_update.py
victor-champonnois: Yes it was a mistake, I juste removed them.

remytms (approved):
LGTM

rvalyi (approved):

rvalyi:
/ocabot merge nobump
(trivial module)

PR #2293, opened 2024-06-04
[16.0][ADD] purchase_order_invoicing_picking_filter

carolina-domatix:
This module is designed to replicate the functionality of the following module: https://github.com/OCA/sale-workflow/tree/16.0/sale_order_invoicing_picking_filter but within the context of purchase orders. The original contributors of the sale-workflow module are acknowledged in the Contributors section of the readme, as this just intends to add the same functionality to purchases.


syci:
Hello, you need to add "purchase_discount" to depends, invoice creation failed in

`File "/opt/odoo/auto/addons/purchase_order_invoicing_picking_filter/models/purchase_order_line.py", line 17, in _prepare_stock_move_invoice_line
"discount": self.discount,
AttributeError: 'purchase.order.line' object has no attribute 'discount'`

PR #2295, opened 2024-06-05
purchase_discount: Map first discount to new field on account.move.line

grindtildeath:
Depends on:
 - [ ] https://github.com/OCA/account-invoicing/pull/1638


grindtildeath:
Closed in favour of #2296

PR #2296, opened 2024-06-05
[FIX] purchase_discount : transmits correctly discount value to account.move.line, if account_invoice_triple_discount is installed.

legalsylvain:
Adapt purchase_discount to pending refactoring done in account_invoice_triple_discount.

(See refactoring https://github.com/OCA/account-invoicing/pull/1638)

Note : this PR should be merged before https://github.com/OCA/account-invoicing/pull/1638.

CC : @grindtildeath

### DIFF
--- a/purchase_discount/models/purchase_order.py
+++ b/purchase_discount/models/purchase_order.py
@@ -110,8 +110,12 @@ class PurchaseOrderLine(models.Model):
         self.discount = seller.discount
 
     def _prepare_account_move_line(self, move=False):
-        vals = super(PurchaseOrderLine, self)._prepare_account_move_line(move)
-        vals["discount"] = self.discount
+        vals = super()._prepare_account_move_line(move)
+        if self.env["account.move.line"]._fields.get("discount1", False):
+            # OCA/account_invoice_triple_discount is installed
+            vals["discount1"] = self.discount
+        else:
+            vals["discount"] = self.discount
         return vals
 
     @api.model


purchase_discount/models/purchase_order.py
```diff
@@ -110,8 +110,12 @@ def _apply_value_from_seller(self, seller):
         self.discount = seller.discount
 
     def _prepare_account_move_line(self, move=False):
-        vals = super(PurchaseOrderLine, self)._prepare_account_move_line(move)
-        vals["discount"] = self.discount
+        vals = super()._prepare_account_move_line(move)
+        if self.env["account.move.line"]._fields.get("discount1", False):
+            # OCA/account_invoice_triple_discount is installed
```
grindtildeath, line 115: It's a dependency of this module, so it will be anyway.

IMO it would make sense to keep your version so that in case someone only updates `purchase-workflow` it won't break. But I would add a WARNING logger in case `discount1` doesn't exists.

grindtildeath (commented):
Thx for the PR.

I guess we better close https://github.com/OCA/purchase-workflow/pull/2295 ?

purchase_discount/models/purchase_order.py
legalsylvain: > It's a dependency of this module, so it will be anyway.

No ! 

``purchase_discount`` doesn't depend on ``account_invoice_triple_discount``

See : https://github.com/OCA/purchase-workflow/blob/16.0/purchase_discount/__manifest__.py#L16

the point is that in Odoo CE : 
- purchase.order.line doesn't contain discount field, so OCA/purchase_discount implement it
- account.move.line does contain discount field, so OCA/account_invoice_triple_discount implement discount2 and discount3

legalsylvain:
> I guess we better close https://github.com/OCA/purchase-workflow/pull/2295 ?

Indeed.

purchase_discount/models/purchase_order.py
grindtildeath: I got confused checking the manifest of `purchase_triple_discount`, my bad.

grindtildeath (approved):

legalsylvain:
Damn'it ! a @OCA/procurement-purchase-maintainers can merge this trivial PR ? 

thanks !

HviorForgeFlow (approved):

PR #2303, opened 2024-06-11
[17.0][IMP] purchase_request (Lines readonly if is_editable==False)

patrickrwilson:
Majority of the fields are readonly once the PR gets approved except the product lines which can be edited even though the PR was approved or done. This PR adds the same readonly="is_editable == False" that the other fields use on the form to the line_ids to not allow modifications once the PR is approved/done.


HviorForgeFlow (approved):
Functional review

PR #2310, opened 2024-06-14
[14.0][FIX] purchase_order_type: preserve order type when matches company

dalonsod:
Before this fix, when onchange event for `company_id` was fired, type for purchase order was always set, then sometimes overwritten, even if former order type was compatible with the new company.

This causes an incompatibility with e.g. `purchase_order_type_dashboard`. When PO creation form is accessed coming from PO dashboard, selected order type was actually ignored, because `onchange` for company was fired and order type overwritten. This fixes it.

### DIFF
--- a/purchase_order_type/models/purchase_order.py
+++ b/purchase_order_type/models/purchase_order.py
@@ -64,4 +64,8 @@ class PurchaseOrder(models.Model):
 
     @api.onchange("company_id")
     def _onchange_company(self):
-        self.order_type = self._default_order_type()
+        if not self.order_type or (
+            self.order_type
+            and self.order_type.company_id not in [self.company_id, False]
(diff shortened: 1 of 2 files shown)


dalonsod:
@kittiu @celm1990 could you review? When merged, I'll port it to v15/v16/v17. Thanks!

celm1990 (approved):
LGTM. 
Functionality tested.

dalonsod:
@vvrossem could you review? Thanks!

vvrossem (approved):
LGTM, thank you

dalonsod:
@HviorForgeFlow is this PR ready to be merged? Thanks!

HviorForgeFlow (approved):

PR #2315, opened 2024-06-20
[16.0] purchase_packaging_default: default packaging return first pac…

cyrilmanuel:
…kaging with purchase=True

This PR allow to return the first packaging on the list of packaging_ids, that meet the condition purchase = true

### DIFF
--- a/purchase_packaging_default/models/purchase_order_line.py
+++ b/purchase_packaging_default/models/purchase_order_line.py
@@ -3,7 +3,7 @@
 
 import math
 
-from odoo import api, fields, models
+from odoo import api, models
 
 
 class PurchaseOrderLine(models.Model):
@@ -12,7 +12,10 @@ class PurchaseOrderLine(models.Model):
     def _get_default_packaging(self):
         """From product get 1st packaging found ordered by sequence"""
         product_template = self.product_id.product_tmpl_id
-        return fields.first(product_template.packaging_ids)
+        for pkg in product_template.packaging_ids:
+            if pkg.purchase:
+                return pkg
+        return product_template.packaging_ids.browse()
 
     @api.onchange("product_id")
     def _onchange_product_id(self):
--- a/purchase_packaging_default/models/purchase_order_line.py
+++ b/purchase_packaging_default/models/purchase_order_line.py
@@ -17,17 +17,14 @@ class PurchaseOrderLine(models.Model):
                 return pkg
         return product_template.packaging_ids.browse()
 
-    @api.onchange("product_id")
-    def _onchange_product_id(self):
-        if self.env.user.company_id.purchase_packaging_default_enabled:
-            product_packaging = self._get_default_packaging()
-            if product_packaging:
-                self.product_packaging_id = product_packaging
-
     @api.depends("product_id", "product_qty", "product_uom")
     def _compute_product_packaging_id(self):
         _self = self
         if self.env.user.company_id.purchase_packaging_default_enabled:
+            for line in self:
+                product_packaging = line._get_default_packaging()
+                if product_packaging:
+                    line.product_packaging_id = product_packaging
             _self = _self.with_context(keep_product_packaging=True)
         return super(PurchaseOrderLine, _self)._compute_product_packaging_id()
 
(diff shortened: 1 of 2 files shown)


santostelmo (approved):
LG. 
It would be a plus if you write a quick unit test: Test that only purchasable packaging are added to PO line

cyrilmanuel:
Hello, 

today the test contain the right order to trigger the carton as first :)

i can add a test with a wrong example to check if its not working correctly. what do you think ? @santostelmo

lmignon (approved):
Great. Can you add a test to ensure that this feature will not be lost in the future.

purchase_packaging_default/models/purchase_order_line.py
```diff
@@ -12,7 +12,9 @@ class PurchaseOrderLine(models.Model):
     def _get_default_packaging(self):
         """From product get 1st packaging found ordered by sequence"""
         product_template = self.product_id.product_tmpl_id
-        return fields.first(product_template.packaging_ids)
+        return fields.first(
+            product_template.packaging_ids.filtered(lambda p: p.purchase)
+        )
```
lmignon: This approach is not optimal... You first loop over all the packaging and then return the first one.... 
```suggestion
        for pkg in  product_template.packaging_ids:
            if pkg.purchase:
                return pkg
        return product_template.packaging_ids.browse()
```

HviorForgeFlow (changes requested):
@cyrilmanuel can you attend comment, thanks: Attend https://github.com/OCA/purchase-workflow/pull/2315#discussion_r1662613500

santostelmo:
@cyrilmanuel could you please have a look on this one ?

santostelmo:
@HviorForgeFlow I reworked this PR:
1. Add [change suggested](https://github.com/OCA/purchase-workflow/pull/2315#discussion_r1662613500) by @lmignon 
2. Add a new commit: purchase-packaging-default-packaging: computed default product packaging. We assign default product packaging via api.depends as we need to compute product packaging even if PO is not created from form view. For example reordering rules create PO's and onchange is not triggered

Could you please review again ?

lmignon (changes requested):
Thank you for the PR but IMO this kind of changes deserve some unit tests. 
* Only purchase packaging are used and set on onchange 
* If a packing is specified into the data given to the create method, it's used otherwise if a default one exists this last one is used.
* ???

santostelmo:
> Thank you for the PR but IMO this kind of changes deserve some unit tests.
> 
> * Only purchase packaging are used and set on onchange
> * If a packing is specified into the data given to the create method, it's used otherwise if a default one exists this last one is used.
> * ???

Preparing some unit tests

santostelmo:
@lmignon added a unit test to cover the following:
1. setting is enabling desired behavior => get the first product packaging in the sequence 
2. PO created from directly from api still have default packaging computed

Let me know if you still see any other case that should be covered

purchase_packaging_default/tests/test_purchase_packaging_default.py
```diff
@@ -39,3 +41,56 @@ def test_purchase_packaging_default(self):
...
+    # def _find_suitable_product_packaging(self, product_qty, uom_id):
+    #         """ try find in `self` if a packaging's qty in given uom is a divisor of
+    #         the given product_qty. If so, return the one with greatest divisor.
+    #         """
+    #         packagings = self.sorted(lambda p: p.qty, reverse=True)
+    #         for packaging in packagings:
+    #             new_qty = packaging._check_qty(product_qty, uom_id)
+    #             if new_qty == product_qty:
+    #                 return packaging
+    #         return self.env['product.packaging']
```
Ricardoalso:
```suggestion

```

Ricardoalso (approved):
LGTM

purchase_packaging_default/tests/test_purchase_packaging_default.py
santostelmo: Thank you @Ricardoalso . Fixed

lmignon (approved):
LGTM (Code + Functionnal)

ivantodorovich (approved):
LG

santostelmo:
@HviorForgeFlow good for you ?

santostelmo:
> @HviorForgeFlow good for you ?

@HviorForgeFlow Could you kindly review the change request?

HviorForgeFlow (approved):

PR #2316, opened 2024-06-21
[17.0][FIX] purchase_order_type: preserve order type when matches company

dalonsod:
Fw-port of #2310 

Before this fix, when onchange event for company_id was fired, type for purchase order was always set, then sometimes overwritten, even if former order type was compatible with the new company.

This cause an incompatibility with e.g. purchase_order_type_dashboard. When PO creation form is accesed coming from PO dashboard, selected order type was actually igored, because onchange for company was fired and order type overwritten. This fix it.

### DIFF
--- a/purchase_order_type/models/purchase_order.py
+++ b/purchase_order_type/models/purchase_order.py
@@ -64,4 +64,8 @@ class PurchaseOrder(models.Model):
 
     @api.onchange("company_id")
     def _onchange_company(self):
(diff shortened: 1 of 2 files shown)


dalonsod:
@HviorForgeFlow  @celm1990 @vvrossem could you review? Thanks!

vvrossem (approved):
Thank you for this PR :pray:

HviorForgeFlow (approved):

PR #2317, opened 2024-06-23
[16.0][REF] purchase_triple_discount: Consolidate discount in std field

ramiadavid:
Until now, the triple discount feature did use the standard Odoo field as the first discount field, adding only extra fields for the second and the third discount. This implied we had to override any function using discount to consider the other discounts properly.

By adding an extra field discount1 to store the first discount, it allows to redefine the standard discount field to a computed field that will consolidate the triple discount from the other fields, and avoid the need to redefine anything relying on the discount field as it will already consider the triple discounts.

depends:
- [x] https://github.com/OCA/account-invoicing/pull/1840


purchase_triple_discount/models/triple_discount_mixin.py
```diff
@@ -0,0 +1,89 @@
+# Copyright 2019 Tecnativa - David Vidal
+# Copyright 2019 Tecnativa - Pedro M. Baeza
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+import functools
+
+from odoo import api, fields, models
+
+
+class TripleDiscount(models.AbstractModel):
```
legalsylvain, line 9: this mixin should be in account_invoice_triple_discount module. so it can be used for the account.move.line model. don't you think ?

ramiadavid: Maybe, we have to look in detail to make sure it is compatible with both modules.

pedrobaeza:
Trying to finish it in 2599

PR #1928, opened 2023-07-03
[15.0][ADD] purchase_order_line_menu: New views and stored fields

Shide:
New views and fields added to the purchase order line menu

MT-2028 @moduon @rafaelbn @yajo @dreispt please review and comment if you want  :)

PR #2330, opened 2024-07-02
[15.0][IMP] purchase_open_qty: add purchase_order_line_menu dependency

ThiagoMForgeFlow:
If the purchase_order_line_menu and purchase_open_qty modules are installed, they both add the same fields to the purchase_order_line_tree view inheritance. This dependency avoids declaring duplicate fields.

@ForgeFlow

### DIFF
--- a/purchase_open_qty/__manifest__.py
+++ b/purchase_open_qty/__manifest__.py
@@ -10,7 +10,10 @@
     "author": "ForgeFlow, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/purchase-workflow",
     "category": "Purchases",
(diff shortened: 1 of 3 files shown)


LoisRForgeFlow (approved):
This was needed after the [latest change sin purchase_order_line_menu including a bunch of new fields to the view](https://github.com/OCA/purchase-workflow/pull/1928). We thought that the easiest was to just include it as a dependency as we think is very common to use both modules together anyway.

HviorForgeFlow (approved):

LoisRForgeFlow:
@ThiagoMForgeFlow could you fwport this one?

PR #2332, opened 2024-07-02
[15.0][FIX] purchase_delivery_split_date: do not create empty pickings

LoisRForgeFlow:
When there was a grouping key (aka a date_planned) that only corresponds
to PO lines that are services, a picking was created and left empty.

PR #2334, opened 2024-07-03
[16.0][FWP][FIX] purchase_delivery_split_date: do not create empty pickings

HviorForgeFlow:
When there was a grouping key (aka a date_planned) that only corresponds to PO lines that are services, a picking was created and left empty.

Forward port #2332 

CC @ForgeFlow


LoisRForgeFlow (approved):
Thanks :+1:

ThiagoMForgeFlow (approved):
LGTM :+1:

PR #2336, opened 2024-07-04
[17.0][FIX] purchase_delivery_split_date: do not create empty pickings

LoisRForgeFlow:
When there was a grouping key (aka a date_planned) that only corresponds to PO lines that are services, a picking was created and left empty.

Fwport of #2332


ThiagoMForgeFlow (approved):
LGTM :+1:

HviorForgeFlow (approved):

PR #2344, opened 2024-07-08
[17.0] [MIG] purchase_line_procurement_group

Deriman-Alonso:


StefanRijnhart (changes requested):
Please remove the pandoc binary. It was added because of how pre-commit worked (should be fixed now) in the linting commit. You have to squash it out of that commit to prevent it from being added to the git history.

PR #2349, opened 2024-07-08
[17.0][MIG] purchase_rfq_number: Migration to 17.0

StefanRijnhart:
Based on unmerged migration to 16.0 at https://github.com/OCA/purchase-workflow/pull/1936

* Fixed warning about ambiguous variable 'l'
* Moved res.config.settings into its own model file
* Clarified test that caught an exception that was not raised anymore
* Adapt to view changes in Odoo 17


Saran440 (commented):
Functional Test :+1:

However, the `RFQ Reference` and `PO Reference` retain different values when we check or uncheck "Use Same Enumeration."

What do you think about improving it so that they always match?

reference: https://github.com/OCA/purchase-workflow/pull/2471

BhaveshHeliconia (approved):
Functional review LGTM!

PR #2357, opened 2024-07-12
[16.0][ADD] purchase_order_packaging

mathieudelva:


kevinkhao:
@mathieudelva I took the liberty to add a mini-fix commit for a typo in the code that was raising KeyErrors when the values were absent

PR #2358, opened 2024-07-15
[16.0][ADD] purchase_force_invoiced_quantity: add module

dessanhemrayev:
This module implements additional featured for the Purchase Force Invoiced [https://github.com/OCA/purchase-workflow/tree/16.0/purchase_force_invoiced](https://github.com/OCA/purchase-workflow/tree/16.0/purchase_force_invoiced%EF%BB%BF) module.

This module add a new "Force Invoiced" field to purchase order lines. This field is used as modifier when computing quantity to be invoiced.

Current behaviour: quantity to invoice = delivered -  invoiced

Implemented behaviour: quantity to invoice = delivered -  invoiced - force invoiced


dessanhemrayev:
@OCA/procurement-purchase-maintainers please check)

Aldeigja (approved):
Functional LGTM

ivs-cetmix (approved):
Code review LGTM

purchase_force_invoiced_quantity/models/purchase_order_line.py
```diff
@@ -0,0 +1,33 @@
...
+        """
+        Compute the quantity to invoice.
+        """
+        res = super()._compute_qty_invoiced()
+        for line in self:
+            # compute qty_to_invoice
+            if (
+                line.order_id.state in ["purchase", "done"]
+                and line.product_id.purchase_method != "purchase"
+            ):
```
hoangtrann: I think we should either filter for lines with `force_invoiced` set in the `for` loop or in the `if` condition

purchase_force_invoiced_quantity/views/purchase_order.xml
```diff
@@ -0,0 +1,21 @@
...
+        <field name="priority" eval="99" />
+        <field name="arch" type="xml">
+            <xpath
+                expr="//field[@name='order_line']/tree/field[@name='qty_invoiced']"
+                position="after"
+            >
+                <field
+                    name="force_invoiced"
+                    optional="show"
+                    attrs="{'column_invisible': [('parent.state', 'not in', ['purchase', 'done'])]}"
```
hoangtrann: we should set `readonly` attribute with the same condition despite it already being invisible

purchase_force_invoiced_quantity/models/purchase_order_line.py
dessanhemrayev: Thanks, good point
I'll fix it

purchase_force_invoiced_quantity/views/purchase_order.xml
dessanhemrayev: ok, i will add this
but why do we need this?
could you explain, please?

hoangtrann: it's for not letting user modify the value in any case when it isn't confirmed or done

maybe considering not letting user edit field when it's done as well (meaning it's already locked)

hoangtrann (approved):
Code LGTM

PR #2367, opened 2024-07-19
[16.0][IMP] purchase_manual_delivery: more specific menu name

JordiMForgeFlow:
To avoid confusions when used with https://github.com/OCA/purchase-workflow/blob/16.0/purchase_order_line_menu


GuillemCForgeFlow (approved):
makes sense 👍🏿

MiquelRForgeFlow (approved):
🤷‍♂️

(you can distinguish them by looking into their full xmlid anyway)

JordiMForgeFlow:
@MiquelRForgeFlow tell this to the user 😆

MateuGForgeFlow (approved):

PR #2369, opened 2024-07-22
[16.0][IMP] purchase_request: add 'In Progress' state

LauraCForgeFlow:
This improvement add a new state 'In Progress' to Purchase Requests. 

It can be useful for users in charge of managing Purchase Requests and creating their respective RFQs, to know if they have started processing that Purchase Request's related purchases or if it's still pending, once approved.

A Purchase Request can be set to the 'In Progress' state manually, but it also is set automatically when the first RFQ is created or assigned to the Purchase Request.

### DIFF
--- a/purchase_request/models/purchase_request.py
+++ b/purchase_request/models/purchase_request.py
@@ -8,8 +8,9 @@ _STATES = [
     ("draft", "Draft"),
     ("to_approve", "To be approved"),
     ("approved", "Approved"),
-    ("rejected", "Rejected"),
+    ("in_progress", "In progress"),
     ("done", "Done"),
+    ("rejected", "Rejected"),
 ]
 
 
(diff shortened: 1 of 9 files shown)


gbrito (approved):

purchase_request/models/purchase_request.py
```diff
@@ -8,8 +8,9 @@
     ("draft", "Draft"),
     ("to_approve", "To be approved"),
     ("approved", "Approved"),
-    ("rejected", "Rejected"),
+    ("in_progress", "In progress"),
```
AungKoKoLin1997, line 11: Non-blocking. This state should come after the draft, isn't it?

LauraCForgeFlow: Not really. The new 'In Progress' state is designed to help the person managing purchases track the progress of purchase requests, specifically the final step, which is purchasing the products on approved requests.

That's why they are automatically set to 'In Progress' when an RFQ is created, indicating that the purchasing process has begun. This ensures that the person in charge knows that action has been taken on the purchase request, in contrast to those in the 'Approved' state, where the purchasing process is still pending.

HviorForgeFlow (approved):

PR #2375, opened 2024-07-25
[FIX]purchase_discount:revert change

jcadhoc:
We create this commit because it enter in conflict with custom modules we have. field "discount1" exist in account.move.line for a diferent funcionality in our case.

We don´t need to merge.


StefanRijnhart:
Can this be closed? The commit was pushed in https://github.com/OCA/purchase-workflow/pull/2296 for forward compatibility with https://github.com/OCA/account-invoicing/pull/1638

PR #2377, opened 2024-07-31
[14.0] [ADD] purchase_line_service_qty_received

raumschmiede-joshuaL:
Automatically receive a service product with purchase_method receive when a product in the same PO is received.

I talked with @mt-software-de, copied https://github.com/OCA/sale-workflow/pull/2717 and changed it to work with POs


raumschmiede-joshuaL:
@jbaudoux

purchase_line_service_qty_received/models/purchase_order_line.py
```diff
@@ -0,0 +1,72 @@
...
+
+            received = True
+            if "stock_moves" in lines:
+                # True if at least any PO line with a stock product has a qty_received
+                received = any(
+                    not float_is_zero(line.qty_received, precision)
+                    for line in lines["stock_moves"]
+                )
+
+            for line in lines[METHOD_RECEPTION]:
```
jbaudoux: I think the complexity can be reduced by first filtering _[O(n)]_ instead of grouping twice _[O(n²)]_. Also better to not redefine _
```suggestion
        lines_grouped = groupby(
            self.order_id.order_line.filtered(lambda line: line.qty_received_method == METHOD_RECEPTION),
            lambda l: l.order_id
        )
        for order, lines in lines_grouped:
            # True if at least any PO line with a stock product has a qty_received
            received = any(
                    line.qty_received_method == "stock_moves"
                    and not float_is_zero(line.qty_received, precision)
                    for line in order.order_line
            )
            for line in lines:
```

raumschmiede-joshuaL: I applied the changes. They are better. I tested them and found out that without the changes a PO with only service products the PO lines with a reception service product have `qty_received` automatically set to 1. With your code not.  

And your code fixes the failing tests in the other modules :D  
Thanks a lot

mt-software-de (approved):
LGTM

jbaudoux (approved):

PR #2263, opened 2024-05-02
[FIX] purchase_request: message_post with html fixed

matiasperalta1:
A new argument has no been taken into account in message_post method.
More info: https://github.com/odoo/odoo/commit/7f4da03e3627be9c424160e4a131a279be021695

PR #2379, opened 2024-07-31
[17.0][MIG] purchase_blanket_order

ArnauCForgeFlow:
Standard Migration to 17.0


purchase_blanket_order/__manifest__.py
```diff
@@ -0,0 +1,29 @@
...
+    "name": "Purchase Blanket Orders",
+    "category": "Purchase",
+    "license": "AGPL-3",
+    "author": "ForgeFlow, Odoo Community Association (OCA)",
+    "version": "17.0.1.0.0",
+    "website": "https://github.com/OCA/purchase-workflow",
+    "summary": "Purchase Blanket Orders",
+    "depends": [
+        "purchase",
+        "web_action_conditionable",
```
LoisRForgeFlow: Looking at the module, this dependency looks like it is not needed. Can you check if it is really the case? If not, we probably can get rid of it during this migration.

ArnauCForgeFlow: You are right, I tested it and it looks like the module no longer needs this dependency. Removing it.

HviorForgeFlow (commented):
I noticed some error when not configuring UOM settings and try to confirm a purchase blanket order.

image

ArnauCForgeFlow:
@HviorForgeFlow solved!

LoisRForgeFlow (changes requested):
The order confirmation message seems that is not parsed appropiately as html? 

image

ArnauCForgeFlow:
> The order confirmation message seems that is not parsed appropiately as html?
> 
> image

Hey @LoisRForgeFlow 

The confirmation message is generated in purchase_request module, and there is an existing pr solving this problem: 
https://github.com/OCA/purchase-workflow/pull/2263

LoisRForgeFlow (approved):
Thanks @ArnauCForgeFlow I tested the wrong module :rofl: 

This time I reviewed blanket order, functional review is good :+1:

HviorForgeFlow (approved):
Thanks for attending the comments.

PR #2383, opened 2024-08-07
[16.0][FIX]purchase_manual_delivery: pending in forecast

ChristophAbenthungCibex:
Purchase Order Lines that are pending to receive are not considered for stock forecast. Thus, the manual and automatic reordering of products can be done despite having already create a Purchase Order and surpassing the max quantity. Confirmed Purchase Orders with pending to receive Purchase Order Lines are now considered at the forecast and the forecast report.

How to replicate at the runboat:
Assumption: Manual Delivery is already enabled in the settings
1. Login as Admin
2. Go to Inventory/Configuration/Reordering Rules
3. Create a new Rule
4. Set "[E-COM10] Pedal bin" as Product, Min Quantity to 50 and Max Quantity to 100
5. The forecast should be 22
6. Click the button "Order once"
7. The forecast changed to 100
8. Click on the button "Forecast Report" for the product "[E-COM10] Pedal bin" and then on the Purchase Order that just got created
9. Confirm the Purchase Order
10. Go back to the "Reordering Rules" or the "Forecast Report"
The Purchase Order is no part of the forecast anymore. Therefore, the "Order Once" button is shown again and when automatic reordering would trigger as well.

### DIFF
--- a/purchase_manual_delivery/__init__.py
+++ b/purchase_manual_delivery/__init__.py
@@ -1,2 +1,3 @@
 from . import models
+from . import report
 from . import wizard
--- a/purchase_manual_delivery/__manifest__.py
+++ b/purchase_manual_delivery/__manifest__.py
@@ -19,4 +19,9 @@
         "views/purchase_order_views.xml",
         "views/res_config_view.xml",
     ],
+    "assets": {
+        "web.assets_backend": [
+            "purchase_manual_delivery/static/src/**/*",
+        ]
+    },
 }
--- a/purchase_manual_delivery/models/__init__.py
+++ b/purchase_manual_delivery/models/__init__.py
@@ -1,3 +1,4 @@
+from . import product_product
 from . import res_company
 from . import res_config
 from . import purchase_order
--- a/purchase_manual_delivery/models/product_product.py
+++ b/purchase_manual_delivery/models/product_product.py
@@ -1,0 +1,58 @@
+from odoo import models
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    def _get_pending_lines_domain(self, location_ids=False, warehouse_ids=False):
+        domain = self._get_lines_domain(location_ids, warehouse_ids)
+        for domain_part in domain:
+            if (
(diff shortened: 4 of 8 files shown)


StefanRijnhart (approved):
Really nice work on the reporting!

purchase_manual_delivery/report/report_stock_forecast.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <template
+        id="purchase_report_product_product_replenishment"
+        inherit_id="stock.report_product_product_replenishment"
+    />
+</odoo>
```
StefanRijnhart: Is this override strictly necessary?

StefanRijnhart:
Picked in my 17.0 migration https://github.com/OCA/purchase-workflow/pull/2417

purchase_manual_delivery/report/report_stock_forecast.xml
ChristophAbenthungCibex: As far as I can tell: No. I tested it and it seems to work fine without it. I kept it in the code to see that this template or rather a part of it is gets changed. Odoo has the same override in its modules. See https://github.com/odoo/odoo/blob/16.0/addons/purchase_stock/report/report_stock_forecasted.xml
I can remove it if it is a problem.

StefanRijnhart: Looks like the file in Odoo is a remainder from this refactoring: https://github.com/odoo/odoo/commit/2bece51b24a960527c0609c4192e177f710e9bfd#diff-be21cb1e02bbe01d1a5bc4701e265ec107ff9717d3f272071c1bfb4822351196

I removed it in the 17.0 migration without any issue. You can do the same here.

ChristophAbenthungCibex: Gotcha. I removed it. 
Thank you very much.

StefanRijnhart (approved):

HviorForgeFlow (approved):
Code review :+1:

PR #2395, opened 2024-08-22
[16.0][IMP] sale_purchase_force_vendor: Add method for values to be used in product.supplierinfo record

victoralmau:
Add method for values to be used in `product.supplierinfo` record

@Tecnativa TT46573

### DIFF
--- a/sale_purchase_force_vendor/models/sale_order_line.py
+++ b/sale_purchase_force_vendor/models/sale_order_line.py
@@ -1,4 +1,4 @@
-# Copyright 2022 Tecnativa - Víctor Martínez
+# Copyright 2022-2024 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from odoo import api, fields, models
@@ -35,6 +35,17 @@ class SaleOrderLine(models.Model):
             )
             item.vendor_id_domain = domain
 
+    def _prepare_force_vendor_product_supplierinfo_vals(self):
+        """We use this method so that we can overwrite it if we need to modify or
+        add a value.
+        """
+        return {
+            "product_tmpl_id": self.product_id.product_tmpl_id.id,
+            "partner_id": self.vendor_id.id,
+            "min_qty": 0,
+            "company_id": self.company_id.id,
+        }
+
     def _prepare_procurement_values(self, group_id=False):
         """Inject in the procurement values the preferred vendor if any, and create
         supplierinfo record for it if it doesn't exist.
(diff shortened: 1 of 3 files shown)


sale_purchase_force_vendor/models/sale_order_line.py
```diff
@@ -35,6 +35,17 @@ def _compute_vendor_id_domain(self):
             )
             item.vendor_id_domain = domain
 
+    def _prepare_product_supplierinfo_vals(self):
```
pedrobaeza: Too generic name, as other module may create also a product.supplierinfo from here:

```suggestion
    def _prepare_force_vendor_product_supplierinfo_vals(self):
```

sale_purchase_force_vendor/models/sale_order_line.py
```diff
@@ -43,7 +54,7 @@ def _prepare_procurement_values(self, group_id=False):
         if self.vendor_id:
             product = self.product_id
             suppinfo = product.with_company(self.company_id.id)._select_seller(
-                partner_id=self.vendor_id,
+                partner_id=self.vendor_id.commercial_partner_id,
```
pedrobaeza: Don't change this, as there can be product.supplierinfo with contacts put and this will avoid its selection and will create over and over a new record:

https://github.com/OCA/purchase-workflow/blob/0c10e083ec9459b4d07c795f1cc7577d5974e8f2/sale_purchase_force_vendor/models/sale_order_line.py#L29

Maybe we should limit in the else case only to those with parent_id = False.

victoralmau: Example of use case:
- There is no limit to the vendors that can be selected.
- A child partner is selected: Vendor A, Child
- When creating the `product.supplierinfo` we want "_Vendor A_" to be created.

Modifying `vendor_id_domain` to not be able to select “_children_” will be changing the behavior for some who are already using it.

Perhaps another possibility is to create a method to get the vendor and if someone needs to extend it they can do so. Any idea for the method name?

pedrobaeza: There are reasons to have children contacts in product.supplierinfo (for example, for generating purchase orders to that children contact for communication purposes), so both the selection and the generation must be done to the children contact.

victoralmau: Ok, the introduced change has been removed.

pedrobaeza (approved):

carlos-lopez-tecnativa (approved):
LGTM

PR #2403, opened 2024-09-09
[16.0][FIX] purchase_return: restore display_type logic on purchase return orders

AlexPForgeFlow:
On migration commit a new display_type was introduced, 'product', but logic was not fully adapted and I was having issues when printing purchase return order report or when creating a refund. I have also added a migration script in order to restore those lines with display_type = 'product'.

### DIFF
--- a/purchase_return/migrations/16.0.1.0.2/post-migration.py
+++ b/purchase_return/migrations/16.0.1.0.2/post-migration.py
@@ -1,0 +1,13 @@
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE purchase_return_order_line
+        SET display_type = NULL
+        WHERE display_type = 'product'
+        """,
+    )
--- a/purchase_return/models/purchase_return_order_line.py
+++ b/purchase_return/models/purchase_return_order_line.py
@@ -118,7 +118,7 @@ class PurchaseReturnOrderLine(models.Model):
     )
 
     display_type = fields.Selection(
-        [("product", "Product"), ("line_section", "Section"), ("line_note", "Note")],
+        [("line_section", "Section"), ("line_note", "Note")],
         default=False,
         help="Technical field for UX purpose.",
     )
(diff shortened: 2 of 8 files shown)


purchase_return/migrations/16.0.1.0.2/post-migration.py
```diff
@@ -0,0 +1,13 @@
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE purchase_return_order_line
+        SET display_type = ''
```
ChrisOForgeFlow: Can you confirm empty value is null? or is a empty string?

ChrisOForgeFlow: Can you confirm empty value is null? or is a empty string?

AaronHForgeFlow (commented):
Can you explain a bit more the solution? Are you disabling the field display_type?  If there are records with display_type='line_section' or 'line_note' this migration script can cause issues I think.

purchase_return/migrations/16.0.1.0.2/post-migration.py
AlexPForgeFlow: @ChrisOForgeFlow if I set NULL value constrain https://github.com/OCA/purchase-workflow/pull/2403/files#diff-bdc7784d763e1a1bf97b6efc189067b7622b07550e82df61d197f5504f902218L134 raises, so I think should be either empty string or false? Or maybe we could remove that constrain

AlexPForgeFlow:
> Can you explain a bit more the solution? Are you disabling the field display_type? If there are records with display_type='line_section' or 'line_note' this migration script can cause issues I think.

@AaronHForgeFlow I'm not disabling the field. I just remove an option 'product' on display type selection, that was added in migration to v16. This module was not working properly, I think was not fully adapted, but I don't see the need of adding the display type product. As you can see, there was a proposal in v14 https://github.com/OCA/purchase-workflow/pull/1273, that do not introduce that change.

With the migration script, I just pretend to remove from the purchase return order lines the display type 'product', so the other types  'line_section' or 'line_note' shouldn't be affected

AaronHForgeFlow:
@AlexPForgeFlow , ok thanks. I thought that "product" was the normal display_type when creating lines. But if it is not a section nor a note the display_type is empty then.

AlexPForgeFlow:
@JordiBForgeFlow can we get it merged?

purchase_return/models/purchase_return_order_line.py
```diff
@@ -118,21 +118,21 @@ class PurchaseReturnOrderLine(models.Model):
     )
 
     display_type = fields.Selection(
-        [("product", "Product"), ("line_section", "Section"), ("line_note", "Note")],
+        [("line_section", "Section"), ("line_note", "Note")],
         default=False,
         help="Technical field for UX purpose.",
     )
```
LoisRForgeFlow, line 124: Should this be removed completely?

LoisRForgeFlow: nevermind, this is purchase.return.order.line. It is correct

PR #2405, opened 2024-09-13
[15.0][ADD] stock_move_purchase_price_update: New module

carlosdauden:
@Tecnativa TT50828

ping @sergio-teruel @CarlosRoca13


stock_move_purchase_price_update/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+    "category": "stock",
+    "website": "https://github.com/OCA/purchase-workflow",
+    "author": "Tecnativa, Odoo Community Association (OCA)",
+    "maintainers": ["carlosdauden"],
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": [
+        "purchase_stock",
+        "stock_picking_batch_extended",
```
sergio-teruel: You should create other module to add this function to batch pickings

stock_move_purchase_price_update/__manifest__.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2024 Tecnativa - Carlos Dauden
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Stock Move Line Purchase Price Update",
```
sergio-teruel: Stock move line or Stock move ???

sergio-teruel (changes requested):
Minor changes... But please create a new module for batch pickins.. I'm sorry for this

carlosdauden:
Changes done

sergio-teruel (approved):
Thanks!!!!

stock_move_purchase_price_update/readme/USAGE.md
```diff
@@ -0,0 +1,8 @@
+- Go to Sales \> Product \> Products and edit/create one
```
carlos-lopez-tecnativa: Why is it necessary to go to the Sales menu? If a user doesn't have the group, it can be done in the Purchase menu.
```suggestion
- Go to Purchase \> Product \> Products and edit/create one
```

carlosdauden: Changes done

carlos-lopez-tecnativa (approved):
LGTM

carlos-lopez-tecnativa:
Only a remark to include as a note in the README (if applicable):
If I receive a partial first picking and change the price to 100, and later receive the remaining and change the price to 110, all pickings will show a unit price of 110 (including the first picking, because it is taken from the purchase line, which is correct).

However, you can add to the README that the price on the picking reflects the purchase price for all pickings and is not individual for each picking. This can be confusing for users who expect each picking to save its unit price independently.

carlosdauden:
@carlos-lopez-tecnativa Added notice in description

HviorForgeFlow:
@carlosdauden what is the status of this PR?

PR #2409, opened 2024-09-19
[17.0][IMP] purchase_advance_payment: auto reconcile adv payments

vvrossem:


purchase_advance_payment/models/account_move.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def action_post(self):
+        # Automatic reconciliation of payment when bill confirmed.
+        res = super().action_post()
+        if bool(
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param("purchase_advance_payment.auto_reconcile_advance_payments")
+        ):
+            purchase_order = self.mapped("line_ids.purchase_order_id")
```
SilvioC2C: `mapped()` is automatically called by `Field.__get__()` for relational fields
```suggestion
            purchase_order = self.line_ids.purchase_order_id
```

purchase_advance_payment/data/ir_config_parameter.xml
```diff
@@ -1,9 +1,17 @@
...
-
+    <record
+        id="auto_reconcile_advance_payments"
+        model="ir.config_parameter"
+        forcecreate="0"
+    >
+        <field
+            name="key"
+        >purchase_advance_payment.auto_reconcile_advance_payments</field>
+        <field name="value">True</field>
```
SilvioC2C: I would set it to `False` to avoid the risk of a breaking change in preexisting workflows for instances that already have this module installed.

purchase_advance_payment/models/account_move.py
```diff
@@ -0,0 +1,32 @@
...
+            ):
+                json_invoice_outstanding_data = (
+                    self.invoice_outstanding_credits_debits_widget.get("content", [])
+                )
+                for data in json_invoice_outstanding_data:
+                    if (
+                        data.get("move_id")
+                        in purchase_order.account_payment_ids.move_id.ids
+                    ):
+                        self.js_assign_outstanding_line(line_id=data.get("id"))
```
SilvioC2C: `action_post()` accepts a recordset as `self`, while `js_assign_outstanding_line` only accepts a single record (calls `self.ensure_one()` as soon as it starts).
Please fix this inconsistency to avoid crashes.

purchase_advance_payment/models/account_move.py
```diff
@@ -0,0 +1,32 @@
...
+        res = super().action_post()
+        if bool(
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param("purchase_advance_payment.auto_reconcile_advance_payments")
+        ):
+            purchase_order = self.mapped("line_ids.purchase_order_id")
+            if (
+                purchase_order
+                and self.invoice_outstanding_credits_debits_widget is not False
```
SilvioC2C: `self` can be a recordset here, and in that case `self.invoice_outstanding_credits_debits_widget` will raise a `ValueError`

SilvioC2C (commented):
Few fixes and updates needed

purchase_advance_payment/models/account_move.py
vvrossem: Note: it should be fixed in `sale_advance_payment` too: https://github.com/OCA/sale-workflow/blob/17.0/sale_advance_payment/models/account_move.py#L10-L21

SilvioC2C (approved):
Overall LGTM from a tech PoV

LoisRForgeFlow (approved):
Hi, sorry for the late review here.

It looks great, functional and technical review :+1:

PR #2189, opened 2024-02-28
[16.0][FIX] purchase_manual_delivery: manual delivery with double validation

astirpe:
When "Purchase Order Approval" is enabled and the PO exceeds the Minimum Amount, the manual delivery is not applied at the moment that the PO is approved.
This PR fixes this case.

PR #2417, opened 2024-09-26
[17.0][MIG] purchase_manual_delivery

StefanRijnhart:
* Implement setUpClass in tests
* Adapt to use stock.move quantity field in Odoo 17
* Fix default value if company of the order is not the user's default company
* Increase test coverage

Includes https://github.com/OCA/purchase-workflow/pull/2189, https://github.com/OCA/purchase-workflow/pull/2383

We're also proposing a small data model change. Previously, purchase order line's `existing_qty` included quantities from done stock moves. We're renaming the field to `qty_in_receipt` and changing the semantics to only include quanties from pending stock moves. The `pending_to_receive` will still be set under the same condition: if the ordered quantity is larger than the received quantity plus the quantities on pending stock moves. Data migration is included.


astirpe (approved):
Thank you also for including #2189

HviorForgeFlow (approved):
Code review :+1: Awesome job with the cherry-picks

Issue #1855, opened 2023-04-25
[14.0] [purchase_advance_payment] wrong residual amount when reconciling credit note with invoice


ygol:
## Module
purchase_advance_payment

## Describe the bug
When reconciling an bill with a vendor credit note, the residual amount is false. As the vendor credit note amount is subtracted twice.

## To Reproduce
tested 14.0

1. create a PO, confirm + create an invoice
2. edit the above PO and add a new line with negative amount, confirm + create bill
3. reconcile the available vendor credit note amount on the vendor bill
4. Get wrong residual amount = 360000
![image](https://user-images.githubusercontent.com/2945174/234322485-14f78896-89ef-4351-95dd-ff7dc3823f41.png)

**Expected behavior**
Residual amount to be 480000

![image](https://user-images.githubusercontent.com/2945174/234322607-9f118d8f-714e-45f7-98b1-031d62dfdddf.png)
**Additional context**
Add any other context about the problem here. (e.g. OS, Python version, ...)

ygol:
the problem is that the PO total amount has already discounted the value of the credit note and when we reconcile the credit note  with the invoice it is thus deducted a 2nd time when https://github.com/OCA/purchase-workflow/blob/c186f0808b932416bb5e1cab563cfd87c2ee4d3b/purchase_advance_payment/models/purchase_order.py#L87

HviorForgeFlow:
ping @MateuGForgeFlow

HviorForgeFlow:
@ygol We will be more than excited to review your fix proposal

djtao1:
I am also very troubled by this issue. Is there a plan or possibility to fix this issue? thanks for support!

rvalyi:
cc @clementmbr @PierrickBrun @NathanQj

djtao1:
I have tested the latest version and it appears to be functioning correctly. but how to correct the erroneous entry caused by this issue earlier. They don't seem to have been automatically corrected.

ygol:
> I have tested the latest version and it appears to be functioning correctly. but how to correct the erroneous entry caused by this issue earlier. They don't seem to have been automatically corrected.

hi @djtao1 working on a solution in #1862

PR #2420, opened 2024-10-01
[14.0][FIX] use commercial_partner_id on prepare_payment_vals method

njeudy:
- If partner_id is used when creating payment, invoice will not detect advance payment to reconcile. so replace by commercial_partner_id.
- code for invoice_paid_amount is not correct, just use amount_total and amount_residual to keep it simple !

### DIFF
--- a/purchase_advance_payment/models/purchase_order.py
+++ b/purchase_advance_payment/models/purchase_order.py
@@ -83,12 +83,10 @@ class PurchaseOrder(models.Model):
             # Consider payments in related invoices.
             invoice_paid_amount = 0.0
             for inv in order.invoice_ids:
-                # use the reconciled payment amounts instead of the invoice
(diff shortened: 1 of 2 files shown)


njeudy:
@MateuGForgeFlow @HviorForgeFlow can you review it and merge ?

njeudy:
I teste it with invocie, advance and credit note and with partial payment, and now it works .. think we should update V15, V16 and V17 as well ?

HviorForgeFlow:
@njeudy I'm not sure about this change because in the past as you can see [here](https://github.com/OCA/purchase-workflow/issues/1855)  there was an issue regarding the change that you revert on da77183be96bc66a6bf1df2dcd4dcdbe937e69b4


@ygol @djtao1 can you give some feedback and review please, thanks :)

I personally never had the chance to use this module.

PR #2424, opened 2024-10-04
ADD purchase_order_downpayment_payment_order module

mourad-ehm:
The module allows to add down payments from purchase
orders to the payment order. Thus we can easily generate
SEPA payment from payment order.


purchase_order_downpayment_payment_order/wizard/purchase_order_downpayment_wizard.py
```diff
@@ -0,0 +1,113 @@
...
+        check_company=True,
+    )
+    company_id = fields.Many2one(
+        "res.company",
+        related="order_id.company_id",
+        store=True,
+    )
+
+    @api.depends("bank_partner_id")
+    def _compute_partner_bank_id(self):
```
florian-dacosta: Where is it used ? 
I am not sure this method is needed

purchase_order_downpayment_payment_order/models/purchase_order.py
```diff
@@ -0,0 +1,51 @@
...
+        return communication_type, communication
+
+    def _prepare_payment_line_vals(self, payment_order, amount_advance):
+        self.ensure_one()
+        communication_type, communication = self._get_communication()
+        if self.currency_id:
+            currency_id = self.currency_id.id
+        else:
+            currency_id = self.company_id.currency_id.id
+        partner_bank_id = first(self.partner_id.bank_ids).id
```
florian-dacosta: I think it would be nice to have a hook here, so the logic can be changed in case of multiple accounts per partner.

purchase_order_downpayment_payment_order/wizard/purchase_order_downpayment_wizard.py
mourad-ehm: I'll remove it

purchase_order_downpayment_payment_order/models/purchase_order.py
mourad-ehm: OK

PaulGoubert (approved):
Functionnal review done. LGTM

PR #2428, opened 2024-10-08
[17.0][MIG] purchase_order_line_price_history

david-s73:


david-s73:
@pedrobaeza You can review the pr

pedrobaeza:
/ocabot migration purchase_order_line_price_history

Isn't this now a native feature? Anyway, others should review first.

david-s73:
@pedrobaeza I have reviewed it with a consultant and we have found that it does the same as the core functionality, so I would close the pr for me.

PR #2430, opened 2024-10-10
[18.0][MIG] purchase_order_line_menu: Migration to 18.0

StefanRijnhart:
Trivial migration.

PR #2432, opened 2024-10-10
[18.0][MIG] purchase_manual_delivery

StefanRijnhart:
From unmerged https://github.com/OCA/purchase-workflow/pull/2417

* Move models and views into their own model file
* Integrate with purchase_order_line_menu (https://github.com/OCA/purchase-workflow/pull/2430)
* Refactor out duplicate buttons with additional context key. Introduce a context key to *avoid* the manual delivery flow instead.
* Fix faulty behaviour if order quantities are increased on a confirmed order.


purchase_manual_delivery/wizard/create_manual_stock_picking.py
```diff
@@ -0,0 +1,251 @@
...
+        res = super().default_get(fields)
+
+        active_model = self.env.context["active_model"]
+        if active_model == "purchase.order.line":
+            po_line_ids = self.env.context["active_ids"] or []
+            purchase_lines = (
+                self.env["purchase.order.line"]
+                .browse(po_line_ids)
+                .filtered(
+                    lambda p: p.product_id.type in ["product", "consu"]
```
astirpe: I think here it could be just:
```suggestion
                    lambda p: p.product_id.type == "consu"
```
as `"product"` is removed in `18.0`

purchase_manual_delivery/wizard/create_manual_stock_picking.py
```diff
@@ -0,0 +1,251 @@
...
+                )
+            )
+        elif active_model == "purchase.order":
+            po_ids = self.env.context["active_ids"] or []
+            purchase_lines = (
+                self.env["purchase.order"]
+                .browse(po_ids)
+                .mapped("order_line")
+                .filtered(
+                    lambda p: p.product_id.type in ["product", "consu"]
```
astirpe: Same here

purchase_manual_delivery/tests/test_purchase_manual_delivery.py
```diff
@@ -0,0 +1,490 @@
...
+        Create a new Product and Purchase Order.
+        Confirm Purchase Order and create a Picking with only a partial amount of
+        the selected amount of the Purchase Order Line. Confirm the Picking.
+        The quantity in progress is the pending to receive quantity of the Purchase
+        Order Line.
+        """
+        product_in_progress = self.env["product.product"].create(
+            {
+                "name": "Test product pending",
+                "type": "consu",
```
astirpe, line 372: I think there should be a `"is_storable"` here:

```suggestion
                "type": "consu",
                "is_storable": True,
```

astirpe (changes requested):

purchase_manual_delivery/wizard/create_manual_stock_picking.py
StefanRijnhart: Thanks, all suggestions taken!

astirpe (approved):

purchase_manual_delivery/models/purchase_order.py
```diff
@@ -0,0 +1,48 @@
...
+    def _compute_pending_to_receive(self):
+        for order in self:
+            order.pending_to_receive = True
+            if not any(order.mapped("order_line.pending_to_receive")):
+                order.pending_to_receive = False
+
+    def button_confirm_manual(self):
+        return super(
+            PurchaseOrder, self.with_context(manual_delivery=True)
+        ).button_confirm()
```
len-foss: Doesn't it seem like an obvious bug?
That it should be return `self.with_context(manual_delivery=True)
        ).button_confirm()`? Otherwise, you go back to purchase, and skip all overrides defined there. I'm not even sure that there's a good reason for this design (I frankly doubt it), but having all overrides in one case and not the other seems quite dangerous.

StefanRijnhart: If you mean, we should avoid creating a picking even when button_confirm is called directly in some way, I agree. I suggest we turn the mechanism around and make _create_picking do nothing for POs with manual_delivery=True *unless* a specific context key is present. We can then pass that context key in the wizard to force creating pickings for manualy delivery POs. Do you agree?

len-foss: I agree, that's the part where I suggested that I don't think it's a very good design (if the code of button_confirm_manual was directly put into an override of button_confirm that depends on a context key, you can still have 2 different buttons by putting the context in the attributes).

To make my point clearer, suppose you have module purchase_x that depends on purchase, and that it does an override of button_confirm to call `fx`.
Because `purchase_x` and `purchase_manual_delivery` have no relation to each other, we could either have `purchase_manual_delivery => purchase_x => purchase`, or `purchase_x => purchase_manual_delivery => purchase`.

If you call button confirm, in all cases you will call `fx` once. If you call button_confirm_manual, in one case you will call `fx` but not the other, depending on how the ORM loaded modules. If you decide that you want an easy solution and add an override of `button_confirm_manual` to also call `fx`, then you have the possibility to call `fx` twice, so you'd better be sure to make it idempotent.

StefanRijnhart: Ah, I see what you mean and you are right, that is a straightforward bug and can be fixed with 
```suggestion
        return self.with_context(manual_delivery=True).button_confirm()
```
Do you see additional value in the refactoring that I suggested?

len-foss: I sure do! Since a migration is the most appropriate time to do it, I  will make a PR with the simple fix for 17.0 and give you a review when you're done.

StefanRijnhart: I can do it, if you like.

StefanRijnhart: I need to rebase anyway.

len-foss: Just ping me when your refactoring is ready for review :-)

17.0 fix: https://github.com/OCA/purchase-workflow/pull/2573

StefanRijnhart: Here you go. I also added an additional fix for the fact that stock moves were still created when increasing quantities on order lines of a confirmed (manual delivery) order. Test failures are unrelated and fixed in https://github.com/OCA/purchase-workflow/pull/2576

purchase_manual_delivery/models/purchase_order.py
```diff
@@ -0,0 +1,39 @@
...
+    def _compute_manual_delivery(self):
+        """The manual delivery option is derived from the company of the order"""
+        for po in self:
+            po.manual_delivery = po.company_id.purchase_manual_delivery
+
+    def _compute_pending_to_receive(self):
+        for order in self:
+            order.pending_to_receive = True
+            if not any(order.mapped("order_line.pending_to_receive")):
+                order.pending_to_receive = False
```
len-foss: `order.pending_to_receive = any(order.mapped("order_line.pending_to_receive")` would be simpler? Also I don't think you want orders with no line to be pending_to_receive.

len-foss (approved):
LGTM with a minor comment

purchase_manual_delivery/models/purchase_order.py
StefanRijnhart: Agreed!

HviorForgeFlow (approved):

PR #2382, opened 2024-08-05
[17.0][FIX] purchase_reception_notify: use markup when posting HTML message

StefanRijnhart:
Fixes warning
```
odoo.addons.purchase_reception_notify.tests.test_purchase_reception_notify: Starting TestPurchaseReceptionNotify.test_reception_notification ...
odoo.addons.mail.models.mail_thread: Posting HTML message using body_is_html=True, use a Markup object instead (user: 1)
```

PR #2433, opened 2024-10-12
[18.0][MIG] purchase_request: Migration to 18.0

quoc-pn:


StefanRijnhart (changes requested):
Please check the tests. There is no product type `product` anymore. It's now product type `consu` + is_storable = True.

StefanRijnhart:
@quoc-pn Tests still fail. You could try running them locally to see if all is well before you push. There is also a markup warning. I solved a similar warning in https://github.com/OCA/purchase-workflow/pull/2382/files that you could apply here as well.

purchase_request/tests/test_purchase_request_procurement.py
```diff
@@ -104,7 +104,6 @@ def test_procure_purchase_request(self):
         self.env["mail.activity"].search(
             [("activity_type_id", "=", activity.id)]
         ).unlink()
-        activity.unlink()
```
StefanRijnhart: The comments above these lines say 'An activity is created on the request even if the activity type was deleted'. If we don't delete the activity type anymore, we don't test the original requirement. Did this functionality break?

purchase_request/wizard/purchase_request_line_make_purchase_order.py
```diff
@@ -200,8 +200,16 @@ def _get_order_line_search_domain(self, order, item):
             ("name", "=", name),
             ("product_id", "=", item.product_id.id),
             ("product_uom", "=", vals["product_uom"]),
-            ("analytic_distribution", "=?", item.line_id.analytic_distribution),
         ]
+
+        if item.line_id.analytic_distribution:
+            analytic_account_ids = list(item.line_id.analytic_distribution.keys())
+            order_line_data.append(
+                ("analytic_distribution", "in", analytic_account_ids)
```
StefanRijnhart, line 208: I think matching the exact distribution, not just overlapping accounts is required.

purchase_request/wizard/purchase_request_line_make_purchase_order.py
```diff
@@ -200,8 +200,16 @@ def _get_order_line_search_domain(self, order, item):
...
+            )
+        else:
+            order_line_data.append(("analytic_distribution", "=", False))
```
StefanRijnhart, line 211: The `=?` operator evaluates as True if the right hand side is falsy. That means that a request line without analytics can be linked to any order line with or without analytics. Did you mean to change that behaviour?

StefanRijnhart (changes requested):

purchase_request/tests/test_purchase_request_procurement.py
quoc-pn: This error will be raised when unlinking the activity type.
```
@api.ondelete(at_uninstall=False)
    def _unlink_except_todo(self):
        if self.env.ref('mail.mail_activity_data_todo') in self:
            raise UserError(_("The 'To-Do' activity type is used to create reminders from the top bar menu and the command palette. Consequently, it cannot be archived or deleted."))
```

`odoo/addons/mail/models/mail_activity_type.py`

purchase_request/wizard/purchase_request_line_make_purchase_order.py
quoc-pn: The `=?` operator doesn't support when searching the field `analytic_distribution` anymore.

```
    def _condition_to_sql(self, alias: str, fname: str, operator: str, value, query: Query) -> SQL:
        if fname != 'analytic_distribution' or self.env.context.get('account_report_analytic_groupby'):
            return super()._condition_to_sql(alias, fname, operator, value, query)

        if operator not in ('=', '!=', 'ilike', 'not ilike', 'in', 'not in'):
            raise UserError(_('Operation not supported'))
```

`odoo/addons/analytic/models/analytic_mixin.py`

purchase_request/tests/test_purchase_request_procurement.py
StefanRijnhart: That is great! That means that the activity type is guaranteed to exist. We can then drop the safeguarding from https://github.com/OCA/purchase-workflow/pull/2433/files#diff-556ada671b2bd79c2398ed16f2f6ad66bc3720e47b87d1ca4c232a367e72a122R44-R47. Can you also remove or adjust the lines of this test that are about the situation in which this particular activity type is not found?

purchase_request/wizard/purchase_request_line_make_purchase_order.py
StefanRijnhart: Yes indeed! Do you agree that we should still keep the original behaviour, which means ignoring the analytics of existing purchase lines in case of a request line that has no analytics set?

purchase_request/wizard/purchase_request_line_make_purchase_order.py
quoc-pn: When you use `=` instead of `in`, you will see this warning.

`WARNING odoo.osv.expression: The domain term '('analytic_distribution', '=', ['65'])' should use the 'in' or 'not in' operator.`

purchase_request/wizard/purchase_request_line_make_purchase_order.py
quoc-pn: The `=?` is unset or equal to, which means including the analytics of existing purchase lines has no analytics set.

I am keeping the original behavior.

luc-adhoc (approved):
functional review

purchase_request/wizard/purchase_request_line_make_purchase_order.py
StefanRijnhart: I'm sorry, you are absolutely right!

StefanRijnhart (approved):

luc-adhoc:
@StefanRijnhart could you please merge this PR? tks!

purchase_request/tests/test_purchase_request_allocation.py
```diff
@@ -19,7 +19,7 @@ def setUp(self):
         self.product_product = self.env["product.product"].create(
             {
                 "name": "Product Product Test",
-                "type": "product",
+                "type": "consu",
```
HviorForgeFlow: Should is_storable flag be set to True as well?

purchase_request/tests/test_purchase_request_to_rfq.py
```diff
@@ -27,7 +27,7 @@ def setUp(self):
         self.product_product = self.env["product.product"].create(
             {
                 "name": "Product Product Test",
-                "type": "product",
+                "type": "consu",
```
HviorForgeFlow: Same

purchase_request/tests/test_purchase_request_to_rfq.py
```diff
@@ -307,7 +307,7 @@ def test_purchase_request_to_rfq_minimum_order_qty_existing_po(self):
         product = self.env["product.product"].create(
             {
                 "name": "Widget",
-                "type": "product",
+                "type": "consu",
```
HviorForgeFlow: Same

HviorForgeFlow (commented):
@quoc-pn

purchase_request/tests/test_purchase_request_allocation.py
quoc-pn: I've updated the code. Thank you for your contribution.

HviorForgeFlow (approved):
Code and functional review :+1:

PR #2434, opened 2024-10-14
[15.0][ADD] purchase_reception_status_line

DavidJForgeFlow:
This module works with purchase_reception_status by adding the "force reception" check from the PO to the PO line.
This gives more flexibility when forcing some quantities.

@ForgeFlow


StefanRijnhart (commented):
Makes sense. What would you think of `purchase_line_reception_status` as a module name instead of `purchase_reception_status_line`? And please add a test.

DavidJForgeFlow:
Hi @StefanRijnhart,
My idea with the name was maintain the two related modules `purchase_reception_status` and `purchase_reception_status_line` consecutive in the modules list to make easier to know that are related. Changing the naming will remove this relation and split them in the list. What do you think about that?
And I will add tests ASAP.

StefanRijnhart:
@DavidJForgeFlow I guess that makes sense as well.

DavidJForgeFlow:
@StefanRijnhart tests done!

StefanRijnhart (approved):
Thanks for adding the tests!

HviorForgeFlow (approved):

PR #2435, opened 2024-10-17
[16.0] [FIX] purchase_request: Invalid Operator - Typo "=?" change to "="

lekha-sydney-tools:
[16.0] [FIX] purchase_request: Invalid Operator - Typo "=?" change to "="

### DIFF
--- a/purchase_request/wizard/purchase_request_line_make_purchase_order.py
+++ b/purchase_request/wizard/purchase_request_line_make_purchase_order.py
@@ -199,7 +199,7 @@ class PurchaseRequestLineMakePurchaseOrder(models.TransientModel):
             ("name", "=", name),
             ("product_id", "=", item.product_id.id),
             ("product_uom", "=", vals["product_uom"]),
-            ("analytic_distribution", "=?", item.line_id.analytic_distribution),
+            ("analytic_distribution", "=", item.line_id.analytic_distribution),
         ]
         if self.sync_data_planned:
             date_required = item.line_id.date_required


luke-stdev001:
@lekha-sydney-tools from our team has come across this operator typo in the code which results in an 'operation not supported' error when trying to trigger the Create RFQ button in the Create RFQ wizard as shown (apologies for the messy screenshot):

Screenshot 2024-10-17 at 1 12 56 pm

If anyone is available to give this a quick review to get this typo fix merged? It would be great to get this fixed upstream here so we can get rid of our local copy of this fix and just pull the latest from this repo.

purchase_request/wizard/purchase_request_line_make_purchase_order.py
```diff
@@ -199,7 +199,7 @@ def _get_order_line_search_domain(self, order, item):
             ("name", "=", name),
             ("product_id", "=", item.product_id.id),
             ("product_uom", "=", vals["product_uom"]),
-            ("analytic_distribution", "=?", item.line_id.analytic_distribution),
+            ("analytic_distribution", "=", item.line_id.analytic_distribution),
```
StefanRijnhart, line 202: It is not a typo. `=?` is a valid operator: https://github.com/odoo/odoo/blob/16.0/odoo/osv/expression.py#L1120-L1122

But the analytic distribution is a tricky field, and I assume the operator may well be buggy in this case. To emulate the behaviour of the operator, you need to include this clause in the domain optionally (i.e. if `item.line_id.analytic_distribution` has any content).

And as always in case of a bugfix, please try and reproduce the bug in a test.

StefanRijnhart (changes requested):
See comment below.

PR #2086, opened 2023-11-06
[16.0] [IMP] purchase_delivery_split_date: Fix key access to date dict

yassine-rhimi:

PR #2436, opened 2024-10-17
[15.0][FIX] purchase_delivery_split_date: prevent error when adding a PO line on confirmed POs

AaronHForgeFlow:
I reproduced it by using the purchase request module to add a new line on an existing PO using the flag:
"Merge on PO lines with equal Scheduled Date"

image

cc @ForgeFlow

### DIFF
--- a/purchase_delivery_split_date/models/purchase.py
+++ b/purchase_delivery_split_date/models/purchase.py
@@ -132,7 +132,7 @@ class PurchaseOrder(models.Model):
                             continue
                         if (
                             move.picking_id.scheduled_date.date() != date_key
-                            or pickings_by_date[date_key] != move.picking_id
+                            or pickings_by_date.get(date_key) != move.picking_id
                         ):
                             if date_key not in pickings_by_date:
                                 copy_vals = line._first_picking_copy_vals(key, line)


purchase_delivery_split_date/models/purchase.py
```diff
@@ -132,7 +132,7 @@ def _check_split_pickings(self):
                             continue
                         if (
                             move.picking_id.scheduled_date.date() != date_key
-                            or pickings_by_date[date_key] != move.picking_id
+                            or pickings_by_date.get(date_key, []) != move.picking_id
```
StefanRijnhart: Just a nitpick: I find using a list as a default value confusing and I would not be surprised if this comparison was going to give a TypeError at some point. You can just leave it out, because comparing with None (`get`'s default) is at least just as valid.
```
In [2]: picking
Out[2]: stock.picking(1,)

In [3]: picking != []
Out[3]: True

In [4]: picking != None
Out[4]: True
```

AaronHForgeFlow: Hi @StefanRijnhart , thanks for the suggestion!

StefanRijnhart: You resolved the conversation but the line has remained unchanged.

AaronHForgeFlow: Oops! Change done now.

StefanRijnhart: Alright, thanks!

StefanRijnhart (approved):

JordiMForgeFlow (approved):

MiquelRForgeFlow (approved):

StefanRijnhart:
Port forward not necessary I see. Already fixed as of 16 (https://github.com/OCA/purchase-workflow/pull/2086)

AaronHForgeFlow:
@StefanRijnhart thank you!

PR #2440, opened 2024-10-18
[16.0][IMP] purchase_request + purchase_request_type: Misc changes (hide company_id field + set groups to group_id field)

victoralmau:
Changes done:
- [x] `purchase_request`: Hide the `company_id` field from the lines + add index.
- [x] `purchase_request`: Set groups to `group_id` field
- [x] `purchase_request`: Add company_id field (invisible) to form view.
- [x] `purchase_request_type`: Show request_type field always (not only when you have Multi-company permission.
- [x] `purchase_request_type`: Leave the company_id field empty.
- [x] `purchase_request`: Leave the company_id field empty in the sequence.
- [x] `purchase_request_tier_validation`: Leave the company_id field empty in the definition.
- [x] `purchase_order_uninvoiced_amount`: Avoid warning log.
- [x] `purchase_product_packaging_container_deposit`: Change method to `invalidate_cache()` to `invalidate_model()` to avoid warning log.
- [x] `purchase_request_department`: Change `Environment.manage()` method.
- [x] `purchase_request_tier_validation`: Change tier.definition data to demo data.
- [x] `purchase_request_type`: Change purchase.request.type data to demo data.
- [x] `purchase_request`: Add kanban view.
ejemplo

Changes done:
- [x] `purchase_request_type`: Change toggle_active to web_ribbon + Archived filter.

**Before**
ejemplo

**After**
despues

@Tecnativa TT51248 TT55213


purchase_request/models/purchase_request.py
```diff
@@ -160,6 +174,18 @@ def _compute_is_editable(self):
         store=True,
     )
 
+    @api.depends("approver_user_id")
```
gbrito: Add the state so it trigger also when the state changes
```suggestion
    @api.depends("approver_user_id", "state")
```

purchase_request/models/purchase_request.py
```diff
@@ -285,11 +321,15 @@ def button_to_approve(self):
...
+        if not self.is_approvable:
+            raise UserError(_("You are not allowed to approve this request."))
+        return self.sudo().write({"state": "approved"})
 
     def button_rejected(self):
-        self.mapped("line_ids").do_cancel()
-        return self.write({"state": "rejected"})
+        if not self.is_approvable:
+            raise UserError(_("You are not allowed to reject this request."))
+        self.mapped("line_ids").sudo().do_cancel()
```
gbrito: Relational fields don't need mapped
```suggestion
        self.line_ids.sudo().do_cancel()
```

gbrito (changes requested):

victoralmau:
Locked by purchase_triple_discount https://github.com/OCA/purchase-workflow/pull/2579#issuecomment-2681177175

Finally the change of the approver has been removed because the focus is changed to `purchase_request_tier_validation`.

Please @carlos-lopez-tecnativa can you review it?

purchase_request/models/purchase_request_line.py
```diff
@@ -44,7 +44,6 @@ class PurchaseRequestLine(models.Model):
         comodel_name="res.company",
         related="request_id.company_id",
         string="Company",
-        store=True,
```
pedrobaeza, line 47: Why removing the store? In fact, I would do it stored with index for ir.rule faster application.

purchase_request/views/purchase_request_line_view.xml
```diff
@@ -449,12 +449,6 @@
                                 readonly="1"
                             />
                             <field name="date_planned" widget="date" readonly="1" />
-                            <field
```
pedrobaeza: Why removing this?

purchase_request/views/purchase_request_view.xml
```diff
@@ -196,11 +202,6 @@
                                     <field name="date_required" />
                                     <field name="estimated_cost" widget="monetary" />
                                     <field name="currency_id" invisible="1" />
-                                    <field
```
pedrobaeza, line 199: Why removing it?

purchase_request_tier_validation/data/purchase_request_tier_definition.xml
```diff
@@ -11,5 +11,6 @@
             name="reviewer_group_id"
             ref='purchase_request.group_purchase_request_manager'
         />
+        <field name="company_id" />
```
pedrobaeza: Add the multi-company group (for all the occurrences).

purchase_request_type/tests/test_purchase_request_type.py
```diff
@@ -74,6 +74,7 @@ def test_compute_purchase_request_type(self):
         self.assertEqual(request2.reduce_step, self.type2.reduce_step)
 
     def test_purchase_request_type_company_error(self):
+        self.type1.company_id = self.env.company
```
pedrobaeza, line 77: This sounds weird if you need to assign the company explicitly. Isn't this revealing a problem?

purchase_request_type/views/purchase_request_view.xml
```diff
@@ -18,7 +18,6 @@
                     name="request_type"
                     required="True"
                     attrs="{'readonly': [('is_editable','=', False)]}"
-                    groups="base.group_multi_company"
```
pedrobaeza, line 21: This is not correct you should add 2 times the same field, one invisible without group, and the other with the group.

purchase_request/views/purchase_request_view.xml
```diff
@@ -159,6 +160,11 @@
                                 widget="selection"
                                 attrs="{'readonly': [('is_editable','=', False)]}"
                             />
+                            <field
+                                name="company_id"
+                                groups="!base.group_multi_company"
```
pedrobaeza: No need of this `groups`. Odoo supports several times the same field.

pedrobaeza (changes requested):

purchase_request/models/purchase_request_line.py
victoralmau: IMO there is no sense in a related store field (in that case it should be a compute store), the `company_id` value of the lines will always be that of the header (a line will not be defined for a different company) therefore, the appropriate thing to do is to define the value as related no store.

purchase_request/views/purchase_request_line_view.xml
victoralmau: Because the company field does not add any value, it can only confuse.

purchase_request/models/purchase_request_line.py
pedrobaeza: The sense is for doing searches over that field in a performant way. For example, for the record rules application (the multi-company rule), so please restore it and put the index.

purchase_request/views/purchase_request_view.xml
victoralmau: What is the purpose of displaying the `company_id` field in the lines? None, let alone allowing to define that field (with something different than what the header has).

purchase_request/views/purchase_request_line_view.xml
pedrobaeza: I don't agree. It serves for knowing the company of the record.

purchase_request_tier_validation/data/purchase_request_tier_definition.xml
victoralmau: It is not necessary, this is a `tier.definition` data.

purchase_request/views/purchase_request_view.xml
pedrobaeza: When you see that lines directly, you can know the company. It's a common practice. If this view is only for being used inside the o2m, then I agree.

purchase_request/views/purchase_request_line_view.xml
victoralmau: Ok, so we keep it, it doesn't make sense IMO but I'll undo it.

purchase_request_tier_validation/data/purchase_request_tier_definition.xml
pedrobaeza: It doesn't matter. Any `company_id` should have it.

purchase_request_type/tests/test_purchase_request_type.py
victoralmau: That's right, the data `tier.definition` has no `company_id` defined and this test is intended to check what happens if a tier.definition has company_id defined but, again, if you want this to work differently comment it and I will modify it.

purchase_request_type/views/purchase_request_view.xml
victoralmau: The `request_type` field must always be shown, not only if you have the group_multi_company permission or the `purchase_request_type` module will be useless.

pedrobaeza: Ah, sorry, I haven't seen that the field is not `company_id`.

purchase_request_tier_validation/data/purchase_request_tier_definition.xml
victoralmau: It is not a view, it is a tier.definition record https://github.com/OCA/purchase-workflow/blob/16.0/purchase_request_tier_validation/data/purchase_request_tier_definition.xml#L5 where is the multi-company group supposed to be added?

pedrobaeza: I mean `<field name="company_id" groups="base.group_multi_company" />` (and maybe `<field name="company_id" invisible="1" />` as well)

victoralmau: Related to my previous comment (**NOT** a view) but applying what you suggest: https://github.com/OCA/purchase-workflow/actions/runs/13782923839/job/38544392532?pr=2440#step:8:1299

You can also go to runboat on a base database and try to install `purchase_request_tier_validation`.

pedrobaeza: Oh, sorry, I didn't understand you the first time. My bad. Indeed, that shouldn't have the group. To reinforce, you can do `<field name="company_id" /> <!-- no company set by default -->` for knowing also the reason.

pedrobaeza (approved):

carlos-lopez-tecnativa (approved):
LGTM

PR #2446, opened 2024-10-30
[UPD] purchase_order_approved: Update purchase order states

julia-factorlibre:
### Module
purchase_order_approved

### Describe the bug
The way states are uploaded is incorrect, if another module wants to include more states this will generate conflicts.
Fields as partner_id will be able to be edited in any state, when this should happen only in specifics states. 

### Affected versions:

16.0


lmchlfl:
Code review LGTM!

dalonsofl (approved):

dalonsofl:
LGTM!

PR #2449, opened 2024-11-04
[14.0'][ADD] Add purchase_propagate_qty_mrp

TDu:


purchase_propagate_qty_mrp/models/purchase_order_line.py
```diff
@@ -0,0 +1,51 @@
...
+            if line.state != "purchase":
+                continue
+            kit_line = line._propagate_qty_to_moves_mrp()
+            if kit_line:
+                kit_purchase_lines |= line
+        super(PurchaseOrderLine, self - kit_purchase_lines)._propagage_qty_to_moves()
+
+    def _propagate_qty_to_moves_mrp(self):
+        self.ensure_one()
+        bom = self.env["mrp.bom"].sudo()._bom_find(product=self.product_id)
```
mt-software-de: Yes _bom_find should be used but at first the moves should be checked. 
If there is no move with a bom line which linked to a bom for the purchased product than _bom_find should be used. 
This is the case for a kits in kit scenario then there is no bom for the main product. 
```suggestion
        boms = self.move_ids.bom_line_id.bom_id
        relevant_bom = False
        for bom in boms:
            if bom.type == "phantom" and (bom.product_id == self.product_id or bom.product_tmpl_id == self.product_id.product_tmpl_id and not bom.product_id):
                relevant_bom = bom
                break
        if not relevant_bom:
            relevant_bom = self.env["mrp.bom"].sudo()._bom_find(product=self.product_id, bom_type="phantom")
        bom = relevant_bom
```

purchase_propagate_qty_mrp/models/purchase_order_line.py
```diff
@@ -0,0 +1,51 @@
...
+        new_kit_quantity = self.product_uom_qty
+        boms, bom_sub_lines = bom.explode(self.product_id, new_kit_quantity)
+        for bom_line, bom_line_data in bom_sub_lines:
+            bom_line_uom = bom_line.product_uom_id
+            quant_uom = bom_line.product_id.uom_id
+            new_component_qty, procurement_uom = bom_line_uom._adjust_uom_quantities(
+                bom_line_data["qty"], quant_uom
+            )
+            moves = self.move_ids.filtered(
+                lambda move: move.product_id == bom_line.product_id
```
mt-software-de:
```suggestion
                lambda move: move.bom_line_id == bom_line
```

mt-software-de (changes requested):

purchase_propagate_qty_mrp/models/purchase_order_line.py
TDu: Thanks for your review. Using the `bom_line_id` on stock.move to find the related bom sounds much better.

Do you think using `_bom_find` is still necessary, because anyway no matching will be possible between the stock.move and the bom.line from the explode.

I will not include it yet.

mt-software-de: Yes it is still necessary for a kit in kit scenario. 
For example:
P-Kit-1 -> includes P1 and P2
P-Kit-2 -> includes P3 and P4
P-Kit-3 -> includes P-Kit-1 and P-Kit-2

The you will end update with moves for P1, P2, P3 and P4 no move will be link to the bom of P-Kit-3, the will be linked to P-Kit-1 bomlines and P-Kit-2. You will never find a bom for P-Kit-3.

PR #2459, opened 2024-11-20
[17.0] [ADD] purchase_partner_terms_conditions

Tisho99:
Adds a new terms and condition field on partner for purchases

T-6754


Tisho99:
@HaraldPanten @ValentinVinagre

PR #2461, opened 2024-11-21
[MIG][17.0] purchase_default_terms_conditions

Tisho99:


Tisho99:
@HaraldPanten @ValentinVinagre

HaraldPanten (approved):
Moving it forward as it's working properly and it's only a standard migration.

PR #2474, opened 2024-11-29
[14.0][ADD] new module purchase_order_qty_change_no_zero

clementmbr:
Prevent unit price to be zero when purchase order line quantity is changed

Variant/alternative of [purchase_order_qty_change_no_recompute](https://github.com/OCA/purchase-workflow/tree/14.0/purchase_order_qty_change_no_recompute)

cc @Kev-Roche @sebastienbeau


clementmbr:
This the purchase version of https://github.com/OCA/sale-workflow/pull/3439

PR #2484, opened 2024-12-13
[18.0][MIG] purchase_cancel_reason: Migration to 18.0

BhaveshHeliconia:


purchase_cancel_reason/tests/test_purchase_cancel_reason.py
```diff
@@ -0,0 +1,54 @@
+# Copyright 2013 Guewen Baconnier, Camptocamp SA
+# Copyright 2017 Okia SPRL
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import fields
+from odoo.tests.common import TransactionCase
+
+
+class TestPurchaseCancelReason(TransactionCase):
```
sebalix: We could inherit from `BaseCommon` class to disable tracking on all records created in tests.

purchase_cancel_reason/tests/test_purchase_cancel_reason.py
```diff
@@ -0,0 +1,54 @@
...
+    def setUp(self):
+        super().setUp()
```
sebalix: Small improvement, but as there is only one test this won't change anything here, so not blocking.
```suggestion
    @classmethod
    def setUpClass(cls):
        super().setUpClass()
```

sebalix (approved):
LG, proposing small improvements but not blocking

BhaveshHeliconia:
@sebalix , Thank you for your insightful comment, which has helped improve the code.

StefanRijnhart (changes requested):
Thanks! I did a functional test. The spacing looks off on a cancelled purchase order. It seems to say

> Cancellation reason:Service no longer needed

A space would be nice, to make it say

> Cancellation reason: Service no longer needed

Then again, I'm not fond of this unconventional way of displaying what is really a regular field. Would you consider just showing the field, under the condition that it has a value, for instance under `Vendor Reference`?


image

BhaveshHeliconia:
@StefanRijnhart , Thank you for your insightful comment, which has helped improve the code.

StefanRijnhart (commented):
Thanks for the update, but you can just make this a simple `<field name="cancel_reason_id" />`. No need for inline layout in this part of the form. That way, you can get rid of the layout tweak distinction between debug and non-debug mode (although it's well caught, and in other circumstances would be a sophisticated solution).

sheragar (commented):
LGTM - Functional Review 👍
It has been verified that purchase orders can be canceled individually, with a cancellation reason correctly assigned within the order. However, when canceling multiple purchase orders simultaneously, the option to include a cancellation reason is not available.
Could we add functionality in this view to optionally include a reason when canceling? This would enhance the traceability and flexibility of the process.

StefanRijnhart (commented):
Ah yes, that's right! Odoo 18 supports action buttons defined in list views as well, and `button_cancel` is added to two of those list views in the purchase module, e.g. https://github.com/odoo/odoo/blob/18.0/addons/purchase/views/purchase_views.xml#L596-L600. They would have to be replaced by this module's `action_purchase_order_cancel` as well.

BhaveshHeliconia:
@StefanRijnhart , @sheragar , Thanks for the suggestion to improve traceability! It was a great idea, and it helped streamline the process. I’ve fixed the issue and pushed the changes.

StefanRijnhart (approved):
Very nice, thanks!

PR #2490, opened 2024-12-17
[18.0][MIG] purchase_default_terms_conditions: Migration to 18.0 

BhaveshHeliconia:


purchase_default_terms_conditions/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,2 @@
+This module allows purchase default terms & conditions from Suppler or
```
sebalix:
```suggestion
This module allows purchase default terms & conditions from Supplier or
```

purchase_default_terms_conditions/tests/test_purchase_order.py
```diff
@@ -0,0 +1,64 @@
+# Copyright (C) 2022 Open Source Integrators
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.tests import tagged
+
+from odoo.addons.account.tests.common import AccountTestInvoicingCommon
+
+
+@tagged("post_install", "-at_install")
+class TestPurchase(AccountTestInvoicingCommon):
```
sebalix: We could probably be able to test this feature without depending on `AccountTestInvoicingCommon`. We could inherit from the new `BaseCommon` class so `tracking_disable` context key used in the test could be dropped.

purchase_default_terms_conditions/models/purchase.py
```diff
@@ -0,0 +1,21 @@
...
+
+    @api.onchange("partner_id", "company_id")
+    def onchange_partner_id(self):
+        if not is_html_empty(self.partner_id.purchase_note):
+            self.notes = self.partner_id.purchase_note
+        elif (
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param("purchase.use_purchase_note")
+        ):
```
sebalix: From what I see, the parameter should be named `purchase.use_company_purchase_note`.
Also, putting a value of `0` will still enable this feature, why not using `odoo.tools.str2bool`?

purchase_default_terms_conditions/tests/test_purchase_order.py
```diff
@@ -0,0 +1,64 @@
...
+                "order_id": purchase_order.id,
+                "taxes_id": False,
+            }
+        )
+
+        self.partner_a.write({"purchase_note": False})
+
+        self.env["ir.config_parameter"].set_param("purchase.use_purchase_note", "Test")
+
+        purchase_order.onchange_partner_id()
```
sebalix: Nowhere we are testing the value of the `notes` field from the PO (no assert). IMO the test doesn't test anything, this has to be fixed.

To trigger automatically onchanges when encoding a record, better to use `odoo.tests.common.Form` class too.

sebalix (commented):
Some remarks.
In fact, these remarks apply to previous version of this module as well, and some of them could be part of a separate commit so it'll be easier to backport.

BhaveshHeliconia:
@sebalix , Thank you for your insightful comment, which has helped improve the code. Could you kindly review it again?

BhaveshHeliconia:
@sebalix, Please review.!

HviorForgeFlow (approved):
code review :+1: 
looks good, what about squashing 2 last commits with the migration one. Not mandatory

marcelsavegnago (approved):

PR #2495, opened 2024-12-19
[18.0][MIG] purchase_order_archive: Migration to 18.0

BhaveshHeliconia:


StefanRijnhart (commented):
Not your fault, but this module seems to have a pretty bad flaw: you can only archive a cancelled or locked order, but once archived you can set the cancelled order to draft and confirm it etc. etc.
Would you consider adding a constraint that prevents archived orders in any other states than cancel or done? If you do, a simple test would be good to have as well.

BhaveshHeliconia:
@StefanRijnhart , Thank you for pointing this out and for the valuable suggestion! I've implemented a constraint to prevent archived orders from being in states other than "cancel" or "done." I’ve also added a simple test case to validate this behavior.

I’d greatly appreciate it if you could review the changes and let me know if there's anything else to improve.

Thank you again for your input!

StefanRijnhart (approved):
Great, thank you!

HviorForgeFlow (approved):
Code review :+1:

PR #2504, opened 2025-01-06
[18.0][MIG] purchase_analytic_global: Migration to 18.0

yankinmax:
The way analytic accounts are handled in v18 is the “analytic distribution“, meaning we can have several analytic accounts in a order line distribution with a specified percentage factor.

What we propose is to have a “global“ analytic account on purchase order if we have the same analytic account across the order lines distributions.


purchase_analytic_global/models/purchase.py
```diff
@@ -0,0 +1,33 @@
...
+    _inherit = "purchase.order"
+
+    account_analytic_id = fields.Many2one(
+        "account.analytic.account",
+        string="Analytic Account",
+        compute="_compute_analytic_account",
+        inverse="_inverse_analytic_account",
+        help="This account will be propagated to all lines distributions, if you need "
+        "to use different distribution accounts, define the accounts at line level.",
+    )
```
gurneyalex, line 17: We want to add an analytic distribution here, and propagate it to all the lines. This is feasible by adding the analytic.mixin class (see https://github.com/odoo/odoo/blob/18.0/addons/analytic/models/analytic_mixin.py#L8) on the purchase.order model.

gurneyalex (changes requested):

yankinmax:
Closed in favor of:
- [ ] https://github.com/OCA/purchase-workflow/pull/2558

PR #2512, opened 2025-01-09
[18.0][FIX] purchase_request: Copy in batch to prevent singleton error

celm1990:
Before this commit, duplicating a picking with more than one stock move would raise a singleton error. The copy_data method is now processed in batch. See https://github.com/odoo/odoo/pull/154132

image

@StefanRijnhart @HviorForgeFlow Could you please take a look?

@pedrobaeza Could you update the migration guide to consider this case during the migration process? https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-18.0

### DIFF
--- a/purchase_request/models/stock_move.py
+++ b/purchase_request/models/stock_move.py
@@ -108,41 +108,43 @@ class StockMove(models.Model):
         """
         if default is None:
             default = {}
-        if not default.get("purchase_request_allocation_ids") and (
-            default.get("product_uom_qty") or self.state in ("done", "cancel")
-        ):
-            default["purchase_request_allocation_ids"] = []
-            new_move_qty = default.get("product_uom_qty") or self.product_uom_qty
-            rounding = self.product_id.uom_id.rounding
-            for alloc in self.purchase_request_allocation_ids.filtered(
-                "open_product_qty"
+        vals_list = super().copy_data(default)
+        for move, vals in zip(self, vals_list, strict=False):
+            if not default.get("purchase_request_allocation_ids") and (
+                default.get("product_uom_qty") or move.state in ("done", "cancel")
             ):
-                if (
-                    float_compare(
-                        new_move_qty,
-                        0,
-                        precision_rounding=self.product_id.uom_id.rounding,
-                    )
-                    <= 0
-                    or float_compare(
-                        alloc.open_product_qty, 0, precision_rounding=rounding
-                    )
-                    <= 0
+                vals["purchase_request_allocation_ids"] = []
+                new_move_qty = default.get("product_uom_qty") or move.product_uom_qty
+                rounding = move.product_id.uom_id.rounding
+                for alloc in move.purchase_request_allocation_ids.filtered(
+                    "open_product_qty"
                 ):
-                    break
-                open_qty = min(new_move_qty, alloc.open_product_qty)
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

purchase_request/models/stock_move.py
```diff
@@ -108,41 +108,43 @@ def copy_data(self, default=None):
...
-        if not default.get("purchase_request_allocation_ids") and (
-            default.get("product_uom_qty") or self.state in ("done", "cancel")
-        ):
-            default["purchase_request_allocation_ids"] = []
-            new_move_qty = default.get("product_uom_qty") or self.product_uom_qty
-            rounding = self.product_id.uom_id.rounding
-            for alloc in self.purchase_request_allocation_ids.filtered(
-                "open_product_qty"
+        vals_list = super().copy_data(default)
+        for move, vals in zip(self, vals_list, strict=False):
```
StefanRijnhart, line 112: Isn't `strict=True` more logical?

purchase_request/models/stock_move.py
```diff
@@ -108,41 +108,43 @@ def copy_data(self, default=None):
...
+                            alloc.open_product_qty, 0, precision_rounding=rounding
+                        )
+                        <= 0
+                    ):
+                        break
+                    open_qty = min(new_move_qty, alloc.open_product_qty)
+                    new_move_qty -= open_qty
+                    vals["purchase_request_allocation_ids"].append(
+                        (
+                            0,
```
StefanRijnhart, line 139: Prefer `fields.Command.create` here.

purchase_request/models/stock_move.py
```diff
@@ -108,41 +108,43 @@ def copy_data(self, default=None):
...
-                        new_move_qty,
-                        0,
-                        precision_rounding=self.product_id.uom_id.rounding,
-                    )
-                    <= 0
-                    or float_compare(
-                        alloc.open_product_qty, 0, precision_rounding=rounding
-                    )
-                    <= 0
+                vals["purchase_request_allocation_ids"] = []
```
StefanRijnhart, line 116: In the previous situation it was `default` that was populated with the allocations, meaning that if a value for `purchase_request_allocation_ids` was passed, this mechanism would be circumvented. Because you set up the allocations directly in `vals`, this is no longer possible. You might want to check if a value for `purchase_request_allocation_ids` is not already present in vals and do nothing in that case.

StefanRijnhart (commented):
Some small comments. And a test would be nice.

StefanRijnhart:
BTW I added a note about the related change in Odoo to https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-18.0

purchase_request/models/stock_move.py
StefanRijnhart: Apologies, that is not how the `default` argument works. I was mislead by its name!

StefanRijnhart (approved):
Downgrading my remaining comments to 'nice to have'

purchase_request/models/stock_move.py
StefanRijnhart: Sorry, you already check this. Ignore.

HviorForgeFlow (approved):

PR #2516, opened 2025-01-15
[16.0][ADD] purchase_duplicate_check: Module added

geomer198:
## This module adds the following features to the Purchase App:

If there are pending RFQs or Purchase orders with associated incoming stock picking not in the "Done" state for the Product in the PO Line:
- Add Confirmation Wizard if there's an existing RFQ/undelivered PO for the same products on the RFQ confirmation
- Add the "Pending orders" field to the PO Line
- Assign Acitivty for the responsible user to check the associated orders


purchase_duplicate_check/models/purchase_order.py
```diff
@@ -0,0 +1,113 @@
...
+            product_qty = sum(product_line.mapped("product_qty"))
+            order_date = order.create_date.date()
+            order_href = (
+                f"<a href='/web#id={order.id}&model={order._name}'>{order.name}</a>"
+            )
+            type_ = order.state in ["draft", "sent"] and "RFQ" or "PO"
+            message += (
+                f"{type_}: {order_href} date: {order_date} Qty: {product_qty}<br/>"
+            )
+        return message
```
GabbasovDinar:
```suggestion
    def _prepare_pending_orders_message(self, product_id):
        """
        Prepare pending order line message
        :param product_id: product.product record id
        :return str: message
        """
        self.ensure_one()
        order_lines = self.env["purchase.order.line"].search([
            ('product_id', '=', product_id),
            ('order_id', 'in', self.ids)
        ])
        message_parts = []
        for line in order_lines:
            order = line.order_id
            order_href = f"<a href='/web#id={order.id}&model={order._name}'>{order.name}</a>"
            type_ = "RFQ" if order.state in ["draft", "sent"] else "PO"
            message_parts.append(
                f"{type_}: {order_href} date: {order.create_date.date()} Qty: {line.product_qty}<br/>"
            )
        return "".join(message_parts)
```

purchase_duplicate_check/models/purchase_order.py
```diff
@@ -0,0 +1,113 @@
...
+            )
+        ):
+            return
+        if not self._context.get("skip_rfq_confirmation"):
+            return (
+                self.env["confirmation.wizard"]
+                .with_context(skip_rfq_confirmation=True)
+                .confirm_pending_order(self)
+            )
+        return
```
GabbasovDinar: please, create separate method to check settings, e.g.:

```python
def _is_activity_enabled(self):
    """
    Check if activity for repeating orders is enabled.
    :return: bool
    """
    return self.env["ir.config_parameter"].sudo().get_param(
        "purchase_duplicate_check.create_activity_repeating_orders", False
    )
```
and
```python
def _check_pending_order(self):
    """
    Check for pending orders and trigger confirmation wizard if needed.
    """
    if self._is_activity_enabled() and not self._context.get("skip_rfq_confirmation"):
        return self.env["confirmation.wizard"].with_context(
            skip_rfq_confirmation=True
        ).confirm_pending_order(self)
```

purchase_duplicate_check/models/purchase_order.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def button_confirm(self):
+        action = self._check_pending_order()
+        if action is not None:
+            return action
+        return super().button_confirm()
```
GabbasovDinar:
```python
def button_confirm(self):
    """
    Confirm the purchase order.
    :return: action or super
    """
    action = self._check_pending_order()
    return action or super().button_confirm()
```

purchase_duplicate_check/models/purchase_order.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+class PurchaseOrderLine(models.Model):
```
GabbasovDinar: please, create separate file for this model: `purchase_order_line.py`

purchase_duplicate_check/models/purchase_order.py
```diff
@@ -0,0 +1,113 @@
...
+                    ("state", "not in", ["draft", "sent"]),
+                    ("picking_ids.picking_type_id.code", "=", "incoming"),
+                    ("picking_ids.state", "not in", ["done", "cancel"]),
+                ]
+            )
+            rec.pending_order_ids = rfq_orders
+
+    def _get_order_confirm_message(self):
+        """Get order confirmation message for pending orders"""
+        message = ""
```
GabbasovDinar: message_parts = []

purchase_duplicate_check/models/purchase_order.py
```diff
@@ -0,0 +1,113 @@
...
+        for line in self:
+            pending_orders = line.pending_order_ids
+            if not pending_orders:
+                continue
+            product_line_msg = pending_orders._prepare_pending_orders_message(
+                line.product_id.id
+            )
+            message += f"""
```
GabbasovDinar: message_parts.append(

purchase_duplicate_check/models/purchase_order.py
```diff
@@ -0,0 +1,113 @@
...
+            Product <b>{line.product_id.name}</b><br/>
+            {product_line_msg}<br/>
+            """
+        return message
```
GabbasovDinar: "".join(message_parts)

purchase_duplicate_check/models/purchase_order.py
```diff
@@ -0,0 +1,113 @@
...
+        compute="_compute_pending_order_ids",
+    )
+
+    def _compute_pending_order_ids(self):
+        purchase_order_obj = self.env["purchase.order"]
+        for rec in self:
+            if rec.product_type != "product":
+                rec.pending_order_ids = False
+                continue
+            rfq_orders = purchase_order_obj.search(
```
GabbasovDinar: it is necessary to use the `read_group` method instead of searching for records in a loop, which will reduce the load and optimize the query into the database.

purchase_duplicate_check/models/res_config_settings.py
```diff
@@ -0,0 +1,19 @@
+from odoo import api, fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    create_activity_repeating_orders = fields.Boolean(
```
GabbasovDinar: please, rename `create_activity_repeating_orders` -> `allow_create_activity_repeating_orders`

GabbasovDinar (changes requested):

purchase_duplicate_check/wizard/confirmation_wizard.py
```diff
@@ -0,0 +1,51 @@
...
+        for record in records:
+            self.env["mail.activity"].create(
+                {
+                    "user_id": activity_type.default_user_id.id or self.env.user.id,
+                    "activity_type_id": activity_type.id,
+                    "res_id": record.id,
+                    "res_model_id": model_id,
+                    "note": self.message,
+                }
+            )
```
GabbasovDinar:
```suggestion
        message = self.message
        user_id = activity_type.default_user_id.id or self.env.user.id
        activity_type_id = activity_type.id
        activity_vals_list = []
      
        for record in records:
            activity_vals_list.append({
                    "user_id": user_id,
                    "activity_type_id": activity_type_id,
                    "res_id": record.id,
                    "res_model_id": model_id,
                    "note":message,
            })
        if activity_vals_list:
            self.env["mail.activity"].create(activity_vals_list)
```

purchase_duplicate_check/__manifest__.py
```diff
@@ -0,0 +1,21 @@
+{
+    "name": "Purchase Duplicate Check",
+    "version": "16.0.1.0.0",
+    "summary": "Prevents overordering in the Purchase app with a Confirmation Wizard, "
+    "'Pending Orders' field, and activity tracking for repeated orders.",
+    "author": "Cetmix, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "category": "Inventory/Purchase",
+    "website": "https://github.com/OCA/purchase-workflow",
+    "live_test_url": "https://demo.cetmix.com",
```
hoangtrann: to remove? we use runbot for testing

purchase_duplicate_check/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+    "depends": ["purchase_stock", "confirmation_wizard"],
+    "external_dependencies": {},
```
hoangtrann: remove if have no value

purchase_duplicate_check/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+    "data": [
+        "views/purchase_order_views.xml",
+        "views/res_config_settings_views.xml",
+        "wizard/confirmation_wizard_views.xml",
+    ],
+    "assets": {},
```
hoangtrann: remove if have no value

purchase_duplicate_check/models/purchase_order.py
```diff
@@ -0,0 +1,57 @@
...
+        """
+        Prepare pending order line message
+
+        :param product_id: product.product record id
+        :return str: message
+        """
+        message_parts = []
+        order_lines = self.env["purchase.order.line"].search(
+            [("product_id", "=", product_id), ("order_id", "in", self.ids)]
+        )
```
hoangtrann: why do we need to search here? could you just use `self.order_line.filtered(lambda l: l.product_id.id == product_id)`?

purchase_duplicate_check/models/purchase_order.py
```diff
@@ -0,0 +1,57 @@
...
+                f"<a href='/web#id={order.id}&model={order._name}'>{order.name}</a>"
+            )
+            type_ = order.state in ["draft", "sent"] and "RFQ" or "PO"
+            message_parts.append(
+                f"{type_}: {order_href} date: {order.create_date.date()} Qty: {line.product_qty}<br/>"  # noqa
+            )
+        return "".join(message_parts)
+
+    def _is_activity_enabled(self) -> bool:
+        """Check if activity for repeating orders is enabled"""
```
hoangtrann: is repeating orders here mean duplicated orders?

purchase_duplicate_check/models/purchase_order_line.py
```diff
@@ -0,0 +1,88 @@
...
+class PurchaseOrderLine(models.Model):
+    _inherit = "purchase.order.line"
+
+    pending_order_ids = fields.Many2many(
+        "purchase.order",
+        string="Pending Orders",
+        compute="_compute_pending_order_ids",
+    )
+
+    def _compute_pending_order_ids(self):
```
hoangtrann: I think you do need at least `@api.depends('product_id')` here

purchase_duplicate_check/tests/test_purchase_order.py
```diff
@@ -0,0 +1,103 @@
...
+        with form.order_line.new() as line:
+            line.product_id = self.product_2
+            line.product_qty = 5.0
+        return form.save()
+
+    def test_prepare_pending_orders_message(self):
+        """Test flow where prepare message for purchase order"""
+        message = self.order1._prepare_pending_orders_message(self.product_2.id)
+        self.assertFalse(message, "Message must be empty")
+        expected_message = f"RFQ: <a href='/web#id={self.order1.id}&model=purchase.order'>{self.order1.name}</a> date: {Date.today()} Qty: 10.0<br/>"  # noqa
```
hoangtrann: shouldn't you get the date from the order instead of using Date.today() ?

purchase_duplicate_check/models/purchase_order.py
```diff
@@ -0,0 +1,57 @@
...
+            [("product_id", "=", product_id), ("order_id", "in", self.ids)]
+        )
+        for line in order_lines:
+            order = line.order_id
+            order_href = (
+                f"<a href='/web#id={order.id}&model={order._name}'>{order.name}</a>"
+            )
+            type_ = order.state in ["draft", "sent"] and "RFQ" or "PO"
+            message_parts.append(
+                f"{type_}: {order_href} date: {order.create_date.date()} Qty: {line.product_qty}<br/>"  # noqa
```
hoangtrann: This will looks like `RFQ: PO001 date: 01/01/2025 Qty: 10.0`, it would be much better to be shown as `RFQ: PO001 orders 10.0 unit, to confirm by 01/01/2025` and `PO: PO001 orders 10.0 unit, expected to received by 01/01/2025`

hoangtrann (changes requested):

hoangtrann:
test is also failing, please check

purchase_duplicate_check/models/purchase_order.py
GabbasovDinar: @hoangtrann in case recordset has a large set of records the filtered method is slow in this case search will work faster.

GabbasovDinar (approved):

purchase_duplicate_check/models/purchase_order.py
geomer198: > is repeating orders here mean duplicated orders?

Yes

purchase_duplicate_check/models/purchase_order.py
geomer198: > This will looks like `RFQ: PO001 date: 01/01/2025 Qty: 10.0`, it would be much better to be shown as `RFQ: PO001 orders 10.0 unit, to confirm by 01/01/2025` and `PO: PO001 orders 10.0 unit, expected to received by 01/01/2025`

I think this is too much.

hoangtrann: then at least separate the components with `-` or `;`

purchase_duplicate_check/models/purchase_order.py
hoangtrann: I think let's use `duplicated` instead of `repeating` then

purchase_duplicate_check/models/purchase_order_line.py
hoangtrann: @geomer198 any comment?

geomer198: This compute method have many dependencies, that is why I don't write `depends` decorator here.

geomer198: > @geomer198 any comment?

This compute method have many dependencies, that is why I don't write `depends` decorator here.

hoangtrann (approved):
Code LGTM

Bearnard21 (approved):
Functionality LGTM

ivs-cetmix:
Hey @OCA/procurement-purchase-maintainers would be great if someone could have a look at this module.
It's alleys being used in production and community could definitely benefit from it too.

HviorForgeFlow:
@ivs-cetmix It appears that this module's implementation of the confirmation feature is overly strict. Many tests (where users create similar purchase orders to validate features in their own modules) are failing as a result.

I recommend using config["test_enable"] as a conditional check within the confirmation flow. This would prevent duplicate checks outside the scope of this module, avoiding unintended test failures in other modules. The confirmation flow could then be explicitly tested in this module's own test suite by directly calling the relevant methods.

It’s also odd that the "Detect unreleased dependencies" check is failing, especially since the confirmation_wizard module was already merged. 

Could you rebase the PR? thanks for the contribution!

geomer198:
@HviorForgeFlow Could you check changes?

test-requirements.txt
```diff
@@ -1 +1,2 @@
 odoo_test_helper
+odoo-addon-confirmation_wizard @ git+https://github.com/OCA/server-ux.git@refs/pull/942/head#subdirectory=setup/confirmation_wizard
```
HviorForgeFlow: Seems that this dependency is already merged. It can be removed now, most people usually link the open PR in the description to give a quick overview on the status.

Also it would have been nice to keep it in a separate PR usually called [TMP] Add blablalb: Don't merge me

purchase_order_duplicate_check/models/purchase_order.py
```diff
@@ -0,0 +1,41 @@
...
+                f"{type_}: {order_href}; date: {order.create_date.date()}; Qty: {line.product_qty};<br/>"  # noqa
+            )
+        return "".join(message_parts)
+
+    def _check_pending_order(self):
+        """Check for pending orders and trigger confirmation wizard if needed."""
+        if not tools.config["test_enable"]:
+            action = self.env["confirmation.wizard"].confirm_pending_order(self)
+            if action:
+                return action
```
HviorForgeFlow: Adding the test_enable will allow other modules to be tested properly, usually, you can see in many places of Odoo code after adding that a context is checked. So maybe you can avoid changing the config['test_enable'] to True to run your test just adding the following context to the test.

```suggestion
    def _check_pending_order(self):
        """Check for pending orders and trigger confirmation wizard if needed."""
        if not tools.config["test_enable"] or self._context.get('test_purchase_duplicate_check') :
            action = self.env["confirmation.wizard"].confirm_pending_order(self)
            if action:
                return action
        return {}        
```

HviorForgeFlow (changes requested):
Code review, tests are fixed, CI green, just some minor comment and it is ready to go.

geomer198:
@HviorForgeFlow Could you check again?

HviorForgeFlow (approved):
Thanks for the quick fix, do you want I merge it like this or are you going to squash the commits?

geomer198:
@HviorForgeFlow Squashed the commits together, it's done!

PR #2540, opened 2025-02-04
[16.0][IMP]purchase_discount: add name in th for inheritance

flalexg:
Added to the th discount a name in order to do an inheritence easily, otherwise, its very difficult to do an inheritance.


pedrobaeza (approved):

dalonsofl (approved):

pedrobaeza:
Please change the commit message to `[IMP] purchase_discount: add name in th for inheritance`.

- Don't put the version in the commit message (but yes on the PR title).
- This is not adding a module, but enhancing an existing one.

flalexg:
@pedrobaeza changed! Thank you!

PR #2542, opened 2025-02-05
[16.0] [IMP] purchase_discount: apply default discount when no seller found

Borruso:


legalsylvain (changes requested):
It's a breaking change.
if a supplier doesn't make any discount for a product ( discount 0%) it will be put the default partner discount (X%), so the amount will be wrong.

PR #2543, opened 2025-02-05
[16.0] [IMP] purchase_triple_discount: apply default discount when no seller found

Borruso:


legalsylvain (changes requested):
same remarks as in https://github.com/OCA/purchase-workflow/pull/2542#pullrequestreview-2595775891

PR #2554, opened 2025-02-12
[18.0][MIG] purchase_sale_link_by_origin: Migration to version 18.0

Ricardo-MC:


purchase_sale_link_by_origin/models/sale_order.py
```diff
@@ -0,0 +1,20 @@
...
+
+    @api.depends("origin")
+    def _compute_purchase_order_count(self):
+        return super()._compute_purchase_order_count()
+
+    def _get_purchase_orders(self):
+        po_related = self.env["purchase.order"]
+        for order in self:
+            po_related |= self.env["purchase.order"].search(
+                [("origin", "ilike", order.name)]
```
AurelijaNorvaisaite:
```suggestion
                [("origin", "=", order.name)]
```
With ilike, all similar purchase order will also be found. Case:
Sale Order S000007, Purchase order with Source document S0000712 will also be included.

Ricardo-MC: Hi @AurelijaNorvaisaite, Thank you for your comment. The suggested correction has been made.

purchase_sale_link_by_origin/tests/test_sale_purchase.py
```diff
@@ -0,0 +1,75 @@
+# Copyright 2023 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import Command
+from odoo.tests.common import TransactionCase
+
+
+class TestSalePurchaseOrder(TransactionCase):
```
dalonsod: Could you use `BaseCommon` instead, as migration guidelines suggests?

dalonsod (changes requested):
See minor comment

purchase_sale_link_by_origin/tests/test_sale_purchase.py
Ricardo-MC: @dalonsod, hi Thanks for pointing that out! I’ve updated the code to use BaseCommon as per the migration guidelines.

HviorForgeFlow (approved):

marcelsavegnago (approved):

PR #2566, opened 2025-02-16
[16.0][MIG] purchase_order_approval_block

etobella:
Standard Migration


kluna1998 (approved):

purchase_order_approval_block/models/purchase_order.py
```diff
@@ -0,0 +1,65 @@
...
+    )
+
+    @api.depends("approval_block_id")
+    def _compute_approval_blocked(self):
+        for rec in self:
+            rec.approval_blocked = rec.approval_block_id
+
+    @api.model_create_multi
+    def create(self, mvals):
+        records = super(PurchaseOrder, self).create(mvals)
```
luisDIXMIT, line 25: When I try to create a purchase order with an approval block reason, that field is automatically unset, and no message is posted. What is wrong?

purchase_order_approval_block/models/purchase_order.py
```diff
@@ -0,0 +1,65 @@
...
+        string="Approval Block Reason",
+    )
+    approval_blocked = fields.Boolean(
+        compute="_compute_approval_blocked",
+    )
+
+    @api.depends("approval_block_id")
+    def _compute_approval_blocked(self):
+        for rec in self:
+            rec.approval_blocked = rec.approval_block_id
```
luisDIXMIT, line 21: Is it correct to assign the value of a Many2one field to a Boolean field?

etobella: Yes, the cache_to_value will transform it to bool for you

purchase_order_approval_block/models/purchase_order.py
etobella: Mmmh, this is weird. I found the same issue on runboat :disappointed: 

Can you test on a local environment with minimal modules please? I try to find if this is a problem that comes from this module or from something else

luisDIXMIT: Sure

luisDIXMIT: After debugging the issue, I realized that the problem occurs because approval_block_id depends on approval_blocked to be set as read-only, and approval_blocked depends on approval_block_id to be set. If you always set approval_blocked to False or remove the readonly attribute in the view to make approval_block_id always editable, the issue is fixed.

I also discovered that, in this case, approval_block_id is not passed as a parameter in the create method. Sorry, but I haven't found why this situation causes a conflict, but I hope this information will be useful.

etobella: I think we can add a force_save="1" in the view in order to fix it :thinking:

luisDIXMIT: It works.

luisDIXMIT (approved):
LGTM!

PR #2568, opened 2025-02-17
[17.0][ADD] procurement_purchase_requisition_generation: New module

victoralmau:
New module

@Tecnativa TT54940


procurement_purchase_requisition_generation/models/stock_rule.py
```diff
@@ -0,0 +1,89 @@
...
+
+from odoo import SUPERUSER_ID, api, models
+from odoo.tools import config
+
+
+class StockRule(models.Model):
+    _inherit = "stock.rule"
+
+    def _make_pr_get_domain(self, company_id, values):
+        group = values.get("group_id") or False
```
pedrobaeza: No need of the other part, as by default it returns `None`, which is a falsy value. 

```suggestion
        group = values.get("group_id")
```

procurement_purchase_requisition_generation/readme/DESCRIPTION.md
```diff
@@ -0,0 +1 @@
+This module restores the purchase_requisition functionality of v15 in which purchas.requisition is generated instead of purchase.order.
```
pedrobaeza: This is a user description, so better to avoid technical names

```suggestion
This module restores the v15 functionality in which you can configure products for generating purchase tenders instead of draft purchase orders.
```

procurement_purchase_requisition_generation/readme/USAGE.md
```diff
@@ -0,0 +1,6 @@
+#. Go to Purchases > Products > Products.
+#. Create a Storable product, set the 'Buy' route and some vendor in the 'Purchase' tab.
+#. Click on the 'Replenish' button.
+#. The 'Buy' route and 'Vendor' field will be defined automatically.
```
pedrobaeza, line 2: I don't get this part of the explanation, but I don't see it relevant for this module. The important part that I'm missing is that you have to configure this by product, and you have simply replaced the feature by always creating requisitions.

procurement_purchase_requisition_generation/tests/test_procurement_purchase_requisition_generation.py
```diff
@@ -0,0 +1,97 @@
...
+        cls.productA.write(
+            {
+                "seller_ids": [(0, 0, {"partner_id": cls.vendor.id})],
+                "route_ids": [(6, 0, [cls.buy.id, cls.mto.id])],
+            }
+        )
+        cls.productB.write(
+            {
+                "route_ids": [(6, 0, [cls.buy.id, cls.mto.id])],
+            }
```
pedrobaeza: Remember to not put trailing commas for avoiding extra lines:

```suggestion
            {"route_ids": [(6, 0, [cls.buy.id, cls.mto.id])]}
```

procurement_purchase_requisition_generation/wizards/product_replenish.py
```diff
@@ -0,0 +1,26 @@
+# Copyright 2025 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class ProductReplenish(models.TransientModel):
+    _inherit = "product.replenish"
+
+    def _get_record_to_notify(self, date):
```
pedrobaeza, line 10: Why do you need to override this? The usual use case flow is through MTO from sales order (anyway, this one should work also of course).

pedrobaeza (changes requested):

procurement_purchase_requisition_generation/readme/USAGE.md
victoralmau: Ok, sorry, I thought this behavior was something global, I will add then the `purchase_requisition` field to the product to decide if you want to create one thing or another (requisition or purchase).

procurement_purchase_requisition_generation/wizards/product_replenish.py
victoralmau: The process of a sales order will work without this with the `_run_buy()` method, but it is required in the Replenish process.
ejemplo

procurement_purchase_requisition_generation/readme/USAGE.md
```diff
@@ -0,0 +1,7 @@
+#. Go to Purchases > Products > Products.
+#. Create a Storable product, set the 'Buy' route and some vendor in the 'Purchase' tab.
+#. Set the Procurement: 'Propose a call for tenders' option in the 'Reordering' section of the 'Purchase' tab.
```
pedrobaeza: These steps should go in CONFIGURE.md

pedrobaeza (approved):

procurement_purchase_requisition_generation/wizards/product_replenish.py
```diff
@@ -0,0 +1,26 @@
...
+        if requisition_line._name == "purchase.requisition.line":
+            action = self.env.ref("purchase_requisition.action_purchase_requisition")
+            r = requisition_line.requisition_id
+            return [
+                {
+                    "label": r.display_name,
+                    "url": f"#action={action.id}&id={r.id}&model={r._name}",
+                }
+            ]
+        return super()._get_replenishment_order_notification_link(requisition_line)
```
carlos-lopez-tecnativa: I see more records than expected (including unrelated products). Please check if this is working correctly.
image


image

image

procurement_purchase_requisition_generation/models/product_template.py
```diff
@@ -0,0 +1,21 @@
...
+        selection=[
+            ("rfq", "Create a draft purchase order"),
+            ("tenders", "Propose a call for tenders"),
+        ],
+        string="Procurement",
+        default="rfq",
+        help="Create a draft purchase order: Based on your product configuration,"
+        "the system will create a draft purchase order.Propose a call for tender : "
+        "If the 'purchase_requisition' module is installed and this option is "
+        "selected, the system will create a draft call for tender.",
```
carlos-lopez-tecnativa:
```suggestion
        help="Create a draft purchase order: Based on your product configuration,"
        "the system will create a draft purchase order. \nPropose a call for tender : "
        "If the 'purchase_requisition' module is installed and this option is "
        "selected, the system will create a draft call for tender.",
```

procurement_purchase_requisition_generation/models/purchase_requisition.py
```diff
@@ -0,0 +1,12 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class PurchaseRequisition(models.Model):
+    _inherit = "purchase.requisition"
+
+    procurement_group_id = fields.Many2one(
+        comodel_name="procurement.group", string="Procurement Group"
```
carlos-lopez-tecnativa: Could it be
```suggestion
        comodel_name="procurement.group", string="Procurement Group", copy=False
```

carlos-lopez-tecnativa (changes requested):
Tested functionally.

I have a question related to dropshipping.

Is linking the sale order with the purchase order generated from a requisition out of scope for this module?

I'm just thinking out loud to avoid needing a glue module like
https://github.com/odoo/odoo/tree/15.0/addons/purchase_requisition_stock_dropshipping

**Procurement=Create a draft purchase order**
image

**Procurement=Propose a call for tenders.** (No picking related to the sale order)
image

procurement_purchase_requisition_generation/wizards/product_replenish.py
victoralmau: You are right, thank you, it is solved.

victoralmau:
Reviewing the dropshipping use case I confirm that an extension for sales would be necessary to take into account this casuistry: create the `sale_line_id` field in `purchase.requisition.line` and set it when it is created from a `sale.order` + transfer that value (`sale_line_id`) to the purchase order line that would be generated.

procurement_purchase_requisition_generation/models/stock_rule.py
```diff
@@ -0,0 +1,85 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from collections import defaultdict
+
+from odoo import SUPERUSER_ID, api, models
+
+
+class StockRule(models.Model):
+    _inherit = "stock.rule"
+
+    def _make_pr_get_domain(self, company_id, values):
```
carlos-lopez-tecnativa: Please use other function name to avoid conflict with `purchase_request` module
https://github.com/OCA/purchase-workflow/blob/1ff6fee1645b9fc42596d2509f9f0268c54e9529/purchase_request/models/stock_rule.py#L49

```suggestion
    def _make_purchase_requisition_domain(self, company_id, values):
```
image

procurement_purchase_requisition_generation/models/stock_rule.py
```diff
@@ -0,0 +1,85 @@
...
+            requistion_line_values = []
+            for procurement in procurements:
+                line_values = rules[0]._prepare_purchase_requisition_line(
+                    procurement.product_id,
+                    procurement.product_qty,
+                    procurement.product_uom,
+                    procurement.values,
+                )
+                requistion_line_values.append((0, 0, line_values))
+            requisition.write({"line_ids": requistion_line_values})
```
carlos-lopez-tecnativa: Is it possible to split this function so that all the code related to the creation of the purchase requisition is moved to a separate function, such as `def create_purchase_requisition`?

carlos-lopez-tecnativa (changes requested):

procurement_purchase_requisition_generation/models/stock_rule.py
pedrobaeza: The hooks should be only for preparing values. I don't get the added value for doing that split. Can you explain?

carlos-lopez-tecnativa: It's OK for me, it's just a wishlist. I tried to split the complex functions, that's all. Please ignore my previous comment.

pedrobaeza: Well, a method with less than 40 lines and 2 deep levels can't be considered complex IMO.

carlos-lopez-tecnativa (approved):
LGTM

PR #2569, opened 2025-02-17
[17.0][FIX] purchase_request: Add Markup to messages

victoralmau:
Add Markup to messages

Remove warning log
`WARNING odoo odoo.addons.mail.models.mail_thread: Posting HTML message using body_is_html=True, use a Markup object instead (user: 1)`

@Tecnativa

### DIFF
--- a/purchase_request/models/purchase_order.py
+++ b/purchase_request/models/purchase_order.py
@@ -1,7 +1,10 @@
 # Copyright 2018-2019 ForgeFlow, S.L.
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl-3.0)
 
+from markupsafe import Markup
+
 from odoo import _, api, exceptions, fields, models
+from odoo.tools import html_escape
 
 
 class PurchaseOrder(models.Model):
@@ -29,13 +32,13 @@ class PurchaseOrder(models.Model):
                 "<li><b>%(prl_name)s</b>: Ordered quantity %(prl_qty)s %(prl_uom)s, "
                 "Planned date %(prl_date_planned)s</li>"
             ) % {
-                "prl_name": line["name"],
+                "prl_name": html_escape(line["name"]),
                 "prl_qty": line["product_qty"],
                 "prl_uom": line["product_uom"],
                 "prl_date_planned": line["date_planned"],
             }
         message += "</ul>"
-        return message
+        return Markup(message)
 
     def _purchase_request_confirm_message(self):
         request_obj = self.env["purchase.request"]
@@ -62,7 +65,6 @@ class PurchaseOrder(models.Model):
                 request.message_post(
                     body=message,
                     subtype_id=self.env.ref("mail.mt_comment").id,
-                    body_is_html=True,
                 )
         return True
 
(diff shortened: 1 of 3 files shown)


pedrobaeza:
It seems this is not adding value as the parameter is_html_body=True is doing the same. Am I wrong?

victoralmau:
I have updated the commit message and the description to indicate that it is to avoid the warning log.

purchase_request/models/purchase_order.py
```diff
@@ -214,7 +214,7 @@ def _purchase_request_confirm_done_message_content(self, message_data):
             "product_uom": message_data["product_uom"],
         }
         message += "</ul>"
-        return message
+        return Markup(message)
```
StefanRijnhart, line 218: While you are fixing the treatment of html in this module, can you apply `odoo.tools.misc.html_escape` to all string values in message_data above? And the same in the other files.

victoralmau: Sorry, I don't understand the need to use html_escape, can you explain a use case (for understanding) where it is necessary to use Markup + html_escape?

StefanRijnhart: When you cast text as Markup, you are claiming that its contents are valid and safe HTML. Therefore, you need to escape all the raw strings you insert into it.
image
image

victoralmau: But, do we need to control that? I think we should not, there are many sites in odoo core where they use Markup() and not html_escape, if for example there is “incorrect html code” in the partner name the same thing would also happen, so IMO if there is an incorrect code somewhere (in the example you indicate in the product name) what should be done is to correct the name, but it is just my opinion.

StefanRijnhart: Do you have an example where a partner name is inserted in a HTML markup message without escaping?

pedrobaeza: I would say more to use `html2text" than escaping, isn't it? In any case, I think it's a common security practice to avoid any kind of HTML injection.

pedrobaeza:
In general about the PR, isn't the parameter `body_is_html=True` doing the same as you are doing manually here?

purchase_request/models/purchase_order.py
StefanRijnhart: For the record, I'm calling it 'escape' because the method for this provided by Odoo is `html_escape` which is an alias for Markupsafe.escape: https://github.com/odoo/odoo/blob/18.0/odoo/tools/misc.py#L1288

StefanRijnhart: BTW another option is to use Markup formatting, which applies escaping itself: https://pypi.org/project/MarkupSafe/. You can then concatenate the Markup safe fragments that combine the bullet list in a message.

pedrobaeza: `html2text` is also a method in Odoo in `tools`. I didn't know about the `Markupsafe.escape` before.

victoralmau:
I think it is important to remember the reason for this PR, to remove the warning log for using `body_is_html=True` https://github.com/odoo/odoo/blob/17.0/addons/mail/models/mail_thread.py#L2118 https://github.com/odoo/odoo/blob/1c896c5da71b58b382269fe6cca556f4a61e4d5d/addons/mail/models/mail_thread.py#L2200 Actually done is to use `Markup` so I understand that the change is correct and nothing extra is needed.

pedrobaeza:
OK, didn't know, but you can now assure no injection through data fields with the advices given here.

StefanRijnhart:
Thanks for the update. You can squash commits because they are closely related.

StefanRijnhart (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #2577, opened 2025-02-21
[18.0][FIX] purchase_request: Add Markup to messages

victoralmau:
FWP from 17.0: https://github.com/OCA/purchase-workflow/pull/2569

Add Markup to messages

Remove warning log
`WARNING odoo odoo.addons.mail.models.mail_thread: Posting HTML message using body_is_html=True, use a Markup object instead (user: 1)`

@Tecnativa


pedrobaeza (approved):
Fast-tracking it as being a fw-port:

/ocabot merge patch

PR #2579, opened 2025-02-24
[16.0][FIX] purchase_discount: Post-install test + fallback to load CoA

pedrobaeza:
Since odoo/odoo@d0342c8, the default existing company is not getting a CoA automatically, provoking than the current tests fail with error:

```
odoo.exceptions.UserError: No journal could be found in company My Company (San Francisco) for any of those types: purchase
```

provoked by the lack of a CoA installed.

Thus, we put tests post-install for being sure localization modules are installed, the same as AccountTestInvoicingCommon does, but we don't inherit from it, as it creates an overhead creating 2 new companies and loading their CoA, and some more stuff, while we don't need all of that.

Besides, if you don't have `l10n_generic_coa` installed, you can't use another CoA (like `l10n_es`) easily, so we put little code to select the first available CoA.

@Tecnativa

### DIFF
--- a/purchase_discount/tests/test_purchase_discount.py
+++ b/purchase_discount/tests/test_purchase_discount.py
@@ -3,16 +3,26 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import fields
-from odoo.tests.common import Form, TransactionCase
+from odoo.tests.common import Form, TransactionCase, tagged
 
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Tested on integration tests and working:

/ocabot merge patch

pedrobaeza:
@legalsylvain the error is very weird with `purchase_triple_discount` and I'm not able to reproduce it locally. Do you know what's going on?

pedrobaeza:
Already solved.

PR #2581, opened 2025-02-25
[16.0][FIX] purchase_landed_cost + purchase_allowed_product + purchase_order_uninvoiced_amount: Post-install test + fallback to load CoA

victoralmau:
Since odoo/odoo@d0342c8, the default existing company is not getting a CoA automatically, provoking than the current tests fail with the error:

`odoo.exceptions.UserError: No journal could be found in company My Company (San Francisco) for any of those types: sale`

Thus, we put tests post-install for being sure localization modules are installed, the same as AccountTestInvoicingCommon does, but we don't inherit from it, as it creates an overhead creating 2 new companies and loading their CoA and some more stuff, while we don't need all of that.

Besides, if you don't have `l10n_generic_coa` installed, you can't use another CoA (like `l10n_es`) easily, so we put little code to select the first available CoA.

@Tecnativa

### DIFF
--- a/purchase_allowed_product/tests/test_purchase_allowed_product.py
+++ b/purchase_allowed_product/tests/test_purchase_allowed_product.py
@@ -3,24 +3,36 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 
+from odoo.tests import tagged
 from odoo.tests.common import Form, TransactionCase
 
 
+@tagged("-at_install", "post_install")
 class TestPurchaseAllowedProduct(TransactionCase):
-    def setUp(self):
-        super().setUp()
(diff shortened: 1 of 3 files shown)


victoralmau:
Locked by `purchase_triple_discount` https://github.com/OCA/purchase-workflow/pull/2579#issuecomment-2681177175

rvalyi:
@victoralmau @pedrobaeza seems like Odoo is doing something about it: https://github.com/odoo/odoo/pull/200997

pedrobaeza (approved):
@rvalyi kind of. They are fixing upper versions, after the merge of the generic CoA into account.

For 16.0, it seems we are going to continue with this behavior.

Merging this one meanwhile, as it will work the same even if a core patch is applied.

/ocabot merge patch

PR #2587, opened 2025-02-27
[18.0][MIG] purchase_packaging_default: Migration to 18.0

YassineTal:
Migration of addon from v16.0 to v18.0
Only changes needed were:
Change to the purchase res.config.settings inheritance (block instead of div).
Use purchase line company_id instead of env.user.company_id (error prone in case of a multi-company env).


rousseldenis (changes requested):
@YassineTal Thanks for this.

You should keep commits history.

Please follow this: 

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-18.0#how-to

PR #2599, opened 2025-03-09
[16.0][REF] purchase_triple_discount: Consolidate discount in std field

pedrobaeza:
Continuation of #2317 and following https://github.com/OCA/account-invoicing/pull/1840

Try to fix the current red CI state.

-----

Until now, the triple discount feature did use the standard Odoo field as the first discount field, adding only extra fields for the second and the third discount. This implied we had to override any function using discount to consider the other discounts properly.

By adding an extra field discount1 to store the first discount, it allows to redefine the standard discount field to a computed field that will consolidate the triple discount from the other fields, and avoid the need to redefine anything relying on the discount field as it will already consider the triple discounts.

@Tecnativa

### DIFF
--- a/purchase_discount/tests/test_purchase_discount.py
+++ b/purchase_discount/tests/test_purchase_discount.py
@@ -199,12 +199,12 @@ class TestPurchaseOrder(TransactionCase):
         line = invoice.invoice_line_ids.filtered(
             lambda x: x.purchase_line_id == self.po_line_1
         )
-        self.assertEqual(line.discount, 50)
+        self.assertAlmostEqual(line.discount, 50)
         line = invoice.invoice_line_ids.filtered(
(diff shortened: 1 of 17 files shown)


pedrobaeza:
As suspected, both sets of changes are needed, so let's merge this for fixing the CI that is blocking other PRs. @legalsylvain please check later if you see any problem due to your initial reluctance. I have put a TODO to move the mixin to the upper module.

/ocabot merge patch

legalsylvain:
> merge this for fixing the CI that is blocking other PRs. @legalsylvain please check later if you see any problem due to your initial reluctance

Do you talk about mixin refactoring ?
It's a Nice to have indeed. The most important is to have green CI in all repos.

Thanks for your work.

PR #2600, opened 2025-03-09
[16.0][FIX] purchase_discount: Relax float comparison

pedrobaeza:
Having conversions with currency rate, leads to a situation where what we want to test is correct, but with a rounding error, so let's use `assertAlmostEqual` to compensate it.

@Tecnativa


pedrobaeza:
We need both patchs that are in #2599, so closing this one.

PR #2605, opened 2025-03-10
[16.0][IMP] purchase_request_tier_validation: Hide Confirm button if the validation is pending or rejected.

victoralmau:
Hide Confirm button if the validation is pending or rejected.

**Before**
antes

**After**
despues

@Tecnativa TT55409


pedrobaeza (approved):

sergio-teruel (approved):
If you do it in this module you would have to do it in other modules like ‘sale_order_tier_validation’.

PR #2611, opened 2025-03-17
[16.0][IMP] purchase_request: Define analytic_distribution field hidden by default

victoralmau:
Define `analytic_distribution` field hidden by default similar to line tree view (https://github.com/OCA/purchase-workflow/blob/b4670f4f387c96648a8609111a6758f71d5e595d/purchase_request/views/purchase_request_line_view.xml#L46)

Please @carlos-lopez-tecnativa can you review it?

@Tecnativa TT55563

### DIFF
--- a/purchase_request/views/purchase_request_view.xml
+++ b/purchase_request/views/purchase_request_view.xml
@@ -194,6 +194,7 @@
                                         widget="analytic_distribution"
                                         groups="analytic.group_analytic_accounting"
                                         options="{'product_field': 'product_id', 'business_domain': 'purchase_order'}"
+                                        optional="hide"
                                     />
                                     <field name="date_required" />
                                     <field name="estimated_cost" widget="monetary" />


pedrobaeza:
"Optional" wording is misleading. It's simply hidden by default.

carlos-lopez-tecnativa (approved):

purchase_request/views/purchase_request_view.xml
```diff
@@ -194,6 +194,7 @@
                                         widget="analytic_distribution"
                                         groups="analytic.group_analytic_accounting"
                                         options="{'product_field': 'product_id', 'business_domain': 'purchase_order'}"
+                                        optional="hide"
```
pedrobaeza, line 197: I'm seeing that in invoices and other documents, the field can be hidden, but it's shown by default.

```suggestion
                                        optional="show"
```

victoralmau: Already exists in the tree view of the lines `optional=hide` (https://github.com/OCA/purchase-workflow/blob/b4670f4f387c96648a8609111a6758f71d5e595d/purchase_request/views/purchase_request_line_view.xml#L46), and in `purchase` too https://github.com/odoo/odoo/blob/1d605fe49aa3a6ed45943312627f0d56b1fb7aee/addons/purchase/views/purchase_views.xml#L241

pedrobaeza: OK, let's go with that approach.

pedrobaeza (approved):
/ocabot merge patch

PR #1439, opened 2022-04-27
[14.0][FIX] purchase_request: merging two lines had wrong quantities

dreispt:
For example, a Request has tow lines for the same product.
One with quantity 2 and another with quantity 3.
When these lines are merged into a single PO line,
final quantity was 4 instead of the expected 5.

The problem could be tracked down to the _calc_new_qty method.
Undoing a change in https://github.com/OCA/purchase-workflow/commit/8b527c4efaaf56ea33621abf81a2ddc1e0e9d27d
resolved this basic case.

Closes #1271

PR #2615, opened 2025-03-17
[16.0][FIX] bring fix from https://github.com/OCA/purchase-workflow/pull/1439

Miguel-Granda1121:


Miguel-Granda1121:
Please ignore this PR, the correct one is https://github.com/OCA/purchase-workflow/pull/2616

PR #2622, opened 2025-03-19
[16.0][FIX] purchase_manual_delivery: Do not use demo data in tests

victoralmau:
Do not use demo data in tests

Using demo data from core modules can have side effects, e.g. the product used is linked to company A, and the record you are trying to create (purchase order) is from another company causing confusing permission errors.

@Tecnativa


pedrobaeza (approved):
Improvement for not having problems in integration tests, so fast-tracking, as it doesn't affect business code:

/ocabot merg patch

PR #2619, opened 2025-03-18
[18.0][MIG] purchase_advance_payment: Migration to 18.0

rjaraspearhead:

PR #2626, opened 2025-03-26
[18.0][MIG] purchase_advance_payment

ThiagoMForgeFlow:
Supersede:
* https://github.com/OCA/purchase-workflow/pull/2619

@ForgeFlow


LoisRForgeFlow (changes requested):
@ThiagoMForgeFlow I cannot uncheck this setting:

image

Everytime I uncheck and save, it keeps enabled.

could you check?

purchase_advance_payment/views/account_payment_views.xml
```diff
@@ -14,33 +14,20 @@
...
-                    name="state_not_sent"
-                    domain="[('is_move_sent', '=', False)]"
-                />
-            </filter>
-            <filter name="reconciled" position="after">
-                <filter
-                    string="Not Reconciled"
-                    name="reconciled"
-                    domain="[('is_reconciled', '=', False)]"
-                />
```
LoisRForgeFlow: This filter is still needed I think

LoisRForgeFlow:
I just realized that this change from v16 was not fwported, could add it here? https://github.com/OCA/purchase-workflow/pull/2629

purchase_advance_payment/views/account_payment_views.xml
ThiagoMForgeFlow: Filter added!

ThiagoMForgeFlow:
> I just realized that this change from v16 was not fwported, could add it here? #2629

Added, Thanks!

ThiagoMForgeFlow:
> @ThiagoMForgeFlow I cannot uncheck this setting:
> 
> image
> 
> Everytime I uncheck and save, it keeps enabled.
> 
> could you check?

I see that is a config_parameter field, and when the module is installed, there are discrepancies between the system parameter value and the UI from the settings. I see that when the value is False odoo removes the system parameter. If we remove the record from the data folder, the error is fixed.

LoisRForgeFlow (approved):
LGTM :ok_hand:

DavidJForgeFlow (approved):

LoisRForgeFlow:
@HviorForgeFlow could you merge this one?

PR #2635, opened 2025-04-07
[FIX] purchase_manual_delivery: use _product_purchase_domain on _compute_draft_quantity_count

AlexPForgeFlow:
Raising error wrong domain
`ValueError: Invalid field purchase.order.line.product_tmpl_id in leaf ('product_tmpl_id', 'in', [30])`

Method _compute_draft_quantity_count was calling wrong domain function _product_domain, instead of _product_purchase_domain


HviorForgeFlow (approved):

AaronHForgeFlow (approved):
LGTM!

PR #2572, opened 2025-02-17
[14.0][FIX] purchase_advance_payment: fix payment_state

WesleyOliveira98:
Having an invoice does not necessarily mean that the payment is partial. Changed to when there is `account_payment_ids` or there is a paid amount on the invoice (`invoice_paid_amount`)

Creating a draft invoice before the changes:
image

Creating a draft invoice after the changes in this PR:
image

When we confirm and pay the invoice, the status will be updated:
image

Ticket: HT00980

cc: @marcelsavegnago @kaynnan

PR #2637, opened 2025-04-08
[15.0][FIX] purchase_advance_payment: fix advance_payment_status with bill

WesleyOliveira98:
Port from #2572 and #2631 

cc: @marcelsavegnago @kaynnan @LoisRForgeFlow @dreispt

### DIFF
--- a/purchase_advance_payment/models/purchase_order.py
+++ b/purchase_advance_payment/models/purchase_order.py
@@ -86,7 +86,7 @@ class PurchaseOrder(models.Model):
                 invoice_paid_amount += inv.amount_total - inv.amount_residual
             amount_residual = order.amount_total - advance_amount - invoice_paid_amount
             payment_state = "not_paid"
-            if mls or order.invoice_ids:
+            if mls or not order.currency_id.is_zero(invoice_paid_amount):
                 has_due_amount = float_compare(
                     amount_residual, 0.0, precision_rounding=order.currency_id.rounding
                 )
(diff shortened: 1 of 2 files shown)


rvalyi (approved):

marcelsavegnago (approved):

LoisRForgeFlow (changes requested):
Just singnaling that i want to review

purchase_advance_payment/models/purchase_order.py
```diff
@@ -86,7 +86,7 @@ def _compute_purchase_advance_payment(self):
                 invoice_paid_amount += inv.amount_total - inv.amount_residual
             amount_residual = order.amount_total - advance_amount - invoice_paid_amount
             payment_state = "not_paid"
-            if mls or order.invoice_ids:
+            if mls or invoice_paid_amount != 0.0:
```
LoisRForgeFlow: I finally tested it, it is indeed correct and the test case is good :+1: 

Just one detail, could you use the `is_zero` method of res.currency instead of doing a raw comparison?

Thanks!

LoisRForgeFlow (changes requested):

purchase_advance_payment/models/purchase_order.py
WesleyOliveira98: Hello @LoisRForgeFlow, thanks a lot for reviewing and testing it! I change to use the `currency_id.is_zero` method as you suggested, this way is definitely better for precision handling. Could you review again, please?

kaynnan (approved):

purchase_advance_payment/models/purchase_order.py
LoisRForgeFlow: Thanks!

PR #2641, opened 2025-04-09
[18.0][MIG] purchase_discount: Migration to 18.0

arantxa-s73:
Module migrated to version 18.0


pedrobaeza:
AFAIK, this one is not needed, as Odoo has already included the discount field in purchase orders.

Issue #2644, opened 2025-04-10
purchase_lot: Removing sale_order_lot_selection dependency


Tisho99:
**Is your feature request related to a problem?**
The purchase_lot module depends on sale_order_lot_selection for the lots being propagated from sales to purchases. This is a very strong dependency and the module could be more versatile without it (also mentioned in https://github.com/OCA/purchase-workflow/issues/1538). 

**Describe the solution you'd like**
I'd like to replace the sale_order_lot_selection dependence for purchase_stock and stock_restrict_lot. With this, we would have lots in purchase order lines with less module dependences.

I think removing the readonly of the lot_id field is also useful to let users filling and creating the lot from the purchase order, now the module only propagates the lots from sales to purchases

**Describe alternatives you've considered**
I have considered creating another bridge module that depends sale_order_lot_selection and the refactored purchase_lot, but i think it would not be neccesary because all the functions the module inherits are created in the purchase_stock module.

Tisho99:
As I saw in the issue https://github.com/OCA/purchase-workflow/issues/1538, you agreed to remove the sale_order_lot_selection dependency from the purchase_lot module.

Do you still agree? We can do the PR.

Would you also agree to make the lot field editable?

Thank you for your time.

Tisho99:
Hello @florian-dacosta 

As I saw in the issue https://github.com/OCA/purchase-workflow/issues/1538, you agreed to remove the sale_order_lot_selection dependency from the purchase_lot module.

Do you still agree? We can do the PR.

Would you also agree to make the lot field editable?

Thank you for your time.

PR #2646, opened 2025-04-14
[16.0][IMP] purchase_lot: remove sale_order_lot_selection dependency

Tisho99:
Explained in the https://github.com/OCA/purchase-workflow/issues/2644 issue

T-7552

### DIFF
--- a/purchase_lot/README.rst
+++ b/purchase_lot/README.rst
@@ -28,8 +28,11 @@ Purchase Lot
 
 |badge1| |badge2| |badge3| |badge4| |badge5|
 
-This module propagate lot from sale orders to purchase orders.
-It allows to buy specific lot for an sale order (may be usefull for fully configurable products)
+This module adds the lot_id field in the purchase order lines,
+and propagates it to stock pickings in the order confirmation.
+
+This module can also be used along with sale_order_lot_selection to propagate lot from sale orders to purchase orders.
+It allows to buy specific lot for a sale order (may be useful for fully configurable products)
 
 **Table of contents**
 
--- a/purchase_lot/__manifest__.py
+++ b/purchase_lot/__manifest__.py
@@ -9,10 +9,7 @@
     "author": "Akretion, Odoo Community Association (OCA)",
     "maintainers": ["florian-dacosta"],
     "website": "https://github.com/OCA/purchase-workflow",
-    "depends": [
-        "purchase",
-        "sale_order_lot_selection",
-    ],
+    "depends": ["purchase_stock", "stock_restrict_lot"],
     "data": [
         "views/purchase_order_view.xml",
     ],
--- a/purchase_lot/models/purchase_order_line.py
+++ b/purchase_lot/models/purchase_order_line.py
@@ -8,9 +8,8 @@ from odoo import api, fields, models
 class PurchaseOrderLine(models.Model):
     _inherit = "purchase.order.line"
 
-    lot_id = fields.Many2one(
-        "stock.lot", string="Serial Number", readonly=True, copy=False
-    )
+    lot_id = fields.Many2one("stock.lot", string="Serial Number", copy=False)
+    tracking = fields.Selection(related="product_id.tracking")
 
     @api.model
     def _prepare_purchase_order_line_from_procurement(
--- a/purchase_lot/pyproject.toml
+++ b/purchase_lot/pyproject.toml
@@ -1,0 +1,3 @@
+[build-system]
+requires = ["whool"]
+build-backend = "whool.buildapi"
--- a/purchase_lot/readme/DESCRIPTION.rst
+++ b/purchase_lot/readme/DESCRIPTION.rst
@@ -1,2 +1,5 @@
-This module propagate lot from sale orders to purchase orders.
-It allows to buy specific lot for an sale order (may be usefull for fully configurable products)
+This module adds the lot_id field in the purchase order lines,
+and propagates it to stock pickings in the order confirmation.
+
+This module can also be used along with sale_order_lot_selection to propagate lot from sale orders to purchase orders.
+It allows to buy specific lot for a sale order (may be useful for fully configurable products)
--- a/purchase_lot/static/description/index.html
+++ b/purchase_lot/static/description/index.html
@@ -1,4 +1,3 @@
-<?xml version="1.0" encoding="utf-8"?>
 <!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
 <html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
 <head>
@@ -9,10 +8,11 @@
 
 /*
 :Author: David Goodger (goodger@python.org)
-:Id: $Id: html4css1.css 8954 2022-01-20 10:10:25Z milde $
+:Id: $Id: html4css1.css 9511 2024-01-13 09:50:07Z milde $
 :Copyright: This stylesheet has been placed in the public domain.
 
 Default cascading style sheet for the HTML output of Docutils.
+Despite the name, some widely supported CSS2 features are used.
 
 See https://docutils.sourceforge.io/docs/howto/html-stylesheets.html for how to
 customize this style sheet.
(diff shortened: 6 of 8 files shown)


florian-dacosta (approved):
Hello @Tisho99 
Thanks for the PR, I indeed still think sale_order_lot_selection should not be a dependency. 
Although I am a bit bothered because this enter in conflict with another PR : https://github.com/OCA/purchase-workflow/pull/2247

I approve as the code review seems ok and the change legit, but I'd like to wait at least for @metaminux to check and give its thoughts.

metaminux:
Hello @florian-dacosta and @Tisho99

I'm answering on both #2247 and #2646 to be sure everyone is OK.

I totally agree with @Tisho99's vision.

The underlying concept is to have control on lot / serial numbers in stock moves and that's what **stock_restrict_lot** is offering.
AFAIK, for the moment, only **sale_order_lot_selection** is triggering the restriction on moves (setting the `restrict_lot_id` field)
From the beginning, **purchase_lot** is aiming to get the lot from the sale order lines that created the purchase... maybe it's not enough :)

We could imagine a module being the equivalent of **sale_order_lot_selection** on purchase side (maybe the need is not so common, but why not)
Let's call it **purchase_order_lot_selection**, it would set restrictions on stock moves from the purchase order lines.
The same way **sale_order_lot_selection** is depending only on **sale_stock** and **stock_restrict_lot**, **purchase_order_lot_selection** should depend on **purchase_stock** and **stock_restrict_lot**.

Then, to get the link between sales and purchases (provided by native Odoo glue modules **sale_purchase** and **sale_purchase_stock**), we can have a glue module called **sale_purchase_lot_selection** or whatever.
This glue module would be what **purchase_lot** is doing nowadays, but would depend on **sale_order_lot_selection** and **purchase_order_lot_selection** with `'auto_install': True.`

In that case, `lot_id` fields in sale and purchase order lines would become computed writable fields that should be kept in sync with `restrict_lot_id` fields on related stock moves.

Two questions stay in my mind :
- how do we deal with updates occurring after confirmation of one or more objects involved (sale, purchase) ?
- where should we store the glue module ? should it stay in **OCA/purchase-workflow** as **purchase_lot** ? 

For the first question, the idea of opened PRs for "lot propagation" is to explicitly allow propagation or not and then if allowed, propagate on every objects as long as no stock move in the whole "chain of moves" is done.
For the second one, keeping it in **OCA/purchase-workflow** and only renaming it seems good to me.

What do you think ? @emiliesoutiras @Kev-Roche

florian-dacosta:
Thanks for your opinion @metaminux 
> This glue module would be what purchase_lot is doing nowadays, but would depend on sale_order_lot_selection and purchase_order_lot_selection

today purchase_lot allows to set a lot on the purchase from a from stock moves / procurements.
The main use case (may the only one in OCA) is that this propagation is made from sale_order but nothing prevent to do it from a procurement (that does not come from a sale order) or from stock move, etc.
I don't think we want to restrict this propagation logic in a module that depends on sale_order_lot_selection.

I mean, from a code point of view, the current module has no link with sale_order_lot_selection whatsoever. So I would prefer to keep the big part of the code in the base module and only what really depends on sale_order_lot_selection on the glue module.

Let's say I manufacture products, and depending on the lot I am producing, I want to buy a specific lot for my component. 
I want to be able to use the base module to create purchase orders with the restrict lot, and maybe I am not using sale order / sale_order_lot_selection.

I would rather have purchase_lot as it is (and eventually rename it if we find a better name), adding the feature of propatation
And an auto-install glue module with only what is missing to make it work well with sale_order_lot_selection.

I mean, the module is small enough and works well enough for different cases without the need of the sale_order_lot_selection dependecy. it would be a shame to move most of it in a module that depends on sale_order_lot_selection IMO. (even now, the module works well with sale_order_lot_selection, without the need of the dependency).

That said, we should not change the name in this v16 version, if any change of name should occur, I guess it would be best to keep it for v18

Tisho99:
Hello @metaminux 

I agree with you it would be more useful to have a similar module to sale_order_lot_selection.
But as @florian-dacosta  says, I think it is better to edit the purchase_lot module to do that, removing the sale_order_lot_selection dependency, improving the lot propagation, and maybe renaming it in future versions; but keeping the compatible functions with sale_order_lot_selection.

Then, this base module will cover most use cases, even propagating lots from sale orders to purchase orders, and the new glue module should be smaller.

I'll try to improve this PR, and port it to v17

Thank you both

Tisho99:
Test added

florian-dacosta:
@metaminux I understand we all agree that having 2 modules would be better, to avoid the sale dependency on the base module.
I am not sure we fully agree on how it should be splitted including the other PR, maybe we kind of are, not sure. But I think we have to move forward soon anyway, so we can have something cleaner for version 18, not yet migrated.

My statements are the following : 
- We can't rename the module merged on a stable version. I think we should do it (if we decide to rename something) for v18.
- The current module (future base module) contains presently 4 methods. 
Considering the procurement source could be a manufacturing order or an orderpoint instead of  a sale order, , all 4 methos may be usefull without the `sale` module installed, so they all belong to the module that does not depends on sale the base module.

Therefore I suggest that : 
1)  Ce merge this PR.
2) We rework the other improvement one : https://github.com/OCA/purchase-workflow/pull/2247 :
- Decide on a new module name which will depends on `sale_order_lot_selection`
- Decide on what (from this other PR) should go on which module
- Eventually add some missing code (option on company to allow or not the propagation?)
- Eventually decide if and how the base module should be rename in v18 or not.
Once decided/agreed, I may help on the PR to get it done /merged quickly enough.

What do you think ? If you are ok with the approach, can you review / approve this one ?

florian-dacosta:
Following my last comment, I think we should merge it now. 
This change is quite small and is already merged in v17. We should merge it now to avoid divergences and because this change is quite simple/logical when checking what is presently merged into the 16.0.

@metaminux I'll merge it next week if it is ok. It should not really impact https://github.com/OCA/purchase-workflow/pull/2247 on a project. But this PR should at least be lighly refactored to add a new module (could be auto-install) that manage the code dependent to sale.

metaminux:
Hello @florian-dacosta, Hello @Tisho99

Sorry, I prepared an answer but never took time to send it...

For the moment, installing purchase_lot alone would be totally useless because there is no other way than sale_order_lot_selection to set restrict_lot_id field in stock moves.
So, there is a FUNCTIONAL dependency for now if using only OCA modules... but anyone can use custom modules too, right ?
So, I totally agree with both of you : there is no need for the technical dependency on sale_order_lot_selection.

In fact, the use of sale_line_id field in #2247, is not relevant... I just saw during tests we could set lot_id on a product with type = service. In that case there is no stock moves, but we could have some purchase generated using subcontracting, so I set the lot on purchase order line using only sale_line_id

Anyway, I think we're discussing 2 different cases here :
- the one retrieving lots on purchase from underlying stock moves / procurements
- the one wanting to set lots on purchase to restrict stock moves

It seems to me the first one is what purchase_lot was originally designed for (back to v7.0 in 2014 : see https://github.com/akretion/odoo-embed-external-configuration/tree/7.0/purchase_from_configuration)
The second one is what @Tisho99 would like to achieve here, making lot_id not readonly on purchase order lines.

florian-dacosta:
> For the moment, installing purchase_lot alone would be totally useless because there is no other way than sale_order_lot_selection to set restrict_lot_id field in stock moves. So, there is a FUNCTIONAL dependency for now if using only OCA modules... but anyone can use custom modules too, right ? So, I totally agree with both of you : there is no need for the technical dependency on sale_order_lot_selection.
> 
> In fact, the use of sale_line_id field in #2247, is not relevant... I just saw during tests we could set lot_id on a product with type = service. In that case there is no stock moves, but we could have some purchase generated using subcontracting, so I set the lot on purchase order line using only sale_line_id
> 

I am not sure I get what you mean. "the use of sale_line_id field is not relevant" but "I set the lot on purchase order line using only sale_line_id".
So is it or is it not usefull ? Because if it is not, we can remove it from the PR 2247 and close the debate (keep one module without dependency on sale for now)

> Anyway, I think we're discussing 2 different cases here :
> 
>     * the one retrieving lots on purchase from underlying stock moves / procurements
> 
>     * the one wanting to set lots on purchase to restrict stock moves
> 
> 
> It seems to me the first one is what purchase_lot was originally designed for (back to v7.0 in 2014 : see https://github.com/akretion/odoo-embed-external-configuration/tree/7.0/purchase_from_configuration) The second one is what @Tisho99 would like to achieve here, making lot_id not readonly on purchase order lines.

Yes, I believe there are these 2 uses cases indeed. And I think there are not incompatible.  If the only thing is that in one case we'd prefer to have the lot readonly and the other we need it editable, maybe we can live with it beeing always editable ?
The more annoying option could be to add an extra field to make the lot editable or not. (the extra boolean field could be set to True if lot_id is set automatically from a procurement.)

Also not clear for me from your message : Are you ok to merge this PR as it is ?
If not, what is blocking (the readonly attribute ?)

metaminux:
Hi @florian-dacosta,

Yes, the readonly attribute is what makes me feel uncomfortable here...
Anyway, as you said earlier, the same PR has already been merged in 17.0, so let's merge this one in 16.0 too and work on the others PRs.

metaminux (approved):

PR #1666, opened 2022-12-02
[16][MIG] purchase_lot

florian-dacosta:
standard migration

Depends on https://github.com/OCA/sale-workflow/pull/2282

@Kev-Roche

PR #2647, opened 2025-04-16
[17.0] [MIG + IMP] purchase_lot

Tisho99:
Supersedes #1666  +  add improvement https://github.com/OCA/purchase-workflow/issues/2644


Tisho99:
@manuelregidor

manuelregidor:
@Tisho99 Could you add a test in which a purchase order is created using lots and then it is validated so it is checked that the lots have been tranfered to the picking? Thank you

manuelregidor (changes requested):
Could you add a test in which a purchase order is created using lots and then it is validated so it is checked that the lots have been tranfered to the picking? Thank you

PR #2655, opened 2025-04-23
[17.0][FIX] purchase_lot: super call

Tisho99:
The _find_candidate() function seems to be dead code right now because the recordset filter is not propagated to super().

Correct me if i am wrong

### DIFF
--- a/purchase_lot/models/purchase_order_line.py
+++ b/purchase_lot/models/purchase_order_line.py
@@ -44,8 +44,9 @@ class PurchaseOrderLine(models.Model):
         values,
     ):
         lot_id = values.get("restrict_lot_id", False)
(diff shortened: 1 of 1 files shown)


purchase_lot/models/purchase_order_line.py
```diff
@@ -44,8 +44,9 @@ def _find_candidate(
         values,
     ):
         lot_id = values.get("restrict_lot_id", False)
-        self = self.filtered(lambda line: line.lot_id.id == lot_id)
```
florian-dacosta, line 47: Did you experiment a bug about this ?
I do believe the proposed change do not change anything, as without argument, super is called on self which is modified (filtered) here.

Tisho99:
@florian-dacosta You are right, the module was correct, i close this PR


Thank you

PR #2663, opened 2025-04-28
[18.0][FIX] purchase_all_shipments: Fix the test cases.

bizzappdev:
In the recent change in Odoo 18 PR: https://github.com/odoo/odoo/pull/205346
The product with ID: product.product_product_27 has been used in purchase_order_1, which was previously enabled to use lot tracking from the MRP module, but now, because of the above PR, it has been enabled from the stock module, which is causing the failure of all new PR test cases in the repo (18.0 branch).


AungKoKoLin1997 (approved):
LGTM

HviorForgeFlow (approved):

PR #2666, opened 2025-05-05
[16.0][FIX] purchase_request: Change the types of messages that are created to notes

victoralmau:
Change the types of messages that are created to notes.

Change the subtype from `mail.mt_comment` to `mail.mt_note` to prevent an email being sent to the vendor.

@Tecnativa TT56225

### DIFF
--- a/purchase_request/models/purchase_order.py
+++ b/purchase_request/models/purchase_order.py
@@ -60,7 +60,7 @@ class PurchaseOrder(models.Model):
                     request, requests_dict[request_id]
                 )
                 request.message_post(
-                    body=message, subtype_id=self.env.ref("mail.mt_comment").id
+                    body=message, subtype_id=self.env.ref("mail.mt_note").id
                 )
         return True
 
@@ -181,7 +181,7 @@ class PurchaseOrderLine(models.Model):
                 )
                 if message:
                     alloc.purchase_request_line_id.request_id.message_post(
-                        body=message, subtype_id=self.env.ref("mail.mt_comment").id
+                        body=message, subtype_id=self.env.ref("mail.mt_note").id
                     )
 
                 alloc.purchase_request_line_id._compute_qty()
--- a/purchase_request/models/purchase_request_allocation.py
+++ b/purchase_request/models/purchase_request_allocation.py
@@ -128,5 +128,5 @@ class PurchaseRequestAllocation(models.Model):
             message_data = self._prepare_message_data(po_line, request, allocated_qty)
             message = self._purchase_request_confirm_done_message_content(message_data)
             request.message_post(
-                body=message, subtype_id=self.env.ref("mail.mt_comment").id
+                body=message, subtype_id=self.env.ref("mail.mt_note").id
             )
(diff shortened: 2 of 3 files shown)


pedrobaeza:
Please indicate always the reasons. You change it because messages are arriving to vendors, which is not correct.

pedrobaeza (approved):

carlos-lopez-tecnativa (approved):

pedrobaeza:
@victoralmau please change the commit message as indicated.

purchase_request/models/purchase_order.py
```diff
@@ -60,7 +60,7 @@ def _purchase_request_confirm_message(self):
                     request, requests_dict[request_id]
                 )
                 request.message_post(
-                    body=message, subtype_id=self.env.ref("mail.mt_comment").id
+                    body=message, subtype_id=self.env.ref("mail.mt_note").id
```
carlos-lopez-tecnativa, line 63: According to this code, if you don’t pass `subtype_id`, it defaults to `mail.mt_note`, so I think this parameter can be removed. Or is there another reason to explicitly pass the parameter?
https://github.com/odoo/odoo/blob/6f30dee4aef8e0078d1a1da8de1878397f872a90/addons/mail/models/mail_thread.py#L2079-L2080

pedrobaeza: Let's follow the Python zen: "Better explicit than implicit", so we avoid regressions in case Odoo (or any custom) changes this default.

victoralmau:
> @victoralmau please change the commit message as indicated.

Commit message updated.

luc-adhoc:
Hello, Is possible to revert this PR?
First, this change affected a stable version (16) and affected our clients. 
Scond, the vendor is not a follower in the PR so it should not receive the emails in first place.. so I do not undertand the reason of the change and the benefits..
thanks

victoralmau:
I think it is important to explain the need for this change.

Previously, there were three different notifications.

1- Message in the purchase request when confirming the order.

ejemplo-1

This was done as “mail.mt_comment,” that is, Conversations, “public” message instead of internal note; and therefore sending the email to all followers.

2- Message in the purchase request upon receiving the order (done picking).

ejemplo-2

Same explanation as in the previous point.

3- Message in the delivery note upon receiving the order (done picking).

ejemplo-3

Same explanation as in the previous points, in this case the partner (the supplier) was a follower of the picking and therefore received that message. https://github.com/OCA/purchase-workflow/commit/fcb866c1892422c6bbc7c260e63ccb2cf9fbf775# diff-1d0a325182ee389a4b623eb623b0bb74e1c0447dff7a05f6ed5c45184749443bR120

Please note that the screenshots I have provided are an “_example_” of how things worked before this change (If, for example, `purchase_tier_validation` is used, there will be more followers in some places).

In addition to this, specific subtypes were later added for each of these things, in order to “_model_” who receives the notifications, so that each message is of a specific subtype and is sent to the corresponding followers. https://github.com/OCA/purchase-workflow/pull/2722

PR #2393, opened 2024-08-19
[16.0][MIG] purchase analytic global: Migration to 16.0

sistecem:

PR #2668, opened 2025-05-06
[17.0][MIG] purchase_analytic_global: Migration to 17.0 

Luca-Policastro:
Migration done from https://github.com/OCA/purchase-workflow/pull/2393


purchase_analytic_global/models/purchase.py
```diff
@@ -0,0 +1,47 @@
...
+        If no lines, respect value given by the user.
+        """
+        for rec in self:
+            if rec.order_line:
+                al = rec.order_line[0].analytic_distribution or False
+                for ol in rec.order_line:
+                    if ol.analytic_distribution != al:
+                        al = False
+                        break
+                rec.analytic_distribution = al
```
yankinmax: hello @Luca-Policastro 
what do you think about such an approach from v18 solving the same issue:
https://github.com/OCA/purchase-workflow/blob/18.0/purchase_analytic_global/models/purchase_order.py

SilvioC2C:
```suggestion
        for po in self:
            if lines := po.order_line:
                distributions = set(ln.analytic_distribution for ln in lines)
                if len(distributions) == 1:
                    po.analytic_distribution = tuple(distributions)[0]
                else:
                    po.analytic_distribution = False
```

purchase_analytic_global/models/purchase.py
```diff
@@ -0,0 +1,47 @@
...
+                )
+
+    @api.onchange("analytic_distribution")
+    def _onchange_analytic_distribution(self):
+        """When change analytic_distribution
+        set analytic distribution on all order lines"""
+        if self.analytic_distribution:
+            self.order_line.update(
+                {"analytic_distribution": self.analytic_distribution}
+            )
```
SilvioC2C: Isn't this equivalent to the `inverse` method that's already in place?

SilvioC2C (commented):
Please squash the last 2 commits

PR #2670, opened 2025-05-07
[IMP] purchase_request: allow editing the request reference without developer mode

mav-adhoc:
`is_name_editable` defaults to `base.group_no_one`, so the request reference is
only editable while developer mode is on. A user who manages purchase requests
cannot rename one, not even in draft, and the reference is the only field that
identifies a request in the list view, the search and the report — users
routinely want a descriptive title there rather than the sequence.

This drops the `readonly` modifier so the reference follows the write access to
the record. The sequence still fills it in on create when the default is left
untouched: `create()` only replaces the name while it still equals `New`.

Test: `test_name_editable_without_developer_mode` asserts the form view does not
gate the reference behind a modifier. It fails on `18.0` and passes with this
change; the rest of the module suite stays green (31 tests).

Note: the original description of this PR blamed the readonly for a validation
error on save. That was wrong — see the comment below.

### DIFF
--- a/purchase_request/__manifest__.py
+++ b/purchase_request/__manifest__.py
@@ -4,7 +4,7 @@
 {
     "name": "Purchase Request",
     "author": "ForgeFlow, Odoo Community Association (OCA)",
-    "version": "18.0.2.5.1",
+    "version": "18.0.2.6.0",
     "summary": "Use this module to have notification of requirements of "
     "materials and/or external services and keep track of such "
     "requirements.",
--- a/purchase_request/tests/test_purchase_request.py
+++ b/purchase_request/tests/test_purchase_request.py
@@ -1,6 +1,8 @@
 # Copyright 2018-2019 ForgeFlow, S.L.
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl-3.0)
 
+from lxml import etree
+
 from odoo import SUPERUSER_ID, exceptions
 from odoo.exceptions import UserError
 from odoo.tests import Form, TransactionCase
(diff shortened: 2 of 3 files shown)


mav-adhoc:
@pedrobaeza @HviorForgeFlow Hi! Can you review this PR? Thanks!

HviorForgeFlow:
@mav-adhoc I cannot reproduce the issue that you mention on runboat. can you elaborate it a bit? Thanks.

You can always for the edition adding a ir.default that sets the is_name_editable to True for all users in the system.

Please, try not to apply auto formatting out of the scope of the change.

For reference:
https://github.com/OCA/purchase-workflow/pull/1120

mav-adhoc:
@HviorForgeFlow you were right, and I owe you a correction: the readonly is not
what caused the error I mentioned, and it does not reproduce. Sorry for the noise.

What the user actually hit was a `NotNullViolation` on `purchase_request.name`
("Create/update: a mandatory field is not set", field `Request Reference`). I went
back and reproduced the three paths on 18.0 with the current code:

- create **without** `name` in the values (what the web client sends while the
  field is readonly) → works, the record gets `PR00001`;
- create with an empty `name` → `NotNullViolation`;
- create without `name` while `next_by_code("purchase.request")` returns `False`
  (no sequence reachable for the active company) → same `NotNullViolation`.

So the readonly was never the cause. Making the field editable only lets the user
type a value and hides a problem that is on the sequence side.

What is left is the actual request, and it is a UX one rather than a bug: the
reference is the only field that identifies a request in the list view, the search
and the report, and users want a descriptive title there. Gating that behind
developer mode means a purchase request manager cannot rename a request, not even
in draft. That is what the PR does now, and I retitled it to `[IMP]` so it stops
advertising a bug that is not there.

On the `ir.default` suggestion: it works for new records, but `is_name_editable`
is stored and evaluated at creation time, so every request already in the database
keeps `False` and stays readonly. If you would rather keep a gate than drop it, I
am happy to turn `is_name_editable` into a **non-stored computed** field over
`purchase_request.group_purchase_request_manager` instead of `base.group_no_one`:
it keeps the intent of #1120 (not everybody renames), works without developer mode
and also covers existing records. Say which one you prefer and I will push it.

I also dropped the README/`index.html` reformatting you flagged — the diff is now
the view change, a small test and the version bump (happy to drop the bump too if
you prefer the bot to handle it).

PR #2643, opened 2025-04-10
[MIG] purchase_order_approved: Migration to 18.0

ellbristow:
Migrated from 16.0

Views updates
ResConfigSetting option renamed slightly for clarity
Removed "states" and replaced with readonly attributes

Otherwise a straight migration

PR #2671, opened 2025-05-08
[18.0] [MIG] purchase_merge: Migration to 18.0

bizzappdev:
Dependency PR for module purchase_order_approved:  https://github.com/OCA/purchase-workflow/pull/2643


amh-mw (changes requested):
@bizzappdev Since `purchase_order_approved` was merged in https://github.com/OCA/purchase-workflow/pull/2732, would you kindly rebase, drop the test requirements commit, and force push? I'm testing locally this morning with that setup and it is generally looking viable.

PR #2675, opened 2025-05-12
[17.0][MIG] purchase_v12_control_menu: Migration to version 17.0

david-banon-tecnativa:
image

@tecnativa @pedrobaeza @victoralmau 

TT52351


legalsylvain:
Hi. Thanks for porting this module !

I dont see the interest to keep V12, in the name of the module.
Any reason to keep it ?

Thanks !

pedrobaeza:
The reason is to state that it's what it was in v12.

legalsylvain:
> The reason is to state that it's what it was in v12.

Yes, That's trivial. 
The question is, does it make sense to mention a version that most people will not remember ? 
I mean v 12 bas been release 6 years ago. Why not just purchase_control_menu ?

pedrobaeza:
Because the name is not specific enough, and mostly we don't want to do all the stuff to rename it.

pedrobaeza (approved):
Code and functional review

victoralmau (approved):
Code review OK

PR #2679, opened 2025-05-13
[14.0][FIX] purchase_quick: price_unit when qty_to_process is less than seller’s qty_min

Chionne27:
BWP #2069


Chionne27:
Hi @angelgarciadelachica

I did a backport of your commit but it doesn't solve the issue in version 14. Do you know how can I solve it. I don't know the differences between 14 and 16 about this functionality.

PR #2682, opened 2025-05-14
[ADD] purchase_return_picking_type

AlexPForgeFlow:
This module allows to define an Operation Type in Purchase Return Orders.


AaronHForgeFlow (commented):
Code LG. I just a have a couple of questions:

- How will the picking_type_id defined in the return order is used?
- I really don't mind it but perhaps it can be called purchase_return_picking_type?

AlexPForgeFlow:
Hi @AaronHForgeFlow !
The purpose is to have information about where the products being returned were originally received from. This is useful, for example, if the module product_supplierinfo_stock_picking_type is installed, which allows defining different prices for the same vendor depending on the Operation Type. For instance, the vendor might offer a different price for WH1:Receipts than for WH2:Receipts.

Taking this into account, and with the module from https://github.com/OCA/purchase-workflow/pull/2683, we can manage the different prices offered by the same vendor depending on the destination on the purchase return orders.

AlexPForgeFlow:
> Code LG. I just a have a couple of questions:
> 
>     * How will the picking_type_id defined in the return order is used?
> 
>     * I really don't mind it but perhaps it can be called purchase_return_picking_type?

Must be called picking_type_id because logic from _prepare_sellers from https://github.com/OCA/product-attribute/blob/16.0/product_supplierinfo_stock_picking_type/models/product_product.py#L9

purchase_return_stock/models/purchase_return_order.py
```diff
@@ -0,0 +1,37 @@
...
+            )
+        return picking_type[:1]
+
+    @api.model
+    def _default_picking_type(self):
+        return self._get_picking_type(
+            self.env.context.get("company_id") or self.env.company.id
+        )
+
+    picking_type_id = fields.Many2one(
```
rousseldenis: Please put fields on top

purchase_return_stock/models/purchase_return_order.py
```diff
@@ -0,0 +1,37 @@
...
+        "stock.picking.type",
+        "Deliver To",
+        states=PurchaseReturn.READONLY_STATES,
+        required=True,
+        default=_default_picking_type,
```
rousseldenis: Change this to lambda

AaronHForgeFlow:
ok! The second question was about the name of the module, but it does not matter

rousseldenis (commented):
Could you add some tests?

I would have named this module with something more understandable from a technical/functional point of view. Like `purchase_return_picking_type`

rousseldenis:
> Could you add some tests?
> 
> I would have named this module with something more understandable from a technical/functional point of view. Like `purchase_return_picking_type`

Didn't see @AaronHForgeFlow comment but we are synced about the name :sweat_smile:

AlexPForgeFlow:
> > Code LG. I just a have a couple of questions:
> > ```
> > * How will the picking_type_id defined in the return order is used?
> > 
> > * I really don't mind it but perhaps it can be called purchase_return_picking_type?
> > ```
> 
> Must be called picking_type_id because logic from _prepare_sellers from https://github.com/OCA/product-attribute/blob/16.0/product_supplierinfo_stock_picking_type/models/product_product.py#L9

@rousseldenis I named the field that way for a specific reason, if it's renamed we will have conflicts when the product_supplierinfo_stock_picking_type module is installed. That's why I created this module.

rousseldenis:
> @rousseldenis I named the field that way for a specific reason, if it's renamed we will have conflicts when the product_supplierinfo_stock_picking_type module is installed. That's why I created this module.

It's not about field name but module one.

AlexPForgeFlow:
> > @rousseldenis I named the field that way for a specific reason, if it's renamed we will have conflicts when the product_supplierinfo_stock_picking_type module is installed. That's why I created this module.
> 
> It's not about field name but module one.

ah, you were talking about module name..

AlexPForgeFlow:
@AaronHForgeFlow @rousseldenis 
module name changed and applied your suggestions

AaronHForgeFlow (approved):
:ok:

rousseldenis:
> @AaronHForgeFlow @rousseldenis module name changed and applied your suggestions

Missing a little test

AlexPForgeFlow:
> > @AaronHForgeFlow @rousseldenis module name changed and applied your suggestions
> 
> Missing a little test

@rousseldenis maybe tests are more interesting adding them here https://github.com/OCA/purchase-workflow/pull/2683, wdyt?

AlexPForgeFlow:
@rousseldenis or you want me to test de defaults to the picking type on purchase return order creation?

PR #2523, opened 2025-01-22
[18.0][MIG] sale_purchase_force_vendor: Migration to 18.0

jaydeep32:

PR #2689, opened 2025-05-21
[17.0][FIX] sale_purchase_force_vendor: Change the done status of sales orders (does not exist) to sale

victoralmau:
Change the `done` status of sales orders (does not exist) to `sale`

Related to https://github.com/OCA/purchase-workflow/pull/2523#discussion_r2099637436

@Tecnativa

### DIFF
--- a/sale_purchase_force_vendor/views/sale_order_view.xml
+++ b/sale_purchase_force_vendor/views/sale_order_view.xml
@@ -26,7 +26,7 @@
                     <field
                         name="vendor_id"
                         domain="vendor_id_domain"
-                        readonly="'state' in ('done', 'cancel')"
+                        readonly="'state' in ('sale', 'cancel')"
                         options='{"no_create": True}'
                     />
                 </group>


sale_purchase_force_vendor/views/sale_order_view.xml
```diff
@@ -26,7 +26,7 @@
                     <field
                         name="vendor_id"
                         domain="vendor_id_domain"
-                        readonly="'state' in ('done', 'cancel')"
+                        readonly="'state' in ('sale', 'cancel')"
```
pedrobaeza, line 29: I think we should add the condition `is_locked` as well.

sale_purchase_force_vendor/views/sale_order_view.xml
```diff
@@ -26,7 +26,7 @@
                     <field
                         name="vendor_id"
                         domain="vendor_id_domain"
-                        readonly="'state' in ('done', 'cancel')"
+                        readonly="'state' in ('sale', 'cancel') or parent.locked"
```
pedrobaeza: Uhm, thinking twice, it was incorrect in previous versions, as you can't change the vendor once the quotation is confirmed. It's enough with the state.

pedrobaeza (approved):
/ocabot merge patch

PR #2694, opened 2025-05-27
[16.0][IMP]Purchase workflow added to review state & exception fix

marcos-mendez:
This fixes the groups in the purchase-request-exception view for the Odoo>v14

Also we have to notice, some time when the user need something that is not registered as a product in the Odoo Database he has de possibility to request the description of something. 

For this reason the approver has to input the product in the database acording to the user's description of the requested product.

This stages maintains the Draft state (so the approver can insert the product in the database)


marcos-mendez:
@mileo @marcelsavegnago @renatonlima podem me ajudar a revisar?

norlinhenrik (approved):
I tested that the "to review" state exists, and the exception view works.

PR #2707, opened 2025-06-04
[18.0][IMP] purchase_delivery_split_date: Split improvements

jbaudoux:
fwd port of https://github.com/OCA/purchase-workflow/pull/2049 and fixes
replaces:
* https://github.com/OCA/purchase-workflow/pull/1326

cc @rousseldenis @sebalix 

* In order to not browse the purchase order when splitting, do the split directly on purchase order lines as we have the modified recordset (for date_planned field).
* Fix po line addition. Reduce the amount of canceled picking.
* Ensure there is one picking per date even when different time.
* Ensure po line planned time is kept on move deadline.
* Make proper date comparison timezone aware.
* Changing the deadline on a move linked to a PO will also trigger the split by date.
* Only rely on picking assignation domain for checking if the picking is still right.
* Use the standard move picking assignation (_assign_picking()) and add the grouping date in the domain.


rousseldenis (approved):
LGTM

sebalix (approved):

HviorForgeFlow (approved):

PR #2715, opened 2025-06-09
[18.0][MIG] purchase_lot: Migration to 18.0

BT-jcolmeiro:
Basic migration v18.0 appliying minimum changes

PR #2721, opened 2025-06-14
[IMP] purchase_lot : lot_id is computed from stock moves so that it c…

dreispt:
Forward port of https://github.com/OCA/purchase-workflow/pull/2247
<s>Depends on https://github.com/OCA/purchase-workflow/pull/2715</s>

### DIFF
--- a/purchase_lot/models/purchase_order_line.py
+++ b/purchase_lot/models/purchase_order_line.py
@@ -8,8 +8,14 @@ from odoo import api, fields, models
 class PurchaseOrderLine(models.Model):
     _inherit = "purchase.order.line"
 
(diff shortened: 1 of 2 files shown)


purchase_lot/models/purchase_order_line.py
```diff
@@ -8,9 +8,15 @@
 class PurchaseOrderLine(models.Model):
     _inherit = "purchase.order.line"
 
-    lot_id = fields.Many2one("stock.lot", string="Serial Number", copy=False)
```
florian-dacosta, line 11: The field is totally removed, I think it is a mistake !

florian-dacosta (changes requested):
@dreispt Could you rebase and check my comment please ?

PR #2722, opened 2025-06-16
[16.0][IMP] purchase_request: Add specific subtypes for using subscribed partners in some messages

victoralmau:
Add specific subtypes for using subscribed partners in some messages

The following subtypes have been created:
- Purchase Order confirmation: The email will be sent only to the followers of this subtype.
- Purchase receipt confirmation: The email will be sent only to followers of this subtype.

@Tecnativa TT56541

### DIFF
--- a/purchase_request/data/purchase_request_data.xml
+++ b/purchase_request/data/purchase_request_data.xml
@@ -27,4 +27,18 @@
         <field name="default" eval="True" />
         <field name="description">Purchase Request is done</field>
     </record>
+    <record id="mt_request_po_confirmed" model="mail.message.subtype">
+        <field name="name">Purchase Order confirmation</field>
+        <field name="res_model">purchase.request</field>
+        <field name="default" eval="True" />
+        <field name="internal" eval="True" />
+        <field name="description">Purchase Order is confirmed</field>
+    </record>
+    <record id="mt_request_picking_done" model="mail.message.subtype">
+        <field name="name">Purchase receipt confirmation</field>
+        <field name="res_model">purchase.request</field>
+        <field name="default" eval="True" />
+        <field name="internal" eval="True" />
+        <field name="description">Receipt is done</field>
+    </record>
 </odoo>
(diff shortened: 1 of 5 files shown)


purchase_request/models/purchase_order.py
```diff
@@ -56,11 +56,19 @@ def _purchase_request_confirm_message(self):
...
+                partners = request.message_follower_ids.filtered(
+                    lambda x: mt_request_po_confirmed in x.subtype_ids
+                ).mapped("partner_id")
                 message = po._purchase_request_confirm_message_content(
                     request, requests_dict[request_id]
                 )
                 request.message_post(
-                    body=message, subtype_id=self.env.ref("mail.mt_note").id
+                    body=message,
+                    subtype_id=self.env.ref("mail.mt_note").id,
```
pedrobaeza: Why don't you put the specific subtype? And I think Odoo will handle it thus without getting partners.

pedrobaeza (approved):

purchase_request/data/purchase_request_data.xml
```diff
@@ -27,4 +27,18 @@
         <field name="default" eval="True" />
         <field name="description">Purchase Request is done</field>
     </record>
+    <record id="mt_request_po_confirmed" model="mail.message.subtype">
+        <field name="name">Purchase Order confirmation</field>
+        <field name="res_model">purchase.request</field>
+        <field name="default" eval="True" />
+        <field name="internal" eval="True" />
```
carlos-lopez-tecnativa, line 34: If the subtype is set to internal, I notice that no email is sent, is this the expected behavior?

I manually changed it to False, and the email was sent. But if the field is set to True, no email is sent.
image

victoralmau: Actually it is not correct, that is to say, although the icons of the recipients are **NOT** seen, actually the emails are sent (you can check it in mailhog). This is not related to `internal=True`, it is because we are indicating a subtype but NOT the `partner_ids` (the same behavior happens if you add a note).

carlos-lopez-tecnativa (approved):

marcelsavegnago (approved):
LGTM

PR #2724, opened 2025-06-18
[16.0][FIX] purchase_request: Add noupdate=1 to mail message subtypes

victoralmau:
Add noupdate=1 to mail message subtypes

Please @pedrobaeza and @pilarvargas-tecnativa can you review it?

@Tecnativa TT56541


pedrobaeza:
This requires migration script

pedrobaeza (approved):

victoralmau:
Ping @carlos-lopez-tecnativa

pilarvargas-tecnativa (approved):

carlos-lopez-tecnativa (approved):

purchase_request/README.rst
```diff
@@ -1,7 +1,3 @@
-.. image:: https://odoo-community.org/readme-banner-image
```
pedrobaeza, line 1: @sbidoul look at this change. The template in the pre-commit definition is not the same as the one being applied "on ocabot" part, provoking this eternal removal and introduction each time a manifest is touched. IMO, ocabot should execute the pre-commit version of the repository, not one fixed.

sbidoul: This is the result of a compromise between many different requirements. If you want to change this, please open an issue on https://github.com/OCA/oca-addons-repo-template/, but before that please spend sometime re-reading all the arguments that were discussed back then.

pedrobaeza: Can you point to such discussion, please? I don't remember it at all. Anyway, I suggest then to update the copier template massively on all the repos from 16.0 to 18.0. What do you think?

sbidoul: > Can you point to such discussion, please?

I found https://github.com/OCA/oca-addons-repo-template/pull/220 but I'm sure there are other places.

> Can you point to such discussion, please? I don't remember it at all. Anyway, I suggest then to update the copier template massively on all the repos from 16.0 to 18.0. What do you think?

Yes we can do that. I'll try to do it somewhere in July.

pedrobaeza: I think such PR is not exactly about that. I don't think it has been discussed in any place that ocabot generates the README not using the pre-commit local config. That can be because the README generation was "previous" to include it in the pre-commit thing, but now that is included, it should be respected IMO.

PR #2725, opened 2025-06-19
[17.0][FIX] purchase_request: Add noupdate=1 to mail message subtypes

victoralmau:
FWP from 16.0: https://github.com/OCA/purchase-workflow/pull/2724

Add noupdate=1 to mail message subtypes

@Tecnativa TT56541

PR #2727, opened 2025-06-20
[18.0][FIX] purchase_request: Add noupdate=1 to mail message subtypes

victoralmau:
FWP from 17.0: https://github.com/OCA/purchase-workflow/pull/2725

Add noupdate=1 to mail message subtypes

@Tecnativa TT56541

### DIFF
--- a/purchase_request/README.rst
+++ b/purchase_request/README.rst
@@ -85,9 +85,9 @@ To configure the product follow this steps:
 Note that MTO route is archived by default, you should unarchive it
 first:
 
-- Go to Inventory > Configuration > Routes, in the Search dropdown menu,
-  click *Archived* to show *Replenish on Order (MTO)* route and
-  *Unarchive* it.
+-  Go to Inventory > Configuration > Routes, in the Search dropdown
+   menu, click *Archived* to show *Replenish on Order (MTO)* route and
+   *Unarchive* it.
 
 With this configuration, whenever a procurement order is created and the
 supply rule selected is 'Buy' the application will create a Purchase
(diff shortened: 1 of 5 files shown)


pedrobaeza (approved):

victoralmau:
Ping @carlos-lopez-tecnativa

carlos-lopez-tecnativa (approved):

carlos-lopez-tecnativa (commented):
@victoralmau When I update this module, I get the following error:
image

In Odoo 18, there's a strict validation on the migration function signature.
It seems that the function must use `(cr, version)` as parameters instead of `(env, version)`.

Please review it:
https://github.com/odoo/odoo/blob/2a839ef1ed09c36f27ce7536ca3052d9f65ceed9/odoo/modules/migration.py#L252-L256

victoralmau:
> @victoralmau When I update this module, I get the following error: image
> 
> In Odoo 18, there's a strict validation on the migration function signature. It seems that the function must use `(cr, version)` as parameters instead of `(env, version)`.
> 
> Please review it: https://github.com/odoo/odoo/blob/2a839ef1ed09c36f27ce7536ca3052d9f65ceed9/odoo/modules/migration.py#L252-L256

I find it very strange, to be honest. There are many migration scripts like this:

```
@openupgrade.migrate()
def migrate(env, version):
```


Can you clarify @pedrobaeza ?

pedrobaeza:
It's nullified with `openupgrade_framework`, but being a intermediate migration script, that module is not server-wide loaded, so it triggers the exception, so we have to stick to the variable name `cr`. Not sure if we can still use the decorator, and assume `cr = env` (putting the corresponding comment).

PR #2728, opened 2025-06-20
[18.0][FIX] purchase_force_invoiced: restore force invoiced flag editable status

dalonsod:
Fw-port of #2726 

Before v17 migration, if you wanted to force invoice status purchase order must be locked. After that migration, condition was changed to the opposite one (purchase order must be unlocked), probably due to a migration error.

This restores original behavior, that shouldn't be changed. Also, restoring old behavior makes the addon coherent with current usage documentation, that's been never changed.


pedrobaeza (approved):
Fast-tracking as already approved in the previous version

/ocabot merge patch

PR #2059, opened 2023-10-25
[16.0][MIG] purchase_stock_secondary_unit

acsonefho:
Migration (from v15) of `purchase_stock_secondary_unit`

Need:
- https://github.com/OCA/purchase-workflow/pull/2068
- https://github.com/OCA/stock-logistics-warehouse/pull/1769#discussion_r1381476739 (with the related comment/fix)

PR #2731, opened 2025-06-25
[17.0][MIG] purchase_stock_secondary_unit

edescalona:
@BinhexTeam

The PR was used as a base #2059 

Depends on

- https://github.com/OCA/stock-logistics-warehouse/pull/2432


purchase_stock_secondary_unit/tests/test_purchase_stock_secondary_unit.py
```diff
@@ -86,8 +86,9 @@ def test_update_confirmed_purchase_order_with_move_validated(self):
         self.assertEqual(picking.move_ids.secondary_uom_id, self.secondary_product_uom)
         self.assertEqual(picking.move_ids.product_uom_qty, qty)
         # Assigned move line
+        picking.move_line_ids.secondary_uom_qty = 3
```
pedrobaeza: This seems to trick the test. Why this data is not filled already?

pedrobaeza (changes requested):

edescalona:
ping @pedrobaeza

purchase_stock_secondary_unit/models/purchase_order_line.py
```diff
@@ -33,15 +33,18 @@ def write(self, vals):
...
+                    {"secondary_uom_qty": secondary_uom_qty}
                 )
             elif moves and vals.get("secondary_uom_qty"):
                 previous_secondary_qty = sum(m.secondary_uom_qty for m in moves[:-1])
+                secondary_uom_qty = vals["secondary_uom_qty"] - previous_secondary_qty
                 moves[-1:].filtered(
                     lambda sm: sm.state not in ["done", "cancel"]
-                ).secondary_uom_qty = vals["secondary_uom_qty"] - previous_secondary_qty
+                ).secondary_uom_qty = secondary_uom_qty
+                moves[-1:].move_line_ids.write({"secondary_uom_qty": secondary_uom_qty})
```
pedrobaeza: This is not correct, as you may have more than one move line per move and the secondary quantity should be the proportional one according the quantity in the sml.

edescalona:
Hi @pedrobaeza , after a closer look, the issue isn't in the `purchase_stock_secondary_unit` module, but in `stock_secondary_unit` in the `stock.move.line` model. The `_secondary_unit_fields` class attribute and the `qty_field` definition list the `qty_done` field, which should be the `quantity` field. An error isn't displayed because the `qty_done` field is redefined. Well, since it doesn't exist, it's created, but no changes occur when creating the picking movements.

I tried changing those values ​​in `stock_secondary_unit` and their respective changes in `purchase_stock_secondary_unit`, and the values ​​are correctly created in `stock.move.line`, and the tests pass without any issues.

Now, could the correction be made in this module or could a PR be created for `stock_secondary_unit`? What do you think? If there's anything wrong with my explanation, please let me know. Thanks.

pedrobaeza:
Please create a new PR for the fix for the already merged modules.

edescalona:
ping @carlosdauden @CarlosRoca13 @pedrobaeza

sergio-teruel (approved):

pedrobaeza:
Please remove the temporary commit.

edescalona:
ready @pedrobaeza

pedrobaeza (approved):
/ocabot merge nobump

PR #2735, opened 2025-06-30
[16.0][FIX] purchase_last_price_info: access rights issue

victoralmau:
Backport from 17.0: https://github.com/OCA/purchase-workflow/pull/2281

On purchase.order.line creation, `_update_inverses` is called before the value of purchase.order.line.company_id is computed. `_update_inverses` causes Odoo to filter purchase.order.line to find the correct lines to put into the cache, which in turn causes access rights check. And because company_id is not yet computed it fails.
Fixing by removing domain from `last_purchase_line_ids` to avoid filtering and filter it later manually.

@Tecnativa

### DIFF
--- a/purchase_last_price_info/models/product_product.py
+++ b/purchase_last_price_info/models/product_product.py
@@ -12,10 +12,6 @@ class ProductProduct(models.Model):
     last_purchase_line_ids = fields.One2many(
         comodel_name="purchase.order.line",
         inverse_name="product_id",
-        domain=lambda self: [
-            ("state", "in", ["purchase", "done"]),
-            ("company_id", "in", self.env.companies.ids),
-        ],
         string="Last Purchase Order Lines",
     )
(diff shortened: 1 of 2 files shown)


LoisRForgeFlow (approved):
LGTM :+1:

LoisRForgeFlow:
@victoralmau side question regarding the use of `filtered_domain`, do you know if it is more performant that `filtered`? Any reason to prefer it?

victoralmau:
Normally I don't use `filtered_domain`, in this case it is because it is what has been used in v17 to correct the same problem; in theory `filtered_domain` is more efficient than `filtered`, although it is true that we don't use it too much.

pedrobaeza (approved):
`filtered_domain` builds a query to do the search, including the `id in ...` clause, so it avoids to iterate per each record, obtain values, and evaluate in Python, so yes, it's more performant.

/ocabot merge patch

PR #2738, opened 2025-07-01
[17.0][ADD] purchase_billing_address

edescalona:
@Binhexteam

This add-on improves purchase order functionality:

1. Add a new partner type (Purchase invoice address) for billing addresses.
2. It allows you to select this billing address on purchase orders based on the customer hierarchy.

The order for obtaining the billing address is as follows:

1. Contacts of the selected provider.
2. Contacts of the provider's contacts.
3. Provider


pedrobaeza (commented):
You have to remove the word "invoice" form all the places, as you are using that contact for purchase orders, not invoices.

edescalona:
> You have to remove the word "invoice" form all the places, as you are using that contact for purchase orders, not invoices.

But in the case of the name of the new type of partner, I should leave it so the user knows, isn't that correct?

christian-ramos-tecnativa:
IMO the name should be something like purchase_billing_address as this is the real purpose of it. And the purchase_invoice type name I think is correct, just purchase doesn't reflect the idea for this new type.

christian-ramos-tecnativa:
@edescalona Just like in the Sales module, it would be great if the new field could be automatically populated when the contact has a child record of this new type.

edescalona:
Hi @Christian-RB , thanks for your suggestion. I made the necessary changes so that secondary contacts could be retrieved. The order for obtaining the billing address is as follows:

1. Contacts of the selected provider.
2. Contacts of the provider's contacts.
3. Provider

Let me know what you think when you can, thanks.

edescalona:
ping @pedrobaeza @Christian-RB

HviorForgeFlow (approved):
code review

christian-ramos-tecnativa (approved):

PR #2754, opened 2025-07-14
[15.0][IMP] purchase_order_line_menu: apply purchase and done filter by default

JoanSForgeFlow:
[IMP] purchase_order_line_menu: apply purchase and done filter by default


JordiMForgeFlow (approved):

LoisRForgeFlow (approved):
Thanks :+1:

PR #2751, opened 2025-07-10
[15.0][ADD] purchase_invoice_status_line

JoanSForgeFlow:
[15.0][ADD] purchase_invoice_status_line

PR #2758, opened 2025-07-18
[15.0][ADD] purchase_uninvoiced_amount_force_invoiced_line

JoanSForgeFlow:
Glue module

Depends on: #2751 and #2757

## Changes
- Modified to set amount_uninvoiced to 0 when:
  - force_invoiced is True or`invoice_status` is 'invoiced'
  - Simplified calculation to always use `product_qty - qty_invoiced` regardless of purchase method

## Justification
This change ensures consistent behavior where:
1. `invoiced + uninvoiced = total_ordered` (no more "limbo" amounts)
2. Force-invoiced fully invoiced lines show 0 uninvoiced amount

The logic is now more intuitive and aligns with the principle that purchase orders track what was ordered vs what was invoiced, independent of receipt status.


purchase_uninvoiced_amount_force_invoiced_line/models/purchase_order_line.py
```diff
@@ -0,0 +1,32 @@
...
+        "force_invoiced",
+        "invoice_status",
+    )
+    def _compute_amount_uninvoiced(self):
+        super()._compute_amount_uninvoiced()
+        for line in self:
+            if line.force_invoiced or line.invoice_status == "invoiced":
+                line.amount_uninvoiced = 0.0
+            else:
+                qty = line.product_qty - line.qty_invoiced
```
AaronHForgeFlow, line 26: sorry for the laziness, but can you explain why this else condition is not covered in super()

AaronHForgeFlow (commented):
Code LG

purchase_uninvoiced_amount_force_invoiced_line/models/purchase_order_line.py
JoanSForgeFlow: Hi @AaronHForgeFlow,

Is in order to ensure invoiced + uninvoiced = ordered (ignore qty_received to avoid "limbo" amounts). This glue module adds a more "accounting" logic to the uninvoiced_amount

I will add a comment in the code. 

Thank you

AaronHForgeFlow: Ok! Thank you for the explanation.

AaronHForgeFlow (approved):
Code LGTM

RicardCForgeFlow (approved):

PR #2759, opened 2025-07-21
[16.0][FIX] purchase_delivery_split_date

DavidJForgeFlow:
When the PO is already confirmed and you add a new line with a different date than the existing one, the new receipt is created but the old receipt and move have the dates updated:

image
image
image
image

The call on the method _update_picking_from_group_key is the method that is messing it up as it updates the previous picking. It actually only updates the previous picking/move. But later in the execution, the picking/move is also updated, so the call on the method is not necessary. 

Also, this method is called by _purchase_location_by_line_ and neither _purchase_delivery_split_date_ nor _purchase_location_by_line_ misses this method as it's not called in any other place (at least I did not found any other case) and the tests are still working fine.

If the changes are approved and merged, I'll remove the hole method _update_picking_from_group_key from both appearances.


LoisRForgeFlow:
The case that is now not covered is updating the date of an existing line:

image


I think a test case for each scenario should be added:
1. update existing line date.
2. create a new line.

LoisRForgeFlow (approved):
Tested, it works good now. Thanks for the test cases.

LoisRForgeFlow:
@OCA/procurement-purchase-maintainers could you have a look?

HviorForgeFlow (approved):

LoisRForgeFlow:
@DavidJForgeFlow could you fwport this?

DavidJForgeFlow:
Forward ports done!

PR #2765, opened 2025-07-25
[18.0][FIX] purchase_advance_payment

DavidJForgeFlow:
The residual amount was not being computed correctly in the scenario where the payment generated does not have an account.move related (when the auto_post_advance_payments was disabled).

Also added test cases to ensure this not happens again.

@LoisRForgeFlow


LoisRForgeFlow (approved):
LGTM

ThiagoMForgeFlow (approved):

PR #2771, opened 2025-08-02
[16.0][FIX] purchase_order_supplierinfo_update: take company from PO line if not available in supplierinfo

giarve:


giarve:
@ernestotejeda Steps to reproduce:

1) enable multi company
2) leave the company field empty in supplierinfo
3) create and confirm a PO for that supplier and product
4) an error message appears because company id is not set

giarve:
Actually it is not working because there is no float compare for the price comparison it seems.

PR #2773, opened 2025-08-06
Update pre-migration.py

habbaba:
odoo.sh package error fix.


habbaba (commented):
please review the changes as of now on some odoo.sh servers are failing to update.

PR #2684, opened 2025-05-15
[18.0] [MIG] purchase_allowed_product: Migration to 18.0

Robrechtc:

PR #2776, opened 2025-08-26
[18.0] [MIG] purchase_allowed_product: Migration to 18.0 

apami0111:
- Updated _search method signature in product.product (removed access_rights_uid).
- Migrated views: replaced <tree> with <list> in account.move and purchase.order lines.
- Updated tests: migrated from TransactionCase → BaseCommon.
- Manifest bumped to version 18.0.1.0.0.


apami0111:
Supersedes: https://github.com/OCA/purchase-workflow/pull/2684

purchase_allowed_product/tests/test_purchase_allowed_product.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+from odoo.tests import Form
+
+from odoo.addons.base.tests.common import BaseCommon
+
+
+class TestPurchaseAllowedProduct(BaseCommon):
+    def setUp(self):
+        super().setUp()
```
carlos-lopez-tecnativa: Please change this method to use `setUpClass` instead of `setUp`.

purchase_allowed_product/views/account_move_views.xml
```diff
@@ -0,0 +1,43 @@
...
+        <field name="model">account.move</field>
+        <field name="inherit_id" ref="account.view_move_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='partner_id']/.." position="inside">
+                <field
+                    invisible="state != 'draft' or move_type not in ('in_invoice', 'in_refund', 'in_receipt')"
+                    name="use_only_supplied_product"
+                    widget="boolean_toggle"
+                />
+            </xpath>
```
carlos-lopez-tecnativa: This is not related to this PR, but I think it’s better to display the field label.

In V16, the field is displayed as:
image

In V17 and V18, this field does not display the label, so as a user I can be confused about what this field represents.
image

carlos-lopez-tecnativa (changes requested):

carlos-lopez-tecnativa (approved):
LGTM

TT57223
@pedrobaeza @victoralmau @AEstLo @dreispt could you please review this?

AEstLo (approved):
Code LGTM

purchase_allowed_product/tests/test_purchase_allowed_product.py
```diff
@@ -0,0 +1,78 @@
...
+from odoo.addons.base.tests.common import BaseCommon
+
+
+class TestPurchaseAllowedProduct(BaseCommon):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.supplierinfo_model = cls.env["product.supplierinfo"]
+        cls.product_model = cls.env["product.product"]
+        cls.partner_4 = cls.env.ref("base.res_partner_4")
```
victoralmau:
```suggestion
        cls.partner = cls.env["res.partner"].create({"name": "Test supplier"})
```
Let's try to avoid using demo data to prevent possible side effects.

victoralmau (changes requested):
Code review OK

purchase_allowed_product/tests/test_purchase_allowed_product.py
apami0111: Thanks @victoralmau for the suggestion. I updated the test to create the partner directly instead of using demo data, as recommended.

victoralmau (approved):
Code review OK

PR #2781, opened 2025-09-01
[17.0][MIG] purchase_reception_status: Split into base + glue modules for purchase_stock compatibility

JoanSForgeFlow:
This PR **supersedes** #2201 and implements the proposal discussed in https://github.com/OCA/purchase-workflow/pull/2201#issuecomment-3160414941

Odoo `purchase_stock` module has a native `receipt_status` field, which conflicts with the OCA's  approach in `purchase_reception_status`.

To avoid this conflict and follow the same approach as `sale_delivery_state`,  https://github.com/OCA/sale-workflow/pull/2770 the module has been split in two:
- A **base module** (`purchase_reception_status`) with the OCA  logic.
- A **glue module** (`purchase_stock_reception_status`) that override its native compute when `purchase_stock` is installed.


purchase_stock_reception_status/models/purchase_order.py
```diff
@@ -0,0 +1,15 @@
...
+
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+
+    @api.depends(
+        "state", "force_received", "order_line.qty_received", "order_line.product_qty"
+    )
+    def _compute_receipt_status(self):
+        # Use the OCA logic instead of the native picking-based logic
+        self._compute_oca_receipt_status()
```
yostashiro: This may be cleaner?

```suggestion
    # Use the OCA logic instead of the native picking-based logic
    receipt_status = fields.Selection(compute="_compute_oca_receipt_status")
```

JoanSForgeFlow: Hi @yostashiro,

Thank you. comments attended.

yostashiro (approved):
Code review and functional test. LGTM. 👍

AaronHForgeFlow (approved):
Ok with the split, thank you. Functional is good. Just super nitpicking, if you can add the widget toggle_button in the PO form for forcing the reception it'll be nice.

JoanSForgeFlow:
@AaronHForgeFlow Thank you. Comment attended!

BhaveshHeliconia (approved):
Looks good to me!

MiquelRForgeFlow (approved):

HviorForgeFlow (approved):

JordiMForgeFlow:
@JoanSForgeFlow @HviorForgeFlow shouldn't migration scripts be added for the change in the field name?

yostashiro:
> shouldn't migration scripts be added for the change in the field name?

Yes, good catch. Something equivalent to https://github.com/OCA/sale-workflow/blob/16.0/sale_stock_delivery_state/hooks.py should be good, indeed.

PR #2784, opened 2025-09-03
[16.0][MIG] purchase_stock_picking_show_currency_rate: Migration to 16.0

AungKoKoLin1997:
This PR includes:

- Migration to 16.0
- Adds an option to display the exchange rate either as company currency per unit or unit per company currency.

Example (company currency: JPY; 1 USD = 150 JPY):

(Default) Unit per Company Currency → 0.0066 (USD/JPY)
Company Currency per Unit → 150 (JPY/USD)

@qrtl QT5684


purchase_stock_picking_show_currency_rate/models/stock_picking.py
```diff
@@ -0,0 +1,76 @@
...
+        return fields.Date.context_today(self)
+
+    @api.depends(
+        "purchase_currency_id",
+        "purchase_id.invoice_ids.state",
+        "show_currency_rate_amount",
+        "company_id",
+        "state",
+    )
+    def _compute_currency_rate_amount(self):
```
yostashiro, line 42: Something appears to be wrong with the computation as I tried the functionality in runboat. What I expect to see is 87.396167 but what actually shows is 133.62 which seems to be the rate per EUR (the company currency is USD).

image

image

Also, we should get the rate from the transaction for "done" pickings (or record the rate when picking is validated), not from the rate master.

purchase_stock_picking_show_currency_rate/readme/CONFIGURE.rst
```diff
@@ -7,3 +7,14 @@ Rates for non-company currencies must be available and be distinct to 1.0. For e
...
+To set up the date field used for calculating the currency exchange rate on pickings:
+
+#. Navigate to *Inventory > Configuration > Settings*.
+#. Find and set the 'Picking Date Field for Currency Rate' to be used in calculating 
+   the currency exchange rate for completed pickings.
+
+To change the currency rate display style on pickings:
+
+- Go to Accounting (or Invoicing) → Configuration → Settings.
+- In the Picking Rate Display Type section, select the default display type for all currencies.
```
yostashiro: It doesn't make much sense to have this config in accounting. It should be moved next to (before) 'Picking Date Field for Currency Rate'.

Please also make the list style consistent with the previous sections.

purchase_stock_picking_show_currency_rate/models/stock_picking.py
```diff
@@ -25,11 +29,48 @@ def _compute_show_currency_rate_amount(self):
                 and item.purchase_id.currency_id != item.company_id.currency_id
             )
 
-    @api.depends("purchase_currency_id", "show_currency_rate_amount", "company_id")
+    def _get_rate_date(self):
```
yostashiro, line 25: This method only matters for pickings not done.

yostashiro (changes requested):

purchase_stock_picking_show_currency_rate/models/stock_picking.py
AungKoKoLin1997: > What I expect to see is 87.396167 but what actually shows is 133.62 which seems to be the rate per EUR (the company currency is USD).

@yostashiro I just realized that is happened because of wrong `technical rate` in base company.
I think it is because of demo data but can't change the same value with `Unit per USD` in base company for some reason.
In new or other companies, this field value is the same with `Unit per USD` field.

https://github.com/odoo/odoo/blob/ff396d4877d6e42c612b8f78cf15efe2b32ebbd9/odoo/addons/base/data/res_currency_rate_demo.xml#L154-L158

Screenshot 2025-09-09 113501

Screenshot 2025-09-09 113554

purchase_stock_picking_show_currency_rate/models/stock_move.py
```diff
@@ -0,0 +1,12 @@
...
+
+from odoo import fields, models
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    price_unit_currency = fields.Float(
+        help="Techinal field to store the price unit of purchase order."
+    )
```
yostashiro: Wouldn't it be better to extend _action_done() of stock.picking to keep the rate in a technical field (e.g., rate_at_done)? That way we don't need to touch stock.move and purchase.order.line models.

AungKoKoLin1997: @yostashiro Can you please review my latest approach?
I didn't use stock.move and purchase.order.line models and any extra methods.
I make it complete with just compute method.

purchase_stock_picking_show_currency_rate/models/stock_picking.py
```diff
@@ -0,0 +1,96 @@
...
+                price_unit_positive = sum(
+                    [abs(price) for price in move_ids.mapped("price_unit")]
+                )
+                # Assume `date_order` is used to compute the exchange rate applied to the
+                # move line's `price_unit` via `_get_stock_move_price_unit`.
+                rates = item.purchase_currency_id._get_rates(
+                    item.company_id, item.purchase_id.date_order
+                )
+                price_unit_rate = rates.get(item.purchase_currency_id.id)
+                base = price_unit_positive * price_unit_rate
```
yostashiro: We can't rely on this logic as `price_unit` of stock move gets updated even after confirming the purchase order when user updates the unit price of the order line. We also don't want to rely on the rate master.

I still stand by my earlier suggestion to add a technical field that captures the rate when the receipt is validated. The logic should be much simpler that way.

purchase_stock_picking_show_currency_rate/models/stock_picking.py
yostashiro: I think we can drop this method and related config if we keep the rate in a technical field.

purchase_stock_picking_show_currency_rate/models/stock_picking.py
```diff
@@ -20,6 +20,12 @@ class StockPicking(models.Model):
         compute="_compute_currency_rate_amount",
         digits=0,
     )
+    rate_date_at_done = fields.Date(
+        help="Technical Field to store the rate date at done."
+    )
```
yostashiro: @AungKoKoLin1997 What's the advantage of keeping the date instead of the actual rate? I've been suggesting that we should keep the actual rate (as we don't want to rely on the rate master for done transactions). Please provide your reasons if you think otherwise.

purchase_stock_picking_show_currency_rate/models/res_currency.py
```diff
@@ -0,0 +1,16 @@
...
+class ResCurrency(models.Model):
+    _inherit = "res.currency"
+
+    @api.model
+    def _get_conversion_rate(self, from_currency, to_currency, company, date):
+        res = super()._get_conversion_rate(from_currency, to_currency, company, date)
+        picking = self.env.context.get("picking")
+        if picking:
+            picking.rate_at_done = res
+        return res
```
yostashiro: This approach doesn't look optimal. I think we should keep it simple by just extending _action_done() of the picking. res_currency.py and stock_move.py can be removed.

purchase_stock_picking_show_currency_rate/models/stock_picking.py
```diff
@@ -29,22 +39,36 @@ def _compute_show_currency_rate_amount(self):
                 and item.purchase_id.currency_id != item.company_id.currency_id
             )
 
-    @api.depends("purchase_currency_id", "show_currency_rate_amount", "company_id")
+    @api.depends(
+        "rate_at_done",
+        "purchase_currency_id",
+        "show_currency_rate_amount",
+        "company_id",
+    )
```
yostashiro: It doesn't seem necessary to have these trigger fields.

yostashiro (approved):
Code review and functional test.

yostashiro (approved):
LGTM. We may squash the last two commits or add the module name in the last commit message.

kanda999 (commented):
Code and functional review

dreispt (approved):

AungKoKoLin1997:
@dreispt Regarding ocabot failed, is there anything I need to follow up on?

HviorForgeFlow (approved):

PR #2788, opened 2025-09-05
[18.0][IMP] purchase_deposit: Do not allow register deposit over PO amount

TheerayutEncoder:
**Improvement:**

If the user tries to register a deposit over the PO amount (including a percentage and a fixed amount)
image

This code will raise UserError as below
image

### DIFF
--- a/purchase_deposit/wizard/purchase_make_invoice_advance.py
+++ b/purchase_deposit/wizard/purchase_make_invoice_advance.py
@@ -5,9 +5,9 @@
 import time
 from datetime import datetime
 
-from odoo import _, api, fields, models
+from odoo import Command, api, fields, models
 from odoo.exceptions import UserError
-from odoo.tools import DEFAULT_SERVER_DATETIME_FORMAT
+from odoo.tools import DEFAULT_SERVER_DATETIME_FORMAT, float_compare
 
 
 class PurchaseAdvancePaymentInv(models.TransientModel):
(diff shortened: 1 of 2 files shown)


TheerayutEncoder:
Dear @AaronHForgeFlow 

Could I get your thoughts on this?

purchase_deposit/wizard/purchase_make_invoice_advance.py
```diff
@@ -169,6 +169,23 @@ def create_invoices(self):
...
+                sum(
+                    line.price_unit
+                    for line in order.order_line
+                    if (line.is_deposit and line.invoice_lines)
+                )
+                or 0.00
+            )
+            if (sum_deposit + amount) > order.amount_total:
+                raise UserError(
+                    _(
```
Saran440:
```suggestion
                    self.env._(
```

Saran440: - Should we change all file from `_` to `self.env._` ?
- use `f""` instead `%`

purchase_deposit/wizard/purchase_make_invoice_advance.py
```diff
@@ -169,6 +169,23 @@ def create_invoices(self):
             amount = self.amount
             if self.advance_payment_method == "percentage":  # Case percent
                 amount = self.amount / 100 * order.amount_untaxed
+            # calculate all deposit lines
+            sum_deposit = (
+                sum(
+                    line.price_unit
```
Saran440, line 177:
```suggestion
                    line.price_unit
```
I'm not sure price_unit is include / exclude tax.
Maybe we should have config to check include or exclude tax. What do you think?

purchase_deposit/wizard/purchase_make_invoice_advance.py
```diff
@@ -169,6 +169,23 @@ def create_invoices(self):
...
+                    for line in order.order_line
+                    if (line.is_deposit and line.invoice_lines)
+                )
+                or 0.00
+            )
+            if (sum_deposit + amount) > order.amount_total:
```
Saran440:
```suggestion
            if (sum_deposit + amount) > order.amount_total:
```

Use `float_compare` instead.

Saran440 (commented):
I agree to check condition.
Have a little bit comment code

TheerayutEncoder:
Thanks @Saran440 I think this should be ready for review!

Saran440 (approved):
Code Review 👍

BhaveshHeliconia (approved):
LGTM!

PR #2792, opened 2025-09-08
[16.0][MIG] purchase_invoice_status_line

JoanSForgeFlow:
Standard migration of purchase_invoice_status_line to 16.0

Test failing because same name for a field has been added for different purpose . This PRs fixes the problem: https://github.com/OCA/purchase-workflow/pull/2793


purchase_invoice_status_line/hooks.py
```diff
@@ -1,18 +0,0 @@
-"""Post initialization hooks for purchase_invoice_status_line module."""
-
-from odoo import SUPERUSER_ID, api
-
-
-def post_init_hook(cr, registry):
```
LoisRForgeFlow: why removing the init hook, not useful anymore?

JoanSForgeFlow: This is something that always makes me doubt a bit. If you are coming from v15, there are no changes here — meaning there is nothing to update or rename during migration.  

Also, that script is meant to run only when the module has never been installed before, and it does:  

```
lines_to_update = env["purchase.order.line"].search(
    [
        ("force_invoiced", "=", False),
        ("order_id.force_invoiced", "=", True),
    ]
)
lines_to_update.write({"force_invoiced": True})
```
It sets force_invoiced = True on all lines where the order is force_invoiced.
However, once you start using it, you may have a purchase order with force_invoiced = True but only some lines marked as force_invoiced. For that reason, I think it’s better to let whatever data you have in v15 carry over directly to v16 without running this script.

LoisRForgeFlow: This hook is only meant for installing the module, it doesn't matter which version you are or if you are migrating. It is only for installation, if someone start to use this module in v16 for the first time, I think we hook still makes sense, isn't it?

JoanSForgeFlow:
@HviorForgeFlow  Could you check this migration?

Thank you

purchase_invoice_status_line/hooks.py
HviorForgeFlow: Hook is added again

HviorForgeFlow (approved):

DavidJForgeFlow (approved):

PR #2793, opened 2025-09-09
[16.0][UPD] purchase_force_invoiced_quantity : rename force_invoiced to force_invoiced_qty

JoanSForgeFlow:
This PR adds a pre-migration script to rename the field `force_invoiced` to `force_invoiced_qty` in `purchase_order_line` in this module.
The reason is to avoid having two fields with the same name across different modules.   https://github.com/OCA/purchase-workflow/pull/2792
Since the field stores a float value, it is more appropriate to use the `_qty` suffix in this one.

As this field was originally introduced by @dessanhemrayev , tagging you here in case you’d like to review and confirm that this change makes sense from your perspective.

### DIFF
--- a/purchase_force_invoiced_quantity/migrations/16.0.1.1.0/pre-migration.py
+++ b/purchase_force_invoiced_quantity/migrations/16.0.1.1.0/pre-migration.py
@@ -1,0 +1,58 @@
+# Copyright 2025 ForgeFlow
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+import logging
+
+_logger = logging.getLogger(__name__)
+
+
+def migrate_force_invoiced_to_force_invoiced_qty(cr):
+    cr.execute(
+        """
+        SELECT column_name
+        FROM information_schema.columns
+        WHERE table_name = 'purchase_order_line'
+        AND column_name = 'force_invoiced'
+        """
+    )
+    old_field_exists = cr.fetchone()
+    cr.execute(
+        """
+        SELECT column_name
+        FROM information_schema.columns
+        WHERE table_name = 'purchase_order_line'
+        AND column_name = 'force_invoiced_qty'
+        """
+    )
+    new_field_exists = cr.fetchone() is not None
+    if old_field_exists and not new_field_exists:
+        _logger.info(
+            "Migrating force_invoiced field to force_invoiced_qty in purchase_order_line"
+        )
+        cr.execute(
+            """
+            ALTER TABLE purchase_order_line
+            ADD COLUMN IF NOT EXISTS force_invoiced_qty numeric DEFAULT 0.0
+            """
+        )
+        cr.execute(
+            """
+            UPDATE purchase_order_line
+            SET force_invoiced_qty = COALESCE(force_invoiced, 0.0)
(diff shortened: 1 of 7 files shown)


LoisRForgeFlow:
@JoanSForgeFlow there is a small typo on the title "purchase_force_invoiced_quanity"

purchase_force_invoiced_quantity/migrations/16.0.1.1.0/pre-migration.py
```diff
@@ -0,0 +1,46 @@
...
+def migrate_force_invoiced_to_force_invoiced_qty(cr):
+    cr.execute(
+        """
+        SELECT column_name
+        FROM information_schema.columns
+        WHERE table_name = 'purchase_order_line'
+        AND column_name = 'force_invoiced'
+        """
+    )
+    if cr.fetchone():
```
LoisRForgeFlow: I would also check if force_invoiced_qty exist, if it exist, also no need to do anything. At the end of the day you are going to take the field name force_invoiced in other module

JoanSForgeFlow: thanks for your advice. Comment attended!

LoisRForgeFlow: That looks better, thanks!

LoisRForgeFlow (approved):
LGTM :+1:

purchase_force_invoiced_quantity/migrations/16.0.1.1.0/pre-migration.py
```diff
@@ -0,0 +1,58 @@
...
+def migrate_force_invoiced_to_force_invoiced_qty(cr):
+    cr.execute(
+        """
+        SELECT column_name
+        FROM information_schema.columns
+        WHERE table_name = 'purchase_order_line'
+        AND column_name = 'force_invoiced'
+        """
+    )
+    old_field_exists = cr.fetchone()
```
MiquelRForgeFlow: You could use `openupgrade.column_exists(env.cr, "purchase_order_line", "force_invoiced")` or `odoo.tools.sql.column_exists(env.cr, "purchase_order_line", "force_invoiced")` instead.

purchase_force_invoiced_quantity/migrations/16.0.1.1.0/pre-migration.py
```diff
@@ -0,0 +1,58 @@
...
+            ADD COLUMN IF NOT EXISTS force_invoiced_qty numeric DEFAULT 0.0
+            """
+        )
+        cr.execute(
+            """
+            UPDATE purchase_order_line
+            SET force_invoiced_qty = COALESCE(force_invoiced, 0.0)
+            WHERE force_invoiced IS NOT NULL
+            """
+        )
```
MiquelRForgeFlow: Why not rename the column?

JoanSForgeFlow:
Thanks @MiquelRForgeFlow  ! I’ve updated it following your suggestion. Could you check if it looks good now?

MiquelRForgeFlow (approved):

PR #2795, opened 2025-09-09
[17.0][MIG] purchase_reception_status_line

JoanSForgeFlow:
Migration of `purchase_reception_status_line`


AaronHForgeFlow:
I am not sure about the dependency, should it be added to previous versions as well?

JoanSForgeFlow:
Hi @AaronHForgeFlow,

From my perspective, previous versions could also benefit from having this dependency, since the receipt status is indeed managed in the PO line tree view.  

That said, it’s not something strictly required — the previous versions remain valid without it. I just believe it makes sense to include it here for consistency, but in earlier versions it can simply be installed separately if needed.

Do you suggest we remove it here, or rather add it to the other versions as well? Maybe add it in a separate PR?

AaronHForgeFlow:
If the dependency is not strictly technical, then I would not add the dependency. Maybe put in the readme that the other module is highly recommended.

JoanSForgeFlow:
I agree, done. Thank you @AaronHForgeFlow

AaronHForgeFlow (approved):
code review :+1:

JordiMForgeFlow:
code LGTM, however I think migration scripts are needed for the field name change

JoanSForgeFlow:
@JordiMForgeFlow thanks for your review. Comments attended!

purchase_reception_status_line/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 
 {
     "name": "Purchase Reception Status Line",
-    "version": "17.0.1.0.0",
+    "version": "17.0.1.1.0",
```
JordiMForgeFlow: as the module is not yet merged I think you don't need to bump a new version, you can trigger the scripts for the 17.0.1.0.0

purchase_reception_status_line/migrations/17.0.1.1.0/pre-migration.py
```diff
@@ -0,0 +1,19 @@
...
+        "reception_status",
+        "receipt_status",
+    ),
+]
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    if not openupgrade.column_exists(env.cr, "purchase_order_line", "receipt_status"):
+        openupgrade.rename_fields(env, _field_renames)
```
JordiMForgeFlow: note that the options in the selection field are not the same, so it is not simply renaming the field but also transitioning correctly the new values

JordiMForgeFlow (approved):

HviorForgeFlow (approved):

BhaveshHeliconia (approved):
LGTM!

PR #2816, opened 2025-09-17
[16.0][FIX] Avoid repetition of lines at reception

CILC98:
This change prevents the creation of lines when receiving products. This will also prevent inventory valuation from being corrupted, such that if there is more than one line (stock.move), the valuation will be incorrect when the purchase order has been invoiced previously.

Before
https://github.com/user-attachments/assets/64041144-a51f-47d4-a401-bc69f9844341

After
https://github.com/user-attachments/assets/be2eb62a-dff3-4794-8bf9-ef5c509cfd73

### DIFF
--- a/purchase_request/tests/test_purchase_request.py
+++ b/purchase_request/tests/test_purchase_request.py
@@ -118,6 +118,7 @@ class TestPurchaseRequest(TransactionCase):
         with self.assertRaises(UserError):
             purchase_request_line.unlink()
         purchase = purchase_request_line.purchase_lines.order_id
+        purchase_request_line.purchase_lines.price_unit = 100.0
         purchase.button_done()
         self.assertEqual(purchase.state, "done")
 
--- a/purchase_request/tests/test_purchase_request_allocation.py
+++ b/purchase_request/tests/test_purchase_request_allocation.py
@@ -70,10 +70,11 @@ class TestPurchaseRequestToRfq(common.TransactionCase):
             active_ids=[purchase_request_line1.id, purchase_request_line2.id],
         ).create(vals)
         wiz_id.make_purchase_order()
+        (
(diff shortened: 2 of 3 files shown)


CILC98:

pedrobaeza:
Tests are in red.

CILC98:
@pedrobaeza Hi Pedro, there's a constraint in the [purchase_order_no_zero_price](https://github.com/OCA/purchase-workflow/blob/16.0/purchase_order_no_zero_price/models/purchase_order_line.py) module that's preventing the tests from running properly. In my opinion, this module shouldn't be involved in the purchase_request module tests. Could you help me by telling me why this might be happening? I'd be very grateful in advance.

pedrobaeza:
Well, why don't you assure your PO lines are not 0 price unit in the tests to avoid it?

CILC98:
@pedrobaeza I followed your recommendation and it worked. Thank you. Can you review it please?

Is smiliar to #2819 but the time zone is not a problem

purchase_request/tests/test_purchase_request.py
```diff
@@ -117,6 +117,8 @@ def test_purchase_request_status(self):
         # Unlink purchase_lines from state approved
         with self.assertRaises(UserError):
             purchase_request_line.unlink()
+        purchase = purchase_request_line.purchase_lines.write({"price_unit": 100.0})
+        self.assertEqual(purchase, True)
```
pedrobaeza: And the rest for the same.

```suggestion
        purchase_request_line.purchase_lines.price_unit = 100.0
```

CILC98: Thanks, I've already adjusted it.

CILC98:
Hi @victoralmau this PR is similar to #2819 which one you helped me. Can you help again with the review please?

PR #2787, opened 2025-09-04
[18.0][FIX] purchase_request: merging two lines had wrong quantities

AaronHForgeFlow:
FWP of https://github.com/OCA/purchase-workflow/pull/2616

That PR is not merged yet, but IMHO it is an important bug and I am creating the FWP in advance

PR #2819, opened 2025-09-17
[18.0][FIX] Duplicate stock.move damages inventory valuation

CILC98:
This change prevents the creation of lines when receiving products. This will also prevent inventory valuation from being corrupted, such that if there is more than one line (stock.move), the valuation will be incorrect when the purchase order has been invoiced previously.

Before

https://github.com/user-attachments/assets/006491c1-7d10-4df4-9231-daf382b834ee

After

https://github.com/user-attachments/assets/eb35eb3e-bae2-4963-8e52-63d64f034b73

### DIFF
--- a/purchase_request/wizard/purchase_request_line_make_purchase_order.py
+++ b/purchase_request/wizard/purchase_request_line_make_purchase_order.py
@@ -150,6 +150,14 @@ class PurchaseRequestLineMakePurchaseOrder(models.TransientModel):
         }
         return self.env["purchase.request.allocation"].create(vals)
 
+    def _get_date_with_user_tz(self, date):
+        user_tz = pytz.timezone(self.env.user.tz or "UTC")
+        return (
+            user_tz.localize(datetime(date.year, date.month, date.day))
+            .astimezone(pytz.utc)
+            .replace(tzinfo=None)
+        )
+
     @api.model
     def _prepare_purchase_order_line(self, po, item):
         if not item.product_id:
@@ -173,9 +181,7 @@ class PurchaseRequestLineMakePurchaseOrder(models.TransientModel):
             "product_qty": qty,
             "analytic_distribution": item.line_id.analytic_distribution,
             "purchase_request_lines": [(4, item.line_id.id)],
-            "date_planned": datetime(
-                date_required.year, date_required.month, date_required.day
-            ),
+            "date_planned": self._get_date_with_user_tz(date_required),
             "move_dest_ids": [(4, x.id) for x in item.line_id.move_dest_ids],
         }
 
@@ -214,13 +220,7 @@ class PurchaseRequestLineMakePurchaseOrder(models.TransientModel):
         if self.sync_data_planned:
             date_required = item.line_id.date_required
             order_line_data += [
-                (
-                    "date_planned",
-                    "=",
-                    datetime(
-                        date_required.year, date_required.month, date_required.day
-                    ),
-                )
+                ("date_planned", "=", self._get_date_with_user_tz(date_required))
             ]
         if not item.product_id:
             order_line_data.append(("name", "=", item.name))
@@ -231,7 +231,6 @@ class PurchaseRequestLineMakePurchaseOrder(models.TransientModel):
         purchase_obj = self.env["purchase.order"]
         po_line_obj = self.env["purchase.order.line"]
         pr_line_obj = self.env["purchase.request.line"]
-        user_tz = pytz.timezone(self.env.user.tz or "UTC")
         purchase = False
 
         for item in self.item_ids:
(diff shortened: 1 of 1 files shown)


CILC98:
@TheerayutEncoder Please could you check the fix for this bug please?

TheerayutEncoder:
Hi @CILC98 

It look like they already fixed.

Could you check this I'm not sure does it relate to this PR https://github.com/OCA/purchase-workflow/pull/2787#issuecomment-3275241377

CILC98:
Hi @TheerayutEncoder, thanks for your reply.

I don't think it's the same issue. I'm trying to correct the valuation when, in the receipt, two lines of the same product appear associated with a purchase order line that has been fully invoiced.

Could you please watch the first video in the description?

CILC98:

pedrobaeza:
@victoralmau can you check?

purchase_request/wizard/purchase_request_line_make_purchase_order.py
```diff
@@ -291,7 +291,9 @@ def make_purchase_order(self):
             new_qty = pr_line_obj._calc_new_qty(
                 line, po_line=po_line, new_pr_line=new_pr_line
             )
-            po_line.product_qty = new_qty
+            po_line.write(
+                {"product_qty": new_qty, "date_planned": fields.Datetime.now()}
+            )
```
victoralmau:
```suggestion
            if not po_line_data:
                po_line_data = self._prepare_purchase_order_line(purchase, item)
            po_line.write({
                "product_qty": new_qty,
                "date_planned": po_line_data["date_planned"]
            })
```

victoralmau (changes requested):
Looking at it in detail, I see that the problem is as follows:

- The way `date_planned` is defined in https://github.com/OCA/purchase-workflow/blob/18.0/purchase_request/wizard/purchase_request_line_make_purchase_order.py# L300 should be the one used in https://github.com/OCA/purchase-workflow/blob/18.0/purchase_request/wizard/purchase_request_line_make_purchase_order.py#L176.
- The quantity assignment in https://github.com/OCA/purchase-workflow/blob/18.0/purchase_request/wizard/purchase_request_line_make_purchase_order.py# L294 and `date_planned` in https://github.com/OCA/purchase-workflow/blob/18.0/purchase_request/wizard/purchase_request_line_make_purchase_order.py#L300C21-L300C33 should be removed in favor of what I have suggested.

The code may seem a little confusing, but the goal is to always define the same `date_planned` value to avoid side effects.

CILC98:
@victoralmau Thanks!!! I have accepted your suggestion.

victoralmau (changes requested):
Can you merge all commits into one?

Image

The appropriate commit could be: _[FIX] purchase_reqest: Duplicate stock.move damages inventory valuation_

CILC98:
@victoralmau  ok, is done

victoralmau (changes requested):
https://github.com/OCA/purchase-workflow/blob/18.0/purchase_request/wizard/purchase_request_line_make_purchase_order.py#L295-L306

This part is no longer necessary because we are already defining the value of `date_planned` earlier.

It may be appropriate to modify the `_prepare_purchase_order_line()` method and define the value of `date_planned` there exactly as it was defined here:
```

user_tz.localize(
                    datetime(date_required.year, date_required.month, date_required.day)
                )
                .astimezone(pytz.utc)
                .replace(tzinfo=None)
```

CILC98:
I think the date based on the user's time zone is important. I combined our solution with date assignment. What do you think? @victoralmau

CILC98:
@victoralmau Hi!! Can you help me please?

victoralmau:
According to my suggestion https://github.com/OCA/purchase-workflow/pull/2819#pullrequestreview-3266076110, the change I believe would be appropriate is https://github.com/Tecnativa/purchase-workflow/commit/35d73f854c6f1f0207c380df7246dcc648c1f4ea

CILC98:
@victoralmau Ok, but you call a method for a date. What do you thing about create new method to change de date with the user time zone? This new method can be use in other issues

victoralmau (approved):

carlos-lopez-tecnativa (changes requested):
I think this method needs to be adapted according to this change:

https://github.com/OCA/purchase-workflow/blob/6953cc5a3c1d5b02f692799615153d4e4b3c4710/purchase_request/wizard/purchase_request_line_make_purchase_order.py#L214-L224

When I select this field in the wizard, I expect that if a purchase order line with the same product and date exists, the quantity will be updated in that line.
However, this does not currently happen, a new line is created instead.
Please review this as well.

image

image

Additionally, please squash the commits.

CILC98:
@victoralmau @carlos-lopez-tecnativa Problem fixed! Can you help me by reviewing it please?

carlos-lopez-tecnativa (approved):
LGTM

CILC98:
@HviorForgeFlow Hi! Maybe you could help me? This PR is ready

PR #2822, opened 2025-09-22
[17.0][FIX] purchase_last_price_info: filter last purchase

david-s73:
- A change has been made to how the last purchase date was filtered, as in some cases it was not taking the last date but the first, so I have modified how the date is filtered.

### DIFF
--- a/purchase_last_price_info/models/product_product.py
+++ b/purchase_last_price_info/models/product_product.py
@@ -48,14 +48,16 @@ class ProductProduct(models.Model):
     @api.depends("last_purchase_line_ids.state")
     def _compute_last_purchase_line_id(self):
         for item in self:
-            item.last_purchase_line_id = fields.first(
-                item.last_purchase_line_ids.sudo().filtered_domain(
-                    [
-                        ("state", "in", ["purchase", "done"]),
-                        ("company_id", "in", self.env.companies.ids),
-                    ]
-                )
+            candidate_lines = item.last_purchase_line_ids.sudo().filtered_domain(
+                [
+                    ("state", "in", ["purchase", "done"]),
+                    ("company_id", "in", self.env.companies.ids),
+                ]
             )
+            sorted_lines = candidate_lines.sorted(
+                key=lambda r: (r.date_order, r.id), reverse=True
+            )
+            item.last_purchase_line_id = fields.first(sorted_lines)
 
     @api.depends("last_purchase_line_id")
     def _compute_last_purchase_line_id_info(self):
--- a/purchase_last_price_info/tests/test_purchase_last_price_info.py
+++ b/purchase_last_price_info/tests/test_purchase_last_price_info.py
@@ -2,6 +2,8 @@
 # Copyright 2021 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+import datetime
+
 import odoo.tests.common as common
 from odoo import fields
 
@@ -16,6 +18,16 @@ class TestPurchaseLastPriceInfo(common.TransactionCase):
         self.purchase_model = self.env["purchase.order"]
         self.purchase_line_model = self.env["purchase.order.line"]
         self.product = self.env.ref("product.consu_delivery_01")
+        self.product1 = self.env["product.product"].create(
+            {
+                "name": "Test product",
+                "type": "product",
+                "standard_price": 10.0,
+                "list_price": 20.0,
+                "uom_id": self.env.ref("uom.product_uom_unit").id,
+                "uom_po_id": self.env.ref("uom.product_uom_unit").id,
+            }
+        )
         self.partner = self.env.ref("base.res_partner_1")
         # Create custom rates to currency + currency_extra
         self._create_currency_rate(self.currency, "2000-01-01", 1.0)
@@ -57,7 +69,7 @@ class TestPurchaseLastPriceInfo(common.TransactionCase):
         self.assertEqual(self.product.last_purchase_price_currency, 1.0)
 
     def test_purchase_last_price_info_new_order(self):
-        purchase_order = self.purchase_model.create(
+        purchase_order1 = self.purchase_model.create(
             {
                 "date_order": "2000-01-01",
                 "currency_id": self.currency_extra.id,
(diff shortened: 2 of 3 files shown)


LoisRForgeFlow (changes requested):
Thanks @david-s73 for contributing. Could you please add a test case covering your fix?

david-s73:
@LoisRForgeFlow What do you mean? Once the purchase order is confirmed in the tests, it will be added to my changes.

LoisRForgeFlow:
@david-s73 I'm referring to add a new automated test case (https://github.com/OCA/purchase-workflow/blob/17.0/purchase_last_price_info/tests/test_purchase_last_price_info.py). The test case should be such, that it fails without your code fix and passes with your patch.

david-s73:
@LoisRForgeFlow Regarding what you mentioned, I'm not sure I can reproduce it in the tests, since it only happened in some cases and I'm not quite sure why it happened.

LoisRForgeFlow:
@david-s73 If you cannot reproduce it, what am I supposed to test? I don't think we can get this merged without a way to reproduce the issue being fixed.

david-s73:
@LoisRForgeFlow I have been able to reproduce the error, and it comes from the “purchase_order_line_sequence” module, as I have tested it with a completely clean database and once you install the module I just mentioned, it changes the sequence of the purchase order lines and does not take the last purchase date.

LoisRForgeFlow:
@david-s73 That's great! then you can now write an automated test that represents exactly what you did functionally and check that the PO line used to get the last price is the expected one.

david-s73:
@LoisRForgeFlow In order to reproduce the case in the tests, you have to rely on the “purchase_order_line_sequence” module, and it doesn't seem quite right to have to rely on it.

LoisRForgeFlow:
@david-s73 You can create a normal test case, that as soon purchase_order_line_sequence is installed (even without dependency), the test fails.

david-s73:
@LoisRForgeFlow I have already run the tests, reproducing what happens when you have the “purchase_order_line_sequence” module installed, as well as filtering the tests to show that using “fields.first” does not work correctly and that it is better to check the purchase order date.

david-s73:
@LoisRForgeFlow You can check the pr now since I've done the tests you asked me to do.

purchase_last_price_info/tests/test_purchase_last_price_info.py
```diff
@@ -67,30 +77,81 @@ def test_purchase_last_price_info_new_order(self):
...
+        candidate_lines = self.product1.last_purchase_line_ids.sudo().filtered_domain(
+            [
+                ("state", "in", ["purchase", "done"]),
+                ("company_id", "in", self.env.companies.ids),
+            ]
+        )
+        sorted_lines = candidate_lines.sorted(
+            key=lambda r: (r.date_order, r.id), reverse=True
+        )
+        real_last_purchase_line = fields.first(sorted_lines)
```
LoisRForgeFlow: @david-s73 Thanks for adding the tests. However, you should not execute the same code you are adding in the compute to get the expected value, obviously they will match. You are defining here a business case so you can simply set and `expected` value and compare.

david-s73: @LoisRForgeFlow I'm not quite sure what you mean, because in the tests I did, the case was malfunctioning, and with my change it works correctly.

david-s73:
@LoisRForgeFlow You can review the pr

david-s73:
@LoisRForgeFlow You can review the pr now after the changes.

purchase_last_price_info/tests/test_purchase_last_price_info.py
LoisRForgeFlow: @david-s73 The issue is that the test reuses the same production logic, because it compute both the actual result and the expected result. This means the test can never fail if that logic is wrong

That makes the test tautological, not a verification of behavior.

image

image

It is like doing this:

```
def add(a,b)
    return a + b
```

```
def test_add()
    res = add(2,3)
    expected = 2 + 3
    assertequal(res, expected
```

You are computing the expected result the same way you compute the function. It is better conceptually to signal the expected result yourself:

```
def test_add()
    res = add(2,3)
    expected = 5
    assertequal(res, expected
```

LoisRForgeFlow:
@david-s73 Sorry for missing your reply, I have answered in the thread https://github.com/OCA/purchase-workflow/pull/2822#discussion_r2713215908. I hope this clarifies my point.

purchase_last_price_info/tests/test_purchase_last_price_info.py
david-s73: @LoisRForgeFlow I have already made some changes to the tests as you requested in your comment.

david-s73:
@LoisRForgeFlow Don't worry, you've made it clear what you wanted to tell me, and I hope my changes are now correct, or at least what you expected.

purchase_last_price_info/tests/test_purchase_last_price_info.py
```diff
@@ -67,30 +77,72 @@ def test_purchase_last_price_info_new_order(self):
...
+                    )
+                ],
+            }
+        )
+        purchase_order1.button_confirm()
+        purchase_order2.button_confirm()
         self.assertEqual(
-            fields.Datetime.from_string(purchase_order.date_order).date(),
-            fields.Datetime.from_string(self.product.last_purchase_date).date(),
+            fields.Datetime.from_string(purchase_order2.date_order).date(),
```
LoisRForgeFlow: All this `fields.Datetime.from_string(` are unneded I would say, You can directly compare dates

purchase_last_price_info/models/product_product.py
```diff
@@ -48,14 +48,16 @@ class ProductProduct(models.Model):
...
     def _compute_last_purchase_line_id(self):
         for item in self:
-            item.last_purchase_line_id = fields.first(
-                item.last_purchase_line_ids.sudo().filtered_domain(
-                    [
-                        ("state", "in", ["purchase", "done"]),
-                        ("company_id", "in", self.env.companies.ids),
-                    ]
-                )
+            candidate_lines = item.last_purchase_line_ids.sudo().filtered_domain(
```
LoisRForgeFlow, line 51: You will need to apply the same patch to product.template (https://github.com/OCA/purchase-workflow/blob/d987b5893cb14ab092929c1d9de7f26eb6143ffe/purchase_last_price_info/models/product_template.py#L52), right?

LoisRForgeFlow (commented):
Looks better! Some more comments

david-s73:
@LoisRForgeFlow The changes you requested have been made. Please check if everything is correct now.

LoisRForgeFlow (approved):
Thanks! LGTM :+1:

Reyes4711-S73 (approved):
LGTM

david-s73:
@OCA/procurement-purchase-maintainers You can review and merge the pr

HviorForgeFlow (approved):

LoisRForgeFlow:
@david-s73 could you fwport to 18.0?

PR #2825, opened 2025-09-24
[17.0][IMP] purchase_stock_price_unit_sync: Avoid create adjustment svl when product_cost_price_avco_sync is installed

CarlosRoca13:
When product_cost_price_avco_sync is installed, the synchronization is done by that module, so it is not necessary to create an adjustment svl to correct the price using the invoices.

cc @Tecnativa TT58144

ping @pedrobaeza @carlosdauden

### DIFF
--- a/purchase_stock_price_unit_sync/migrations/17.0.1.0.1/pre-migration.py
+++ b/purchase_stock_price_unit_sync/migrations/17.0.1.0.1/pre-migration.py
@@ -1,0 +1,14 @@
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html)
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    if env["ir.module.module"].search(
+        [("name", "=", "product_cost_price_avco_sync"), ("state", "=", "installed")]
+    ):
+        # Delete adjustment svl's when product_cost_price_avco_sync is installed
+        openupgrade.logged_query(
+            env.cr,
+            "DELETE FROM stock_valuation_layer WHERE account_move_id IS NOT NULL",
+        )
--- a/purchase_stock_price_unit_sync/models/account_move_line.py
+++ b/purchase_stock_price_unit_sync/models/account_move_line.py
@@ -1,0 +1,18 @@
+# Copyright 2025 Tecnativa - Carlos Roca
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class AccountMoveLine(models.Model):
+    _inherit = "account.move.line"
+
+    def _generate_price_difference_vals(self, layers):
+        # Remove svl_vals_list returned by the original module when
+        # product_cost_price_avco_sync is installed because it will
+        # make the synchronization.
+        if self.env["ir.module.module"].search(
+            [("name", "=", "product_cost_price_avco_sync"), ("state", "=", "installed")]
+        ):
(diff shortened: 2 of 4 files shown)


purchase_stock_price_unit_sync/migrations/17.0.1.0.1/pre-migration.py
```diff
@@ -0,0 +1,16 @@
...
+def migrate(env, version):
+    if env["ir.module.module"].search(
+        [("name", "=", "product_cost_price_avco_sync"), ("state", "=", "installed")]
+    ):
+        # Delete adjustment svl's when product_cost_price_avco_sync is installed
+        openupgrade.logged_query(
+            env.cr,
+            """
+            DELETE FROM stock_valuation_layer WHERE account_move_id IS NOT NULL
+        """,
```
pedrobaeza:
```suggestion
            "DELETE FROM stock_valuation_layer WHERE account_move_id IS NOT NULL",
```

purchase_stock_price_unit_sync/models/account_move_line.py
```diff
@@ -0,0 +1,19 @@
...
+
+    def _generate_price_difference_vals(self, layers):
+        # Remove svl_vals_list returned by the original module when
+        # product_cost_price_avco_sync is installed because it will
+        # make the synchronization.
+        res = super()._generate_price_difference_vals(layers)
+        if self.env["ir.module.module"].search(
+            [("name", "=", "product_cost_price_avco_sync"), ("state", "=", "installed")]
+        ):
+            res[0].clear()
```
pedrobaeza: Isn't better to not call super in this case?

pedrobaeza (approved):
Although the solution is not ideal looking for installed module, it's the only way according current dependency scheme.

CarlosRoca13:
Changes done @pedrobaeza

HviorForgeFlow:
@carlosdauden should we move it forward?

purchase_stock_price_unit_sync/migrations/17.0.1.0.1/pre-migration.py
```diff
@@ -0,0 +1,14 @@
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html)
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    if env["ir.module.module"].search(
+        [("name", "=", "product_cost_price_avco_sync"), ("state", "=", "installed")]
+    ):
```
carlosdauden, line 9:
```suggestion
    if "product_cost_price_avco_sync" in self.env.registry._init_modules:
```

purchase_stock_price_unit_sync/models/account_move_line.py
```diff
@@ -0,0 +1,18 @@
...
+class AccountMoveLine(models.Model):
+    _inherit = "account.move.line"
+
+    def _generate_price_difference_vals(self, layers):
+        # Remove svl_vals_list returned by the original module when
+        # product_cost_price_avco_sync is installed because it will
+        # make the synchronization.
+        if self.env["ir.module.module"].search(
+            [("name", "=", "product_cost_price_avco_sync"), ("state", "=", "installed")]
+        ):
```
carlosdauden, line 16:
```suggestion
        if "product_cost_price_avco_sync" in self.env.registry._init_modules:
```

carlosdauden (changes requested):
Use registry._init_modules instead of search

PR #2841, opened 2025-10-07
[16.0][ADD] purchase_stock_cost_update: new module

chienandalu:
https://www.loom.com/share/a2a41bae6b3c4baaaa8cbc21b3a655df?sid=a1230ca6-b60a-4c3b-828f-00b51b9f95d4

cc @moduon MT-12014

please review @Gelojr @rafaelbn


Gelojr (changes requested):
Thank you very much for your contribution and the work done on this module @chienandalu .
However, a review is requested since Test 5 has not passed correctly. The module is revaluing quantities that are no longer in stock, which causes inconsistencies in the valuation.

Test summary:

- Test 1: Purchase order with two products and return without changes → OK
- Test 2: Price change and adjustment with “Fix valuation” → OK
- Test 3: Partial return after adjustment → OK
- Test 4: Price change from vendor bill → OK

- Test 5: Revaluation after price adjustment with partially consumed stock (NOK)
- Description: Reproduce the exact sequence carried out to verify that, when applying “Fix valuation” after one unit of product by lots 1 has been delivered, only the units still in stock are revalued and not those already consumed.

Steps performed:

1. Created a purchase order with product by lots 1 (10 units at €10).
2. Received product by lots 1 in lot 001 (10 units).
3. Created a second purchase order with product by lots 1 (10 units at €20).
4. Received product by lots 1 in lot 002 (10 units).
5. Verified that after these receipts the product by lots 1 had 20 units valued at €300, with an average cost of €15/unit.
6. Sold one unit of product by lots 1 and delivered one unit from lot 001.
7. Verified that after the delivery the product by lots 1 had 19 units valued at €285, with an average cost of €15/unit.
8. Went back to the first purchase order and changed the price of product by lots 1 from €10 to €5.
9. Pressed the “Fix valuation” button and confirmed the action.

Expected result: Only the quantity still in stock from the first receipt (9 units, since 1 unit had been delivered) should be revalued. Accounting equivalence expected: write off 9 units at €10 (−€90) and add 9 units at €5 (+€45). The expected total value after adjustment should be 19 units valued at €240, giving an average cost of €12.63/unit. The unit already consumed should not be revalued.
image
image


Actual result: The system replaced the entire initial valuation (10 units at €10 → 10 units at €5), equivalent to writing off 10 units at €10 (−€100) and adding 10 units at €5 (+€50), resulting in 19 units valued at €235 and an average cost of €12.37/unit. The module incorrectly revalued the unit that had already been delivered, indicating that it is not excluding consumed quantities from the valuation adjustment.

Result: NOK

Gelojr (changes requested):
Nice job @chienandalu but I tkink a review is needed

**Preliminary comment:**
Although the final result obtained is the same both with the development and with Odoo’s standard behavior, the procedure is not identical.
With the development, when applying the revaluation from the purchase order, the system modifies the original valuation line and sometimes adds an extra valuation line to reflect the adjustment.
Odoo standard, however, performs the revaluation differently, without directly altering the original receipt line.
Even so, the accounting impact and average cost result are equivalent in both cases.

**Common context across all tests:**
• A lot-tracked product is created.
• A first purchase of 10 units at €10 is made and received as lot 1.
• A second purchase of 10 units at €20 is made and received as lot 2.
• Each test compares Odoo standard behavior (adjustment from the vendor bill) with the development behavior (adjustment from the purchase order using “Fix valuation”).

Test 1 – Sale of lot 1
• Lot 1 (from the first purchase at €10) is sold.
• The price is modified both in the vendor bill (Odoo standard) and in the purchase order (development).
• The revaluation result is identical in both cases.
• Result: OK

Test 2 – Sale of lot 2
• Lot 2 (from the second purchase at €20) is sold.
• The price is modified both from the bill and the purchase order.
• The final valuation result is the same with both methods.
• Result: OK

Test 3 – Return of lot 1
• After the initial purchases, a return of lot 1 is made.
• The behavior is compared between changing the price from the bill and from the purchase order.
• In both cases, the returned value is revalued equally.
• Result: OK

Test 4 – Return of lot 2
• A return of lot 2 is made after adjusting the purchase price.
• The results from the bill and the purchase order are compared.
• The valuation result is identical.
• Result: OK

Test 5 – Applying a discount on lot 1
• A discount is applied on the line for lot 1.
• The price adjustment is tested both from the bill and from the purchase order.
• The revaluation and average cost results are the same in both cases.
• Result: OK

Test 6 – Applying a discount on lot 2
• A discount is applied on the line for lot 2.
• The price change is performed both from the bill and the purchase order.
• The valuation result is identical in both cases.
• Result: OK

Test 7 – Purchase with a different unit of measure
• A purchase is made using a unit of measure different from the one defined on the product.
• After receiving the goods, the system displays a “valuation adjustment” warning, which should not appear.
• This behavior is not correct, as changing the UoM should not automatically trigger a valuation adjustment warning.
• Result: NOK

Conclusion:
All tests except the unit of measure scenario behaved correctly.
The development produces the same valuation results as Odoo standard, though through a slightly different mechanism (modifying or adding valuation lines).
Only the UoM-related warning requires review.

rafaelbn (approved):
Functional review in staging 👍🏼

Gelojr (approved):
Nice @chienandalu

PR #2850, opened 2025-10-16
[19.0][MIG] purchase_order_line_original_date: Migration to 19.0

Hudson5577:
Standard migration


purchase_order_line_original_date/tests/test_purchase_order_line_original_date.py
```diff
@@ -0,0 +1,69 @@
+# Copyright 2024 ForgeFlow S.L. (https://www.forgeflow.com)
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+from datetime import datetime
+from datetime import timedelta as td
+
+from odoo.tests import common
+
+
+class TestPoLineOriginalDate(common.TransactionCase):
```
BhaveshHeliconia, line 10:
```suggestion
class TestPoLineOriginalDate(Basecommon):
```

purchase_order_line_original_date/tests/test_purchase_order_line_original_date.py
```diff
@@ -0,0 +1,69 @@
...
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.po_model = cls.env["purchase.order"]
+
+        cls.vendor_1 = cls.env["res.partner"].create(
```
BhaveshHeliconia, line 16:
```suggestion
```

BhaveshHeliconia (changes requested):
Consider using BaseCommon instead of TransactionCase. In this case, there’s no need to redefine res.partner is already available via BaseCommon. This would help simplify and optimize the test code.

PR #2855, opened 2025-10-23
[18.0][FIX] purchase_delivery_split_date: fix dropship pick partner

TDu:
Before this change, for dropship purchase orders when new pickings are created the contact on the new picking is not correct. And actually different than the one set on the original picking.

### DIFF
--- a/purchase_delivery_split_date/models/stock_move.py
+++ b/purchase_delivery_split_date/models/stock_move.py
@@ -67,8 +67,18 @@ class StockMove(models.Model):
 
     def _get_new_picking_values(self):
         vals = super()._get_new_picking_values()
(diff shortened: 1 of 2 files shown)


sebalix (approved):

lmignon (commented):
@TDu Can you add a test to avoid future regressions?
Thank you for the fix.

dreispt (approved):

WesleyOliveira98 (approved):

HviorForgeFlow:
@TDu https://github.com/OCA/purchase-workflow/pull/2855#pullrequestreview-3404067746 would be nice. Just let us know if you have time for that.

TDu:
@lmignon @HviorForgeFlow  Test has been added :santa: :gift:

PR #2838, opened 2025-10-03
[19.0][MIG] purchase_tag: Migration to 19.0

bizzappdev:

PR #2860, opened 2025-10-28
[19.0][MIG] purchase_tag: Migration to 19.0

Hudson5577:
Changed 1 line: 

_sql_constraints = [("tag_name_uniq", "unique (name)", "Tag name already exists !")]

into

models.Constraint("unique (name)", message="Tag name already exists !",)


purchase_tag/models/purchase_tag.py
```diff
@@ -0,0 +1,44 @@
...
+    name = fields.Char("Tag Name", required=True, translate=True)
+    color = fields.Integer(default=lambda self: self._get_default_color())
+    parent_id = fields.Many2one("purchase.tag", index=True, ondelete="cascade")
+    child_ids = fields.One2many("purchase.tag", "parent_id")
+    parent_path = fields.Char(index=True)
+
+    _name_uniq = models.Constraint(
+        "unique (name)",
+        message="Tag name already exists !",
+        name="tag_name_uniq",
```
lmignon: The `Constraint` class doesn't expect the `name` attribute into its constructor
```suggestion
```

lmignon (commented):
Thank you for the migration. Nevertheless, even if it appears to be easy, an issuer remains... 😏

lmignon (changes requested):
Tests fails...  see my comment

purchase_tag/models/purchase_tag.py
Hudson5577: Thanks for the feedback! I’ll update the constraint and remove the name attribute.

purchase_tag/models/purchase_tag.py
```diff
@@ -21,9 +21,10 @@ def _get_default_color(self):
     child_ids = fields.One2many("purchase.tag", "parent_id")
     parent_path = fields.Char(index=True)
 
-    _sql_constraints = [
-        ("tag_name_uniq", "unique (name)", "Tag name already exists !"),
-    ]
+    _name_uniq = models.Constraint(
```
bizzappdev, line 24: _tag_name_uniq will be the proper name.
Also we have already created the PR https://github.com/OCA/purchase-workflow/pull/2838/

purchase_tag/models/purchase_tag.py
```diff
@@ -0,0 +1,43 @@
...
+            while current:
+                if current.name:
+                    names.append(current.name)
+                current = current.parent_id
+            display_name = " / ".join(reversed(names))
+            tag.display_name = display_name
+
+    @api.model
+    def _search_display_name(self, operator, value):
+        domain = [("name", operator, value.split(" / ")[-1])]
```
sergijPf, line 42: @Hudson5577 
thanks for your contribution!

this search will fail when 'value' is not a string. 
To reproduce it:
use  'is equal to'/'is set'  (or not)  for custom filter
Image
maybe, it worth to make some checks if operator is 'ilike' and value is str, otherwise call super, 
(please see more examples in Odoo source code)

the rest is LGTM

BhaveshHeliconia (changes requested):
@Hudson5577 it looks like there is already an open PR for this module: https://github.com/OCA/purchase-workflow/pull/2838.
Could you please close this one and review that PR instead?

HviorForgeFlow:
closing in forward to https://github.com/OCA/purchase-workflow/pull/2838 that is already merged

PR #2861, opened 2025-10-28
[18.0][FIX] purchase_request: fix migration script for enterprise migration

DavidJForgeFlow:
If you are migrating using the enterprise service, the table ir_property is already converted and deleted. Also the conversion to company_dependent is done for all the fields in that service too. So when trying to access the table in the openupgrade process it raises an exception. So I'm adding a condition to ensure this table still exists and being able to do the migration with both services!

### DIFF
--- a/purchase_request/migrations/18.0.1.0.0/post-migration.py
+++ b/purchase_request/migrations/18.0.1.0.0/post-migration.py
@@ -5,6 +5,7 @@ from openupgradelib import openupgrade, openupgrade_180
 
 @openupgrade.migrate()
 def migrate(env, version):
-    openupgrade_180.convert_company_dependent(
-        env, "product.template", "purchase_request"
-    )
+    if openupgrade.table_exists(env.cr, "ir_property"):
+        openupgrade_180.convert_company_dependent(
+            env, "product.template", "purchase_request"
+        )


pedrobaeza (changes requested):
If you use enterprise, your should adapt the scripts on your end, not here, and filling everything with ifs is not the solution. In fact, you should convert the same the property fields, as enterprise shouldn't be doing that. AFAIK, the table is renamed to `_ir_property`, so rename it back to the correct one.

DavidJForgeFlow:
> If you use enterprise, your should adapt the scripts on your end, not here, and filling everything with ifs is not the solution. In fact, you should convert the same the property fields, as enterprise shouldn't be doing that. AFAIK, the table is renamed to `_ir_property`, so rename it back to the correct one.

I understand that we can rename the table, execute the script and delete the table. However I don't think that is really needed as simply adding this condition will allow me and all the people migrating with Odoo service avoid doing these operations just to avoid an exception (as the properties are migrated by the service). I don't really think avoid adding a condition justify forcing all the enterprise partners to do a workaround here.

LoisRForgeFlow (approved):
:+1: 

@pedrobaeza I don't think this can cause any inconvenience to anyone, if you don't have the ir_property table for any other reason, it would have failed anyway too.

I think this is a simple approach that can serve for both migration paths.

pedrobaeza:
Sorry, but no, as this is happening across whole OCA (see for example https://github.com/OCA/l10n-spain/pull/4515, https://github.com/OCA/credit-control/pull/513, https://github.com/OCA/sale-workflow/pull/3972), and adding extra code for not "crashing" is a burden for OCA that shouldn't be put, and anyway, putting this if it's not going to solve your problem: you'll find that the company dependent data is not migrated, so you should manually convert them. Better to crash and warn you about an action to do, than silently pass, but not having a correct migrated DB.

DavidJForgeFlow:
I've been personally testing what happens with the resulting data without executing the script and using this PR changes. All the data required is available and the properties are correctly migrated. Maybe all the changes done in the OCA need to be also updated to be usable with the enterprise version. The upgrades in OU users should not affect IMHO in any case to enterprise users.

pedrobaeza:
@hbrunn was there any impossibility to detect property fields to be automatically migrated without having to put explicitly the call to the method?

hbrunn:
I never looked into detecting old property fields automatically.

As for this PR: If that's how enterprise migrations work, I think the right level to add the check for ir_property existing is https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade_180.py#L22

pedrobaeza:
Yeah, but for this specific case, a modification in openupgradelib may fit, but my point is that there's no easy way to converge enterprise migration with OCA migration scripts, as the number of incompatibilities can be overwhelming, and even if not, handling with that code for something that is a blackbox, specially for those community contributors, is not possible.

@hbrunn doing an automatic recognition like we did with translatable fields may save a lot of these PRs / migration scripts.

DavidJForgeFlow:
> I never looked into detecting old property fields automatically.
> 
> As for this PR: If that's how enterprise migrations work, I think the right level to add the check for ir_property existing is https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade_180.py#L22

Great idea, I've done a proposal, please check it out! https://github.com/OCA/openupgradelib/pull/426

pedrobaeza:
Closing this.

PR #2867, opened 2025-10-31
[16.0][FIX] purchase_deposit: fixes account.tax access error in a multi company environment

almumu:
Solves access error when creating a Deposit invoice, because it was getting all the product taxes of the existing companies.
image

### DIFF
--- a/purchase_deposit/wizard/purchase_make_invoice_advance.py
+++ b/purchase_deposit/wizard/purchase_make_invoice_advance.py
@@ -56,7 +56,9 @@ class PurchaseAdvancePaymentInv(models.TransientModel):
     def _compute_deposit_account_id(self):
         product = self.purchase_deposit_product_id
         self.deposit_account_id = product.property_account_expense_id
-        self.deposit_taxes_id = product.supplier_taxes_id
+        self.deposit_taxes_id = product.supplier_taxes_id.filtered(
+            lambda tx: tx.company_id == self.env.company
+        )
 
     def _prepare_deposit_val(self, order, po_line, amount):
         ir_property_obj = self.env["ir.property"]


ljsalvatierra-factorlibre (approved):

purchase_deposit/wizard/purchase_make_invoice_advance.py
```diff
@@ -56,7 +56,9 @@ class PurchaseAdvancePaymentInv(models.TransientModel):
     def _compute_deposit_account_id(self):
         product = self.purchase_deposit_product_id
         self.deposit_account_id = product.property_account_expense_id
-        self.deposit_taxes_id = product.supplier_taxes_id
+        self.deposit_taxes_id = product.supplier_taxes_id.filtered(
+            lambda tx: tx.company_id == self.env.company
```
dreispt, line 60: Shouldn't we honor the document's company?
```suggestion
            lambda tx: tx.company_id == self.company_id
```

almumu: in this case "self" is the wizard, not the document, and the wizard has no company field. This method https://github.com/OCA/purchase-workflow/blob/501ae75e4d30df34e2faf6050e3ef906bccb9329/purchase_deposit/wizard/purchase_make_invoice_advance.py#L56 depends on purchase_deposit_product_id field, and where the purchase_deposit_product_id field is defined, it uses the env company https://github.com/OCA/purchase-workflow/blob/501ae75e4d30df34e2faf6050e3ef906bccb9329/purchase_deposit/wizard/purchase_make_invoice_advance.py#L30

CasVissers-360ERP (approved):
Same issue in v18

astirpe (approved):

dreispt (approved):

CasVissers-360ERP:
@dreispt can you also merge #2923 #2924

PR #2874, opened 2025-11-03
[16.0][ADD] purchase_line_receipt_status: Compute receipt status on purch…

lmignon:
…ase order line

In standard Odoo, the `purchase.order` model provides a `receipt_status` field to track the reception of purchase orders. This field is computed based on the associated pickings:

- None: No pickings created or all pickings cancelled
- `pending`: Not received yet
- `partial`: Some pickings done
- `full`: All pickings done (even if some were cancelled)

However, there is **no equivalent status at the purchase order line level**. This module introduces the missing `line_receipt_status` field on purchase order lines, using the same logic as the standard PO `receipt_status`.

It complements OCA modules such as `purchase_reception_status` without modifying their quantity-based logic. This module preserves the original picking-based status computation, ensuring consistency with Odoo’s standard behavior.

One may wonder about the formulation of possible values for the `line_receipt_status` field. The semantics do not truly reflect the concept of auditing outstanding logistics operations. However, to maintain consistency with the standard field declared at the purchase order level and having the same meaning, I decided to keep the same values.


luisDIXMIT (commented):
I have 2 received lines, but the filter doesn’t seem to be working:

image

image


Is there something wrong, or am I missing something?

luisDIXMIT (commented):
On the other hand, is this a substitute for purchase_reception_status_line? From my point of view, it looks like two different ways of doing the same thing.

BhaveshHeliconia (approved):
LGTM!

PR #2875, opened 2025-11-03
[15.0][FIX] purchase_order_secondary_unit: Purchase secondary uom by default non store in variant

sergio-teruel:
@Tecnativa TT59046

ping @carlosdauden @carlos-lopez-tecnativa 

Depends on:
- [x] https://github.com/OCA/purchase-workflow/pull/3163

### DIFF
--- a/purchase_order_secondary_unit/migrations/15.0.2.0.0/post-migration.py
+++ b/purchase_order_secondary_unit/migrations/15.0.2.0.0/post-migration.py
@@ -1,0 +1,33 @@
+# Copyright 2025 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    # Set default purchase secondary uom from template to variants.
+    # purchase_secondary_uom_id is no longer stored on product.template, so
+    # the legacy values must be read straight from the column, which is
+    # still physically present at this point (Odoo never drops it).
+    env.cr.execute(
+        "SELECT id, purchase_secondary_uom_id FROM product_template"
+        " WHERE purchase_secondary_uom_id IS NOT NULL"
+    )
+    legacy_values = dict(env.cr.fetchall())
+    templates = (
+        env["product.template"].with_context(acive_test=False).browse(legacy_values)
+    )
+    unique_variants = templates.filtered(lambda tmpl: tmpl.product_variant_count == 1)
+    for template in unique_variants:
+        legacy_secondary_uom_id = legacy_values[template.id]
+        if template.product_variant_ids.purchase_secondary_uom_id.id != (
+            legacy_secondary_uom_id
+        ):
+            # Force store purchase_secondary_uom_id for unique variants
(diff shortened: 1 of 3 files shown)


purchase_order_secondary_unit/migrations/15.0.2.0.0/post-migration.py
```diff
@@ -0,0 +1,43 @@
...
+            FROM product_product
+            GROUP BY product_tmpl_id
+            HAVING COUNT(*) > 1
+        )
+        UPDATE product_template pt
+        SET purchase_secondary_uom_id = NULL
+        FROM multi
+        WHERE pt.id = multi.id
+          AND pt.purchase_secondary_uom_id IS NOT NULL;
+    """
```
carlos-lopez-tecnativa: In the compute, you consider that a template can have more than one variant, but if all variants have the same `purchase_secondary_uom_id`, it should be set on the template. Why didn’t you consider this case in the migration script?

purchase_order_secondary_unit/migrations/15.0.2.0.0/post-migration.py
```diff
@@ -0,0 +1,43 @@
...
+        )
+        UPDATE product_product pp
+        SET purchase_secondary_uom_id = pt.purchase_secondary_uom_id
+        FROM product_template pt
+        JOIN single s ON s.id = pt.id
+        WHERE pp.product_tmpl_id = pt.id
+          AND pt.purchase_secondary_uom_id IS NOT NULL
+          -- avoid writing unchanged rows
+          AND COALESCE(pp.purchase_secondary_uom_id, 0) <> pt.purchase_secondary_uom_id;
+    """
```
carlos-lopez-tecnativa: Once you consider the case mentioned above and later set null on templates with more than a single variant, I think the remaining cases don’t require a complex query, because the remaining records have a single variant. So, the `UPDATE` can be done directly without the `WITH` clause.

carlos-lopez-tecnativa (changes requested):
Code review

sergio-teruel:
@carlos-lopez-tecnativa changes done!! Finally I have used orm

carlos-lopez-tecnativa (approved):
LGTM, thanks for the improvement. Please forward-port to the upper versions.

carlosdauden (approved):

purchase_order_secondary_unit/models/product_template.py
```diff
@@ -9,5 +9,66 @@ class ProductTemplate(models.Model):
     purchase_secondary_uom_id = fields.Many2one(
         comodel_name="product.secondary.unit",
         string="Default secondary unit for purchases",
+        compute="_compute_purchase_secondary_uom_id",
+        inverse="_inverse_purchase_secondary_uom_id",
+        help="In order to set a value, please first add at least one record"
+        " in 'Secondary Unit of Measure'",
         domain="[('product_tmpl_id', '=', id), ('product_id', '=', False)]",
+        store=True,
```
pedrobaeza, line 17: Why being stored if it has `inverse`?

sergio-teruel:
@pedrobaeza Changes done. The store attribute was not needed..

sergio-teruel:
Rebase done!!

PR #2813, opened 2025-09-15
[18.0][FIX] purchase_delivery_split_date

DavidJForgeFlow:
Forward port of #2759 to 18.0

PR #2876, opened 2025-11-04
[MIG] purchase_delivery_split_date: Migration to 19.0

Tatider:


purchase_delivery_split_date/models/purchase_order_line.py
```diff
@@ -43,7 +42,7 @@ def _create_stock_moves(self, picking):
             # A picking should be provided
             return super()._create_stock_moves(picking)
         moves = self.env["stock.move"]
-        tz = picking.picking_type_id.warehouse_id.partner_id.tz or self.env.user.tz
+        tz = picking.picking_type_id.warehouse_id.partner_id.tz
```
yankinmax, line 44: Maybe `tz = picking.picking_type_id.warehouse_id.partner_id.tz or self.env.tz`?

purchase_delivery_split_date/models/purchase_order_line.py
```diff
@@ -22,11 +21,11 @@ def _purchase_split_date_get_group_keys(self, picking):
         can add additional keys or replace them by others.
         """
         tz = self.order_id.picking_type_id.warehouse_id.partner_id.tz
-        wh_tz = pytz.timezone(tz or self.env.user.tz or "UTC")
+        wh_tz = pytz.timezone(tz) if tz else self.env.tz
         date_planned_tz = self.date_planned.astimezone(pytz.utc).astimezone(wh_tz)
         date = date_planned_tz.date()
         # Split date value to obtain only the attributes year, month and day
-        key = ({"date_planned": fields.Date.to_string(date)},)
+        key = (("date_planned", fields.Date.to_string(date)),)
```
yankinmax: Can you pls explain this change?

purchase_delivery_split_date/models/purchase_order_line.py
```diff
@@ -3,11 +3,10 @@
 # Copyright 2021 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
-from itertools import groupby
```
yankinmax: This change is also strange:
`from itertools import groupby` to `from odoo.tools import groupby`.
This `gropuby` isn't interchangeable.
Can you pls explain?

purchase_delivery_split_date/models/purchase_order_line.py
Tatider: No, it is correct. The timezone is used in the "_purchase_split_date_assign_domain" function. And that function will use env's timezone if "tz" parameter is not defined.

* picking.picking_type_id.warehouse_id.partner_id.tz - has type string
* self.env.tz - has type pytz.timezone

To prevent code block duplicates, it makes sence to convert or take from env only one time.

Screenshot 2025-12-18 at 11 15 53

yankinmax: Thanks for explanation

purchase_delivery_split_date/models/purchase_order_line.py
Tatider: Good point, at this point I no longer remember why it was done...
I have checked the logic, and runned test w/o this change and it seems to work okay. But I think it is important to mentioned that Odoo had changed the expected type for the argument for "groupby" https://github.com/odoo/odoo/commit/f59fe1190aceef9cdffc6cfe2fe34e6a9ad7d090#diff-61162ac65633a1c7b054fc83ce1813f1a7984e3169ff36021713ef441f62a208L1140

And this module is using different functions in different files:
* `purchase.order.line` was using `itertools.groupby`
* `stock.move` was using `odoo.groupby`

It seems a bit odd and potentially leading to conflicts/errors if different parts of module expecting different format of a "key" ...

With this in mind do you want me to roll back this change still?

yankinmax: Thanks for the explanation. Completely agree with you. Probably it was mixed from the very beginning of module history. Let's stick to one usage. No need to roll back I think.

purchase_delivery_split_date/models/purchase_order_line.py
Tatider: Since this change was related to the format of the key of the "groupby" function and we agreed to keep the new version, I am marking this comment as resolved.

CasVissers-360ERP:
@Tatider can you add #2813

yankinmax (approved):
Thanks for the work done

purchase_delivery_split_date/models/stock_move.py
```diff
@@ -70,5 +70,6 @@ def _get_new_picking_values(self):
         if self.env.context.get("purchase_delivery_split_date") and not vals.get(
             "partner_id"
         ):
-            vals["partner_id"] = fields.first(self.purchase_line_id.partner_id).id
+            partners = self.purchase_line_id.partner_id
+            vals["partner_id"] = next(iter(partners), partners).id
```
ivantodorovich: A better replacement for `fields.first`, IMO:

```
vals["partner_id"] = self.purchase_line_id.partner_id[:1].id
```

purchase_delivery_split_date/models/purchase_order_line.py
```diff
@@ -0,0 +1,96 @@
...
+        """
+        tz = self.order_id.picking_type_id.warehouse_id.partner_id.tz
+        wh_tz = pytz.timezone(tz) if tz else self.env.tz
+        date_planned_tz = self.date_planned.astimezone(pytz.utc).astimezone(wh_tz)
+        date = date_planned_tz.date()
+        # Split date value to obtain only the attributes year, month and day
+        key = (("date_planned", fields.Date.to_string(date)),)
+        return key
+
+    def _purchase_split_date_get_sorted_keys(self):
```
ivantodorovich, line 30: I would drop this method.

sorting was only needed when using itertools.groupby, but it's not needed with `odoo.tools.groupby` nor `model.grouped`

purchase_delivery_split_date/models/purchase_order_line.py
```diff
@@ -0,0 +1,96 @@
...
+        order_lines = sorted(
+            self.filtered(
+                lambda line: not line.display_type and line.product_id.type == "consu"
+            ),
+            key=lambda line: line._purchase_split_date_get_sorted_keys(),
+        )
+        date_groups = groupby(
+            order_lines,
+            lambda line: line._purchase_split_date_get_group_keys(picking),
+        )
```
ivantodorovich: IMO you can simplify this with the "new" `models.grouped` added in 18.0, which just like `odoo.tools.groupby` does not require sorting before hand (unlike `itertools.groupby`)


```suggestion
        order_lines = self.filtered(
            lambda line: not line.display_type and line.product_id.type == "consu"
        )
        date_groups = order_lines.grouped
            lambda line: line._purchase_split_date_get_group_keys(picking),
        )
```

purchase_delivery_split_date/models/purchase_order_line.py
```diff
@@ -0,0 +1,96 @@
...
+
+        for key, lines in date_groups:
+            po_lines = self.browse().concat(*lines)
```
ivantodorovich:
```suggestion
```

With `model.grouped` (see my other suggestion) you don't need to concat afterwards. It'd already be a recordset.

purchase_delivery_split_date/models/purchase_order_line.py
```diff
@@ -0,0 +1,96 @@
...
+        containing a dictionary element with the field that you want to group
+        by. This method is designed for extensibility, so that other modules
+        can add additional keys or replace them by others.
+        """
+        tz = self.order_id.picking_type_id.warehouse_id.partner_id.tz
+        wh_tz = pytz.timezone(tz) if tz else self.env.tz
+        date_planned_tz = self.date_planned.astimezone(pytz.utc).astimezone(wh_tz)
+        date = date_planned_tz.date()
+        # Split date value to obtain only the attributes year, month and day
+        key = (("date_planned", fields.Date.to_string(date)),)
```
ivantodorovich:
```suggestion
        key = (("date_planned", date),)
```

why not using the python date instead?

purchase_delivery_split_date/models/stock_move.py
```diff
@@ -0,0 +1,75 @@
...
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    def _purchase_split_date_get_group_keys(self):
+        tz = self.picking_type_id.warehouse_id.partner_id.tz
+        wh_tz = pytz.timezone(tz) if tz else self.env.tz
+        date_planned_tz = self.date_deadline.astimezone(pytz.utc).astimezone(wh_tz)
+        date = date_planned_tz.date()
+        # Split date value to obtain only the attributes year, month and day
+        key = (("date_planned", fields.Date.to_string(date)),)
```
ivantodorovich:
```suggestion
        key = (("date_planned", date),)
```

same as other suggestion, why not using the python date instead?

purchase_delivery_split_date/models/stock_move.py
```diff
@@ -0,0 +1,75 @@
...
+        )
+        if not po_moves:
+            return
+        po_moves.date = new_deadline
+        for picking, moves_list in groupby(po_moves, lambda m: m.picking_id):
+            if picking.printed:
+                # Do not split by date anymore
+                continue
+            # todo: if all moves on the same date
+            moves = self.browse().concat(*moves_list)
```
ivantodorovich:
```suggestion
        for picking, moves in po_moves.grouped("picking_id"):
            if picking.printed:
                # Do not split by date anymore
                continue
```

purchase_delivery_split_date/models/stock_picking.py
```diff
@@ -0,0 +1,25 @@
...
+    def _purchase_split_date_assign_domain(self, key, tz):
+        date = key[0][1]
+        wh_tz = pytz.timezone(tz) if tz else self.env.tz
+        # The date is in local time
+        dt_start_tz = wh_tz.localize(fields.Datetime.to_datetime(date))
+        dt_start = dt_start_tz.astimezone(pytz.utc).replace(tzinfo=None)
+        dt_end = fields.Datetime.add(dt_start, days=1)
+        return [
+            ("scheduled_date", ">=", fields.Datetime.to_string(dt_start)),
+            ("scheduled_date", "<", fields.Datetime.to_string(dt_end)),
```
ivantodorovich:
```suggestion
            ("scheduled_date", ">=", dt_start),
            ("scheduled_date", "<", dt_end),
```

no need to convert `to_string` here

LoisRForgeFlow (changes requested):
> @Tatider can you add #2813

I drop a review here as a reminder of this

purchase_delivery_split_date/models/purchase_order_line.py
Tatider: Disagree, with 2 arguments:
1. It is a good entry point to modificate the key.
2. It is used in different OCA module https://github.com/OCA/purchase-workflow/blob/18.0/purchase_location_by_line/models/purchase_order_line.py#L27

purchase_delivery_split_date/models/stock_move.py
Tatider: Well, I see your point and probably share a similar opinion... But I did not came up with this replacement on my own, I have checked how the transition was done in Odoo source code...
With this information in mind, if you still think the change is needed,  I can do that.

Tatider:
@ivantodorovich I am grateful for your comments, but they are mostly to the parts of the code, I have not changed. Therefore, they are refactoring changes and not migration ones... I will do them, but I don't feel particularly comfortable doing them...

Tatider:
> @Tatider can you add #2813

Done

Tatider:
> > @Tatider can you add #2813
> 
> I drop a review here as a reminder of this

Thank you, I missed it before. it is added now

Tatider:
@LoisRForgeFlow I have done the change you have requested - could you void your request?

yankinmax:
Hello @LoisRForgeFlow @ivantodorovich can you pls take a look at this PT again and possibly merge?

LoisRForgeFlow (approved):
LGTM, thanks!

yankinmax:
Hello @OCA/procurement-purchase-maintainers can you pls take a look and possibly merge this PR?

HviorForgeFlow (approved):

BhaveshHeliconia (approved):
LGTM!

HviorForgeFlow:
Tests are failing @Tatider

cle-yona:
Any news ?

yankinmax:
Hello @Tatider can you pls rebase (possibly tests were failing during merge because template outdated, just a guess)?

PR #2881, opened 2025-11-07
[ADD][18.0] partner_replenishment_block_archive

mathieudelva:


partner_replenishment_block_archive/models/res_partner.py
```diff
@@ -0,0 +1,20 @@
...
+    _inherit = "res.partner"
+
+    def write(self, vals):
+        if "active" in vals and not vals["active"]:
+            replenishements = self.env["stock.warehouse.orderpoint"].search(
+                [("supplier_id", "=", self.id)]
+            )
+            if replenishements:
+                raise ValidationError(
+                    _("You cannot archive a partner with replenishement")
```
dreispt:
```suggestion
                    _("You cannot archive a partner with replenishement rule.")
```

partner_replenishment_block_archive/models/res_partner.py
```diff
@@ -0,0 +1,20 @@
...
+from odoo.exceptions import ValidationError
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    def write(self, vals):
+        if "active" in vals and not vals["active"]:
+            replenishements = self.env["stock.warehouse.orderpoint"].search(
+                [("supplier_id", "=", self.id)]
```
dreispt: Won't work when updating many records.
You can also make it more efficient.
```suggestion
                [("supplier_id", "in", self.ids)], limit=1
```

dreispt (changes requested):

mathieudelva:
@dreispt thanks for the review, I updated the PR

PR #2883, opened 2025-11-11
[19.0][MIG] purchase_order_supplier_return: Migration to 19.0

sergioM-S73:
While migrating this module to v19, I ran into a test failure. The root cause is a core change in Odoo 19: the to_refund field on stock.move now defaults to True. 
https://github.com/odoo/odoo/commit/98fbff3dcab6cba1ea8905311ed1ae012245b5de

This caused the test to fail on the positive-quantity PO line, as it expected the value to be False. 

My first thought was to adapt the module's logic to do the inverse (setting to_refund = False on standard receipts), however, I believe this is incorrect.

This situation led me to question the module's necessity. 
https://github.com/OCA/purchase-workflow/pull/2116

To verify, I ran tests on Runboat instances for versions 16, 17, 18, and 19. In all cases, when validating a return picking generated from a purchase order with a negative quantity line, the received quantity is correctly updated

Therefore, I conclude that this module is redundant


Reyes4711-S73 (approved):
LGTM

Reyes4711-S73:
@sergioM-S73 I made the functional review and this module in a odoo 19.0 is useless. Odo base made this functionality without this module.
In my humble oppinion, we must close this PR

PR #2886, opened 2025-11-12
[19.0][MIG] purchase_analytic_global

bosd:


bosd:
empty

BhaveshHeliconia:
Hi @bosd ,

When testing this module on the Purchase Order header, I encountered a Javascript crash (Oops! Something went wrong) when interacting with the Analytic Distribution widget (specifically when clicking "New" to save a model inside the widget).

Error Traceback: TypeError: Cannot read properties of undefined (reading 'id') at AnalyticDistribution.onSaveNew

Root Cause: The purchase_order.xml view sets the product_field option of the widget to 'product_id':
options="{'product_field': 'product_id', 'business_domain': 'purchase_order'}"

However, the purchase.order model (the header) does not have a product_id field (unlike purchase.order.line). When the widget attempts to pre-fill a new analytic model using the record's product, it looks for product_id on the header, finds undefined, and crashes when trying to read its .id.

Proposed Fix: Remove the 'product_field': 'product_id' option from the analytic_distribution field in the purchase.order header view. On the header level, there is no single product, so the filtering by product is not applicable.

https://github.com/user-attachments/assets/13da6d51-cb0d-4067-b069-e33af165de1a

purchase_analytic_global/views/purchase_order.xml
```diff
@@ -0,0 +1,18 @@
...
+        <field name="name">purchase.order.form</field>
+        <field name="model">purchase.order</field>
+        <field name="inherit_id" ref="purchase.purchase_order_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='partner_ref']" position="after">
+                <field
+                    name="analytic_distribution"
+                    widget="analytic_distribution"
+                    groups="analytic.group_analytic_accounting"
+                    options="{'product_field': 'product_id', 'business_domain': 'purchase_order'}"
```
BhaveshHeliconia:
```suggestion
                   options="{'business_domain': 'purchase_order'}"
```

BhaveshHeliconia (changes requested):

BhaveshHeliconia (approved):
LGTM!

yostashiro:
@bosd It looks like the scope of this PR is already covered by https://github.com/OCA/account-analytic/tree/19.0/purchase_analytic (in case you were not aware...).

sergiocorato:
> @bosd It looks like the scope of this PR is already covered by https://github.com/OCA/account-analytic/tree/19.0/purchase_analytic (in case you were not aware...).

This was already true in v. 18.0: https://github.com/OCA/account-analytic/tree/18.0/purchase_analytic

vvrossem:
`pre-commit` is unhappy

PR #2891, opened 2025-11-19
[19.0] [MIG] purchase tier validation: Migration to 19.0

moitabenfdz:


pedrobaeza (changes requested):
This should be moved with its history to OCA/tier-validation.

PR #2895, opened 2025-11-21
[16.0][FIX] purchase_cancel_reason: cancel before writing cancel reason.

nicolas-delbovier-acsone:
Ensures the purchase order is cancelled **before** writting the `cancel_reason_id`:
1. This makes more sense.
2. Certain integration modules, such as `base_tier_validation`, rely on the record being in the 'cancel' state when the `write` method is executed to properly skip validation tiers. Cancelling first guarantees the correct state for these integrations.


rousseldenis (approved):
Code review

HviorForgeFlow (approved):

PR #2907, opened 2025-12-09
[18.0][IMP] purchase_partner_incoterm: set default incoterm when PO are created automatically

AaronHForgeFlow:
For example, reordering rule POs does not include the default incoterms


JordiMForgeFlow (approved):

HviorForgeFlow (approved):

bealdav (approved):
LGTM, thanks

PR #2912, opened 2025-12-10
[18.0][IMP] purchase_tag: Allow to add tags on lines that will be put on orders

rousseldenis:
In some flows, users want to put tags on lines. Those tags should be reflected on purcahse order.

### DIFF
--- a/purchase_tag/models/__init__.py
+++ b/purchase_tag/models/__init__.py
@@ -1,2 +1,3 @@
 from . import purchase_tag
 from . import purchase_order
+from . import purchase_order_line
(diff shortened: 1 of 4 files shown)


BhaveshHeliconia (approved):
@rousseldenis 
I reviewed the PR and found a few issues related to Purchase Tags:

1. When a Purchase Tag is added to the Purchase Order, it is not automatically applied to newly created Purchase Order Lines.
2. When a Purchase Tag is added to a Purchase Order Line, the same tag is automatically added to the Purchase Order.
3. If a tag is removed from a Purchase Order Line, the corresponding tag is not removed from the Purchase Order.

Please handle these cases and improve the synchronization between Purchase Orders and Purchase Order Lines to keep tags consistent.

PR #2918, opened 2025-12-16
[FIX]purchase_order_type: try to assign the correct sequence on create

mav-adhoc:
It happens that when the PO is generated by a replenishment rule, the "order_type" value is not defined, so it takes a random sequence to generate the PO, and not the sequence defined on the PO type set for the partner of the PO.


jcadhoc (approved):

ced-adhoc (approved):
LGTM

HviorForgeFlow (approved):

PR #2919, opened 2025-12-16
[FIX]purchase_order_type: PO type in res_partner_view

mav-adhoc:
There are two main changes on this PR:

1.     Remove the supplier rank != 0 condition for visibility of the purchase_type field. The field was invisible when you create a new partner or try to set an old partner as a customer.
2. Relocate the field in view. It is more clear and organized if the PO type field is inside the purchase page.
image

### DIFF
--- a/purchase_order_type/__manifest__.py
+++ b/purchase_order_type/__manifest__.py
@@ -3,7 +3,7 @@
 
 {
     "name": "Purchase Order Type",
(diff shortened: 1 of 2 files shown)


jjscarafia (approved):
LGTM simple but great improvement!

HviorForgeFlow (approved):

PR #2932, opened 2026-01-11
[18.0][IMP] purchase_order_secondary_unit: add vendor pricelist support

yostashiro:
- Extend product.supplierinfo and purchase.order.line to support secondary unit pricing
- ~~Add company setting to control price display in purchase reports (primary only, prioritize secondary, or both)~~ -> Moved to product_secondary_unit
- Update purchase order/quotation reports to show secondary unit prices according to the configuration

@qrtl QT6202


yostashiro:
Secondary Price fields added to purchase order line and vendor pricelist.
image

Added company-level report settings. The default settings generate reports in the same layout but with a small enhancement of showing the UoM for the unit price (as it could otherwise be a bit confusing).
image

If you prioritize secondary unit price and hide secondary UoM column, the report will only show secondary qty and secondary price if the purchase order line has a secondary UoM set, otherwise fall back to primary qty and primary price.
image

If you choose to show both primary and secondary, the report will do so in the same columns.
image

AungKoKoLin1997:
I am reviewing the functional and found some strange behaviors. I am not sure this may be not part of the scope of this PR.

**Test Case 1**

- I create two `Secondary Unit of Measure` for a product.
- Then, I create a purchase order with secondary uom. The secondary uom and secondary uom price are correctly assigned in vendor info.
- After that, I create another purchase order with with the same vendor of first PO and different price unit and different secondary uom.
- When I check the vendor info, I found the price unit of existing vendor line is updated with new price unit. But the secondary uom and secondary uom price are not updated with new values.

**Proudct:**

Screenshot 2026-01-14 101658

**Purchase Order 1:**

Screenshot 2026-01-14 102100

**Seller Info:**

Screenshot 2026-01-14 102146

**Purchase Order 2:**

Screenshot 2026-01-14 102424

**Seller Info:**

Screenshot 2026-01-14 102458



**Test Case 2**

-  I create two `Secondary Unit of Measure` for a product.
- Then, I create a purchase order with secondary uom. The secondary uom and secondary uom price are correctly assigned in vendor info.
- After that, I create another purchase order with with the same vendor of first PO and same price unit and different secondary uom.
- In the vendor info, I found the secondary uom and secondary uom price are not updated.

**Purchase Order 1:**

Screenshot 2026-01-14 104256

**Seller Info:**

Screenshot 2026-01-14 104328

**Purchase Order 2:**

Screenshot 2026-01-14 104843

**Seller Info:**
Screenshot 2026-01-14 104944

AungKoKoLin1997:
Regarding the two test cases I mentioned in my previous comment, they’re out of scope for this PR. I’m also not confident they’re worth addressing right now, and we don’t expect to use those use cases at the moment—so I’ll ignore them for now.

AungKoKoLin1997 (approved):
LGTM

AungKoKoLin1997:
In my last commit, I fixed the layout issue when we use `Prioritize Secondary Unit Price` without hiding Secondary UoM Column.
image

AungKoKoLin1997:
@yostashiro Could you please review my last two commits? I will squash commits after you review.

purchase_order_secondary_unit/reports/purchase_order_templates.xml
```diff
@@ -6,16 +6,82 @@
...
+        <!-- Unit price: hide primary ('secondary' mode) -->
+        <xpath expr="//span[@t-field='line.price_unit']" position="attributes">
+            <attribute
+                name="t-if"
+            >line.get_secondary_uom_display_mode() != 'secondary'</attribute>
+        </xpath>
+        <!-- Unit price: add UoM and secondary price ('secondary' and 'both' modes) -->
+        <xpath expr="//span[@t-field='line.price_unit']" position="after">
+            <t
+                t-if="line.report_show_price_uom() and line.get_secondary_uom_display_mode() != 'secondary'"
```
yostashiro:
```suggestion
                t-if="line.report_show_price_uom()"
```
@AungKoKoLin1997 We should not change the code here. Please update `line.report_show_price_uom()` instead.

purchase_order_secondary_unit/views/purchase_order_portal_templates.xml
```diff
@@ -0,0 +1,94 @@
...
+            <t t-if="line.get_secondary_uom_display_mode() == 'both'">
+                <br />
+                <span class="text-muted">
+                    (<span t-field="line.secondary_uom_price" />
+                    / <span
+                    t-field="line.secondary_uom_id.name"
+                    groups="uom.group_uom"
+                />)
+                </span>
+            </t>
```
yostashiro: I guess we could create a common template and call it from both report and portal view.

AungKoKoLin1997: Updated!

purchase_order_secondary_unit/reports/purchase_order_templates.xml
AungKoKoLin1997: Updated!

purchase_order_secondary_unit/views/purchase_order_portal_templates.xml
```diff
@@ -0,0 +1,89 @@
+<odoo>
+    <!-- Unit price: add UoM and secondary price ('secondary' and 'both' modes) -->
+    <template id="purchase_secondary_uom_price">
```
yostashiro:
```suggestion
    <template id="purchase_uom_price">
```

AungKoKoLin1997: I changed the template name.

yostashiro:
@AungKoKoLin1997 Looks good. 👍  Please go ahead and squash commits.

AungKoKoLin1997:
Depends on https://github.com/OCA/product-attribute/pull/2182.

AungKoKoLin1997:
I added new commit to propagate `secondary_uom_id` from purchase order lines to account move lines when the field exists on account.move.line. This is useful when `account_move_secondary_unit` is installed, and avoids the need for a glue module.

purchase_order_secondary_unit/models/product_supplierinfo.py
```diff
@@ -0,0 +1,42 @@
...
+
+    @api.depends("price", "secondary_uom_id", "secondary_uom_id.factor")
+    def _compute_secondary_uom_price(self):
+        for rec in self:
+            if rec.secondary_uom_id:
+                rec.secondary_uom_price = rec.price * rec.secondary_uom_id.factor
+            else:
+                rec.secondary_uom_price = 0.0
+
+    @api.onchange("secondary_uom_price")
```
nobuQuartile, line 31:
```suggestion
```
I think we don't have to use @api.onchane for _inverse function.

purchase_order_secondary_unit/models/purchase_order.py
```diff
@@ -25,12 +34,34 @@ class PurchaseOrderLine(models.Model):
...
 
+    @api.depends("price_unit", "secondary_uom_id", "secondary_uom_id.factor")
+    def _compute_secondary_uom_price(self):
+        for rec in self:
+            if rec.secondary_uom_id:
+                rec.secondary_uom_price = rec.price_unit * rec.secondary_uom_id.factor
+            else:
+                rec.secondary_uom_price = 0.0
+
+    @api.onchange("secondary_uom_price")
```
nobuQuartile, line 59:
```suggestion
```
I think we don't have to use @api.onchane for _inverse function.

purchase_order_secondary_unit/models/product_supplierinfo.py
nobuQuartile: I removed the onchange myself and tested the behavior, and found that the value isn’t updated dynamically until the record is saved.
With onchange in place, it seems the value updates dynamically as soon as it’s written, even before saving.
So, we should have api.onchane for this _inverse function.

purchase_order_secondary_unit/models/purchase_order.py
nobuQuartile: I removed the onchange myself and tested the behavior, and found that the value isn’t updated dynamically until the record is saved.
With onchange in place, it seems the value updates dynamically as soon as it’s written, even before saving.
So, we should have api.onchane for this _inverse function.

nobuQuartile (approved):
LGTM
Code review and functional review

purchase_order_secondary_unit/models/product_supplierinfo.py
HviorForgeFlow: Maybe you can directly update the price in _onchange_product_id_secondary_uom and skip adding the onchange in the inverse.

purchase_order_secondary_unit/models/purchase_order.py
HviorForgeFlow: Same here

HviorForgeFlow (commented):
Code review, as it looks like adding a api.onchange in the inverse can work, because in the user interface the compute is not done until you save, you could add that price recalculation directly on the secondary uom on change, you can try, if this does not work we can go with your option

purchase_order_secondary_unit/models/product_supplierinfo.py
yostashiro: We could add a new onchange method that replicates the logic of this inverse (I guess _onchange_product_id_secondary_uom is not the one to work on) if we should be strict about not using onchange with an inverse.

I see a lot of these patterns, though, in the standard codebase, so I thought it was a widely accepted practice.

yostashiro:
@HviorForgeFlow Thank you very much for your review. Commented here: https://github.com/OCA/purchase-workflow/pull/2932#discussion_r2727970720

purchase_order_secondary_unit/models/product_supplierinfo.py
HviorForgeFlow: Why not reusing the onchange_product_id_secondary_uom? You update secondary_uom_id and price which is related with it in a single transaction

yostashiro: We could make it work but I don't feel there is a good reason to do so as the intents are different.

HviorForgeFlow: Ok, not blocking the decision, let me know when we can move it forward, appreciate that the module has been functional tested deeply :)

yostashiro: Thanks! You can go ahead and merge this if you are okay with the current code, or should we add new onchange methods to remove the onchange from inverses?

purchase_order_secondary_unit/models/purchase_order.py
```diff
@@ -25,12 +34,34 @@ class PurchaseOrderLine(models.Model):
...
         compute="_compute_product_packaging_id", store=True, precompute=True
     )
+    secondary_uom_price = fields.Float(
+        string="Secondary Price",
+        digits="Product Price",
+        aggregator="avg",
+        compute="_compute_secondary_uom_price",
+        inverse="_inverse_secondary_uom_price",
+        store=True,
+    )
```
HviorForgeFlow, line 44: My next concern is that this compute store can slowdown the update of the module for big databases. 

Maybe we need a migration script where we create the column and fill it manually with a sql query.

HviorForgeFlow (changes requested):
Code is not beautiful as it could be but it works.

My second concern is about adding that compute store fields.

@pedrobaeza what do you think about the onchange in the inverse and the compute store? Thanks in advice for taking your time to respond

yostashiro:
> Maybe we need a migration script where we create the column and fill it manually with a sql query.

Thanks for catching this!

@AungKoKoLin1997 Can you please help add a migration script for the purchase order line field?

AungKoKoLin1997:
> Can you please help add a migration script for the purchase order line field?

@yostashiro Done! 
I already bump module version from `18.0.1.0.1` → `18.0.1.0.2`.

purchase_order_secondary_unit/migrations/18.0.1.0.2/pre-migrate.py
```diff
@@ -0,0 +1,23 @@
...
+            UPDATE purchase_order_line pol
+                SET secondary_uom_price =
+                    COALESCE(
+                        pol.price_unit * (
+                            SELECT su.factor
+                              FROM product_secondary_unit su
+                            WHERE su.id = pol.secondary_uom_id
+                        ),
+                        0.0
+                    );
```
yostashiro: I suppose this would be more performant?
```suggestion
            UPDATE purchase_order_line pol
            SET secondary_uom_price = pol.price_unit * su.factor
            FROM product_secondary_unit su
            WHERE su.id = pol.secondary_uom.id;
```

purchase_order_secondary_unit/migrations/18.0.1.1.0/pre-migrate.py
yostashiro, line 1: The bump should be "minor". i.e. `18.0.1.0.1` -> `18.0.1.1.0`

purchase_order_secondary_unit/migrations/18.0.1.0.2/pre-migrate.py
AungKoKoLin1997: Yes. It will be.

yostashiro:
Looks good now.

purchase_order_secondary_unit/models/purchase_order.py
pedrobaeza: Yes, the Héctor's concern is legit. About being computed or use onchanges, I usually prefer computed fields for not depending on UI, but sometimes the things get messy due to dependencies, and a simple onchange resolves the usability question.

yostashiro: Hi @pedrobaeza , I think Héctor is concerned about adding `@api.onchange` to inverse methods. Do you have any thoughts on this?

pedrobaeza: Well, I meant about making it computed stored. About reusing `inverse` methods as onchange may be legit for reflecting things in UI, but it depends on the logic of inverse method.

HviorForgeFlow (approved):

mav-adhoc:
Hello! I don't get the "hide_secondary_uom_column(o)" in the view. There is not such attribute in res.company.
image

yostashiro:
@mav-adhoc Oops, https://github.com/OCA/product-attribute/pull/2182 needs to be merged, or we need to revert the parts that use hide_secondary_uom_column().

@sergio-teruel Would you be able to take a look at https://github.com/OCA/product-attribute/pull/2182?

purchase_order_secondary_unit/reports/purchase_order_templates.xml
```diff
@@ -6,16 +6,64 @@
...
+            >
                 <span t-field="line.secondary_uom_qty" />
                 <span t-field="line.secondary_uom_id" />
             </td>
         </xpath>
+        <!-- Quantity: hide primary when secondary is prioritized -->
+        <xpath expr="//span[@t-field='line.product_qty']" position="attributes">
+            <attribute
+                name="t-if"
+            >line.get_secondary_uom_display_mode() != 'secondary'</attribute>
```
gdgellatly, line 32: @yostashiro Where is this function? We updated and getting errors as not existing, can't find anywhere.

AungKoKoLin1997: This PR depends on https://github.com/OCA/product-attribute/pull/2182, and the function is introduced in that PR , as I mentioned in https://github.com/OCA/purchase-workflow/pull/2932#issuecomment-3771138318 . Unfortunately, this PR was merged before the dependency PR was merged. @yostashiro also explained the situation in the comment above: https://github.com/OCA/purchase-workflow/pull/2932#issuecomment-3844427117.

gdgellatly: OK, we have pinned to old version anyway after review as we really don't want these changes. In my opinion this should have been a seperate extension module, not because of the breakage, but it is for a feature that not everyone will want, its not really fixing an obvious oversight or error in functionality. In my experience this module is mostly used for straight conversions of qty's and adding price features just makes harder.

yostashiro: @gdgellatly That's a fair point and we did consider a split. The tradeoff is four additional modules across the core apps (xxx_secondary_unit_price for product, purchase, sale, and account), and we thought the maintenance overhead would outweigh the benefit. Our clients who need the secondary unit conversion also tend to need pricing per secondary unit, and all the added features here are effectively opt-in. That said, we are happy to split the module if the PSC opts for it. You can also leave a comment on https://github.com/OCA/product-attribute/pull/2182 as it gets harder to change direction after merge.

gdgellatly: @yostashiro Honestly it is no big deal, we have pinned now and that will get us through the v18 lifecycle which is only a few more months. For us a secondary unit is just a conversion to a primary unit, so price_subtotal is always identical and secondary_price_unit if we had would always just be price_subtotal/secondary_qty

And then look again during migration to decide what to do.

celm1990 (commented):
I think this PR must be reverted until the dependency https://github.com/OCA/product-attribute/pull/2182
 is merged; otherwise, this module is broken when you try to print the report.

yostashiro:
@AungKoKoLin1997 Can you please help create a PR to revert the template adjustments to avoid printing errors?

@celm1990 The dependency is now https://github.com/OCA/product-attribute/pull/2211 by the way. We could also have it merged quickly if it gets support from the community.

yostashiro:
Here is a follow-up: https://github.com/OCA/purchase-workflow/pull/2953

PR #2933, opened 2026-01-12
[19.0][MIG]: purchase_stock_price_unit_sync

sbiosca-s73:
Standard migration module v19 `purchase_stock_price_unit_sync`


Reyes4711-S73 (approved):

pablo-cort-s73 (approved):

Reyes4711-S73:
@OCA/procurement-purchase-maintainers Please, can you review/merge this PR?

purchase_stock_price_unit_sync/__manifest__.py
```diff
@@ -4,11 +4,11 @@
...
     "summary": "Update cost price in stock moves already done",
-    "version": "18.0.1.0.0",
+    "version": "19.0.1.0.0",
     "category": "Purchase",
     "website": "https://github.com/OCA/purchase-workflow",
     "author": "Tecnativa, Odoo Community Association (OCA)",
     "license": "AGPL-3",
     "installable": True,
-    "depends": ["purchase_stock"],
+    "depends": ["mrp", "purchase_stock"],
```
HviorForgeFlow: I guess that mrp should not be added as a dependency.

purchase_stock_price_unit_sync/models/purchase_order.py
```diff
@@ -36,20 +35,3 @@ def stock_price_unit_sync(self):
...
-                        lambda x: not x.stock_valuation_layer_id
-                    )
-                )
-                svls.write(
-                    {
-                        "unit_cost": line.with_context(
-                            skip_stock_price_unit_sync=True
-                        )._get_stock_move_price_unit(),
-                    }
-                )
```
HviorForgeFlow: AFAIK valuation has been moved to the stock.move so I guess that we cannot get rid of the feature but migrate the logic to the new data model

HviorForgeFlow (changes requested):
Code review. Some comments that I would appreciate to be discussed. Thanks for taking the time to migrate this module.

Reyes4711-S73 (approved):
Functional review ok
LGTM

purchase_stock_price_unit_sync/__manifest__.py
sbiosca-s73: To ensure full code coverage, we must include a dependency on `mrp`. Accessing `mrp.bom` without this dependency causes the tests to fail.

purchase_stock_price_unit_sync/models/purchase_order.py
sbiosca-s73: After checking the standard logic for `stock.valuation.layer`, I see that the data is transferred to `stock.move` and then deleted, so I don't see the need for any additional modifications.

HviorForgeFlow:
Adding mrp as dependency is not acceptable, as this module can run without the requirement on having Manufacturing app installed. I know that can look like a overengineering but if you need mrp for a specific case you should create a glue module on the top of this one an mrp.

For the valuation layers, we can remove them and create a ROADMAP were we alert the users that this part is skipped and maybe someone in the future can work on that.

HviorForgeFlow (approved):

PR #2935, opened 2026-01-13
Update translation files - pt_BR

rodmad85:


rodmad85:
Sorry, i will try in weblate.

PR #2936, opened 2026-01-19
[19.0][MIG] purchase_force_invoiced: Migration to 19.0

arantxa-s73:
Module migrated to version 19.0


Reyes4711-S73 (approved):
LGTM

david-s73 (approved):

arantxa-s73:
We have added inheritance to action_create_invoice so that it does not create the invoice, because with the base code from version 19, it does not block the creation of the invoice even if we change the invoice_status.

purchase_force_invoiced/reports/purchase_report.py
```diff
@@ -21,5 +21,5 @@ def _select(self):
             _logger.error(
                 "Query substitution failed. Check 'purchase/report/purchase_report.py "
                 "for changes"
-            )
+            )  # pragma: no cover
```
pedrobaeza: Please don't trick this for the coverage.

purchase_force_invoiced/security/res_groups_data.xml
```diff
@@ -3,8 +3,7 @@
     <data>
         <record model="res.groups" id="group_force_invoiced">
             <field name="name">Set purchase orders to 'Force Invoiced'</field>
-            <field name="category_id" ref="base.module_category_hidden" />
```
pedrobaeza: No equivalent now for the category?

purchase_force_invoiced/tests/test_purchase_force_invoiced.py
```diff
@@ -108,11 +106,13 @@ def test_purchase_order(self):
         )
 
         # We set the force invoiced.
-        po.button_done()
+        po.button_lock()
```
pedrobaeza, line 109: The PO shouldn't be locked for setting the flag, isn't it?

purchase_force_invoiced/tests/test_purchase_force_invoiced.py
```diff
@@ -108,11 +106,13 @@ def test_purchase_order(self):
...
         po.force_invoiced = True
         self.assertEqual(
             po.invoice_status, "invoiced", "The invoice status should be Invoiced"
         )
+        account = po.action_create_invoice()
+        self.assertTrue(account)
```
pedrobaeza: This check is a bit ambiguous.

pedrobaeza (commented):
/ocabot migration purchase_force_invoiced

purchase_force_invoiced/security/res_groups_data.xml
arantxa-s73: No, in fact, they also remove it from the base

https://github.com/odoo/odoo/blob/332b7543cdd22d7cdd7523cebb461e2c8422de62/addons/account/security/account_security.xml#L46

purchase_force_invoiced/tests/test_purchase_force_invoiced.py
arantxa-s73: The PO doesn't have that method now. I've seen in the base view that the lock button that used to be button_done is now button_lock. 
 https://github.com/odoo/odoo/blob/332b7543cdd22d7cdd7523cebb461e2c8422de62/addons/purchase/views/purchase_views.xml#L146

HviorForgeFlow (approved):
Code review

PR #2944, opened 2026-01-30
[IMP] product_supplierinfo_qty_multiplier: added logic for bypassing qty. multiplier

arielbarreiros96:
Proposing this improvement since the current implementation uses an @api.onchange("product_qty") for detecting if the product quantity is multiple or not, and enforces the behaviour or the multiplier.
However in some cases, including vendor negotiation, the multiplier could change, this interface doesn't allow to use any amount if it's not a multiple, therefore I'm proposing to create a bool at purchase order level, allowing to change this behaviour at will, by default this boolean will be False, keeping the default behaviour, so the current users won't need to adjust their current workflows.

I'm also adjusting the documentation to reflect the changes

@BinhexTeam T19318


product_supplierinfo_qty_multiplier/readme/USAGE.md
```diff
@@ -1,9 +1,8 @@
...
-- Assign it to the partner where the multiplier quantity was previously
-  set.
+- Assign it to the partner where the multiplier quantity was previously set.
 - Add a product line with the previously edited product.
-- Quantity is set to 6.0.
-- If you set 9.0 in the quantity field, the quantity will be rounded up
-  to 12.0
+- Quantity is set to 6.0 by default (based on the vendor multiplier).
+- If you set 9.0 in the quantity field, the quantity will be rounded up to
+  12.0 unless the 'Bypass vendor pricelist Qty. Multiplier' flag is enabled on the purchase order.
```
legalsylvain, line 6: Hi. Look great to me. 
But I'd to explicit in which case it's interesting to by pass the setting ? 
I mean, I developp this module to prevent user to buy one (or 2) bottle of milk, when supplier sell only by 6. In the use cases of my company, my users never said me they want to bypass the feature. Could you write some real use case in the USAGE.md (or DESCRIPTION.md file).

thanks !

legalsylvain (commented):
a minor remark inline. otherwise, LGTM. Thanks for your contribution !

epieters:
Hi @legalsylvain, this use case comes from us.
We have a customer where the # of products purchased do not need to be a multiple of the quantity.
Ideally it is always ordered as a multiple of the quantity, but it should be possible to overwrite it.

legalsylvain:
Thanks @epieters. Could you elaborate witch kind of produits ?
Thanks !

PR #2948, opened 2026-02-11
[19.0][MIG] product_supplierinfo_qty_multiplier: Migration to 19.0

therbin200:
Depends on web_notify [#3377](https://github.com/OCA/web/pull/3377)


alexey-pelykh (approved):
Thanks for the migration, looks good!

The 19.0 adaptations are all correct:

- Version bump to 19.0
- `product_uom` renamed to `product_uom_id` (purchase order line and supplierinfo view)
- Translation calls updated from `_()` to `self.env._()` with keyword args passed directly
- Unused `_` import removed

Tests pass on both Odoo and OCB. The only CI failure is the expected unreleased dependency on web_notify (OCA/web#3377).

BhaveshHeliconia (approved):
LGTM!

PR #2949, opened 2026-02-15
[18.0][IMP] purchase_request: Store the purchased_qty field.

celm1990:
I don’t see a reason to keep this field non-stored, and I think it could be a good idea to store it. What do you think?
@victoralmau @pedrobaeza could you please review this?


pedrobaeza (commented):
Ok about the change, but upgrading the module in a existing DB will take a lot of time depending on the existing records. Please create a migraine script.

victoralmau (approved):

pedrobaeza (changes requested):
Migration script pending

PR #2953, opened 2026-02-16
[18.0][FIX] purchase_order_secondary_unit: revert report features requiring unmerged dependency

yostashiro:
Since many implementations are getting hit by report crashes while OCA/product-attribute#2211 has yet to be merged... :sweat:

Partially revert a1eea87f for purchase order and quotation report templates. The reverted portions rely on methods
(hide_secondary_uom_column, get_secondary_uom_display_mode) introduced by OCA/product-attribute#2211, which is not yet merged. Without that dependency, the reports would crash when printing.

The secondary quantity column is kept but the conditional visibility and display mode logic are removed until the dependency is available.

@qrtl

### DIFF
--- a/purchase_order_secondary_unit/reports/purchase_order_templates.xml
+++ b/purchase_order_secondary_unit/reports/purchase_order_templates.xml
@@ -6,64 +6,16 @@
     >
         <!-- Header data -->
         <th name="th_quantity" position="before">
-            <th
-                name="th_secondary_unit"
-                class="text-end"
-                t-if="not o.company_id.hide_secondary_uom_column(o)"
-            >
+            <th name="th_secondary_unit" class="text-end">
                 <strong>Second Qty</strong>
             </th>
         </th>
         <!-- Content data -->
         <xpath expr="//span[@t-field='line.product_qty']/.." position="before">
-            <td
-                id="secondary_unit"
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):
Let's fast track it for fixing the situation

/ocabot merge patch

JochenDeBie:
@yostashiro Is it possible the changes made to the portal template here https://github.com/OCA/purchase-workflow/commit/a1eea87fc933c6d269b8dec7eb83c87a3cc91462#diff-88648ec43d7bc467356eb6a30b89ae0be9e11574ed6d50316772d2a7dae8d50c should also still be reverted? The reverted methods (report_show_price_uom, hide_secondary_uom_column, get_secondary_uom_display_mode, ) are also used in here.

AungKoKoLin1997:
@JochenDeBie I create a PR to revert the portal views changes.
https://github.com/OCA/purchase-workflow/pull/2967

carlos-lopez-tecnativa:
The README was not updated, so I removed this part in https://github.com/OCA/purchase-workflow/pull/3155/changes/bf8e6cc660bf52103bcc1f87c8a86fde00d7f235. I think this could also be updated in V18 because the feature has not been merged yet. What do you think?

PR #2955, opened 2026-02-17
[19.0][MIG] partner_supplierinfo_smartbutton

JasminSForgeFlow:
Standard Migration

@ForgeFlow


partner_supplierinfo_smartbutton/readme/INSTALL.md
```diff
@@ -0,0 +1,2 @@
+To use this module, you need to install "contacts" and "purchase"
```
legalsylvain: That is wrong, dont you think ?
This module can work without contacts / purchase installed. Dont you think ?

JasminSForgeFlow: Right, I have removed it.
Thanks

alexey-pelykh (approved):
Thanks for the migration! Looks good. The `read_group` -> `_read_group` update with tuple unpacking is correct for 19.0, version bump is right, XML was already using inline `invisible` from 18.0, and tests pass. LGTM.

BhaveshHeliconia (approved):
Functional review LGTM!

HviorForgeFlow (approved):

PR #2851, opened 2025-10-17
[19.0][MIG] purchase_last_price_info : migrate to 19.0

therbin200:

PR #2964, opened 2026-02-24
[18.0][IMP] purchase_last_price_info: improve field name for last purchase currency rate

JordiMForgeFlow:
The field is showing the currency rate of the last purchase, the wording was a bit confusing IMHO.

In order to not mess up already existing installations, we add new fields with the right wording and keep the old ones as related to the new ones.

Also FW of https://github.com/OCA/purchase-workflow/pull/2822

### DIFF
--- a/purchase_last_price_info/views/product_views.xml
+++ b/purchase_last_price_info/views/product_views.xml
@@ -14,16 +14,19 @@
                     <field name="last_purchase_supplier_id" />
                     <field name="last_purchase_date" />
                     <field name="last_purchase_price" />
-                    <field name="show_last_purchase_price_currency" invisible="1" />
+                    <field
+                        name="show_last_purchase_price_currency_rate"
+                        invisible="1"
+                    />
                     <field
                         name="last_purchase_currency_id"
                         groups="base.group_multi_currency"
-                        invisible="not show_last_purchase_price_currency"
+                        invisible="not show_last_purchase_price_currency_rate"
                     />
                     <field
-                        name="last_purchase_price_currency"
+                        name="last_purchase_price_currency_rate"
                         groups="base.group_multi_currency"
-                        invisible="not show_last_purchase_price_currency"
+                        invisible="not show_last_purchase_price_currency_rate"
                     />
                 </group>
             </field>
@@ -43,16 +46,19 @@
                     <field name="last_purchase_supplier_id" />
                     <field name="last_purchase_date" />
                     <field name="last_purchase_price" />
-                    <field name="show_last_purchase_price_currency" invisible="1" />
+                    <field
+                        name="show_last_purchase_price_currency_rate"
+                        invisible="1"
+                    />
                     <field
                         name="last_purchase_currency_id"
                         groups="base.group_multi_currency"
-                        invisible="not show_last_purchase_price_currency"
+                        invisible="not show_last_purchase_price_currency_rate"
                     />
                     <field
-                        name="last_purchase_price_currency"
+                        name="last_purchase_price_currency_rate"
                         groups="base.group_multi_currency"
-                        invisible="not show_last_purchase_price_currency"
+                        invisible="not show_last_purchase_price_currency_rate"
                     />
                 </group>
             </field>
--- a/purchase_last_price_info/models/product_product.py
+++ b/purchase_last_price_info/models/product_product.py
@@ -36,11 +36,19 @@ class ProductProduct(models.Model):
         string="Last Purchase Currency",
     )
     show_last_purchase_price_currency = fields.Boolean(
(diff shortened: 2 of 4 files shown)


LoisRForgeFlow:
For reference, the second commit is fwporting this https://github.com/OCA/purchase-workflow/pull/2822

JordiMForgeFlow:
@HviorForgeFlow could you check this one?

HviorForgeFlow (approved):

hugho-ad (changes requested):
I think first commit should land as a fix on the migration PR at https://github.com/OCA/purchase-workflow/pull/2851

Or better ensure a correct migration-script handaling the renaming of the field, you could use openupgradelib

JordiMForgeFlow:
@hugho-ad yes, in the migration PR we should simply add the new fields + renaming migration script. However, in V18 we cannot remove the existing fields to not break already existing modules that may inherit them.

CarmenMiranda:
Hi @JordiMForgeFlow , I think it will be best to merge this first, then the migration (since migration should only have migration changes and not improvements), at least the FW as it is (without any other changes, leaving the changes for renaming to the correct commit)

Maybe if this isn't approved yet, it will be better to separate them to merge the FW that's really needed

And since the fields are all non-stored, no migration script is needed

Cc. @hugho-ad @luisg123v @LoisRForgeFlow @HviorForgeFlow

hugho-ad:
@CarmenMiranda 

what about merge it AsIs, then at v19.0 just remove old fields?

CarmenMiranda:
> @CarmenMiranda
> 
> what about merge it AsIs, then at v19.0 just remove old fields?


@hugho-ad 
That's exactly what I said in the first paragraph, 

> I think it will be best to merge this first, then the migration

I just recommend that the FW of the PR 2822 should be done without any other unrelated changes; those should be applied in the commit "[IMP] purchase_last_price_info: improve field name for last" where they correspond

image
image

luisg123v:
@CarmenMiranda could you submit the PR with just the FP, please? That should be easy to merge.

CarmenMiranda:
> @CarmenMiranda could you submit the PR with just the FP, please? That should be easy to merge.

Hi @luisg123v, here it is the FW  https://github.com/OCA/purchase-workflow/pull/3119

JordiMForgeFlow:
@HviorForgeFlow could we merge this one so it can be added in the V19 migration PR?

JordiMForgeFlow:
@CarmenMiranda @luisg123v could you please review and approve this one so we can add the field rename in V19 migration?

purchase_last_price_info/views/product_views.xml
```diff
@@ -14,16 +14,19 @@
                     <field name="last_purchase_supplier_id" />
                     <field name="last_purchase_date" />
                     <field name="last_purchase_price" />
-                    <field name="show_last_purchase_price_currency" invisible="1" />
+                    <field
+                        name="show_last_purchase_price_currency_rate"
+                        invisible="1"
```
luisg123v, line 19: It's no longer required to add invisible fuields so that they're available in attributes of other fields.

luisg123v (changes requested):
I don't fully understand the rationale for this and there's no commit message.

Could you elaborate, please?

In addition, I think keeping related fields to others is not a clean solution, we will then have to maintain both. IMO, better renaming the fields. They're not stored, so you won't even need migration screipts to rname database columns.

purchase_last_price_info/views/product_views.xml
hugho-ad: FYI
https://github.com/OCA/purchase-workflow/pull/2964#issuecomment-4889780970

JordiMForgeFlow:
@luisg123v as I put in the PR comment, the wording is clearly wrong.

The field is showing last purchase currency rate, not the last purchase currency.

By having a field called "last purchase currency" you would expect to have values like "EUR, USD, etc".

Finally, you have to always consider that there may be custom developed modules that use the already existing fields, so changing a field in an already merged module is not the best practice. As a solution, we just keep the old fields for reference (to prevent breaking anything in V18) and in V19 migration the rename of the fields should be done.

Hope this clarifies your doubts :)

luisg123v (changes requested):
> Finally, you have to always consider that there may be custom developed modules that use the already existing fields

OK then. Just missing filling commit description, it's currently empty.

Please mention there, even though invisible fields are no longer required, it's added for backward compatibility and that they will be removed when migrating to 19.0.

JordiMForgeFlow:
@luisg123v do you think it can finally be approved now?

luisg123v (approved):
@JordiMForgeFlow excellent! Thanks a lot.

LGTM 👍

PR #2839, opened 2025-10-04
[18.0][FIX] purchase_manual_currency: assign amount_total_cc using total_company_currency

TheerayutEncoder:
1. Company currency is USD and CNY rate = 8.755600 CNY per USD
image

2. Create RFQ with Manual Currency as you can see, tax summarization displays total amount in company currency using rate in Currency not the manual currency rate
image

Fixed:
Reassign `tax_totals["amount_total_cc"]` with `total_company_currency`

PR #2984, opened 2026-03-06
[18.0][IMP] purchase_manual_currency / purchase_stock_manual_currency: refactor to use standard currency_rate

Saran440:
Refactor code manual_currency
- delete `is_manual`, `currency_diff` because they are not used
- delete `total_company_currency` since the standard field `amount_total_cc` already exists
- Replace `manual_currency_rate` with the standard `currency_rate`
- Update test script
- Reuse standard logic instead of introducing new custom logic

This PR is an improvement based on https://github.com/OCA/purchase-workflow/pull/2839

cc @TheerayutEncoder @AndreuOForgeFlow @HviorForgeFlow What do you think about this refactor?


TheerayutEncoder:
Does it mean we cannot edit currency rate on PO, so what the purpose of this module?

Saran440:
@TheerayutEncoder The `currency_rate` field already exists in the standard Purchase Order but is hidden.

This module simply exposes that field and allows users to edit it when using `manual currency`.
The computation still relies on the standard `currency_rate` logic instead of introducing a custom rate field.

TheerayutEncoder:
> @TheerayutEncoder The `currency_rate` field already exists in the standard Purchase Order but is hidden.
> 
> This module simply exposes that field and allows users to edit it when using `manual currency`. The computation still relies on the standard `currency_rate` logic instead of introducing a custom rate field.

Oh, I see it, so I think the way you did it is good. :+1:

TheerayutEncoder (approved):

HviorForgeFlow (commented):
I think you've done a great job simplifying those legacy custom fields, would it be required a migration script to don't break existing instances using this module, at least historical data?

Saran440:
@HviorForgeFlow Thank you for your advice. I added migration script and change version to 18.0.2.0.0 (purchase_manual_currency) and 18.0.1.0.1 (purchase_stock_manual_currency)
Can you check please?

HviorForgeFlow (approved):

PR #2988, opened 2026-03-13
[16.0][ADD] purchase_quick_product_supplierinfo_qty_multiplier

quentinDupont:
image

This module extends the OCA `purchase_quick` module to add extra fields to display that comes from `product_supplierinfo_qty_multiplier` 

This module force right input quantity (superior to min and compliant with multiplier). It adds temporarly a background red color warning, if quantity are not compliant with the supplier settings.


quentinDupont:
Shit, logic is different between `purchase_quick`  and this module

Last test of purchase_quick says :

>  Checks that if you enter a qty_to_process lower than the seller's min_qty, the price_unit in the purchase.order.line is the standard_price.

But my module forces the quantity to be compliant with supplier info : if quantity is lower than min_qty → this module changes the quantity.

That's why the CI is red :x:

PR #2992, opened 2026-03-17
[19.0][MIG] purchase_order_no_zero_price: Migration to 19.0

ellbristow:
Adds check on PO confirmation as this no longer works under api.constrains


ellbristow:
Test fails are from purchase_request. It's likely this module will cause issues with unit tests in other modules as it insists on having a unit_price in all POs

BhaveshHeliconia:
@ellbristow could you please rebase this?

PR #2994, opened 2026-03-17
[17.0][FIX] purchase_security: Prevent an access error when selecting a product in a purchase order

### LINKED ISSUE #2905: purchase_security: seller (product.supplerinfo) access can raise contact access error
Due to Contact security restrictions added by this addon, in some situations, restricted users could not save PO lines, when a product with restricted vendor prices are involved. Not sure which the best solution for this could be.

## Module
`purchase_security`

## Describe the bug
For a restricted Purchase user (e.g. "See own documents" one), a read access error is raised when a product with seller records is selected, if one of them belongs to a vendor (contact) restricted for the user.

## To Reproduce
At least at `17.0`, but presumably at other versions as well.

1. From a Purchase "Manager" user, add a Purchase record (Vendor pricelist) to a product A, for a certain vendor V. 
2. Ensure that selected Vendor V has a Purchase Representative set.
3. From a Purchase "User (own orders)" user _that is not vendor V Purchase Representative_, open a PO and try to select product A for a new line => a 'read' access error for Contact is raised.

**Expected behavior**
No error should be raised. Not sure what should Odoo do/select instead.

What should be the proper approach? I've checked that PO line's `_suggest_quantity()` method (that calls `_select_seller()` at `product` module) is the error responsible, so adding here a `sudo()` could solve it. Other approach should be adding record rules for `product.supplierinfo` (and hide those that belong to restricted vendors, for this group). Both solutions are not equivalent anyway....

cc @victoralmau @StefanRijnhart

victoralmau:
Prevent an access error when selecting a product in a purchase order

Fixes #2905 

Please @pedrobaeza and @dalonsod can you review it?

@Tecnativa

### DIFF
--- a/purchase_security/models/__init__.py
+++ b/purchase_security/models/__init__.py
@@ -1,4 +1,5 @@
 from . import ir_rule
+from . import product_supplierinfo
 from . import purchase_order
 from . import purchase_team
 from . import res_partner
--- a/purchase_security/models/product_supplierinfo.py
+++ b/purchase_security/models/product_supplierinfo.py
@@ -1,0 +1,32 @@
+# Copyright 2026 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class SupplierInfo(models.Model):
+    _inherit = "product.supplierinfo"
+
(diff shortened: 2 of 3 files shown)


StefanRijnhart (approved):
Nice, thanks!

pedrobaeza (approved):
/ocabot merge patch

PR #3003, opened 2026-03-20
[18][FIX] purchase_security: Avoid write in not stored compute field

florian-dacosta:
The write could causes issues for a user with readonly rights on purchase orders.
The `self.is_user_id_editable = is_user_id_editable` is working for multiple PO in recordset, maybe it would be cleaner to loop anyway, I can change this if needed.

@joao-p-marques @pilarvargas-tecnativa @dalonsod @Deriman-Alonso @bguillot


dalonsod (approved):
Only code review 👍

pedrobaeza (approved):
/ocabot merge patch

PR #3004, opened 2026-03-24
[19.0][MIG] purchase_order_revision: Migration to 19.0

BhaveshHeliconia:
- Depends on :- https://github.com/OCA/server-ux/pull/1193


CRogos:
@BhaveshHeliconia  Dependent module has been merged. You can remove the last commit.

BhaveshHeliconia:
> @BhaveshHeliconia Dependent module has been merged. You can remove the last commit.

Done. Thanks!

CRogos (changes requested):
1) Could you have a look into the migration of sale_order_revision, that we also adapt to the current revision changes? https://github.com/OCA/sale-workflow/pull/4024

Image

2) When you navigate to the prev. Revisions, I think we should get rid of the summary above the list?

Image

Image

CRogos:
The opened view looks good now, but the list is not filtered on the related purchases. 

image

CRogos:
Could you also squash these: 
image

CRogos (approved):
Strange that the verison 17+18 are not available, and the migration is done from 16.0.

Nevertheless module is working and code LGTM.

If there are two commits with add and update of the same person and the same language, these changes can be merged as well... but only a minor issue.
Image

PR #3005, opened 2026-03-24
[19.0][ADD] purchase_order_line_copier: New module to duplicate PO lines

BhaveshHeliconia:
This module introduces a new feature to easily duplicate Purchase Order Lines directly from the Purchase Order form.

### Key Features:
- One-click duplication: A new copy icon on each PO line allows for instant duplication.
- Bulk duplication: A "Copy PO Lines" button in the header opens a wizard where users can select multiple lines to copy at once.
- User-friendly: Seamless integration with the standard Purchase Order form.


BhaveshHeliconia:
@CristianoMafraJunior  @WesleyOliveira98  @ivantodorovich @yostashiro  could you please review this?

purchase_order_line_copier/models/purchase_order_line.py
```diff
@@ -0,0 +1,70 @@
...
+                "default_order_id": self.id,
+                "default_line_ids": [Command.set(line_ids)] if line_ids else False,
+            },
+        }
+
+
+class PurchaseOrderLine(models.Model):
+    _inherit = "purchase.order.line"
+
+    selected = fields.Boolean(string="Select Line")
```
WesleyOliveira98: I think this selection must not be stored on purchase.order.line. It is shared between users and remains set when the wizard is cancelled. Please keep this state exclusively in the transient wizard.

purchase_order_line_copier/security/ir.model.access.csv
```diff
@@ -0,0 +1,3 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_copy_purchase_line_wizard,access_copy_purchase_line_wizard,model_copy_purchase_line_wizard,base.group_user,1,1,1,1
+access_copy_purchase_line_wizard_line,access_copy_purchase_line_wizard_line,model_copy_purchase_line_wizard_line,base.group_user,1,1,1,1
```
WesleyOliveira98: These transient models should be restricted to purchase.group_purchase_user, not every internal user. Please update both access rules accordingly.

WesleyOliveira98 (changes requested):

purchase_order_line_copier/models/purchase_order_line.py
BhaveshHeliconia: Agreed!

purchase_order_line_copier/security/ir.model.access.csv
BhaveshHeliconia: Updated both access rules. Thanks!

WesleyOliveira98 (approved):

pedrobaeza:
I think you should use https://github.com/OCA/web/tree/19.0/web_widget_one2many_tree_line_duplicate as a generic method which works with any o2m.

BhaveshHeliconia:
> I think you should use https://github.com/OCA/web/tree/19.0/web_widget_one2many_tree_line_duplicate as a generic method which works with any o2m.

Thanks for pointing this out, Pedro! I wasn't aware of web_widget_one2many_tree_line_duplicate.

You're right that it already covers single-line duplication with options="{'allow_clone': True}. This module also has a bulk copy wizard, though, which lets users select and duplicate multiple lines at once.

Do you think that bulk copy feature is enough to justify keeping this as a separate module, or would you prefer to handle it through the generic widget as well?

pedrobaeza:
I don't think it's the usual path, as if someone wants that, they usually copy the order, and remove the undesired lines.

PR #3012, opened 2026-04-06
[18.0][IMP] purchase_exception: Use hook to popup exception after rollback

grindtildeath:
With OCA/server-tools#3590 changing the way exceptions are detected, the error raising doesn't allow to call `_popup_exception` as we used to, but we can use the new hook to have the popup displayed smoothly.

Depends on:
 - [ ] https://github.com/OCA/server-tools/pull/3590

### DIFF
--- a/purchase_exception/tests/test_purchase_exception.py
+++ b/purchase_exception/tests/test_purchase_exception.py
@@ -7,12 +7,17 @@ from datetime import datetime
 from odoo.tools import DEFAULT_SERVER_DATETIME_FORMAT
 
 from odoo.addons.base.tests.common import BaseCommon
+from odoo.addons.base_exception.tests.common import (
+    patch_base_exception_method_env,
+    swallow_base_exception_error,
+)
 
 
 class TestPurchaseException(BaseCommon):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
+        cls.env = cls.env(context=dict(cls.env.context, test_base_exception=True))
         # Useful models
         cls.PurchaseOrder = cls.env["purchase.order"]
         cls.PurchaseOrderLine = cls.env["purchase.order.line"]
@@ -71,6 +76,8 @@ class TestPurchaseException(BaseCommon):
             ],
         }
 
+    @patch_base_exception_method_env
+    @swallow_base_exception_error
     def test_purchase_order_exception(self):
         self.exception_noemail.active = True
         self.exception_qtycheck.active = True
--- a/purchase_request_exception/models/purchase_request.py
+++ b/purchase_request_exception/models/purchase_request.py
@@ -36,9 +36,11 @@ class PurchaseRequest(models.Model):
         if self.state == "to_approve":
             self.ignore_exception = False
 
+    def _must_popup_exception(self):
+        return True
+
     def button_to_approve(self):
-        if self.detect_exceptions() and not self.ignore_exception:
-            return self._popup_exceptions()
+        self.detect_exceptions()
         return super().button_to_approve()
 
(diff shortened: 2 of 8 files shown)


wouitmil (approved):

SilvioC2C (approved):

vvrossem (approved):

purchase_exception/tests/test_purchase_exception.py
```diff
@@ -7,12 +7,17 @@
...
+    patch_base_exception_method_env,
+    swallow_base_exception_error,
+)
 
 
 class TestPurchaseException(BaseCommon):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
+        cls.env = cls.env(context=dict(cls.env.context, test_base_exception=True))
```
divad1196, line 20: What is the standard way between this and [this](https://github.com/OCA/server-tools/pull/3590/changes#diff-703bc80203201f7f7d98c6bb9b350c8d91fd93c23a7aaa93bad1f4f82bf9d405R26)
```python
cls.env.context = dict(cls.env.context, test_base_exception=True)
cls.env = cls.env(context=dict(cls.env.context, test_base_exception=True))
```
?

Today, they are the same, but if tomorrow they differ, I think there should be a common way to do it . the `cls.env(context=...)` seems to be the more idiomatic.

purchase_exception/tests/test_purchase_exception.py
```diff
@@ -71,6 +76,8 @@ def setUpClass(cls):
             ],
         }
 
+    @patch_base_exception_method_env
```
divad1196, line 79: Since the most common use-case is to apply both decorators, shouldn't we have a decorator to combine them both ?

purchase_request_exception/models/purchase_request.py
```diff
@@ -36,9 +36,12 @@ def onchange_ignore_exception(self):
         if self.state == "to_approve":
             self.ignore_exception = False
 
+    def _must_popup_exception(self):
+        return True
+
     def button_to_approve(self):
         if self.detect_exceptions() and not self.ignore_exception:
-            return self._popup_exceptions()
+            return
```
divad1196: are we not supposed to return `action_popup_exceptions` ?

divad1196 (approved):
Except for `button_to_approve` where it looks like a function is missing, the rest looks good. Good work

purchase_exception/tests/test_purchase_exception.py
grindtildeath: Good catch, I didn't even notice I've been doing it differently  :sweat_smile: 

I agree `cls.env = cls.env(...)` looks like a cleaner way to do it though.

purchase_request_exception/models/purchase_request.py
grindtildeath: The popping up of exception is managed by the web client as the BaseExceptionError is returned from the backend (cf https://github.com/OCA/server-tools/pull/3590/changes#diff-da007865bb4c33420b568fd6c3d45c4f9bf94b7c6be09822ac2731a296aace30R11 )

Actually we could probably remove the `if` because `detect_exceptions` is going to raise the error anyway in case one is detected.

purchase_exception/tests/test_purchase_exception.py
grindtildeath: we could but IMO it's ok to have two different ones as it explicits the two things these are doing :shrug:

BhaveshHeliconia (approved):

simahawk:
Depending PR merged. Rebased.

PR #3016, opened 2026-04-14
[16.0][FIX] purchase_order_approved: skip purchase amount validation if it was already approved

JordiMForgeFlow:
When using Odoo standard PO amount validation, the check is done on button_approve. If we are using this module, when doing the button_release the amount validation has already been approved before, so we need to prevent it from being done again. Otherwise, purchase user can't do the release.

CC @HviorForgeFlow @rousseldenis @lmignon


HviorForgeFlow (approved):

GuillemCForgeFlow (approved):
lgtm 👍🏿

PR #3026, opened 2026-04-24
[FIX] Control the multi purchase order case in the pending field

Vicent-C6:
With a case of an invoice that group multiple purchase orders, this compute assumes that the 100% of invoice payments are from a single PO, This new loop handles the possibility that lines from different orders may be distinct and not be added together


LoisRForgeFlow (changes requested):
Can you add a test case?

arantxa-s73:
@LoisRForgeFlow test added

Reyes4711-S73 (approved):
LGTM

PR #3028, opened 2026-04-24
[19.0][MIG] purchase_commercial_partner: Migration to 19.0

BhaveshHeliconia:


dnplkndll (approved):
LGTM

dnplkndll:
I checked the new built in _read_group. the field/index is 1.85 faster @ 10k records.

BhaveshHeliconia:
> I checked the new built in _read_group. the field/index is 1.85 faster @ 10k records.

@dnplkndll Thanks for the review! It's great to confirm that storing and indexing this field still provides a significant 1.85x speedup by avoiding the join overhead in Odoo 19's native _read_group.

PR #3033, opened 2026-04-27
[19.0][MIG] purchase_landed_cost: Migration to version 19.0

Andrii9090-tecnativa:
Migration to 19.0

@pedrobaeza @victoralmau can you review please?
@Tecnativa
TT61958


purchase_landed_cost/models/purchase_expense_type.py
```diff
@@ -9,7 +9,7 @@ class PurchaseExpenseType(models.Model):
     _name = "purchase.expense.type"
     _description = "Purchase cost type"
 
-    name = fields.Char(required=True, translate=True, index=True)
+    name = fields.Char(required=True, translate=True, index="trigram")
```
pedrobaeza, line 12: Don't change this index, as we don't need similar searches.

purchase_landed_cost/wizard/import_landed_cost_pickings_wizard.py
```diff
@@ -24,7 +24,7 @@ class ImportLandedCostPickingsWizard(models.TransientModel):
     def default_get(self, fields_list):
         res = super().default_get(fields_list)
         if "possible_picking_ids" in fields_list:
-            expenses = self.env["purchase.cost.distribution.expense"].search([])
+            expenses = self.env["purchase.cost.distribution.expense"].browse()
```
pedrobaeza: This change is not correct.

Andrii9090-tecnativa: In this case the pre-commit checks failed 
`purchase_landed_cost/wizard/import_landed_cost_pickings_wizard.py:27: [W8163(no-search-all), ImportLandedCostPickingsWizard.default_get] Using an empty domain `search([])` without a `limit` will load all records, may impact performance.`

I think the limit is used incorrectly in this case.

pedrobaeza: Yes, but it's what we want to do, so you have to silent linter in this case, which is a warning anyway.

pedrobaeza (approved):

purchase_landed_cost/models/purchase_expense_type.py
Andrii9090-tecnativa: I need to change this to fix this warning. https://github.com/OCA/purchase-workflow/actions/runs/25308370215/job/74189325924?pr=3033#step:8:377

`WARNING odoo odoo.schema: Index attribute on 'purchase.expense.type.name' ignored, only trigram index is supported for translated fields`

pedrobaeza: Oh, I didn't know about this, thanks.

HviorForgeFlow (approved):
code review

PR #3036, opened 2026-04-28
[19.0][MIG] purchase_invoice_status_line: Migration to 19.0

BhaveshHeliconia:


tonygalmiche (changes requested):
Tested on a local Odoo 19.0 instance:

- Bug found: "Invoice Status" / "Force Invoiced" columns never appear on the
  actual Purchase Order form, only in the separate "Purchase Order Lines" menu
  (from purchase_order_line_menu). Root cause: the module only patches the
  reusable purchase.purchase_order_line_tree view, but the order lines list
  embedded in the PO form is defined inline in purchase.purchase_order_form
  and is not connected to that reusable view. The sibling module
  purchase_reception_status_line handles this correctly via an additional
  view (views/purchase_order.xml) that directly inherits
  purchase.purchase_order_form — purchase_invoice_status_line is missing the
  equivalent patch.

BhaveshHeliconia:
Thanks @tonygalmiche! Added the view patch for purchase.purchase_order_form to include the columns in the embedded PO order lines list.

tonygalmiche (approved):
Tested on a local Odoo 19.0 instance:

- "Invoice Status" / "Force Invoiced" columns now correctly display on the Purchase Order form (confirmed after the latest fix)
- Line status correctly reflects billing progress (Nothing to Bill / Waiting Bills)

Thanks @BhaveshHeliconia for the quick fix on the missing view patch!

No further issues found. Works as expected.

PR #3037, opened 2026-04-29
[17.0][FIX] purchase_order_general_discount: prevent overriding manual line discounts

atgalvez08:
@BinhexTeam

## Problem

When modifying discounts on purchase order lines, the onchange defined in
`purchase_order_general_discount` is triggered due to the dependency on
`order_line`.

This causes the general discount to be reapplied automatically,
overriding any manual changes performed by the user.

## Root cause

The onchange decorator includes `order_line`:

    @api.onchange("general_discount", "order_line")

This causes the method to be triggered on any modification of order lines.

## Solution

Restrict the onchange trigger to `general_discount` only.

## Result

- Changing general discount applies to all lines ✔
- Manual edits on line discounts are preserved ✔
- No unintended overwrites occur ✔

## Tests

A new test has been added to ensure that manual overrides on line discounts
are not lost after editing.

## Notes

This issue becomes especially visible when using modules such as
`purchase_triple_discount`, but can affect any module that modifies
line-level discounts.

### DIFF
--- a/purchase_order_general_discount/models/purchase_order.py
+++ b/purchase_order_general_discount/models/purchase_order.py
@@ -36,7 +36,7 @@ class PurchaseOrder(models.Model):
         discount_field = self.company_id.purchase_general_discount_field
         return discount_field or "discount"
 
-    @api.onchange("general_discount", "order_line")
+    @api.onchange("general_discount")
     def onchange_general_discount(self):
         discount_field = self._get_general_discount_field()
         self.mapped("order_line").update({discount_field: self.general_discount})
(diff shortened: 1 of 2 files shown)


rrebollo (commented):
Given the nature of the issue you’re facing, especially since it also involves interaction with another module that affects discounts—but without an explicit hard dependency between them—it may be worth revisiting whether removing that field as an `onchange` trigger is actually justified.

We should verify the legitimacy of that trigger in the first place, because the addon documentation explicitly mentions *peaceful coexistence* with the triple discount module. That suggests compatibility was at least an intended design goal.

From a first look, the fix being proposed feels somewhat aggressive. A more conservative approach might be preferable before altering trigger behavior globally.

For example, it could be worth exploring whether the `onchange` can selectively ignore or filter purchase order lines in the scenarios where recalculation should not happen, instead of removing the trigger entirely. That would reduce regression risk and better preserve expected behavior for existing users.
https://github.com/OCA/purchase-workflow/blob/18.0/purchase_order_general_discount/README.rst#configuration

PR #3043, opened 2026-04-30
[19.0][MIG] purchase_representative: Migration to 19.0

BhaveshHeliconia:


tonygalmiche (approved):
Tested on Odoo 19.0: 
- a sale order auto-triggered a PO via the buy route (created by __system__). 
- Verified in DB that the PO's "Purchase Representative" correctly resolved to the actual confirming user, not the superuser.

Works as expected.

PR #3048, opened 2026-05-01
[FIX] purchase_discount: match SQL constraint with Postgres format

randall-vx:
Odoo drops and recreates constraints if the Python definition does not exactly match the definition returned by PostgreSQL. Because PostgreSQL normalizes the expression to include extra parentheses, the previous definition 'CHECK (discount <= 100.0)' caused a mismatch.

This mismatch led to a DuplicateObject exception during module update because the drop constraint operation was skipped or failed silently and the subsequent add constraint operation failed, raising a WARNING in the logs and failing strict CI checks.

By matching the exact string returned by PostgreSQL, we prevent Odoo from attempting to recreate the constraint on every registry initialization.

### DIFF
--- a/purchase_discount/models/product_supplierinfo.py
+++ b/purchase_discount/models/product_supplierinfo.py
@@ -25,7 +25,7 @@ class ProductSupplierInfo(models.Model):
     _sql_constraints = [
         (
             "discount_limit",
-            "CHECK (discount <= 100.0)",
+            "CHECK ((discount <= 100.0))",
             "Supplier discount must be lower than 100%.",
         )
     ]


randall-vx:
@moylop260 could you review please ?

regards,

moylop260:
could you request review of maintainers or authors, please?

legalsylvain:
maybe it should be handled upstream. 
I mean, if odoo drop a constraints, then recreate the same constrains, again and again, in each update all, I see it more like a bug (or a non optimization) in Odoo core. 
I don't see changing all the constraint of all the OCA modules as as solution.
Don't you think?

randall-vx:
@legalsylvain  thank you for the review!

I agree that Odoo core should handle SQL normalization better. However, as seen in issues like [odoo/odoo#40740](https://github.com/odoo/odoo/issues/40740), this is a long-standing behavior where Odoo relies on a strict string comparison.

By not matching the PostgreSQL output format, we force Odoo to drop and recreate the constraint on every single module update. This is not just a 'non-optimization'; it causes noise in logs and can lead to CI failures or DuplicateObject exceptions during concurrent updates. Given that an upstream fix is unlikely in the short term, following the PostgreSQL normalization format is the standard pragmatic approach in OCA to ensure stable and silent updates.

While I agree this should ideally be handled by the Odoo ORM, providing the normalized string is the current _**de facto standard**_ in the ecosystem to prevent redundant DDL operations and ensure a clean CI/CD pass, especially considering how sensitive OCA tools are to "WARNINGS" in logs during CI.

https://github.com/odoo/odoo/pull/30334
https://github.com/odoo/odoo/issues/25661
https://github.com/odoo/odoo/issues/28381
https://github.com/odoo/odoo/issues/40740

PR #3055, opened 2026-05-06
[19.0][MIG] purchase_container

theangryangel:
Standard migration bar 2 "issues":

* `weight_uom_id` originally had a domain limiting to the weight uom category - this longer exists, so I've just erased it in absence of a better idea for the moment
* `uom_po_id` no longer exists, I followed the upstream behaviour in moving to `uom_id`


purchase_container/tests/test_module.py
```diff
@@ -0,0 +1,78 @@
+from odoo import Command, fields
+
+from odoo.addons.base.tests.common import BaseCommon
+
+
+class Test(BaseCommon):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.partner = cls.env["res.partner"].create({"name": "Test Supplier"})
```
BhaveshHeliconia:
```suggestion
```

BhaveshHeliconia (changes requested):
No need to create a res.partner record when using BaseCommon, as it is already provided by the base class.

purchase_container/tests/test_module.py
theangryangel: Thanks Bhavesh <3 Updated.

BhaveshHeliconia (approved):
Code review LGTM!

PR #3057, opened 2026-05-07
[18.0][IMP] product_main_seller: Add main_seller_id to product.product product and add to depends date_start, date_end

Andrii9090-tecnativa:
Add field `main_seller_id` to `product.product`

@pedrobaeza can you review, please?
@Tecnativa


product_main_seller/models/product_product.py
```diff
@@ -0,0 +1,41 @@
...
+    @api.depends(
+        "variant_seller_ids.sequence",
+        "variant_seller_ids.partner_id.active",
+        "variant_seller_ids.date_start",
+        "variant_seller_ids.date_end",
+    )
+    def _compute_main_seller_id(self):
+        for product in self:
+            if product.variant_seller_ids:
+                product.main_seller_id = fields.first(
```
pedrobaeza, line 27: You are missing here the filter for only this variant to have priority.

legalsylvain (changes requested):
Hi. 
Unfortunately, it is not possible to compute and store fields, based on  fields.Date.today() as this value is changing all the time.

pedrobaeza:
The idea is to include a cron for updating that daily. @Andrii9090-tecnativa please do it at the same time.

product_main_seller/models/product_supplierinfo.py
```diff
@@ -0,0 +1,14 @@
...
+
+class ProductSupplierinfo(models.Model):
+    _inherit = "product.supplierinfo"
+
+    @api.model
+    def _cron_recompute_main_seller_id(self):
+        products = self.env["product.product"].search([])
+        templates = self.env["product.template"].search([])
+        products._compute_main_seller_id()
+        templates._compute_main_seller_id()
```
legalsylvain, line 14: This will be generate a huge computation, even if most of the value are the same each night.

If we filter, to only recompute for  product that has supplierinfo with date (start and or end) defined, the recomputation will be very quicker. (even no recomputation, for many instances that doesn't use start / end feature.

WDYT ?

product_main_seller/models/product_supplierinfo.py
```diff
@@ -0,0 +1,14 @@
...
+from odoo import api, models
+
+
+class ProductSupplierinfo(models.Model):
+    _inherit = "product.supplierinfo"
+
+    @api.model
+    def _cron_recompute_main_seller_id(self):
+        products = self.env["product.product"].search([])
+        templates = self.env["product.template"].search([])
```
legalsylvain, line 12:
```suggestion
        products = self.env["product.product"].with_context(active_test=false).search([])
        templates = self.env["product.template"].with_context(active_test=false).search([])
```

legalsylvain (commented):
Thanks for adding the cron ! 

some remarks inline.

PR #3067, opened 2026-05-20
[19.0][MIG] purchase_warn_message: Migration to 19.0

dnplkndll:
Port of `purchase_warn_message` from 18.0 to 19.0. Follows the [OCA migration guide](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-19.0).

Non-mechanical adaptations worth flagging:

- Odoo 19 dropped `res.partner.purchase_warn` (the Selection field that gated whether to warn). The module's compute now keys on `purchase_warn_msg` being non-empty instead of the gone Selection value.
- Odoo 19 dropped the `done` value from `purchase.order.state`. The compute now suppresses the warning only on `cancel`; the corresponding test was removed.
- Tests updated to drop `purchase_warn` from the partner `create()` dicts.


dnplkndll:
ready for review  @marcelsavegnago @BhaveshHeliconia

BhaveshHeliconia (commented):
This module’s functionality is already available in Odoo core. You can check it here:
[Odoo Core Purchase Order Implementation](https://github.com/odoo/odoo/blob/6f6aa962c36975f90a78b7fe4263f6e7ab171997/addons/purchase/models/purchase_order.py?utm_source=chatgpt.com#L179)

dnplkndll:
Closing per
  @BhaveshHeliconia — Odoo 19.0 introduced `purchase_warning_text`
  (`purchase_order.py:179`) which computes warnings from partner + parent company
  + **product lines**, going strictly beyond what this module provided
  (partner-only). The only piece core doesn't replicate is the form-banner XML,
  which is now trivially a view-only override rather than a full module. Thanks for
  catching this!

PR #3078, opened 2026-05-27
[18.0][IMP] purchase_invoice_status_line: control policy

JoanSForgeFlow:
Adds the product Control Policy (On ordered / On received) as an optional column on the PO line, to make clear why a line is in Waiting Bills vs Nothing to Bill. The value is a stored snapshot synced with the invoicing recompute, so it stays consistent with qty_to_invoice and is not altered retroactively when the product policy changes.

Includes a minor fix: `_compute_invoice_status` was using `qty_to_invoice` without declaring the dependency.

It also fixes the billing status for over-billed lines (negative qty_to_invoice): these were wrongly shown as Waiting Bills. Now only a strictly positive qty_to_invoice shows Waiting Bills.

### DIFF
--- a/purchase_invoice_status_line/models/purchase_order_line.py
+++ b/purchase_invoice_status_line/models/purchase_order_line.py
@@ -27,7 +27,7 @@ class PurchaseOrderLine(models.Model):
         "even if some quantities are not fully invoiced. ",
     )
 
-    @api.depends("qty_invoiced", "product_qty", "force_invoiced")
+    @api.depends("qty_invoiced", "qty_to_invoice", "product_qty", "force_invoiced")
     def _compute_invoice_status(self):
         precision = self.env["decimal.precision"].precision_get(
             "Product Unit of Measure"
--- a/purchase_invoice_status_line/models/purchase_order_line.py
+++ b/purchase_invoice_status_line/models/purchase_order_line.py
@@ -27,6 +27,25 @@ class PurchaseOrderLine(models.Model):
         "even if some quantities are not fully invoiced. ",
     )
 
+    purchase_method = fields.Selection(
+        selection=[
+            ("purchase", "On ordered"),
+            ("receive", "On received"),
+        ],
+        string="Control Policy",
+        compute="_compute_purchase_method",
+        store=True,
+    )
+
+    @api.depends("product_id", "qty_to_invoice")
+    def _compute_purchase_method(self):
+        # qty_to_invoice is a dependency on purpose: changing the policy on the
+        # product does not recompute existing lines, so this snapshot is
+        # refreshed whenever invoicing is recomputed (the moment the policy is
+        # actually applied), keeping it in sync with the billing status.
+        for line in self:
+            line.purchase_method = line.product_id.purchase_method
+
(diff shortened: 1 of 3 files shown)


LauraCForgeFlow (approved):
Code LGTM 👍

purchase_invoice_status_line/models/purchase_order_line.py
```diff
@@ -27,7 +27,22 @@ class PurchaseOrderLine(models.Model):
...
+        selection=[
+            ("purchase", "On ordered"),
+            ("receive", "On received"),
+        ],
+        string="Control Policy",
+        compute="_compute_purchase_method",
+        store=True,
+    )
+
+    @api.depends("product_id", "qty_to_invoice")
```
JordiMForgeFlow, line 40: why qty to invoice is needed in this compute?

purchase_invoice_status_line/models/purchase_order_line.py
```diff
@@ -27,7 +27,22 @@ class PurchaseOrderLine(models.Model):
...
 
-    @api.depends("qty_invoiced", "product_qty", "force_invoiced")
+    purchase_method = fields.Selection(
+        selection=[
+            ("purchase", "On ordered"),
+            ("receive", "On received"),
+        ],
+        string="Control Policy",
+        compute="_compute_purchase_method",
+        store=True,
```
JordiMForgeFlow, line 37: if you are making it store True then better to also keep it as readonly True so it can't be modified

purchase_invoice_status_line/models/purchase_order_line.py
JoanSForgeFlow: The goal is that the Control Policy shown on the line always matches the one actually used to compute qty_to_invoice and invoice_status.

If someone changes the policy on the product, existing lines do not recompute qty_to_invoice / invoice_status automatically. But the moment something triggers a recompute (a receipt, etc.), those are recalculated with the new policy, and that is exactly when we want the line's policy to update too, because from that point the new one is the one being used. Depending on qty_to_invoice makes the field refresh at those same moments.

JordiMForgeFlow: @JoanSForgeFlow could you then add a comment so that no one gets confused in the future? 🙏🏼

purchase_invoice_status_line/models/purchase_order_line.py
JoanSForgeFlow: In Odoo 18 a computed field without an inverse is already readonly=True by default.

JordiMForgeFlow: TIL, thank you!!

JordiMForgeFlow (approved):

BhaveshHeliconia (approved):
One minor suggestion: Could you add a brief inline comment in the _compute_purchase_method method explaining why qty_to_invoice is included as a dependency? It would help prevent future maintainers from accidentally removing it.

AaronHForgeFlow (approved):
👍 code LGTM

HviorForgeFlow (approved):
code review 👍

PR #3080, opened 2026-05-28
[16.0][IMP][purchase_blanket_order]  add send a mail and put supplier…

Kev-Roche:
… code in line name

This PR Add "Send by Email" button on Purchase Blanket Orders and fix purchase order line name to use supplier code.

### DIFF
--- a/purchase_blanket_order/models/blanket_orders.py
+++ b/purchase_blanket_order/models/blanket_orders.py
@@ -373,6 +373,30 @@ class BlanketOrder(models.Model):
         res.append(("id", "in", order_ids.ids))
         return res
 
+    def action_send_email(self):
+        self.ensure_one()
+        template = self.env.ref(
+            "purchase_blanket_order.email_template_blanket_order",
+            raise_if_not_found=False,
+        )
+        ctx = {
+            "default_model": "purchase.blanket.order",
+            "default_res_id": self.id,
+            "default_use_template": bool(template),
+            "default_template_id": template and template.id or False,
+            "default_composition_mode": "comment",
+            "force_email": True,
+        }
+        return {
+            "type": "ir.actions.act_window",
+            "view_mode": "form",
+            "res_model": "mail.compose.message",
+            "views": [(False, "form")],
+            "view_id": False,
+            "target": "new",
+            "context": ctx,
+        }
+
 
 class BlanketOrderLine(models.Model):
     _name = "purchase.blanket.order.line"
--- a/purchase_blanket_order/__manifest__.py
+++ b/purchase_blanket_order/__manifest__.py
@@ -24,6 +24,7 @@
         "views/purchase_order_views.xml",
         "report/templates.xml",
         "report/report.xml",
+        "data/mail_template_blanket_order.xml",
     ],
     "installable": True,
 }
--- a/purchase_blanket_order/data/mail_template_blanket_order.xml
+++ b/purchase_blanket_order/data/mail_template_blanket_order.xml
@@ -1,0 +1,45 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!-- Copyright (C) 2026 Akretion (<http://www.akretion.com>).
+     @author Kévin Roche <kevin.roche@akretion.com>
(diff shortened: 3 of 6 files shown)


BhaveshHeliconia (commented):
I have completed the review of the code implementation for the Purchase Blanket Order Email Template
Here is a summary of the assessment and the optimizations applied:
1. Email Subject Line Enhancement
 - Current state: {{ object.name }} (e.g., BO00001)
 - Issue: Receiving an email with just a document number can easily look like spam or confuse suppliers, resulting in lower open rates.
 - Fix: Upgraded the subject line to include the company name and document type context:
{{ object.company_id.name }} Blanket Order (Ref {{ object.name or '' }})

2. HTML Structure Fixes (Body HTML)
 - Issue: The original template placed a <tbody> tag directly inside a <div> layer without an enclosing <table> wrapper. This is semantically invalid HTML and causes layout breakage across strict email clients like Microsoft Outlook and Gmail.
 - Fix: Re-structured the markup into an Odoo-standard responsive HTML email table container.

BhaveshHeliconia:
3. Company Branding & Fallbacks
 - Improvement: Included the company context inside the email text body. If a Blanket Order is sent but the internal Odoo user profile (object.user_id) lacks a properly configured name, the signature will seamlessly fall back onto the company name (object.company_id.name) to maintain a professional appearance.

4. Backend Python Optimization
 - Sanitized the context dictionary layout using Odoo's native clean_context() helper method to prevent active view variables from conflicting with the Mail Composer wizard.
 - Streamlined the dynamic variable assignment using modern Python conditional expressions.

Kev-Roche:
Thanks @BhaveshHeliconia for your review, I applied your suggestions about 1,2,3 but did not understand the 4th point.

purchase_blanket_order/models/blanket_orders.py
```diff
@@ -373,6 +373,30 @@ def _search_remaining_uom_qty(self, operator, value):
         res.append(("id", "in", order_ids.ids))
         return res
 
+    def action_send_email(self):
+        self.ensure_one()
+        template = self.env.ref(
+            "purchase_blanket_order.email_template_blanket_order",
+            raise_if_not_found=False,
+        )
+        ctx = {
```
BhaveshHeliconia, line 382: 1. "Sanitized the context dictionary layout using Odoo's native
In your original method, you defined the context dictionary like this:
ctx = {
    "default_model": "purchase.blanket.order",
    "default_res_id": self.id,
    ...
}

When you return an action to open a new wizard window in Odoo, Odoo automatically carries over the current view's context (self.env.context).

If your user clicked the "Send Email" button from a specific Kanban view, a Search filter, or an active form view action, variables like active_id, active_model, or view-specific default values (e.g., default_partner_id) are passed silently along in the background. When the Mail Composer wizard opens up, these hidden context keys can conflict with the email wizard's fields, causing visual bugs, wrong data population, or access errors.

The Optimized Fix:
ctx = dict(clean_context(self.env.context))
ctx.update({
    "default_model": "purchase.blanket.order",
    "default_res_id": self.id,
    # ...
})
Odoo's native clean_context() strips out all action-specific, search-specific, and view-specific keys (like active_id, search_default_..., etc.) from the environment context. It hands you a clean slate. By converting it to a dictionary and updating it, you guarantee that only the specific parameters you intend for the Mail Composer are passed, eliminating unpredictable backend bugs.

2. "Streamlined the dynamic variable assignment using modern Python conditional expressions."
In your original snippet, the template ID assignment was written using old Python/Odoo legacy shorthand notation:
"default_template_id": template and template.id or False,
This is an old-school way of writing an inline fallback condition. It works like this: If template is true, evaluate template.id. If template.id is true, return it; otherwise return False.

While it works most of the time, it introduces a dangerous edge case in Python: if template.id evaluates to 0 (or any falsy value), the expression will mistakenly skip to the or False statement. While Odoo database IDs are rarely 0, it is still structurally unsafe and harder to read at a glance.
"default_template_id": template.id if template else False,

This uses modern Python ternary operators (conditional expressions). It translates perfectly to human-readable logic: "Give me template.id if a template exists, otherwise give me False."

Issue #3070, opened 2026-05-22
[18.0] purchase_allowed_product: EvalError "parent is not defined" when account.move.line form is rendered without a parent move


MartinHeisig:
## Module

`purchase_allowed_product`

## Description

The view `view_move_line_form_supplied_product` (in
`views/account_move_views.xml`) adds a `context` attribute to the
`product_id` field of the generic journal item form
(`account.view_move_line_form`). The context references `parent`:

    {
      'restrict_supplier_id': parent.partner_id,
      'use_only_supplied_product':
          parent.move_type in ('in_invoice', 'in_refund', 'in_receipt')
          and parent.use_only_supplied_product
    }

`parent` is only resolvable while the line is rendered inside the x2many
list of its parent `account.move`. When an `account.move.line` is rendered
without a parent move in scope, `parent` does not exist and Odoo raises:

    EvalError: Name 'parent' is not defined

This aborts the Owl lifecycle and the form cannot be opened.

## Steps to reproduce

1. Install `purchase_allowed_product`.
2. Open any invoice (customer or vendor) and, on an invoice line that has a
   product, open the line as a form (its internal-link / form view).
   The crash occurs.

Alternative reproduction, triggering the same crash:
- Accounting > Journal Items, open any journal item that carries a product.

Journal items without a product (tax / receivable / payable lines) are not
affected, because the `product_id` field is not rendered for them.

## Traceback

    EvalError: Can not evaluate python expression:
    ({'restrict_supplier_id': parent.partner_id,
      'use_only_supplied_product':
          parent.move_type in ('in_invoice', 'in_refund', 'in_receipt')
          and parent.use_only_supplied_product})
    Error: Name 'parent' is not defined
        at evaluateExpr (...)
        at makeContext (...)
        at getFieldContext (...)
        at extractProps (...)
        at Field.template (...)

## Notes

- The other view of the module, `view_move_form_supplied_product`, places
  the same context on `invoice_line_ids/list/field[product_id]`, i.e. inside
  the x2many line, where `parent` resolves correctly. That view is not
  affected.
- This inherited view extends the generic `account.view_move_line_form`,
  which applies to all `account.move.line` records regardless of move type,
  not only to purchase-related documents. The restriction to vendor
  documents happens solely through the `move_type` condition inside the
  context expression.
- The view definition in `views/account_move_views.xml` is identical on the
  18.0 and 19.0 branches, so the 19.0 branch may carry the same defect; this
  has not been verified at runtime.
[...]

PR #3082, opened 2026-05-30
[18.0][FIX] purchase_allowed_product: avoid parent in line form

fsmw:
This fixes the standalone form crash in when the field context is evaluated without an x2many parent scope.

Cause
- the inherited standalone form view used in the context
- standalone move line forms do not define, so the web client raised 
Solution
- replace references with in the standalone move line form view
- add a regression test asserting the merged form architecture uses and no longer references 

Refs #3070

### DIFF
--- a/purchase_allowed_product/views/account_move_views.xml
+++ b/purchase_allowed_product/views/account_move_views.xml
@@ -32,10 +32,10 @@
             <field name="product_id" position="attributes">
                 <attribute name="context">
                     {
-                    'restrict_supplier_id': parent.partner_id,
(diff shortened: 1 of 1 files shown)


pedrobaeza (changes requested):
OK for changing `parent` to `move_id`, but please don't add a test that is not adding value.

fsmw:
@pedrobaeza thanks for the review — done. Dropped the new test in 7a466b8f8; only the XML fix remains. Could you re-review when you have a moment?

pedrobaeza (approved):

BhaveshHeliconia (approved):
LGTM!

PR #3086, opened 2026-06-05
[16.0][IMP] purchase_order_approved: show PO name in approved state

JordiMForgeFlow:
https://github.com/odoo/odoo/blob/16.0/addons/purchase/report/purchase_order_templates.xml#L24

had to add a high priority to prevent conflict with https://github.com/OCA/purchase-workflow/blob/18.0/purchase_no_rfq/reports/purchase_order_template.xml#L6


HviorForgeFlow (approved):

GuillemCForgeFlow (approved):
lgtm 👍🏿

PR #3093, opened 2026-06-06
[18.0][MIG] purchase_deposit_analytic

Saran440:
Change `analytic_account_id` to `analytic_distribution`


BhaveshHeliconia (commented):
I currently reviewing the financial workflow for advance payments (deposits) on Purchase Orders (POs) and want to confirm your preferred accounting behavior regarding Analytic Distributions.
Current Workflow Behavior:
1. RFQ Creation: A Request for Quotation is created where the original product line items are assigned to Analytic Account A (e.g., Department A).
2. Deposit Registration: The PO is confirmed, and the user clicks Register Deposit to open the payment wizard.
3. Wizard Modification: In the wizard, the user explicitly changes the analytic distribution to Analytic Account B (e.g., a corporate prepayment or clearing account) and clicks Create and View Bills.
4. Result:
- The generated Vendor Bill is created using Analytic Account B.
- A new Purchase Deposit Line is added to the PO, also using Analytic Account B.
- The original product line items remain untouched, keeping Analytic Account A.

Is it correct that the newly created Deposit Line and Vendor Bill follow the wizard's distribution (Account B)? Or should they automatically copy the original line's distribution (Account A)?

If a user overrides the distribution in the wizard (Account B), should this change retroactively update the existing, original PO lines, or should those original lines keep their initial values (Account A)?

PR #3095, opened 2026-06-09
[MIG] product_supplierinfo_disable_autocreation: Migration to 19.0

nedaszilinskas:


product_supplierinfo_disable_autocreation/models/res_company.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2025 Juan Alberto Raja<juan.raja@sygel.es>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class Company(models.Model):
+    _inherit = "res.company"
+
+    purchase_pricelist_disable_autocreate = fields.Boolean(
```
legalsylvain, line 9:
```suggestion
    purchase_pricelist_enable_autocreate = fields.Boolean(
```

= when checked, the autocreation is enabled. The name is not correct I think and doesn't match with the string and the help values.

don't you think ?

mlk994:
@nedaszilinskas Hi! are you still working on this migration?

PR #3097, opened 2026-06-10
[19.0][FIX] purchase_request: handle False move_dest_ids in procurement values

### LINKED ISSUE #2985: Purchase Request module 19v Error on Supply Method: Take From Stock, if unavailable, Trigger Another Rule.
An Odoo (Odoo 19 EE)  error occurs when creating a purchase request if the route is configured with Supply Method: Take From Stock, if unavailable, Trigger Another Rule.

If other Supply Methods are configured, this problem doesn't occur and the system correctly creates a purchase request.

Could you help me figure out this problem?

Video: https://monosnap.ai/file/EGgtcaZhRmVrx8r8SbZTdZf7hNVn9S
Link to error: https://self-erp-kik-st-03-27648072.dev.odoo.com/odoo/documents/TSD8ThBNTQqYrHJZtncBTQo8f

## Module
The name of the module that has a bug.
Purchase Request

## Describe the bug
https://self-erp-kik-st-03-27648072.dev.odoo.com/odoo/documents/TSD8ThBNTQqYrHJZtncBTQo8f

## To Reproduce
**Affected versions**:
Odoo 19 EE
1. Product with route where supply method is Take From Stock, if unavailable, Trigger Another Rule.
2. System shows error when we try create Sale Order
3.

**Expected behavior**

**Additional context**
Add any other context about the problem here. (e.g. OS, Python version, ...)

HviorForgeFlow:
## Summary

- `dict.get("move_dest_ids", [])` only falls back to `[]` when the key is **absent** from the dict. Odoo stores empty relational fields as `False`, so when the key is present with value `False` the default is bypassed and iterating over `False` raises `TypeError: 'bool' object is not iterable`.
- Fixed by switching to `(procurement.values.get("move_dest_ids") or [])`, which handles both the absent-key and `False`-value cases — matching the pattern already used by `purchase_stock` core for the same field.

Fixes #2985

### DIFF
--- a/purchase_request/models/stock_rule.py
+++ b/purchase_request/models/stock_rule.py
@@ -23,7 +23,7 @@ class StockRule(models.Model):
             "product_qty": procurement_uom_po_qty,
             "request_id": request_id.id,
             "move_dest_ids": [
-                (4, x.id) for x in procurement.values.get("move_dest_ids", [])
+                (4, x.id) for x in (procurement.values.get("move_dest_ids") or [])
             ],
             "orderpoint_id": procurement.values.get("orderpoint_id", False)
             and procurement.values.get("orderpoint_id").id,


BhaveshHeliconia (approved):
LGTM!

ACBRI (approved):
Verified the edge case: when `move_dest_ids` is present with a `False` value, the previous code raises `TypeError` while `or []` correctly handles it without changing the behavior for missing or populated values. This is consistent with the pattern used in Odoo core. LGTM.

HviorForgeFlow:
Thanks for the accurate testing @ACBRI

PR #3110, opened 2026-06-18
[14.0][ADD] purchase_auto_validation

Kev-Roche:
This module allows to isolate purchase orders for a configurable set
of products into a dedicated purchase order, kept separate from any other
purchase order generated for the same supplier, and confirms it
automatically based on a scheduled weekday and time.

Especially useful in an intercompany context.


purchase_auto_validation/models/purchase_order.py
```diff
@@ -0,0 +1,38 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+
+    auto_purchase_validation_id = fields.Many2one(
```
sebalix: To be consistent, not blocking
```suggestion
    purchase_auto_validation_id = fields.Many2one(
```

purchase_auto_validation/models/stock_rule.py
```diff
@@ -0,0 +1,46 @@
...
+                (
+                    "product_tmpl_ids",
+                    "in",
+                    [procurement.product_id.product_tmpl_id.id],
+                ),
+                "|",
+                ("company_id", "=", False),
+                ("company_id", "=", procurement.company_id.id),
+            ]
+            rule = rule_model.search(domain, limit=1)
```
sebalix, line 19: A product could be added in different rules currently, and here we take the first matching rule. Maybe a constraint should be added on the rule to avoid that, or a `sequence` field to let the user order them if it makes sense.

purchase_auto_validation/models/stock_rule.py
```diff
@@ -0,0 +1,46 @@
...
+            if rule:
+                procurement.values["auto_purchase_validation_id"] = rule.id
+        return super()._run_buy(procurements)
+
+    def _make_po_get_domain(self, company_id, values, partner):
+        domain = super()._make_po_get_domain(company_id, values, partner)
+        rule_id = values.get("auto_purchase_validation_id")
+        if rule_id:
+            rule = self.env["purchase.auto.validation"].sudo().browse(rule_id)
```
sebalix:
```suggestion
            rule = self.env["purchase.auto.validation"].sudo().browse(rule_id).exists()
```

purchase_auto_validation/models/purchase_auto_validation.py
```diff
@@ -0,0 +1,83 @@
...
+    _name = "purchase.auto.validation"
+    _description = "Purchase Auto Validation"
+
+    name = fields.Char(compute="_compute_name", store=True)
+
+    product_tmpl_ids = fields.Many2many(
+        comodel_name="product.template",
+        string="Products",
+        required=True,
+    )
```
sebalix, line 18: Why product templates?
In supplier info data, we can set different suppliers on variants, which could lead to inconsistencies.

purchase_auto_validation/tests/test_purchase_auto_validation.py
```diff
@@ -0,0 +1,264 @@
...
+        return self.env["procurement.group"].Procurement(
+            product,
+            qty,
+            product.uom_id,
+            self.env.ref("stock.stock_location_stock"),
+            product.name,
+            "TEST",
+            self.env.company,
+            values,
+        )
```
sebalix, line 52: In case we want to extend this module and write tests, putting the data init + helper methods in a common class is welcome

sebalix (commented):
LG, some remarks

Kev-Roche:
@sebalix Thank you for your review, I applied your remarks.

sebalix (approved):
LG!

HviorForgeFlow (approved):
code review 👍

PR #3113, opened 2026-06-22
[18.0][FIX] purchase_order_secondary_unit: avoid price recomputation from rounded secondary price

sergio-teruel:
**Before this PR**
When a purchase line uses a secondary unit with a factor that produces a rounded secondary price, the inverse of secondary_uom_price can recompute price_unit from the rounded value.
Example: supplier price 19.95 and secondary factor 0.9 gives 17.955, displayed/stored as 17.96. The inverse then computes 17.96 / 0.9 = 19.955555..., changing the original supplier price.

**After this PR**
Skip the inverse recomputation when secondary_uom_price matches the expected value from price_unit * factor within the Product Price decimal precision.
This preserves the supplier price while still allowing manual edits of the secondary unit price to update the base price.

cc @Tecnativa TT63349

ping @carlosdauden @CarlosRoca13 @juancarlosonate-tecnativa @sysadminmatmoz


carlosdauden (approved):
Perhaps you could avoid using `continue` if you check that it is not equal to 0

CarlosRoca13 (approved):

PR #3114, opened 2026-06-24
[19.0][ADD] purchase_order_line_price_unit_digits: new module

chienandalu:
From this change https://github.com/odoo/odoo/pull/243987 , `price_unit` in purchase order lines has a minimum display digits instead of a rounding in screen as it used to be. This is for compatibility issues with Peppol.

For the cases this isn't needed, the new behavior is quite anoying, as prices computed from the pricelists or currency conversions often throw quite a bunch of decimals (e.g.: 10.000023), which is ok for accuracy sake but is terrible for ux.

cc @moduon MT-15119

please review @Shide  @EmilioPascual


EmilioPascual (approved):
LGTM. Code Review

BhaveshHeliconia (approved):
LGTM!

PR #3120, opened 2026-07-07
[FIX] purchase_quick : base the computation of the seller_price field on the uom_po_id of the related product.

legalsylvain:
Rational : if the uom_po_id is dozen and the uom of the purchase order line is also dozen, the price should be the price of the supplierinfo, without any conversion


BhaveshHeliconia (approved):

petrus-v (approved):
Thanks

legalsylvain:
thanks for your review ! 

/ocabot merge patch

PR #3121, opened 2026-07-09
[MIG] purchase_last_price_info: Migration to 19.0

CarmenMiranda:
### Supersedes https://github.com/OCA/purchase-workflow/pull/2851

Migrating the module from Odoo 18.0 to Odoo 19.0, and the following changes were needed:

- `odoo.fields.first` was removed from core in [1].
- `product_uom` was renamed to `product_uom_id` in the `purchase.order.line` model in [2].
- The purchase unit field (`uom_po_id`) was removed from Odoo according to [3].
- Deprecate the usage of `invisible` in views according to [4].
- Deprecate the fields `show_last_purchase_price_currency`, and `last_purchase_price_currency` as it was expected in [5].
- Adapting the test to avoid relying on the demo data, since it is not installed by default.

Co-Authored-By: Thomas HERBIN <the@apik.cloud>

[1]: https://github.com/odoo/odoo/commit/871427d6c4a3c79228dc3bafc253a550afe5ee58
[2]: https://github.com/odoo/odoo/commit/c8461a5f382b8684a6483a4dbb7914056164e8ee
[3]: https://github.com/odoo/odoo/commit/91032eefdbfb2e48922b2c0a620af5c825871fca
[4]: https://github.com/odoo/odoo/pull/137031
[5]: https://github.com/OCA/purchase-workflow/commit/d5f8a3e39b9a8294a778701e84dc355fb0bbf148


CarmenMiranda:
Hi @luisg123v can you help me review this?

HviorForgeFlow (approved):

purchase_last_price_info/readme/CONTRIBUTORS.md
```diff
@@ -11,3 +11,5 @@
   - Pedro M. Baeza
 - [Bloopark systems](https://www.bloopark.de/):
   - Achraf Mhadhbi \<<machraf@bloopark.de>\>
+- [Apik](https://www.apik.cloud/):
+  - Thomas Herbin \<<the@apik.cloud>\>
```
luisg123v: There has been a lot of discussion lately regarding when people should be added as contributor, e.g. https://github.com/orgs/OCA/discussions/185. This because a lot of people submit PRs only to add themselves.

Since the original author didn't even finished its original work, I'd remove this, because it's probably not enough to qualify.

purchase_last_price_info/README.rst
```diff
@@ -21,13 +21,13 @@ Purchase Product Last Price Info
     :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
     :alt: License: AGPL-3
 .. |badge3| image:: https://img.shields.io/badge/github-OCA%2Fpurchase--workflow-lightgray.png?logo=github
-    :target: https://github.com/OCA/purchase-workflow/tree/18.0/purchase_last_price_info
+    :target: https://github.com/OCA/purchase-workflow/tree/19.0/purchase_last_price_info
```
luisg123v: Let's let the bot to do this post merge.

Same in the HTML index.

purchase_last_price_info/views/product_views.xml
```diff
@@ -0,0 +1,81 @@
...
+            <field name="seller_ids" position="before">
+                <group
+                    string="Purchase"
+                    name="last_purchase_info"
+                    invisible="product_variant_count &lt;= 1"
+                >
+                    <field name="last_purchase_supplier_id" />
+                    <field name="last_purchase_date" />
+                    <field name="last_purchase_price" />
+                    <field name="show_last_purchase_price_currency" invisible="1" />
```
luisg123v: It's no longer required to add invisible fields for making them available in other field's attributes.

Same in similar cases.

luisg123v (changes requested):

purchase_last_price_info/readme/CONTRIBUTORS.md
CarmenMiranda: Removed

purchase_last_price_info/README.rst
CarmenMiranda: Ok, got it, removed

purchase_last_price_info/views/product_views.xml
CarmenMiranda: Done

BhaveshHeliconia (approved):
LGTM!

purchase_last_price_info/views/product_views.xml
CarmenMiranda: Hi @luisg123v, can you review it again? I included the latest changes that were merged in v18

HviorForgeFlow:
@CarmenMiranda is it possible to align with the following change in v18.0 https://github.com/OCA/purchase-workflow/pull/2964

I think it would be nice to change the naming and add a migration script. @JordiMForgeFlow could you give some advice here?

CarmenMiranda:
Hi @HviorForgeFlow, the deprecation of the fields in #2964 was already included in the migration commit

And as mentioned here https://github.com/OCA/purchase-workflow/pull/2964#issuecomment-4896250424, those fields are both non-stored, so no migration script is needed

Cc. @JordiMForgeFlow @luisg123v

luisg123v (changes requested):
Code LGTM, just a couple of point regarding the migration commit:
- In the end, you ended up doing most of the migration changes. I consider then you should become the main author and leave the original PR's author as coauthor. What do you think?
- Before the list, add a text to introduce it, something like "The following changes are implemented" or similar.

luisg123v (approved):
LGTM 👍

PR #3129, opened 2026-07-17
[18.0][IMP] purchase_order_line_menu: add analytic_distribution on line

Saran440:
Add analytic_distribution on line

### DIFF
--- a/purchase_order_line_menu/views/purchase_order_line_views.xml
+++ b/purchase_order_line_menu/views/purchase_order_line_views.xml
@@ -26,6 +26,12 @@
             </xpath>
             <xpath expr="//field[@name='product_id']" position="after">
                 <field name="product_type" optional="hide" />
+                <field
+                    name="analytic_distribution"
+                    widget="analytic_distribution"
+                    optional="hide"
+                    groups="analytic.group_analytic_accounting"
+                />
             </xpath>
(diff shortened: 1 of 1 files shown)


BhaveshHeliconia (approved):
LGTM!

acbripenta (approved):
Posted from my work account by mistake — same reviewer as @ACBRI. Please count these as a single review; the tested evidence is in the review from @ACBRI.

ACBRI (approved):
Tested on a clean 18.0 database with `purchase_order_line_menu` installed.

- The column reaches the list only for users in `analytic.group_analytic_accounting`, hidden by default (`optional="hide"`).
- `web_search_read` with the menu action's domain returns the distribution correctly on a confirmed PO line. `analytic_precision` is fetched automatically through the widget's `fieldDependencies`, so no companion field is needed.
- The list is readonly (`create="false"`), so only the display path of the widget is exercised, which skips the analytic plans fetch in list views.

Small optional nit: core passes `options="{'product_field': 'product_id', 'business_domain': 'purchase_order'}"` wherever it places this field on `purchase.order.line`. Inert in a readonly list, but it would keep this aligned with core if the list ever becomes editable. Not blocking.

LGTM.

HviorForgeFlow (approved):

PR #3139, opened 2026-07-30
[18.0][IMP] purchase_stock_price_unit_sync: Sync the price corrected on the vendor bill

carlosdauden:
This module already pushes a price corrected on the purchase order down to the moves that are already done and to their valuation layers. The vendor bill, which is the other place where the real price of a purchase shows up, was left to Odoo, and Odoo only corrects half of it.

Odoo books a price difference as a child valuation layer worth the difference times the quantity that has not left stock yet, and sends the rest to the expense account. The stock ends up valued right, but the moves that already left keep the cost that turned out to be wrong, so the margin of what was sold, and any stock valuation asked for a date before the bill, stay wrong too. Correcting the same price on the purchase order did fix all of it, so the two ways of doing the very same thing gave different results.

The invoiced price is now applied to the whole layer, which leaves the bill on the same footing as the purchase order. The hook is `_prepare_pdiff_vals`, where Odoo has already worked out, with its own matching of layers and bills, which layer each invoice line pays for and at what price, so none of that logic is duplicated here.

It needs `product_cost_price_avco_sync` installed, checked at runtime rather than declared as a dependency: it is that module which replays the valuation chain once the layer changes, and restating a layer without it would leave the layer inconsistent while also throwing away the correction Odoo does make, which is worse than not doing anything. Refunds are left to Odoo, which compensates them against the original bill with a logic of its own, and so is automated valuation, where the journal entry of the layer is already posted and restating it would pull the two apart.

@Tecnativa TT63558

ping @sergio-teruel @carlos-lopez-tecnativa @CarlosRoca13


CarlosRoca13 (approved):

purchase_stock_price_unit_sync/models/account_move_line.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+class AccountMoveLine(models.Model):
+    _inherit = "account.move.line"
+
+    def _is_price_unit_sync_restated(self, layer):
+        """Whether the price difference of this invoice line has to be applied
+        by restating the layer instead of by correcting only what is left of it.
+
+        It needs `product_cost_price_avco_sync`, which is what replays the chain
```
carlos-lopez-tecnativa, line 15: Based on your docstring, should this module have an explicit dependency on `product_cost_price_avco_sync` in order to work correctly?

PR #3141, opened 2026-08-05
[FIX] purchase_deposit: convert deposit amounts to the order currency

mav-adhoc:
## What

`action_post` copies the deposit amount from the vendor bill back into the purchase
order line. When the bill is issued in a currency other than the order one the figure
was copied **verbatim**, so an order in a strong currency ended up with a deposit line
holding the nominal of a weak one — three orders of magnitude off, in the case that
brought us here.

This converts it to the order currency, at the bill date.

## Why it matters beyond the wrong figure

The deposit line's `price_unit` is not only informative: it is the amount deducted from
the final bill (`_prepare_account_move_line` puts the deposit line in with
`quantity = -qty_invoiced`), and on 18.0 it is also what the wizard compares against the
order total. So a bill figure landing there unconverted both corrupts the deduction and,
on 18.0, makes any further deposit on that order impossible.

Deposits billed in the order currency are unaffected: the conversion rate is exactly 1.

## Test plan

`test_deposit_billed_in_another_currency`: an order in a currency worth 1000x the
company one, a deposit of 300, and a bill edited to the company currency for 300,000.
Without the fix the order line ends up at `300000.0`; with it, at `300.0`.

Full suite green (7 tests on 18.0, 6 on 19.0).

### DIFF
--- a/purchase_deposit/models/account_move.py
+++ b/purchase_deposit/models/account_move.py
@@ -1,7 +1,7 @@
 # Copyright 2023 Quartile Limited (https://www.quartile.co)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
-from odoo import models
+from odoo import fields, models
 
 
 class AccountMove(models.Model):
@@ -12,6 +12,14 @@ class AccountMove(models.Model):
         for line in self.line_ids:
             if not line.purchase_line_id.is_deposit:
                 continue
+            order = line.purchase_line_id.order_id
             line.purchase_line_id.taxes_id = line.tax_ids
-            line.purchase_line_id.price_unit = line.price_unit
+            # The bill may be issued in a currency other than the order one.
+            line.purchase_line_id.price_unit = line.currency_id._convert(
+                line.price_unit,
+                order.currency_id,
+                order.company_id,
+                line.move_id.invoice_date or fields.Date.context_today(self),
+                round=False,
+            )
         return res
--- a/purchase_deposit/tests/__init__.py
+++ b/purchase_deposit/tests/__init__.py
@@ -3,3 +3,4 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
 from . import test_purchase_deposit
+from . import test_purchase_deposit_multicurrency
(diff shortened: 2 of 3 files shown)


BhaveshHeliconia (approved):

mav-adhoc:

les-adhoc (approved):
LGTM

mav-adhoc:
Applied the review suggestions in both branches: the test case now inherits from `BaseCommon` and builds its data with `Command.create`. Thanks @BhaveshHeliconia.

Heads-up, since this was already approved: the fix itself changed. It used to convert the deposit at the bill date, it now converts at the order rate (`purchase.order.currency_rate`), which is the basis the order total already uses. The reasoning is in the updated description — converting at the bill date makes the amount left to bill on an order float with the exchange rate, and a fully advanced order stops adding up. The test was extended so it discriminates between the two. Worth a fresh look rather than carrying the previous approval over.

mav-adhoc:
Sorry for the churn: I briefly pushed a version that converted the deposit at the order rate instead of the bill date, and I have reverted it. It looked like the better basis — it is the one `purchase.order.currency_rate`/`amount_total_cc` use, and it makes an order advanced in fixed instalments of the company currency add up exactly. But measuring an old order at its own rate is only right while that rate is still meaningful: on a long-lived order in a foreign currency, paid over a year or more in a high-inflation company currency, the order rate is stale and it inflates every later deposit. I checked it against real orders and it would have newly blocked several of them, some by a wide margin. There is no single conversion basis that is right for every order, so this PR stays on the bill date and the deposit stays worth what was actually paid.

The only difference from the version you approved is your two test suggestions, now applied in both branches. Thanks again @BhaveshHeliconia.

jcadhoc (approved):

BhaveshHeliconia (approved):
LGTM!

mav-adhoc:
Dropped the manifest version bump here too, per @HviorForgeFlow's review on #3142 — ocabot does it on merge, so no `nobump` needed on any of the three.

mav-adhoc:
For whoever merges: the red `runboat/build` here is not this PR. Every build of the 18.0 branch is failing at init, including the branch build itself (`11184a7d7`), while 19.0 deploys fine; the failures cluster on the heaviest branches across OCA repos and the platform reports its deployment cap reached. Reported upstream in sbidoul/runboat#158 with the details. Everything else — `pre-commit`, `test with Odoo`, `test with OCB`, codecov — is green.

PR #3142, opened 2026-08-06
[FIX] purchase_deposit: convert deposit amounts to the order currency

mav-adhoc:
## What

`action_post` copies the deposit amount from the vendor bill back into the purchase
order line. When the bill is issued in a currency other than the order one the figure
was copied **verbatim**, so an order in a strong currency ended up with a deposit line
holding the nominal of a weak one — three orders of magnitude off, in the case that
brought us here.

This converts it to the order currency, at the bill date.

## Why it matters beyond the wrong figure

The deposit line's `price_unit` is not only informative: it is the amount deducted from
the final bill (`_prepare_account_move_line` puts the deposit line in with
`quantity = -qty_invoiced`), and on 18.0 it is also what the wizard compares against the
order total. So a bill figure landing there unconverted both corrupts the deduction and,
on 18.0, makes any further deposit on that order impossible.

Deposits billed in the order currency are unaffected: the conversion rate is exactly 1.

## Test plan

`test_deposit_billed_in_another_currency`: an order in a currency worth 1000x the
company one, a deposit of 300, and a bill edited to the company currency for 300,000.
Without the fix the order line ends up at `300000.0`; with it, at `300.0`.

Full suite green (7 tests on 18.0, 6 on 19.0).

### DIFF
--- a/purchase_deposit/models/account_move.py
+++ b/purchase_deposit/models/account_move.py
@@ -1,7 +1,7 @@
 # Copyright 2023 Quartile Limited (https://www.quartile.co)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
-from odoo import models
+from odoo import fields, models
 
 
 class AccountMove(models.Model):
@@ -12,6 +12,14 @@ class AccountMove(models.Model):
         for line in self.line_ids:
             if not line.purchase_line_id.is_deposit:
                 continue
+            order = line.purchase_line_id.order_id
             line.purchase_line_id.tax_ids = line.tax_ids
-            line.purchase_line_id.price_unit = line.price_unit
+            # The bill may be issued in a currency other than the order one.
+            line.purchase_line_id.price_unit = line.currency_id._convert(
+                line.price_unit,
+                order.currency_id,
+                order.company_id,
+                line.move_id.invoice_date or fields.Date.context_today(self),
+                round=False,
+            )
         return res
--- a/purchase_deposit/tests/test_purchase_deposit.py
+++ b/purchase_deposit/tests/test_purchase_deposit.py
@@ -2,7 +2,7 @@
 # Copyright 2019 Ecosoft Co., Ltd., Kitti U. <kittiu@ecosoft.co.th>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
-from odoo import fields
+from odoo import Command, fields
 from odoo.exceptions import UserError
 from odoo.tests import Form, TransactionCase
 
@@ -205,3 +205,36 @@ class TestPurchaseDeposit(TransactionCase):
         deposit_line = self.po.order_line.filtered(lambda p: p.is_deposit)
         self.assertEqual(deposit_line.price_unit, 500.0)
         self.assertEqual(deposit_line.tax_ids.id, self.tax.id)
+
+    def test_deposit_billed_in_another_currency(self):
+        """Posting a bill must not leak its currency figure into the order."""
+        # An order currency worth 1000 times the company one
+        self.po.currency_id = self.env["res.currency"].create(
+            {
+                "name": "PDX",
+                "symbol": "PDX",
+                "rate_ids": [
+                    Command.create(
+                        {
+                            "name": "2020-01-01",
+                            "rate": 0.001,
+                            "company_id": self.env.company.id,
+                        },
+                    )
+                ],
+            }
+        )
+        f = self.create_advance_payment_form()
+        f.advance_payment_method = "fixed"
+        f.amount = 300.0
+        f.deposit_account_id = self.account_deposit
+        f.save().create_invoices()
+        deposit_line = self.po.order_line.filtered("is_deposit")
+        self.assertEqual(deposit_line.price_unit, 300.0)
+        # The vendor bills the deposit in the company currency instead
+        bill = self.po.invoice_ids
+        bill.invoice_date = fields.Date.today()
+        bill.currency_id = self.env.company.currency_id
+        bill.invoice_line_ids.price_unit = 300000.0
+        bill.action_post()
+        self.assertEqual(deposit_line.price_unit, 300.0)


purchase_deposit/tests/test_purchase_deposit_multicurrency.py
```diff
@@ -0,0 +1,91 @@
+# Copyright 2026 ADHOC SA (https://www.adhoc.com.ar)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields
+from odoo.tests import TransactionCase
+
+
+class TestPurchaseDepositMultiCurrency(TransactionCase):
```
BhaveshHeliconia, line 8:
```suggestion
class TestPurchaseDepositMultiCurrency(Basecommon):
```

purchase_deposit/tests/test_purchase_deposit_multicurrency.py
```diff
@@ -0,0 +1,91 @@
...
+        super().setUpClass()
+        # Order currency worth 1000 times the company one
+        cls.order_currency = cls.env["res.currency"].create(
+            {
+                "name": "PDX",
+                "symbol": "PDX",
+                "rate_ids": [
+                    (
+                        0,
+                        0,
```
BhaveshHeliconia, line 20:
```suggestion
                    Command.create(
```

purchase_deposit/tests/test_purchase_deposit_multicurrency.py
```diff
@@ -0,0 +1,91 @@
...
+        cls.deposit_product = cls._create_service("Purchase Deposit")
+        product = cls._create_service("Test Product")
+        cls.po = cls.env["purchase.order"].create(
+            {
+                "partner_id": cls.env["res.partner"].create({"name": "Vendor"}).id,
+                "currency_id": cls.order_currency.id,
+                "order_line": [
+                    (
+                        0,
+                        0,
```
BhaveshHeliconia, line 42:
```suggestion
                    Command.create(
```

BhaveshHeliconia (approved):
LGTM!

Just minor comments

mav-adhoc:

mav-adhoc:
Applied the review suggestions in both branches: the test case now inherits from `BaseCommon` and builds its data with `Command.create`. Thanks @BhaveshHeliconia.

Heads-up, since this was already approved: the fix itself changed. It used to convert the deposit at the bill date, it now converts at the order rate (`purchase.order.currency_rate`), which is the basis the order total already uses. The reasoning is in the updated description — converting at the bill date makes the amount left to bill on an order float with the exchange rate, and a fully advanced order stops adding up. The test was extended so it discriminates between the two. Worth a fresh look rather than carrying the previous approval over.

mav-adhoc:
Sorry for the churn: I briefly pushed a version that converted the deposit at the order rate instead of the bill date, and I have reverted it. It looked like the better basis — it is the one `purchase.order.currency_rate`/`amount_total_cc` use, and it makes an order advanced in fixed instalments of the company currency add up exactly. But measuring an old order at its own rate is only right while that rate is still meaningful: on a long-lived order in a foreign currency, paid over a year or more in a high-inflation company currency, the order rate is stale and it inflates every later deposit. I checked it against real orders and it would have newly blocked several of them, some by a wide margin. There is no single conversion basis that is right for every order, so this PR stays on the bill date and the deposit stays worth what was actually paid.

The only difference from the version you approved is your two test suggestions, now applied in both branches. Thanks again @BhaveshHeliconia.

jcadhoc (approved):

BhaveshHeliconia (approved):
LGTM!

les-adhoc (approved):
LGTM!

purchase_deposit/__manifest__.py
```diff
@@ -4,7 +4,7 @@
 
 {
     "name": "Purchase Deposit",
-    "version": "19.0.1.0.0",
+    "version": "19.0.1.0.1",
```
HviorForgeFlow: don't change version, ocabot will do

HviorForgeFlow (changes requested):
Code looks good, thanks for the contribution.

Just a minor alignment

mav-adhoc:
Thanks @HviorForgeFlow. Version bump dropped, the manifest is untouched now — same in the 18.0 counterpart (#3141) and in #3172, so all three can be merged with the plain `/ocabot merge patch`.

About the red `test with Odoo` / `test with OCB`: it is not this PR. `purchase_deposit` runs clean (`purchase_deposit: 10 tests`, no failure); the only failing test is `purchase_order_line_description`'s tour, which is red on every open PR of the 19.0 branch right now — #3152 hits the exact same step.

I tracked it down and it is a core change: since odoo/odoo#285946 the read-only label of `product_label_section_and_note_field` is only rendered as cell text when the label is actually read-only, which on an open order it is not, so the description now lives in a textarea value and `:contains()` no longer matches it. Fix in **#3174** (test-only, two selectors), verified against core `5587c47a865` and current `e809d194b88`. Once that one lands the 19.0 branch goes green again and I will rebase this PR on top.

purchase_deposit/tests/test_purchase_deposit_multicurrency.py
```diff
@@ -0,0 +1,88 @@
...
+        return cls.env["product.product"].create(
+            {
+                "name": name,
+                "type": "service",
+                "purchase_method": "purchase",
+                "property_account_expense_id": cls.account_expense.id,
+            }
+        )
+
+    def test_deposit_billed_in_another_currency(self):
```
Saran440: Could you add this test case to `test_purchase_deposit.py` instead of creating a separate test file?

Also, if AI was used for this contribution, please add the `Assisted-by:` trailer according to the OCA AI policy.

https://github.com/OCA/.github/blob/master/AI_POLICY.md

HviorForgeFlow: Attend comments @mav-adhoc, moving test is enough for now

mav-adhoc:
@Saran440 @HviorForgeFlow Done! Thnaks to both of you!

HviorForgeFlow (approved):

PR #3143, opened 2026-08-07
[FIX] sale_purchase_force_vendor: do not create supplierinfo on non-p…

LuisAlejandroS:
…ositive lines

A sale order line with a non-positive quantity (a return line, for instance) is not a purchase, so it must not register a vendor for the product.

`_prepare_procurement_values` looked the seller up with the raw line quantity. `_select_seller` discards every seller whose `min_qty` is greater than the requested quantity, and every `min_qty` is >= 0, so a negative quantity never matches any seller -- not even one whose `min_qty` is 0. The lookup therefore always came back empty and the fallback branch autocreated a `product.supplierinfo` with no price, no product name and no product code, once per line and per return, even when the correct vendor price was already there.

Two changes:

- The supplierinfo is only created for a line that will actually trigger a purchase. This is what fixes the bug, on its own: it also covers a product with no vendor price at all, where the lookup has nothing to find.
- The seller lookup degrades a non-positive quantity to 0 rather than making up a purchase quantity for a line that is not a purchase. A seller whose `min_qty` is above 0 will not match on such a line, which is harmless: the line triggers no purchase and never consumes `supplierinfo_id`.

`res["supplierinfo_id"]` is still assigned unconditionally (an empty recordset when nothing matched) because `stock.move._prepare_procurement_values` reads that key with direct access and would raise `KeyError` otherwise.

Adds tests for both signs, for a product with and without a vendor price, and for both sides of the `min_qty` filter.

### DIFF
--- a/sale_purchase_force_vendor/models/sale_order_line.py
+++ b/sale_purchase_force_vendor/models/sale_order_line.py
@@ -55,10 +55,24 @@ class SaleOrderLine(models.Model):
             product = self.product_id
             suppinfo = product.with_company(self.company_id.id)._select_seller(
                 partner_id=self.vendor_id,
-                quantity=self.product_uom_qty,
+                # `_select_seller` discards every seller whose `min_qty` is
+                # greater than the requested quantity, and every `min_qty` is
+                # >= 0, so a negative quantity matches no seller at all. A
+                # non-positive line is not a purchase, so degrade it to 0
+                # rather than making up a purchase quantity for it. A seller
+                # with a `min_qty` above 0 will not match either, which is
+                # harmless: such a line creates no purchase and never consumes
+                # `supplierinfo_id`.
+                quantity=max(self.product_uom_qty, 0),
                 uom_id=self.product_uom,
             )
-            if not suppinfo:
+            # Only a line that will actually trigger a purchase may register a
+            # new vendor. Otherwise every return line would autocreate a
+            # supplierinfo with no price, no product name and no product code.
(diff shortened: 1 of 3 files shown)


amarcosg (approved):
Code review LGTM!

sale_purchase_force_vendor/models/sale_order_line.py
```diff
@@ -55,10 +55,24 @@ def _prepare_procurement_values(self, group_id=False):
...
-                quantity=self.product_uom_qty,
+                # `_select_seller` discards every seller whose `min_qty` is
+                # greater than the requested quantity, and every `min_qty` is
+                # >= 0, so a negative quantity matches no seller at all. A
+                # non-positive line is not a purchase, so degrade it to 0
+                # rather than making up a purchase quantity for it. A seller
+                # with a `min_qty` above 0 will not match either, which is
+                # harmless: such a line creates no purchase and never consumes
+                # `supplierinfo_id`.
+                quantity=max(self.product_uom_qty, 0),
```
victoralmau, line 66:
```suggestion
                quantity=self.product_uom_qty,
```
In my opinion, this should remain `self.product_uom_qty`.

I think you're mixing things up; it's one thing that if the quantity is negative, it retrieves a `suppinfo` (it won't find one), and another thing is what happens next with the condition `self.product_uom_qty > 0`, which only creates a `product.supplierinfo` under that condition.

victoralmau (changes requested):

PR #2968, opened 2026-02-26
[19.0][MIG]: purchase_order_secondary_unit

JochenDeBie:
Depends on:

- https://github.com/OCA/product-attribute/pull/2222

This is only upgrading the code in the 18.0 branch as-is + cleanup of https://github.com/OCA/purchase-workflow/pull/2967

Perhaps it's interesting to also look what happens on:

- https://github.com/OCA/product-attribute/pull/2211
- https://github.com/OCA/purchase-workflow/pull/2950

PR #3155, opened 2026-08-20
[19.0][MIG] purchase_order_secondary_unit: Migration to version 19.0

carlos-lopez-tecnativa:
Supersedes https://github.com/OCA/purchase-workflow/pull/2968

TT64122
@Tecnativa @pedrobaeza @carlosdauden @pilarvargas-tecnativa @CarlosRoca13 @BhaveshHeliconia could you please review this?


BhaveshHeliconia (approved):
Code and functional review LGTM!

pilarvargas-tecnativa (approved):

purchase_order_secondary_unit/models/purchase_order.py
```diff
@@ -29,12 +29,6 @@ class PurchaseOrderLine(models.Model):
         copy=True,
         precompute=True,
```
pedrobaeza, line 30: I think this doesn't need to be pre-computed now that core is not computed.

carlos-lopez-tecnativa: If I remove precompute, a warning appears because the mixin defines the secondary_uom_qty field as precompute=True, and due to the dependency, this field also needs to be precomputed:
https://github.com/OCA/product-attribute/blob/74b3b87352e02b3cc42ec45bc80b66651e058db9/product_secondary_unit/models/product_secondary_unit_mixin.py#L40

`WARNING devel py.warnings: /opt/odoo/custom/src/odoo/odoo/orm/fields.py:837: UserWarning: Field purchase.order.line.secondary_uom_qty cannot be precomputed as it depends on non-precomputed field purchase.order.line.product_qty`

pedrobaeza (approved):
/ocabot merge nobump

PR #3156, opened 2026-08-24
[FIX] purchase_location_by_line: avoid location error on line logic

Harold9904:
### Summary
Fixes purchase line location computation to avoid the error in stock picking flow.

### Version
- Target branch: 17.0
### Notes
- Minimal fix focused on `purchase_location_by_line` behavior.
- - No functional scope outside this module.


BhaveshHeliconia (commented):
Thanks for the fix!

squash the 3 commits into a single commit

PR #3161, opened 2026-09-01
[16.0][ADD] purchase_deposit_company_amount

AungKoKoLin1997:
This module books a foreign-currency purchase deposit at the company-currency amount actually paid and carries that figure through to the final bill, so the deposit account clears to zero and the exchange-rate difference lands in the inventory cost of the goods instead of being stranded on the balance sheet.

In detail:

-  Adds a Company Currency Amount field to the deposit line of a deposit
   bill, which forces the line's balance to the amount entered instead
   of converting the foreign currency amount at the exchange rate.
-  Carries the company currency value of a posted deposit bill over to
   the deposit offset line of the final bill, so the deposit account
   closes out at the amount that was actually paid.
-  Books the resulting exchange rate difference into the product lines,
   and reflects it in the stock valuation of the received goods.

The field is only available on the deposit line of a deposit bill raised in a foreign currency, and the column is hidden everywhere else. Everything on the final bill is derived from it, and a deposit nobody entered an amount for keeps the standard rate conversion throughout.

@qrtl QT6131


purchase_deposit_company_amount/readme/USAGE.md
yostashiro, line 1: Can we illustrate differences in journal entries with and without this module at relevant steps by placing them side-by-side?

purchase_deposit_company_amount/models/account_move.py
```diff
@@ -0,0 +1,156 @@
...
+            targets[line] = sign * abs(deposit_amount)
+        delta = sum(
+            line._get_rate_based_balance() - targets[line]
+            for line in offset_lines
+            if line in targets
+        )
+        absorbing_lines = product_lines.filtered(
+            lambda l: l.purchase_line_id and not l.purchase_line_id.is_deposit
+        )
+        targets.update(self._get_absorbed_targets(absorbing_lines, delta))
```
yostashiro, line 109:
```suggestion
        # The goods absorb delta: it is acquisition cost, not an FX gain or loss.
        targets.update(self._get_absorbed_targets(absorbing_lines, delta))
```

purchase_deposit_company_amount/models/account_move.py
```diff
@@ -0,0 +1,156 @@
...
+        self.ensure_one()
+        return self.line_ids.filtered(
+            lambda l: l.display_type == "product"
+            and l.purchase_line_id.is_deposit
+            and l.quantity < 0
+        )
+
+    def _get_absorbed_targets(self, absorbing_lines, delta):
+        self.ensure_one()
+        company_currency = self.company_id.currency_id
```
yostashiro: Do the same for other occurrences.
```suggestion
        company_currency = self.company_currency_id
```

purchase_deposit_company_amount/models/account_move.py
AungKoKoLin1997: Updated!

purchase_deposit_company_amount/models/account_move.py
AungKoKoLin1997: Updated!

purchase_deposit_company_amount/readme/USAGE.md
AungKoKoLin1997: Updated!

yostashiro (approved):
Code review and functional test. LGTM. 👍

nobuQuartile (approved):
LGTM: Code review and functional review

kanda999 (changes requested):
@AungKoKoLin1997 

purchase_deposit relieves the entire deposit on the first bill that follows (quantity = -1 * qty_invoiced), but _get_company_amount_targets allocates the resulting delta only across the goods lines on that bill.

On an instalment order the first partial bill therefore absorbs a deposit that belongs to the whole order.

Example — JPY company, PO 100 units × 10 USD, 30% deposit (300 USD) settled for 39,000 JPY, rate 200 at billing. First bill covers 40 units: delta = −21,000 lands entirely on the goods line, giving 80,000 − 21,000 = 59,000.

first 40 units → 1,475 JPY/unit
remaining 60 units (no offset line left) → 2,000 JPY/unit
correct blended cost → 1,790 JPY/unit


The total is right (179,000) and the move balances, so no error is raised, but each batch enters its own valuation layer at the skewed cost via _get_gross_unit_price, so FIFO/AVCO and COGS are wrong per layer. Sell the first 40 units before the second bill posts and the error is realised in P&L permanently.

Note the tests all use product_qty = 1.0 with full receipt and a single bill, so this path isn't covered.

kanda999:
@AungKoKoLin1997 

Prorating the deposit FX difference across instalments would require carry-forward logic, so I think the module should only support the case it can handle safely.

Apply the pin only when:

- the bill fully invoices every non-deposit PO line (qty_invoiced >= product_qty), and
- the FX delta can be fully absorbed by the goods lines without making balance opposite in sign to amount_currency.

If either condition is not met, I would raise a UserError.

This guard covers the current problems in one place: distorted unit costs on partial bills, the possible sign-flip IntegrityError, and cases where the delta leaks to the payable line because there are no valid absorbing lines.

I would also document this limitation in the README and make the UserError clearly explain that deposit FX pinning is only supported when the whole order is invoiced at once.

AungKoKoLin1997:
@kanda999 I did some adjustments to the PR. Please review again.

PR #3162, opened 2026-09-01
[18.0][FIX] purchase_order_secondary_unit: Purchase secondary uom by default non store in variant

CarlosRoca13:
cc @Tecnativa TT64341

FW-PORT of https://github.com/OCA/purchase-workflow/pull/2875

ping @sergio-teruel @carlosdauden

### DIFF
--- a/purchase_order_secondary_unit/__manifest__.py
+++ b/purchase_order_secondary_unit/__manifest__.py
@@ -3,7 +3,7 @@
 {
     "name": "Purchase Order Secondary Unit",
     "summary": "Purchase product in a secondary unit",
-    "version": "18.0.1.2.3",
+    "version": "18.0.2.0.0",
     "development_status": "Beta",
     "category": "Purchase",
     "website": "https://github.com/OCA/purchase-workflow",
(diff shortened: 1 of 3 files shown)


sergio-teruel:
@CarlosRoca13 Review the PR https://github.com/OCA/purchase-workflow/pull/2875
I have attended to @pedrobaeza comments...

AungKoKoLin1997:
@CarlosRoca13 @sergio-teruel 
We are adding generic template/variant sync helpers to `product_secondary_unit` in https://github.com/OCA/product-attribute/pull/2166, aligned with exactly the logic of this PR. With those helpers, `sale_order_secondary_unit`, `purchase_order_secondary_unit` and `stock_secondary_unit` can share the same implementation instead of keeping duplicate logic in each module (the `create()` override would not be needed either, as it is handled generically there).

Could you please review that PR? If you agree with the approach, would you consider applying the helpers in this PR, so that the logic is not duplicated in the first place? It would mean waiting for that PR to be merged, so if you prefer to get this fix in sooner, the refactoring can be done later in a follow-up PR, together with `sale_order_secondary_unit`.

PR #3167, opened 2026-09-01
[15.0][FIX] purchase_request: do not close allocations on locked orders

les-adhoc:
Fix forwarded from 18.0 to 15.0

Original pull request (18.0): https://github.com/OCA/purchase-workflow/pull/3160

`done` on a purchase order means "Locked", a modification lock, not a
completion, so a locked order must not close the allocation. See the original
PR for the full analysis and the regression history (#1419, #1422).

The test is adapted to 15.0: `qty_done` on the move line, and `mail.mt_comment`
as the subtype, since the dedicated `mt_request_picking_done` subtype only
exists from 16.0 on.

### DIFF
--- a/purchase_request/models/purchase_request_allocation.py
+++ b/purchase_request/models/purchase_request_allocation.py
@@ -82,7 +82,7 @@ class PurchaseRequestAllocation(models.Model):
     )
     def _compute_open_product_qty(self):
         for rec in self:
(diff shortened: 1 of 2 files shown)


BhaveshHeliconia (approved):
LGTM!

HviorForgeFlow (approved):

PR #3168, opened 2026-09-01
[18.0][MIG] purchase_tier_validation_forward: Migration to 18.0

rjaraspearhead:
Migration of `purchase_tier_validation_forward` to 18.0.

Changes:

- Preserve the module history from 13.0 using `format-patch`.
- Update the module version to `18.0.1.0.0`.
- Apply the repository pre-commit formatting.
- Remove the obsolete purchase order form inheritance. In 18.0, `purchase_tier_validation` injects the tier validation controls dynamically, while `base_tier_validation_forward` extends the corresponding template with the **Forward** button.
- Remove the obsolete translation files, as their only term referenced the removed view and XML ID.
- Keep the module as an auto-install bridge between `purchase_tier_validation` and `base_tier_validation_forward`.
- Add integration tests covering:
  - Dynamic insertion of the **Forward** button.
  - Forwarding a purchase order review to another reviewer.
  - The original review changing to `forwarded`.
  - The delegated review changing to `pending`.


rjaraspearhead:
Hi @kittiu, @HviorForgeFlow, @celm1990 

Could you please help review this migration?

HviorForgeFlow:
I think we can skip the migration as now the forward module don't requires a glue module

celm1990:
> I think we can skip the migration as now the forward module don't requires a glue module

I see this was improved in https://github.com/OCA/server-ux/pull/245. So, in OpenUpgrade V14, this module should be marked as merged into base_tier_validation_forward.

https://github.com/OCA/OpenUpgrade/blob/495a099aebf6c4acc0e993037d73732c39f23c62/openupgrade_scripts/apriori.py#L84

rjaraspearhead:
> Thanks for the clarification. We have prepared OCA/OpenUpgrade#5986 to register `purchase_tier_validation_forward` as merged into `base_tier_validation_forward`.
>
> This covers the migration path from 13.0 to 14.0. Since this glue module is no longer required in later versions, I will proceed with closing this 18.0 migration PR.

PR #3169, opened 2026-09-02
[16.0][IMP] purchase_order_line_deep_sort: create records in batch

baguenth:
create() on both purchase.order and purchase.order.line was decorated with @api.model, which makes the ORM split every batched create into one call per record. Converted both to @api.model_create_multi and sort each created order once, so batched imports/EDI/order generation keep a single create() call instead of N.

### DIFF
--- a/purchase_order_line_deep_sort/models/purchase_order.py
+++ b/purchase_order_line_deep_sort/models/purchase_order.py
@@ -63,18 +63,20 @@ class PurchaseOrder(models.Model):
             self._sort_purchase_line()
         return res
 
-    @api.model
-    def create(self, values):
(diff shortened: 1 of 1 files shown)


BhaveshHeliconia (approved):
Thanks for this!

pedrobaeza (changes requested):
Please don't add extra tests (probably generated by AI) that are not adding value, but more computing time each time the CI is triggered.

baguenth:
> Please don't add extra tests (probably generated by AI) that are not adding value, but more computing time each time the CI is triggered.

Updated :)

pedrobaeza (approved):
/ocabot merge patch

PR #3173, opened 2026-09-09
[19.0][FIX] purchase_order_line_description: repair tour selectors

HviorForgeFlow:
https://github.com/odoo/odoo/commit/21ee46537433c3fe398789af05b6304e82c78ca5
narrowed the readonly label branch to locked or cancelled orders, so a draft
order line now renders its description in a textarea instead of a text div.
The tour matched on text content, which a textarea value never exposes, so
`test_manual_description_edit_keeps_product_name_hidden_tour` has failed on
every 19.0 build since 2026-09-03, whatever the PR under test contains.


HviorForgeFlow:
closing in favor of https://github.com/OCA/purchase-workflow/pull/3174

PR #3178, opened 2026-09-17
[16.0][IMP] purchase_order_type: default order type on create, commercial partner fallback, no compute

baguenth:
Default order type on create, commercial partner fallback, no compute

## Problem

Three gaps, all the same shape: the form has a fallback that create()/write() lacks, because
order_type is a compute=/store= field driven by partner_id, while the form layer carries its own,
separate onchange logic that does more than the compute does.

1. **No fallback order type on create.** order_type only ever came from `partner_id.purchase_type`,
   or stayed empty - there's no company-level default for a vendor that doesn't provide one, so an
   order created in the background (MTO, procurement, an import, any script) with such a vendor
   always ends up without a type.
2. **The commercial-partner fallback is form-only.** `onchange_partner_id()` falls back to
   `partner_id.commercial_partner_id.purchase_type` when the specific contact has none; the compute
   does not. A contact with no `purchase_type` but whose parent company has one gets that type in
   the form, and nothing via `create()`.
3. **`payment_term_id`/`incoterm_id` never follow `order_type` on create either.**
   `onchange_order_type()` is the only place that copies them onto the order, and onchange methods
   only run from the form. This becomes more consequential once (1) is fixed, since `order_type`
   will now reliably get populated on `create()` far more often than before.

## What this PR does

- Adds `is_default` (Boolean) to `purchase.order.type`, with a company-scoped uniqueness constraint
  (at most one default per company, and one global default).
- `order_type` becomes a plain field instead of `compute=`/`store=` driven by `partner_id` - a
  compute can't tell *why* it re-ran, which caused every edge case explored while designing this.
  The same priority - vendor's type, then the company's `is_default` type - now applies wherever
  `order_type` gets resolved, stated once:

  ```python
  def _get_order_type(self, partner, company):
      """Return the order type that fits ``partner``, or ``company``'s default."""
      return self._get_vendor_order_type(partner) or self._default_order_type(company)
  ```

  `_onchange_company()` calls it directly on a company-mismatch correction. `onchange_partner_id()`
  follows the same priority but only overwrites when the vendor has a type of its own or
  `order_type` is still empty - an already-set type is never downgraded to the default just because
  the newly-selected vendor has none. `create()`'s `_set_order_type_defaults()` follows it too, for
  every vals missing an explicit `order_type`, but caches the resolved company default per company
  across a batch create instead of searching it again for every order; it also carries the resolved
  type's `payment_term_id`/`incoterm_id` onto the vals if not already given, closing gap 3.
- `_default_order_type()` only ever returns a type explicitly flagged `is_default` - no fallback to
  "lowest sequence" or any other implicit pick. A vendor without its own type, and no `is_default`
  type configured for the company, simply leaves `order_type` empty, same as an uninstalled/
  unconfigured module today. A default configured for the specific company always wins over a global
  one, regardless of which has the lower sequence - the global default exists for companies that
  have none of their own, not to override a more specific choice.
- A migration (`migrations/16.0.1.1.0/`) flags, per company, the type `_default_order_type()` used to
  silently fall back to before this PR (the lowest-sequence type in that company's scope) - so an
  existing installation keeps resolving to the type it already got, while the choice becomes explicit
  and reconfigurable from here on. A fresh production installation is unaffected: it has nothing
  configured and simply leaves `order_type` empty until a default is set, which is the point of this
  change. The demo data flags `po_type_regular` as the default, so demo/test installations - this
  repo's own CI included - keep getting a type the same way they always did.

## Behavior changes (please read before merging)

These are deliberate, not bugs:

- A vendor with no `purchase_type`, on a company with no `is_default` type configured, now leaves
  `order_type` empty instead of silently picking the lowest-sequence type in scope - see above. An
  existing installation is unaffected (the migration flags its current implicit pick), a fresh one
  starts from nothing configured, matching how every other optional default in Odoo behaves.
- `write({"partner_id": ...})` no longer touches `order_type` at all - only the form and `create()`
  derive it now, since the compute that used to do it on any write is gone.
- `_onchange_company()` now checks the partner's type before the company default, instead of
  resolving straight to the company default.
- Clearing `partner_id` in the form now also clears `order_type` (carried over from the old
  compute's behavior, previously invisible from the onchange path specifically).

Assisted-by: Claude Sonnet 5


baguenth:
Hey @HviorForgeFlow @pedrobaeza @HaraldPanten,

kindly asking for a review :)

HviorForgeFlow:
Code looks good, good improvement, could you follow https://github.com/OCA/.github/blob/master/AI_POLICY.md on commit and description.
It would be nice to reduce the diff, I can see there is method moves that make review difficult.
Also no version bump is needed, ocabot provides it, it will take a look on the migration number and merge according.

Thanks for contributing

baguenth:
> Code looks good, good improvement, could you follow https://github.com/OCA/.github/blob/master/AI_POLICY.md on commit and description. It would be nice to reduce the diff, I can see there is method moves that make review difficult. Also no version bump is needed, ocabot provides it, it will take a look on the migration number and merge according.
> 
> Thanks for contributing

Hey @HviorForgeFlow, 

thank you for the feedback. Let me answer your points:

1) Will update description and commit. From my understanding i just need to add "Assisted-by: Claude Sonnet 5 " and need to remove the Co-Authorship.
2) I have rearranged the functions on purpose, to have a better structure (all onchange functions stand together, etc.). Anyway i understand the diff-impact and will revert that.
3) I thought version bump is needed, when migration is there, sorry. I will revert it.

HviorForgeFlow (approved):

purchase_order_type/models/purchase_order.py
```diff
@@ -53,6 +59,40 @@ def create(self, vals_list):
...
+        A vals that already names an order type, or whose partner
+        provides one of its own, is left as is for the type;
+        payment_term_id/incoterm_id are only filled in when the vals
+        doesn't already name them either, mirroring what
+        onchange_order_type does in the form.
+        """
+        default_per_company = {}
+        for values in vals_list:
+            if not values.get("order_type"):
+                partner = self.env["res.partner"].browse(values.get("partner_id"))
```
dalonsod: For a large set of POs to be created (e.g. during an import process), this query is inefficient. A single `browse` outside for statement `for` where fetching every partner involved should be a better option IMO.

purchase_order_type/views/view_purchase_order_type.xml
```diff
@@ -12,6 +12,7 @@
                 <field name="incoterm_id" />
                 <field name="description" />
                 <field name="company_id" groups="base.group_multi_company" />
+                <field name="is_default" />
```
dalonsod: If `boolean_toggle` widget is added, `is_default` change (unset the old one and, then, set the new one), could be achieved without accessing each form view.

dalonsod (changes requested):
Only code review, see comments. Thanks for the contribution!

purchase_order_type/models/purchase_order.py
```diff
@@ -62,26 +102,28 @@ def _check_po_type_company(self):
...
+        return self._get_vendor_order_type(partner) or self._default_order_type(company)
+
+    def _get_vendor_order_type(self, partner):
+        return partner.purchase_type or partner.commercial_partner_id.purchase_type
+
+    def _default_order_type(self, company=None):
+        company = company or self.company_id
+        po_type = self.env["purchase.order.type"]
+        domain = [("is_default", "=", True)]
+        # A default configured for this specific company always wins over a
```
dalonsod, line 138: I understand this, explained as a new feature. `is_default` actually replaces `sequence` management as fallback mechanism. Then, `sequence` won't have meaning and usage at all, if I'm not wrong. Not sure if it should be removed...

baguenth: > I understand this, explained as a new feature. `is_default` actually replaces `sequence` management as fallback mechanism. Then, `sequence` won't have meaning and usage at all, if I'm not wrong. Not sure if it should be removed...

Yes, you are right. Technically sequence could be dropped. However, the sequence is still critical for the migration step and there could still be a use case where the ordering is somehow relevant for users. Sequence also controls the order in a many2one dropdown etc.

So i would keep the field for now.

purchase_order_type/models/purchase_order.py
baguenth: > For a large set of POs to be created (e.g. during an import process), this query is inefficient. A single `browse` outside for statement `for` where fetching every partner involved should be a better option IMO.

I agree. I will change to a batched browse.

purchase_order_type/views/view_purchase_order_type.xml
baguenth: > If `boolean_toggle` widget is added, `is_default` change (unset the old one and, then, set the new one), could be achieved without accessing each form view.

Good point! I have added the toggle widget in the tree view, so we can now easily set defaults without openening each form view.

sebinthomasst (approved):

BhaveshHeliconia (approved):

baguenth:
> Only code review, see comments. Thanks for the contribution!

Hey @dalonsod,

i have updated the pr as per your recommendations. Can you please check again? :)

baguenth:
> > Only code review, see comments. Thanks for the contribution!
> 
> Hey @dalonsod,
> 
> i have updated the pr as per your recommendations. Can you please check again? :)

Hello @HviorForgeFlow,

can we proceed with this? :)

PR #3180, opened 2026-09-21
[16.0][FIX]purchase_date_planned_manual: guard against False date_order

LoisRForgeFlow:
## Summary

`order_id.date_order` can be transiently `False` on a new/virtual record while onchange values are being computed, even though the underlying order already has a date set. This made `_compute_predicted_arrival_late` crash with an `AttributeError` instead of computing the field correctly.

The fix falls back to the origin record's `date_order` in that case, so the field keeps computing correctly instead of erroring out.


marco-lucaf:
Functional test: OK ✅

Tested in our production environment and everything works as expected.

BhaveshHeliconia (approved):

marco-lucaf (approved):

HviorForgeFlow (approved):

HviorForgeFlow:
@LoisRForgeFlow does this need forward-porting?

